PR #4, opened 2017-05-23
cleanup header

crptm:


common/actions/config.js
```diff
@@ -1,22 +1,12 @@
 // @flow
 export const CONFIG_LANGUAGE_CHANGE = 'CONFIG_LANGUAGE_CHANGE';
-export const CONFIG_LANGUAGE_DROPDOWN_TOGGLE = 'CONFIG_LANGUAGE_DROPDOWN_TOGGLE';
 export const CONFIG_NODE_CHANGE = 'CONFIG_NODE_CHANGE';
-export const CONFIG_NODE_DROPDOWN_TOGGLE = 'CONFIG_NODE_DROPDOWN_TOGGLE';
 
 
-export const CHANGE_LANGUAGE = (index: number) => Object({
-    type: CONFIG_LANGUAGE_CHANGE, index: index
+export const CHANGE_LANGUAGE = (value: any) => Object({
```
dternyak: why not `value: string`?

crptm: Its moving whole objects to save some typing, could change later

dternyak: gotcha

common/containers/App/index.jsx
```diff
@@ -87,24 +72,18 @@ function mapStateToProps(state) {
...
 }
 
 function mapDispatchToProps(dispatch) {
     return {
         changeNode: (i: number) => {
-            dispatch(CHANGE_NODE(i))
-        },
-        toggleNodeDropdown: () => {
-            dispatch(TOGGLE_NODE_DROPDOWN())
+            dispatch(CHANGE_NODE(i));
```
dternyak: Is this right? Thought we were passing the whole object now? should be changed to `i: any`?

crptm: Indeed, file has no flow declation so wasnt catched :)

dternyak:
Really good refactors :) thanks for this.

PR #5, opened 2017-05-24
Help view

dternyak:
Initial help view - we can build on this later or decide to completely deprecate in favor of knowledge base.

### DIFF
--- a/common/components/Header/components/TabsOptions.jsx
+++ b/common/components/Header/components/TabsOptions.jsx
@@ -23,7 +23,8 @@ const tabs = [
         link: 'view-wallet'
     },
     {
-        name: 'Help'
+        name: 'Help',
+        link: 'help'
     }
 ]
 
--- a/common/containers/Tabs/Help/index.js
+++ b/common/containers/Tabs/Help/index.js
@@ -1,0 +1,34 @@
+import React, {Component} from "react";
+
+export default class Help extends Component {
+    constructor(props) {
+        super(props)
+    }
+
+    render() {
+        return (
+            <section className="container" style={{minHeight: '50%'}}>
+                <div className="tab-content">
+
+                    <article className="tab-pane help active">
+                        <h1 translate="NAV_Help">Help</h1>
+                        <article className="collapse-container">
+                            <div>
+                                <ul>
+                                    <li><h3><a
+                                        href="https://www.reddit.com/r/ethereum/comments/47nkoi/psa_check_your_ethaddressorg_wallets_and_any/d0eo45o"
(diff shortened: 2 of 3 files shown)


common/components/Header/components/TabsOptions.jsx
```diff
@@ -23,7 +23,8 @@ const tabs = [
         link: 'view-wallet'
     },
     {
-        name: 'Help'
+        name: 'Help',
```
dternyak: Probably a better convention than this - maybe react-router even has a way of listing all routes. 

Can fix in a separate PR.

tayvano: Off the top of my head each 'view' or page or whatever should have a name (Help Me) slug (help-me) which is the url, title (SEO title tag + title shown on page itself), description (OMG wat more SEO?) And ability to hide / show in the nav for different frontends (eg website vs chrome extension) - although this last one should maybe be done on the compile level? 

The ability for me to quickly turn on or off access to a view has come in super handy more than once, and having a single location that controls what views are available for which frontend makes keeping the chrome extension up to date essentially mindless.  

While in the future I hope we don't have 'turn it off and then we can figure it out' moments...it's better to be prepared...especially if I happen to be the only one around.

tayvano: ps: When I comment like this, take it or leave it. I'll make it clear when I feel strongly about something existing in a particular way. For most things, I'm going to share insightsabout what we've learned and you can use those as you proceed. Please don't make a decision based solely on something I've said, especially if it compromises something else.

crptm: So re:RR, we should probably switch to v4. No specific reason besides forward compatibility.

tayvano:
I can do the help entries if someone sets the first one up. If we could have separate files for each item, that would work. I am going to keep the help entries much shorter: the most common "Getting Started" and info regarding error messages.

Guides and other questions will be https://myetherwallet.groovehq.com/help_center

Also keep in mind, help is translated and the "save your key" is important to have translated. We used markdown for the current site but I don't think markdown is necessary though, unless we are using it elsewhere already (not another lib = 👍).

crptm:
I think there are only links and bolds from markdown i will prolly ad-hoc parse them.

common/components/Header/components/TabsOptions.jsx
dternyak: @crptm Sorry, I'm a bit confused. What should we switch in v4?

crptm: @dternyak react-router, we are on 3.x release latest is 4.x

dternyak:
@crptm Yeah, separate PR to start laying out a way for @tayvano to be able to setup help entries.

PR #6, opened 2017-05-25
Baby translations

crptm:
We can start with something like that. Honestly cant think of a good way to do it which doesnt involve babel transforms :D


common/actions/config.js
```diff
@@ -1,12 +1,18 @@
...
+};
 
-export const CHANGE_LANGUAGE = (value: any) => Object({
-    type: CONFIG_LANGUAGE_CHANGE, value
-})
-
-export const CHANGE_NODE = (value: any) => Object({
-    type: CONFIG_NODE_CHANGE, value
-})
+export const CHANGE_NODE = (value: any) =>
```
dternyak: Just whitespace changes? Already a huge PR, would be nice to avoid this.

crptm: @dternyak yeah i am trying to not change files you could be touching, this is a prettier thing. One time formatting war :D

dternyak: 👍

crptm:
Merging

PR #9, opened 2017-05-31
Initial work for "Wallet Decrypt" component

NessDan:
Fixed up the markup and naming of the plugin, next will be adding state and having it function like it used to.


NessDan:
I updated the aria-labels as well since they were being used wrong

https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_aria-labelledby_attribute

common/containers/Tabs/WalletDecrypt/index.jsx
```diff
@@ -0,0 +1,117 @@
+import React, {Component} from 'react';
+import translate from 'translations';
+
+
+export default class WalletDecrypt extends Component {
+    constructor(props) {
```
crptm: My OCD cant stand this :)

common/routing/index.jsx
```diff
@@ -3,15 +3,15 @@ import {browserHistory, Redirect, Route} from 'react-router';
...
+import WalletDecrypt from 'containers/Tabs/WalletDecrypt'
 import Help from 'containers/Tabs/Help'
 
 export const history = getHistory()
 
 export const Routing = () => (
     <Route name="App" path='' component={App}>
         <Route name="GenerateWallet" path="/" component={GenerateWallet}/>
-        <Route name="ViewWallet" path="/view-wallet" component={ViewWallet}/>
+        <Route name="WalletDecrypt" path="/view-wallet" component={WalletDecrypt}/>
```
crptm: It is not route tho

common/containers/Tabs/WalletDecrypt/index.jsx
NessDan: What exactly?

crptm: Constructor that does nothing :(

common/containers/Tabs/WalletDecrypt/index.jsx
```diff
@@ -0,0 +1,117 @@
...
+                                                aria-labelledby="trezor-label"
+                                                type="radio"
+                                                value="trezor"
+                                                name="142"
+                                            />
+                                            <span id="trezor-label">
+                                                {translate('x_Trezor')}
+                                            </span>
+                                        </label>
+                                        <label className="radio">
```
dternyak: Kinda would rather see all these be an array of dicts defined at the top, and then iterated in here.

NBD though, just a suggestion.

common/containers/Tabs/WalletDecrypt/index.jsx
NessDan: Good catch! Haha, removing them now.

common/containers/Tabs/WalletDecrypt/index.jsx
NessDan: @dternyak You're absolutely right. I think I will break these out and do that.

I was so fixated on the markup that I forgot to write it right 😢

NessDan:
Closing this, will get more work done to it and hopefully have it even stronger than it was before!

PR #10, opened 2017-05-31
linkify in markup.js now adds keys to multiple links.

NessDan:
Notably, LID "VIEWWALLET_Subtitle" was causing an issue here, since it
had more than one link. React was complaining about it but now it's all
good.


NessDan:
One thing to note: I had to add a `key` variable to keep track of iterations, the `i` isn't being used in the traditional sense and is keeping count of other information, not `index` info.

davidgreens:
@hektve87 @davidgreens

davidgreens:
@NessDan @dternyak @hektve87 @davidgreens

PR #14, opened 2017-06-12
Cleanup

dternyak:
Fixes active class on nav options and removes some comments.

### DIFF
--- a/common/components/Header/components/TabsOptions.jsx
+++ b/common/components/Header/components/TabsOptions.jsx
@@ -11,7 +11,8 @@ const tabs = [
         name: 'NAV_SendEther'
     },
     {
-        name: 'NAV_Swap'
+        name: 'NAV_Swap',
+        link: 'swap'
     },
     {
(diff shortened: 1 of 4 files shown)


common/components/Header/components/TabsOptions.jsx
```diff
@@ -62,7 +63,7 @@ export default class TabsOptions extends Component {
                             {
                                 tabs.map((object, i) => {
                                         // if the window pathname is the same or similar to the tab objects name, set the active toggle
-                                        const activeOrNot = (window.location.pathname === object.name || window.location.pathname.substring(1) === object.name) ? 'active' : '';
+                                        const activeOrNot = (window.location.pathname === object.link || window.location.pathname.substring(1) === object.link) ? 'active' : '';
```
crptm: better use router props.location object, that way we could make it work in non-browser contexts

NessDan: @crptm is `props.location` something passed from React Router?

PR #17, opened 2017-06-13
Use `fs` library so even Windows users can run build.

NessDan:
Noticed that we were using `exec` to run unix-style commands for clearing up old builds.

This'll use FS so it'll work on everyone's machine :)

image


dternyak:
lgtm

PR #32, opened 2017-07-05
Redux: Persist language selection to local storage via middleware and…

dternyak:
… create pattern for future persistence.

### DIFF
--- a/common/utils/localStorage.js
+++ b/common/utils/localStorage.js
@@ -1,0 +1,33 @@
+export const REDUX_STATE = 'REDUX_STATE';
+
+export const loadState = () => {
+  try {
+    const serializedState = localStorage.getItem(REDUX_STATE);
+    if (serializedState === null) {
+      return undefined;
+    }
+    return JSON.parse(serializedState);
+  } catch (err) {
+    // TODO - log warning? or bubble?
+  }
+};
+
+export const saveState = state => {
+  try {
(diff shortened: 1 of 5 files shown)


common/utils/localStorage.js
```diff
@@ -0,0 +1,33 @@
...
+
+export const loadState = () => {
+  try {
+    const serializedState = localStorage.getItem(REDUX_STATE);
+    if (serializedState === null) {
+      return undefined;
+    }
+    return JSON.parse(serializedState);
+  } catch (err) {
+    // TODO - log warning? or bubble?
```
crptm: we can console.log('Warn: blah')

dont think user cares much about corrupted localStorage

common/utils/localStorage.js
```diff
@@ -18,7 +18,7 @@ export const saveState = (state: Object) => {
     const serializedState = JSON.stringify(state);
     localStorage.setItem(REDUX_STATE, serializedState);
   } catch (err) {
-    // TODO - log warning? or bubble?
+    console.warn(' Warning: corrupted local storage');
```
crptm: leading space OCDs me :D, also i think this message is factually incorrect, we can use something like 'failed to save to local storage' or something. Load case is about getting something broken from storage, but here we either got object that isnt serializable or browser failed to save to localstorage for w/e reason

PR #36, opened 2017-07-07
Fixed trying to improperly delete our dist folder.

NessDan:
Noticed an error when we tried deleting the `dist/` folder, that since it wasn't empty it couldn't do it.

This fixes that :)

snip_20170706231056


crptm:
Can we use rimraf? Also this helper has no business being in app source folder

PR #37, opened 2017-07-09
Sagas: Adjust bity saga to allow for cancelled polling, and stop poll…

dternyak:
…ing when part 2 of swap ends.


common/sagas/bity.js
```diff
@@ -1,23 +1,45 @@
...
+    if (yield cancelled()) {
+      // TODO - implement request cancel if needed
+      // yield put(actions.requestFailure('Request cancelled!'))
+    }
   }
 }
 
 export default function* bitySaga(): Generator<Effect, void, any> {
-  yield takeLatest(SWAP_LOAD_BITY_RATES, loadBityRates);
+  while (yield take(SWAP_LOAD_BITY_RATES)) {
```
dternyak: @crptm I'm not confident about this, used the pattern described here: https://redux-saga.js.org/docs/advanced/TaskCancellation.html

common/containers/Tabs/Swap/components/receivingAddress.js
```diff
@@ -19,6 +20,7 @@ export default class ReceivingAddress extends Component {
   };
 
   onClickPartTwoComplete = () => {
+    this.props.stopLoadBityRates();
```
crptm: what if user never clicks?

.eslintrc.json
```diff
@@ -24,6 +24,7 @@
     "no-underscore-dangle": 0,
     "no-console": 0,
     "no-unused-vars": 0,
+    "no-constant-condition": 0,
```
crptm: if you plan on keeping disabling rules instead of fixing, lets remove eslint.

dternyak: This was specifically because ```
  try {
    while (true) {
 ...
``` was stoping webpack from even starting. I think its setup to the point where we won't have to touch it too much more.

common/containers/Tabs/Swap/components/receivingAddress.js
dternyak: Good question. The existing site seems to just keep polling, so maybe its okay for us too? Do you have any ideas?

.eslintrc.json
crptm: just look at the diff, if we dont want to fix linter issues why have it at all. It would be easier to just remove config and let those interested in linting have their own config.

dternyak:
AK said good to merge in slack

PR #39, opened 2017-07-09
Node's network based header background

rottingcorpse:
Changes MEW header background based on selected node's network.

Default: blue'ish gradient for ETH
ETC: Green gradient (green and blue color values swapped for header gradient colors)

### DIFF
--- a/common/assets/styles/etherwallet-utilities.less
+++ b/common/assets/styles/etherwallet-utilities.less
@@ -22,10 +22,8 @@
 
 .bg-gradient {
   background: @ether-navy;
-  &.eth {
-    background: linear-gradient(149deg, #132a45, #143a56, #21a4ce, #19b4ad);
-  }
-  &.etc {
+  background: linear-gradient(149deg, #132a45, #143a56, #21a4ce, #19b4ad);
+  &.ETC {
     background: linear-gradient(149deg,#13452a,#14563a,#21cea4,#19adb4);
   }
 }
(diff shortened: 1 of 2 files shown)


common/components/Header/index.jsx
```diff
@@ -22,7 +22,7 @@ export default class Header extends Component {
 
     return (
       <div>
-        <section className="bg-gradient header-branding">
+        <section className="bg-gradient header-branding {selectedNode.network}">
```
NessDan: The class name feels too generic, does attaching a `branding-node-{selectedNode.network}` sound good?

common/assets/styles/etherwallet-utilities.less
```diff
@@ -23,8 +23,12 @@
 .bg-gradient {
   background: @ether-navy;
   background: linear-gradient(149deg, #132a45, #143a56, #21a4ce, #19b4ad);
+  &.ETC {
```
NessDan: ETC doesn't exist in the config data file, is there plans to add ETC to MEW?

common/components/Header/index.jsx
crptm: This wouldnt interpolate

common/assets/styles/etherwallet-utilities.less
crptm: yep

dternyak:
@crptm / @NessDan This PR is to claim a bounty against V3 and as such should be submitted at https://github.com/kvhnuke/etherwallet

PR #40, opened 2017-07-11
Docker

dternyak:
Helps people like @NessDan when ethereum-blockies breaks their env :)


dternyak:
Probably don't need to push people to use it unless they need to. Could mention to use it on windows on readme, but probably not needed for now.

Prettier is done on commit, which is outside of the container so not needed here.

PR #47, opened 2017-07-14
support tokens on send page

crptm:


common/containers/Tabs/SendTransaction/index.jsx
```diff
@@ -136,6 +141,10 @@ export class SendTransaction extends React.Component {
                   <AmountField
                     value={value}
                     unit={unit}
+                    tokens={this.props.tokenBalances
```
dternyak: I would prefer if this was separated out to a method.

common/containers/Tabs/SendTransaction/index.jsx
```diff
@@ -251,9 +260,16 @@ export class SendTransaction extends React.Component {
   };
 
   onAmountChange = (value: string, unit: string) => {
-    // TODO: tokens
+    // TODO sub gas for eth
```
dternyak: ?

common/containers/Tabs/SendTransaction/components/AmountField.jsx
```diff
@@ -32,7 +34,7 @@ export default class AmountField extends React.Component {
           />
           <UnitDropdown
             value={unit}
-            options={['ether']}
+            options={['ether'].concat(this.props.tokens)}
```
dternyak: Existing side uses the symbol instead of the name (e.g. ETH)

common/containers/Tabs/SendTransaction/index.jsx
crptm: You cant send ALL eth, bc you wouldnt be able to pay for gas.

common/containers/Tabs/SendTransaction/components/AmountField.jsx
crptm: Yep, it is that way so we dont forget to support ETC/RSK or whatever else we need to support.

dternyak:
LGTM! Nice bite-sized chunk while we continue to work on the send view.

PR #50, opened 2017-07-14
Reusable modal component

wbobeirne:
Here's a proposal for a reusable modal component. I didn't emphasize styling all that much, but it would be _very_ easy to restyle this modal even after it gets used, as it's pretty agnostic about its content.

An example of using the modal would look something like this:

```jsx
import React from "react";
import Modal from "components/ui/Modal";

export default MyComponent extends Component {
  state = {
    isModalOpen: false,
  };

  _openModal() {
    this.setState({ isModalOpen: true });
  }

  _closeModal() {
    this.setState({ isModalOpen: false });
  }

  _saveModal() {
    this.props.doSomeSaveThing();
    this._closeModal();
  }

  render() {
    const modalButtons = [{
      text: 'Save',
      type: 'primary',
      onClick: this._saveModal,
    }, {
      text: 'Close',
      onClick: this._closeModal,
    }];
  
    return (
      <div className="MyComponent">
        <button onClick={this._openModal}>Open Modal</button>
        <Modal
          isOpen={this.state.isModalOpen}
          title="Example Modal"
          size="medium"
          buttons={modalButtons}
          handleClose={this._closeModal}
        >
          <p>I’m a paragraph in a modal! Neat-o.</p>
        </Modal>
      </div>
    );
  }
}
```

Modals look like this when they open: https://gfycat.com/gifs/detail/MelodicDifficultCheetah

They come in four sizes, `small`, `medium` (default), `large`, and `xlarge`. The only thing this affects is the max width:

modal-sizes

Buttons are optional. When omitted, the footer disappears. This can be good for non-interactive, read-only modals. Here's an example:

modal-read-only

Modals are responsive. The width will shrink (With some slight padding from the side of the windows.) The header and footer are always visible, the content will begin to scroll if it gets too short:

modal-responsive

Modals also lock body scrolling while open (Hard to demo something NOT happening, so I didn't record anything for that.)

This branch extends the `css-modules` branch (Being PR'd at #48) so unfortunately this PR looks a lot bigger than it really is. All you need to look at is `common/components/ui/Modal.jsx` and `common/components/ui/Modal.scss`


wbobeirne:
One thing I failed to point out was the behavior of the `handleClose` prop. Modals don't internally handle their opened / closed state, the parent component does. However, they'll communicate when they _should_ close. This can happen by clicking the X in the top right, clicking on the shade behind the modal, or hitting the `Esc` key on your keyboard while not targeting any inputs. All 3 will call the `handleClose` prop.

This is so that things that might need to intercept a close, or perform an action on the close, have a chance to handle it. This could be useful for a confirmation / warning message on a form in a modal, or to save something before closing it.

common/components/ui/Modal.jsx
```diff
@@ -0,0 +1,107 @@
...
+import React, { Component } from 'react';
+import PropTypes from 'prop-types';
+import closeIcon from 'assets/images/icon-x.svg';
+
+import './Modal.scss';
+
+export default class Modal extends Component {
+  static propTypes = {
+    isOpen: PropTypes.bool,
+    size: PropTypes.oneOf(['small', 'medium', 'large', 'xlarge']),
```
crptm: Would prefer content pushing modal size

common/components/ui/Modal.jsx
```diff
@@ -0,0 +1,107 @@
+// @flow
+import React, { Component } from 'react';
+import PropTypes from 'prop-types';
+import closeIcon from 'assets/images/icon-x.svg';
+
+import './Modal.scss';
+
+export default class Modal extends Component {
+  static propTypes = {
+    isOpen: PropTypes.bool,
```
crptm: Not really needed, parent component could conditionally render

common/components/ui/Modal.jsx
```diff
@@ -0,0 +1,107 @@
...
+
+    if (ev.key === 'Escape' || ev.keyCode === 27) {
+      this.props.handleClose();
+    }
+  };
+
+  _renderButtons() {
+    let idx = 0;
+
+    return this.props.buttons.map(btn => {
```
crptm: (btn, idx) =>

common/components/ui/Modal.scss
```diff
@@ -0,0 +1,139 @@
...
+    transform: translate(-50%, -50%) scale(0.88);
+  }
+  100% {
+    opacity: 1;
+    transform: translate(-50%, -50%) scale(1);
+  }
+}
+
+.Modalshade {
+  position: fixed;
```
crptm: Body will still scroll on wheel, right?

common/components/ui/Modal.jsx
wbobeirne: How have I gone on this long without knowing that map had a second argument for index...

common/components/ui/Modal.jsx
wbobeirne: It does for height, but how could that work for width?

common/components/ui/Modal.scss
wbobeirne: See `componentWillReceiveProps` and `componentWillUnmount` in `Modal.jsx`, it adds / removes a body class that prevents scrolling.

tayvano:
Okay so I put this somewhere but basically for the confirmation modals it should always parse the signed TX and display:

- From address + addressicon
- Amount
- Coin Type
- To address + addressicon
- Gas Limit
- Gas Price 
- TX Fee in ETH + USD (USD situation is optional for now)
- Data
- Title / Message (e.g. "You are about to send a transaction" or "you are about to resolve your ENS name" or whatever is passed to it)
- Current chain (ETH vs ETC)
- Who node is provided by

I think what you've built here is great for that and love the table. 

This gives us a really reusable modal and acts as a final check for the user. One of the things I really don't like is, for example, the current contract page modal doesn't give you anything. No address, no idea how much its going to cost to deploy, etc.

---

**Current issues with current conf modals**

- People dont read them (had someone the other day literally tell me we need a confirmation for the confirmation....)

- Too easy to miss the *coin* you are sending (e.g. ETH vs SNT)

- Too easy to miss the chain you are on (ETC vs ETH), although the border I added has helped reduce those

- Not the information the user wants / needs in an easily parsable / skimmible *thing* (table format will help)

---

Separately, I hate how modals sometimes close all by themselves sometimes if you click twice or click outside. I do it on Twitter all the time and I think it's worse with bootstrap v4. I lose tweets ALL THE TIME when I'm trying to highlight something. 

It shouldn't close unless you hit the X or the CANCEL button or hit the esc button (optional)

--

Other than that, I really like the success / error state inside the modal itself, but we'll have to decide how to handle alerts and things. While some successes and errors come from front-end validation, some are message returned directly from the node (e.g. a TX with that hash has already been imported). It would be safer to handle all alerts via the alerts handler, avoiding a potential situation where you sometimes show alert, sometimes put alert in modal, and if you aren't careful, accidentally don't show the alert at all. Making things fail-proof for our future selves is very important so we can build quickly with less worry. 😉

common/components/ui/Modal.scss
crptm: Cant see those in changes :/ Anyhow we can do with on didmount willunmount if we remove closed state

common/components/ui/Modal.jsx
wbobeirne: This was my first instinct as well, but after looking at [react-modal](https://github.com/reactjs/react-modal), they did this to allow for internally managed animations. I didn't have any fancy close animations going on for this one, but I wanted to choose the most flexible option for future changes to the modals.

If you feel like this isn't a worthwhile consideration, I'd be happy to ditch it.

common/components/ui/Modal.jsx
crptm: Just leave max-width decision to those passing children

crptm:
@tayvano we need to track that in separate issue, here we are only concerned with generic reusable modal component

wbobeirne:
This PR isn't for the confirmation modals, it's just so that we have a One True Modal component. After this lands, someone can use this modal to build a ConfirmationModal component or something.

I hear you on the windowshade thing though. I'll go ahead and remove click to close on that.

crptm:
@wbobeirne as per @tayvano comments we need to make esc/click outside opt-in.

common/components/ui/Modal.scss
crptm: I am also pretty sure we dont need to filter event.target in that case (modal is in react tree only if its open)

common/components/ui/Modal.jsx
wbobeirne: I'm not sure how I would do that. If children components define an explicit width, that would work. But that won't work for responsive content, as doing something like `max-width: 600px; width: 100%;` won't correctly set the width. Example:

screen shot 2017-07-14 at 4 04 40 pm

common/components/ui/Modal.jsx
crptm: Dunno, this makes implementation/usage simpler. Moving back to add animations could be tough tho :D

common/components/ui/Modal.scss
wbobeirne: That filtering is so that if you're inside of a form input in a modal, and you hit escape to blur it, you don't also close the modal.

wbobeirne:
I'd prefer convention over configuration. Either all modals should be esc to close, or none. Same goes with windowshade. Users should have a reasonable expectation for how to close modals.

common/components/ui/Modal.jsx
crptm: I am not even sure what `max-width: 600px; width: 100%;` does. 

But if you have chidren wrapped in container with maxwidth it is exact same behaviour as seen in modal right now or not?

wbobeirne: the max-width / width 100% combo makes it so that a container is at most max-width px, but will allow itself to shrink down to the size of its container. It's how most responsive things do what they do.

Unfortunately it doesn't have the same behavior if I do that, as the screenshot above shows. The `width: 100%;` doesn't work because the modal has `width: auto;` (To determine width from its content.) So the only thing that would work, is if the content set an explicit width. But then it wouldn't be responsive.

crptm: screenshot 2017-07-15 02 48 51

this + wrapping children in <div style={{width: desired-max-w}}> seems like replicating existing behaviour, can you please re-check on your end?

PR #52, opened 2017-07-15
Working paper wallet

wbobeirne:
With my first pass at the generate wallet flow, I left some stuff in the "placeholder" phase. I've circled back around and properly implemented the private key and printable paper wallet now. This required some changes to `lib/keystore.js` that I definitely want to make sure were desirable changes.

But on the more fun side, I ended up rewriting the paper wallet as a react component with inline styles. As much as I despised writing them, inline styles were needed for two reasons:
1. In order to print the wallet, you must sock its HTML into a popup window, and print that. This popup window doesn't have our stylesheet. Making them inline causes all of the styles to carry over.
2. Making them inline styles insulates the paper wallet from any other styling changes around the site. We can be guaranteed that the paper wallet's styling will never be broken by any other parts of the site.

Here's what the flow looks like now, the wallet is inlined because it's in react land now, but clicking print works pretty identically to how it did before:

screen shot 2017-07-15 at 1 35 06 am

cc @dternyak @crptm


common/libs/keystore.js
```diff
@@ -11,6 +11,14 @@ export const scryptSettings = {
 
 export const kdf = 'scrypt';
 
+export function genNewPkey(password: string): Buffer {
+  return randomBytes(32);
+}
+
+export function pkeyToAddress(pkey: Buffer): string {
```
dternyak: nice addition

common/libs/keystore.js
```diff
@@ -11,6 +11,14 @@ export const scryptSettings = {
 
 export const kdf = 'scrypt';
 
+export function genNewPkey(password: string): Buffer {
```
dternyak: takes an unused password arg?

PR #57, opened 2017-07-17
Enforce node / npm versions, add package-lock.json

wbobeirne:
## What this does

* Adds a package-lock.json file (Added in NPM v5+) that locks us into specific
node module versions
* Adds pre script hooks to ensure that you're on Node 8+ / NPM 5+ before running
commands

This is what you'll see if you're not on Node 8 when you run:

```
node: 7.7.2
npm: 4.1.2
yarn: 0.23.3
Error: Wanted node version >= 8.0.0 (>=8.0.0)
To install node, run `nvm install >= 8.0.0` or see https://nodejs.org/
Error: Wanted npm version >= 5.0.0 (>=5.0.0)
To install npm, run `npm install -g npm@>= 5.0.0`
```

## Why do this

On the development side, it's a pain to debug issues if we're not all on the
same page. Some issues may present themselves differently, or be limited to
ceertains versions of node.

On the production / build side, it's _extremely_ important that the project be
compiled in identical conditions. Docker is great for this, but this is just
another layer to ensure that everything is the same.

## Caveats

Devs who were not on node 8 before will need to update to it to continue. They
should do so using `nvm` (To be able to jump back to node 7 or lower if they need to
elsewhere.) This may cause some of your node modules to break (Sass did for me)
so you should re-install all of your modules after upgrading, which is a good idea anyway since your package versions might differ from what's in package-lock.json 
(`rm -rf node_modules; npm install`)


package.json
```diff
@@ -41,6 +45,7 @@
     "babel-preset-react-optimize": "^1.0.1",
     "babel-preset-stage-0": "^6.22.0",
     "babel-runtime": "^6.23.0",
+    "check-node-version": "^2.1.0",
```
macbem: Should be a dev-dependency

wbobeirne:
Travis will need to be updated to respect the package.json engines attribute. It looks like it's doing `nvm install 7` right now. Where is that configured, and who has access to it?

package.json
wbobeirne: It is (Expand the diff)

macbem: My bad :|

wbobeirne:
Thank god for Travis CI, the `package-lock.json` I generated had some cached mode modules from my node 7 install. Cleared cache and regenerated, and the package-lock is fixed.

.nvmrc
```diff
@@ -0,0 +1 @@
+8
```
NessDan: May be a silly question, but is +8 the correct syntax, not 8+?

NessDan:
I'll pull this down locally and confirm things are good for Docker environments once I fix my Docker env :)

.nvmrc
wbobeirne: The + is from the diff. The file only contains `8`, which is the correct syntax (It just pulls the latest stable 8 that nvm can grab)

NessDan:
Broke the docker 😢 

```Attaching to myetherwallet_web_1
web_1  | npm info it worked if it ends with ok
web_1  | npm info using npm@5.0.3
web_1  | npm info using node@v8.1.4
web_1  | npm info lifecycle MyEtherWallet@0.0.1~predev: MyEtherWallet@0.0.1
web_1  |
web_1  | > MyEtherWallet@0.0.1 predev /usr/app
web_1  | > check-node-version --package
web_1  |
web_1  | sh: 1: check-node-version: not found
web_1  | npm info lifecycle MyEtherWallet@0.0.1~predev: Failed to exec predev script
web_1  | npm ERR! file sh
web_1  | npm ERR! code ELIFECYCLE
web_1  | npm ERR! errno ENOENT
web_1  | npm ERR! syscall spawn
web_1  | npm ERR! MyEtherWallet@0.0.1 predev: `check-node-version --package`
web_1  | npm ERR! spawn ENOENT
web_1  | npm ERR!
web_1  | npm ERR! Failed at the MyEtherWallet@0.0.1 predev script.
web_1  | npm ERR! This is probably not a problem with npm. There is likely additional logging output above.
web_1  |
web_1  | npm ERR! A complete log of this run can be found in:
web_1  | npm ERR!     /root/.npm/_logs/2017-07-19T00_01_02_364Z-debug.log
myetherwallet_web_1 exited with code 1
PS D:\NessDan\Documents\Code\MyEtherWallet>```

package.json
ny030897: !

PR #61, opened 2017-07-19
Gas price dropdown

skubakdj:
Implements the gas price slider in the header. I'm very new to React, Redux, and Flow, so please point out anything silly I might have done :smile:

### DIFF
--- a/common/components/Header/components/GasPriceDropdown.jsx
+++ b/common/components/Header/components/GasPriceDropdown.jsx
@@ -3,19 +3,17 @@ import React, { Component } from 'react';
 import PropTypes from 'prop-types';
 
 import './GasPriceDropdown.scss';
+import { gasPriceDefaults } from 'config/data';
 
 export default class GasPriceDropdown extends Component {
   state = { expanded: false };
 
   static propTypes = {
-    gasPriceGwei: PropTypes.number,
-    changeGasPrice: PropTypes.func.isRequired
+    value: PropTypes.number,
+    onChange: PropTypes.func.isRequired
   };
 
   render() {
-    const gasPriceMinGwei = 1;
-    const gasPriceMaxGwei = 60;
-
     return (
       <span className="dropdown">
         <a
@@ -24,18 +22,18 @@ export default class GasPriceDropdown extends Component {
           className="dropdown-toggle"
           onClick={this.toggleExpanded}
         >
-          <span>Gas Price</span>: {this.props.gasPriceGwei} Gwei
+          <span>Gas Price</span>: {this.props.value} Gwei
           <i className="caret" />
         </a>
         {this.state.expanded &&
           <ul className="dropdown-menu GasPrice-dropdown-menu">
             <div className="GasPrice-header">
-              <span>Gas Price</span>: {this.props.gasPriceGwei} Gwei
+              <span>Gas Price</span>: {this.props.value} Gwei
               <input
                 type="range"
-                value={this.props.gasPriceGwei}
-                min={gasPriceMinGwei}
-                max={gasPriceMaxGwei}
+                value={this.props.value}
+                min={gasPriceDefaults.gasPriceMinGwei}
+                max={gasPriceDefaults.gasPriceMaxGwei}
                 onChange={this.updateGasPrice}
               />
               <p className="small col-xs-4 text-left GasPrice-padding-reset">
@@ -77,6 +75,6 @@ export default class GasPriceDropdown extends Component {
   };
 
   updateGasPrice = (e: SyntheticInputEvent) => {
-    this.props.changeGasPrice(e.currentTarget.valueAsNumber);
+    this.props.onChange(e.currentTarget.valueAsNumber);
   };
 }
--- a/common/components/Header/components/GasPriceDropdown.jsx
+++ b/common/components/Header/components/GasPriceDropdown.jsx
@@ -5,17 +5,11 @@ import PropTypes from 'prop-types';
 import './GasPriceDropdown.scss';
 
 export default class GasPriceDropdown extends Component {
-  constructor(props) {
-    super(props);
-
-    this.state = {
-      expanded: false
-    };
-  }
+  state = { expanded: false };
 
   static propTypes = {
     gasPriceGwei: PropTypes.number,
-    changeGasPrice: PropTypes.func
+    changeGasPrice: PropTypes.func.isRequired
   };
 
   render() {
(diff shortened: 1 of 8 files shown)


common/components/Header/index.jsx
```diff
@@ -1,6 +1,7 @@
 // @flow
 import React, { Component } from 'react';
 import Navigation from './components/Navigation';
+import GasPriceDropdown from './components/GasPriceDropdown.jsx';
```
crptm: Skip extensions pls

common/reducers/config.js
```diff
@@ -14,7 +15,10 @@ export type State = {
 
 export const initialState: State = {
   languageSelection: languages[0].sign,
-  nodeSelection: Object.keys(NODES)[0]
+  nodeSelection: Object.keys(NODES)[0],
+  gasPriceGwei: 21,
+  gasPriceMaxGwei: 60,
```
crptm: if min/max values do not change during app lifetime, we probably dont want to have them in state

crptm: You also need to add new fields to `type State`

common/components/Header/components/GasPriceDropdown.jsx
```diff
@@ -0,0 +1,100 @@
+// @flow
+import React, { Component } from 'react';
+import PropTypes from 'prop-types';
+
+import './GasPriceDropdown.scss';
+
+export default class GasPriceDropdown extends Component {
+  constructor(props) {
```
crptm: Looks like i am the only one, but i think code w/o constructors (where they not needed) are easier to read.

Here you can delete whole consturctor thing and replace it with 
`
export default class GasPriceDropdown extends Component {
  state = {
   expanded: false
  }
`

common/components/Header/components/GasPriceDropdown.jsx
```diff
@@ -0,0 +1,100 @@
...
+  static propTypes = {
+    gasPriceGwei: PropTypes.number,
+    gasPriceMinGwei: PropTypes.number,
+    gasPriceMaxGwei: PropTypes.number,
+    changeGasPrice: PropTypes.func
+  };
+
+  render() {
+    // col-xs-4 applies padding not observed in MEWv3, this overrides
+    const paddingReset = { paddingLeft: '0px', paddingRight: '0px' };
```
wbobeirne: You've got an scss file, it would be better served to do styling in there.

common/components/Header/components/GasPriceDropdown.jsx
```diff
@@ -0,0 +1,100 @@
...
+          className="dropdown-toggle"
+          onClick={this.toggleExpanded}
+        >
+          <span>Gas Price</span>: {this.props.gasPriceGwei} Gwei
+          <i className="caret" />
+        </a>
+        {this.state.expanded &&
+          <ul
+            className="dropdown-menu"
+            style={{ padding: '.5rem', minWidth: '300px' }}
```
wbobeirne: Move to scss

common/components/Header/components/GasPriceDropdown.jsx
```diff
@@ -0,0 +1,100 @@
...
+              <p style={paddingReset} className="small col-xs-4 text-right">
+                Fast AF
+              </p>
+              <p
+                className="small"
+                style={{
+                  whiteSpace: 'normal',
+                  fontWeight: '300',
+                  margin: '2rem 0 0'
+                }}
```
wbobeirne: Move to scss

common/components/Header/components/GasPriceDropdown.scss
```diff
@@ -0,0 +1,17 @@
+@import "../../../../common/sass/variables";
```
wbobeirne: You can just to `@import "common/sass/variables";`, it looks from the top of the repo.

common/components/Header/components/GasPriceDropdown.jsx
```diff
@@ -0,0 +1,100 @@
...
+  toggleExpanded = () => {
+    this.setState(state => {
+      return {
+        expanded: !state.expanded
+      };
+    });
+  };
+
+  updateGasPrice = (e: SyntheticInputEvent) => {
+    this.props.changeGasPrice(e.currentTarget.valueAsNumber);
```
wbobeirne: If you call a prop function without checking for it first, you'll need to make the propType `PropTypes.func.isRequired`

crptm: lets get rid of proptypes :D

common/components/Header/components/GasPriceDropdown.jsx
skubakdj: Yep, I definitely agree that's easier to read. I'll add that in :smile:

common/components/Header/components/GasPriceDropdown.jsx
skubakdj: Totally agree the scss file is underutilized at the moment. I'll move the styling over as suggested :smile:

common/components/Header/components/GasPriceDropdown.scss
skubakdj: Sounds good! Webstorm was throwing a warning with this path, but indeed it seems to work fine :+1:

common/components/Header/components/GasPriceDropdown.jsx
skubakdj: Looks like there's a Babel plugin for that :grinning: 

https://www.npmjs.com/package/babel-plugin-transform-react-remove-prop-types

crptm: Yeah they do nothing in production anyway.

skubakdj:
@crptm & @wbobeirne, I think I addressed both of your comments with this latest round of commits. Let me know if I missed anything :+1:

common/components/Header/components/GasPriceDropdown.jsx
```diff
@@ -0,0 +1,82 @@
...
+export default class GasPriceDropdown extends Component {
+  state = { expanded: false };
+
+  static propTypes = {
+    gasPriceGwei: PropTypes.number,
+    changeGasPrice: PropTypes.func.isRequired
+  };
+
+  render() {
+    const gasPriceMinGwei = 1;
```
crptm: you can hoist it to module OR move to config/{something}

common/components/Header/index.jsx
```diff
@@ -47,6 +50,11 @@ export default class Header extends Component {
                 Open-Source & Client-Side Ether Wallet · v3.6.0
               </span>
 
+              <GasPriceDropdown
+                gasPriceGwei={this.props.gasPriceGwei}
+                changeGasPrice={this.props.changeGasPrice}
```
crptm: Would prefer value & onChange here

crptm:
🚢  it

PR #67, opened 2017-07-21
Update flow-bin version to 0.50.0

dternyak:


crptm:
Why we have so many changes in lockfile?

crptm:
cc @wbobeirne

dternyak:
Merging, can discuss why so many additions to lock-file async

PR #70, opened 2017-07-23
Contracts Tab Scaffolding

wbobeirne:
This is a first pass at the Contracts tab. It puts most of the UI in place, and a good chunk of the reducer behavior. However, without wallets fully implemented, and I was uncertain of the state of communication with the node, I left that undone for now.

This was my first crack at touching our actions / reducers / sagas, so please provide suggestions there, as I just tried to mesh with the existing styles.

### What this PR Does

* Adds a Contracts Tabs container
* Adds both the Interact and Deploy tabs on that page
* Provides the list of pre-added contracts (Supplied in `config/contracts/*.json`) on the Interact page
* Allows you to load the interface for interacting, given an ABI json and an address

### What this PR _Doesn't_ Do

* Allow you to deploy contracts on the Deploy tab (Just a stubbed out interface)
* Allow you to read constants / run contract functions on the Interact tab
* Allow you to add your own contract to the contract selector (I have no idea if this was in the original, but I didn't add it here)
* Validate any fields in the component (it does in the reducer.) After some discussion, I decided to punt on this until we had figured out some kind of common validated address / abi / gas price / byte code component(s). This affects the interact form, the deploy form, and the interact function running forms.


common/actions/config.js
```diff
@@ -1,41 +1,45 @@
 // @flow
+import {
+  NODE_CHANGE,
+  LANGUAGE_CHANGE,
+  GAS_PRICE_CHANGE
+} from './configConstants';
 
 export type ChangeNodeAction = {
-  type: 'CONFIG_NODE_CHANGE',
+  type: NODE_CHANGE,
```
crptm: Breaks typecheck in reducers.

common/actions/contracts.js
```diff
@@ -0,0 +1,62 @@
...
+};
+
+export function fetchNodeContracts(): FetchNodeContractsAction {
+  return { type: FETCH_NODE_CONTRACTS };
+}
+
+/***** Set Node Contracts *****/
+export type SetNodeContractsAction = {
+  type: SET_NODE_CONTRACTS,
+  contracts: Array
```
crptm: Flow wants generic argument here

common/actions/contracts.js
```diff
@@ -0,0 +1,62 @@
...
+  type: ACCESS_CONTRACT,
+  abiFunctions: Array
+};
+
+export function setInteractiveContract(
+  functions: string
+): SetInteractiveContractAction {
+  return {
+    type: SET_INTERACTIVE_CONTRACT,
+    functions
```
crptm: abiFunction in type

common/sagas/contracts.js
```diff
@@ -0,0 +1,61 @@
...
+import CONTRACTS from 'config/contracts';
+import { NODES } from 'config/data';
+import { getNode } from 'selectors/config';
+import { isValidETHAddress } from 'libs/validators';
+
+function* handleFetchNodeContracts() {
+  const nodeKey = yield select(getNode);
+  const node = NODES[nodeKey] || {};
+  const contracts = CONTRACTS[node.network] || [];
+  yield put(setNodeContracts(contracts));
```
crptm: Why we need to put this in state?

common/sagas/contracts.js
```diff
@@ -0,0 +1,61 @@
...
+  } catch (err) {
+    console.error('Error parsing contract ABI JSON', err);
+    yield put(showNotification('danger', translate('ERROR_26'), 5000));
+  }
+}
+
+export default function* contractsSaga(): Generator<Effect, void, any> {
+  yield takeEvery(ACCESS_CONTRACT, handleAccessContract);
+  yield takeEvery(
+    [FETCH_NODE_CONTRACTS, NODE_CHANGE],
```
crptm: If we keep contracts out of state we dont need this

common/containers/Tabs/Contracts/index.jsx
```diff
@@ -0,0 +1,126 @@
...
+  };
+
+  state = {
+    activeTab: 'interact'
+  };
+
+  constructor(props) {
+    super(props);
+
+    this.props.fetchNodeContracts();
```
crptm: and this

common/components/ui/Identicon.jsx
```diff
@@ -17,6 +17,7 @@ export default function Identicon(props: Props) {
     : '';
   return (
     <div
+      className="Identicon"
```
crptm: Cant see it being used anywhere

common/containers/Tabs/Contracts/index.jsx
```diff
@@ -0,0 +1,126 @@
...
+    selectedABIJson: PropTypes.string,
+    selectedABIFunctions: PropTypes.array,
+    // Actions
+    fetchNodeContracts: PropTypes.func.isRequired,
+    accessContract: PropTypes.func.isRequired,
+    deployContract: PropTypes.func.isRequired
+  };
+
+  state = {
+    activeTab: 'interact'
```
crptm: maybe we can move it to router

common/libs/validators.js
```diff
@@ -83,3 +84,12 @@ export function isPositiveIntegerOrZero(number: number): boolean {
   }
   return number >= 0 && parseInt(number) === number;
 }
+
+export function isValidABIJson(str) {
```
crptm: Cant see this being used

common/sagas/contracts.js
```diff
@@ -0,0 +1,61 @@
...
+import { NODE_CHANGE } from 'actions/configConstants';
+import CONTRACTS from 'config/contracts';
+import { NODES } from 'config/data';
+import { getNode } from 'selectors/config';
+import { isValidETHAddress } from 'libs/validators';
+
+function* handleFetchNodeContracts() {
+  const nodeKey = yield select(getNode);
+  const node = NODES[nodeKey] || {};
+  const contracts = CONTRACTS[node.network] || [];
```
crptm: Strictly prefer to have everything inside network object, in either case this should be behind selector

crptm:
In general: 
- LETS DITCH CONSTANTS
- Explicitly type arrays and functions (no Function and Array)
- Setup flow checking in w/e you use

common/actions/config.js
wbobeirne: Ah, I see what it is. It's because all of these actions had the same keys, and then each action was split into functions in the reducers, so there was a conflict. One action expected value to be a `number`, and the other a `string`.

The other issue is that by splitting each action into its own function, they have inconsistent inputs. The _only_ reason this worked before was they were all an identical layout, `{ type: string, value: string }`. If we added a different action signature, it would have broken the whole thing.

To fix this, I'm going to move everything back into the reducer function, and stop using a generic `value` key, to prevent these kinds of convenient collisions.

wbobeirne: Flow's docs on redux, that I referenced: https://flow.org/en/docs/frameworks/redux/

crptm: Problem here is type of `type` is not string constant but string (since you not using inline value). This approach is definitely working with differently shaped actions, i.e. https://github.com/MyEtherWallet/MyEtherWallet/blob/develop/common/actions/ens.js

common/sagas/contracts.js
wbobeirne: By state, do you mean the reducer store instead of component state? If so, I wanted to keep this logic separated out. I figured this would make the pre-select list more flexible for if we add local storage / some way of having nodes provide their own suggested contracts.

If this seems like too much overhead, I could move it into a mapStateToProps function and do all this there.

common/components/ui/Identicon.jsx
wbobeirne: You're right, leftovers from some early styling stuff I was doing. I'll cut it out.

common/containers/Tabs/Contracts/index.jsx
wbobeirne: Yeah, this makes a lot of sense, to have each have its own route (But still use the same component.) I'll look into this.

common/sagas/contracts.js
crptm: It is code for sake of code (like constants :D), we could cut half^W quarter of PR if we just use that json directly.

common/libs/validators.js
wbobeirne: I had originally written this when I was making generic inputs for all sorts of things. I had since taken that out, but this was left behind. I'll remove it.

common/sagas/contracts.js
wbobeirne: So you'd rather this all be put in the `NODES` constant? That would be fine by me, I just didn't want to tack on something that's only used by one page to that constant.

crptm: we have `network`s, since we could have multiple nodes per network https://github.com/MyEtherWallet/MyEtherWallet/blob/develop/common/config/data.js#L136

you can get it via https://github.com/MyEtherWallet/MyEtherWallet/blob/develop/common/selectors/config.js#L11

common/sagas/contracts.js
wbobeirne: Roger that, makes sense, I'll put it in there and add a selector. Thanks for the advice!

wbobeirne:
@crptm I think I hit all of your feedback, and the PR is much better for it. Thanks for the thorough review!

common/reducers/config.js
```diff
@@ -20,38 +19,28 @@ export const initialState: State = {
   gasPriceGwei: 21
 };
 
-function changeLanguage(state: State, action: ChangeLanguageAction): State {
```
crptm: But why? Cant you just let it sit alone? :) Switch is not scalable, you cant extract it to another file for example, helper variables are awkward because of scope rules, etc

wbobeirne: I'll undo this back to what it was on master.

PR #74, opened 2017-07-25
Contracts Tab Scaffolding - Flow Fixes

wbobeirne:
I busted this out into a separate PR as I wanted to keep all flow-related discussion of how we deal with flow / redux / sagas in one thread. I cleaned up the mess I had in #70, but I wanted to make sure I did this the "right" way.

I think the only majorly contentious thing in here is changing the config reducer to `action: Object` where it was previously `action: ConfigAction`. However, reducers receive all actions, not just the ones from that reducer, so the previous way of defining all actions would not have worked in a reducer that listens to actions meant for other reducers anyway.


wbobeirne:
I forgot to run tests, which hadn't been updated to reflect the action changes I made. I'll circle back to that soon.

.flowconfig
```diff
@@ -7,3 +7,4 @@
 [options]
 module.system.node.resolve_dirname=node_modules
 module.system.node.resolve_dirname=common
+module.name_mapper='.*\(.sass|scss|less|css\)' -> 'empty/object'
```
crptm: fixed in my branch w/o dep, i will resolve when merging

wbobeirne:
From slack discussion: I'll be removing action constants as a part of this. So changing config back to the way it was, and changing contracts to use strings instead of constants.

crptm:
Dont really want to repeat other PR comments, but can we not store contracts in redux for now please

wbobeirne:
Closing this out in favor of the main contracts PR. I think I'm on top of the flow stuff now.

PR #95, opened 2017-07-26
Standardize Redux Actions / Reducers

wbobeirne:
After much bickering, I think we settled on some standards for redux actions and reducers. I went ahead and unified them all, and added a style guide for them to the readme. You can see that guide here: https://github.com/MyEtherWallet/MyEtherWallet/tree/redux-unify#redux-and-actions

Most of the changes are just structure, naming, file organization etc. but I did do some pretty heavy overhauling of the generateWallet reducer to move a lot out of reducer state, and into component state. I also renamed things to make a lot more sense. I apologize for my first pass at it, I think it's much better now.


README.md
```diff
@@ -51,6 +51,90 @@ docker-compose up
...
+* Initial state should match state flow typing, define every key
+* Reducer function should handle all cases for actions. If state does not change
+as a result of an action (Because it merely kicks off side-effects in saga) then
+define the case above default, and have it fall through.
+
+```js
+import type { NamespaceAction } from "actions/namespace";
+
+export type State = { /* Flowtype definition for state object */ };
+export const INITIAL_STATE = { /* Initial state shape */ };
```
crptm: export const INITIAL_STATE: State

README.md
```diff
@@ -51,6 +51,90 @@ docker-compose up
...
+#### Reducer
+
+* State should be explicitly defined and exported
+* Initial state should match state flow typing, define every key
+* Reducer function should handle all cases for actions. If state does not change
+as a result of an action (Because it merely kicks off side-effects in saga) then
+define the case above default, and have it fall through.
+
+```js
+import type { NamespaceAction } from "actions/namespace";
```
crptm: // @flow

so copy-past is codbase friendly :)

Issue #98, opened 2017-07-27
Off by magnitude in sidebar


wbobeirne:
v3 sidebar:
<img width="194" alt="live-site" src="https://user-images.githubusercontent.com/649992/28685174-f0ea5454-72d4-11e7-9f84-aea0f1df3e26.png">

v4 sidebar:
<img width="178" alt="dev-site" src="https://user-images.githubusercontent.com/649992/28685189-f85d1bb8-72d4-11e7-89d4-0fd632d51ec3.png">

From this address: 0xbF1B0b4dAebb78b65e842716936d68F178d5e8D1

wbobeirne:
Fixed with #99.

PR #99, opened 2017-07-27
Fix number formatter

wbobeirne:
Number formatter was failing in the following cases:

Input: 0.0127491, default digits desired
Expected output: 0.013
Actual output: 0.13

Input: 100.48391, 0 digits desired
Expected output: 100
Actual output: Error!

Fixed those cases, commented up the function a bit to make it easier to understand, and added tests.


wbobeirne:
Fixes #98.

PR #100, opened 2017-07-28
WIP: Generating transaction on Send tab (Pt 1)

wbobeirne:
I was going to work on the transaction modal, but realized we didn't even have the ability to generate one. I took a stab at it, going off of https://github.com/kvhnuke/etherwallet/blob/mercury/app/scripts/uiFuncs.js#L96. This isn't meant to be the whole thing, just the basics for now.

### What it Does

* Given a wallet a raw transaction and a wallet, generates transaction JSON
* Also provides the signed transaction using the `signRawTransaction` wallet function
* Does some error handling of node problems, insufficient funds, or bad input (Bad address, TODO bad gas limit / prices)

### What it Doesn't Do (Yet)

* ERC20 tokens
* Hardware wallet signing (Unlock doesn't even work for them though.)
* Offline support
* Broadcasting the transaction
* **Unit tests for generating a transaction**. This one should definitely be looked at before release.

## Screenshot

screen shot 2017-07-28 at 3 57 54 am


common/utils/formatters.js
```diff
@@ -28,3 +28,19 @@ export function formatNumber(number: Big, digits: number = 3): string {
 
   return parts.join('.');
 }
+
+export function padLeftEven(hex: string): string {
```
crptm: available in ethereumjs-util

common/utils/formatters.js
```diff
@@ -28,3 +28,19 @@ export function formatNumber(number: Big, digits: number = 3): string {
...
+  hex = hex.length % 2 != 0 ? '0' + hex : hex;
+  return hex;
+}
+
+export function sanitizeHex(hex: string): string {
```
crptm: available in ethereumjs-util

crptm:
This is unlikely to be merged, i am working on a same thing for quite some time already

wbobeirne:
I'm still working on some cleanup for this, so don't drop any comments for now. I'll ping when I'm ready to have this looked at again.

common/components/BalanceSidebar/BalanceSidebar.jsx
```diff
@@ -28,12 +28,20 @@ type Props = {
 export class BalanceSidebar extends React.Component {
   props: Props;
   state = {
-    showLongBalance: false
+    showLongBalance: false,
+    address: ''
   };
 
+  componentDidMount() {
+    this.props.wallet.getAddress().then(address => {
```
tayvano: Anytime we display an address it needs to be checksumm'd address

Anytime we interact with an address (e.g. hashes, blockies) it needs to be all lowercase address.

https://github.com/ethereumjs/ethereumjs-util/blob/cfeac0fc49c00111a77124bbe76011e4b1129cf9/index.js#L441-L455

When a user types in an address, it should display error if not checksummed and different error if completely invalid.

> This address is not checksummed. If you hand-typed the address, please go back and copy & paste it. Your funds are not recoverable if you send to an invalid or wrong address.

> This address is not valid. The address should look similar to: 0x7cB57B5A97eAbe94205C07890BE4c1aD31E486A8

common/components/BalanceSidebar/BalanceSidebar.jsx
```diff
@@ -94,10 +99,7 @@ export class BalanceSidebar extends React.Component {
               {!!tokenExplorer &&
                 <li>
                   <a
-                    href={tokenExplorer.address.replace(
-                      '[[address]]',
-                      wallet.getAddress()
-                    )}
+                    href={tokenExplorer.address.replace('[[address]]', address)}
```
tayvano: When considering future networks / chains, we should probably have the ability to not display a token explorer if there isn't one. Not sure best place for it (probably not here :P)

common/containers/Tabs/SendTransaction/index.jsx
```diff
@@ -381,16 +404,42 @@ export class SendTransaction extends React.Component {
...
+          to: this.state.to,
+          from: address,
+          value: this.state.value,
+          gasLimit: this.state.gasLimit,
+          gasPrice: this.props.gasPrice,
+          data: this.state.data,
+          chainId: this.props.network.chainId
+        },
+        wallet
+      );
```
tayvano: Checks

**to: valid address**
https://github.com/ethereumjs/ethereumjs-util/blob/cfeac0fc49c00111a77124bbe76011e4b1129cf9/index.js#L432-L434

**from: valid address**

gasLimit: greater than 21000, less than 5M (the network has an ever-shifting limit). integer.

gasPrice: this should be in wei. warn user if over 250 GWEI (250000000000 wei). Do not send if over 1000(1000000000000) GWEI (tell them to contact us to explain why they would want to send that expensively?
) 

https://www.myetherwallet.com/helpers.html is a good place to convert

common/components/BalanceSidebar/BalanceSidebar.jsx
wbobeirne: Just to be clear, the wallet's `getAddress` function's behavior is to return the checksumm'd address, so there's no need to do it here. But that's good to have that rule laid out for where we do what with it.

As for the valid / invalid address, that's being handled in [`generateTransaction`](https://github.com/MyEtherWallet/MyEtherWallet/pull/100/files#diff-c762f03dadee1e4d3252d9aaa56e84a6R77)

common/containers/Tabs/SendTransaction/index.jsx
wbobeirne: Already checking valid to address. I'll add those gas checks to the `generateTransaction` errors though, good call.

I'm not sure how one would get here with an invalid from address though, since this uses the wallet class to do everything, so I'm not concerned about that.

wbobeirne:
OK, feel free to comment away. Going to try to get this merge ready ASAP.

PR #101, opened 2017-07-30
Swap Part 4

dternyak:
### Re-implements:
* min/max validators on initial currency swap selection
* polling of order status
* timer that persists across refreshes via localStorage (computed based on `createdTime` and `validFor` amount)
* swap persists across refreshes once order is created.
* various type refactors

### New additions:
* *SimpleButton* (can be PRd separately on request)
* clear loading state after order create (via SimpleButton and font-awesome)
* buffers for non-BTC swaps (bity does not actually accept 0.01 BTC worth of ETH as they claim they do in their JSON response, so a magic number of 10% is added to the minimum).

### Todo in another PR:

- [ ] Support Request Button
- [ ] QR Code for BTC send
- [ ] Replace rate with destination address in Info Header
- [ ] address important todo about checking status in order status poll when order is not expired (although orders almost never succeed completely before they expire).


common/components/ui/SimpleButton.jsx
```diff
@@ -0,0 +1,60 @@
...
+  type?: ButtonType
+};
+
+export default class SimpleButton extends Component {
+  props: Props;
+
+  computedClass = () => {
+    return `btn btn-${this.props.size || DEFAULT_BUTTON_TYPE} btn-${this.props
+      .type || DEFAULT_BUTTON_SIZE}`;
+  };
```
wbobeirne: I'd strongly suggest we add [`classnames`](https://www.npmjs.com/package/classnames) to the module stack. It makes a lot of this stuff more manageable, and it's really lightweight.

common/components/ui/SimpleButton.jsx
```diff
@@ -0,0 +1,60 @@
+// @flow
+import React, { Component } from 'react';
+
+const DEFAULT_BUTTON_TYPE = 'primary';
+const DEFAULT_BUTTON_SIZE = 'lg';
+
+const Spinner = () => {
+  return <i className="fa fa-spinner fa-spin fa-fw" />;
+};
```
wbobeirne: Would you mind if I made a PR on top of this to do this in CSS? We could kill the added FontAwesome dependency (Unless we've been itching for an icon font, which might not be bad to have.)

common/containers/Tabs/Swap/components/CurrencySwap.jsx
```diff
@@ -0,0 +1,224 @@
...
+import { combineAndUpper } from 'utils/formatters';
+import SimpleDropDown from 'components/ui/SimpleDropdown';
+import SimpleButton from 'components/ui/SimpleButton';
+import type {
+  OriginKindSwapAction,
+  DestinationKindSwapAction,
+  OriginAmountSwapAction,
+  DestinationAmountSwapAction,
+  ChangeStepSwapAction
+} from 'actions/swap';
```
wbobeirne: From above, it looks like the types moved back into actions/swapTypes.js

common/containers/Tabs/Swap/components/SwapProgress.jsx
```diff
@@ -0,0 +1,141 @@
...
+
+  computedClass(step: number) {
+    const {
+      orderStatus,
+      destinationKind,
+      showNotification,
+      destinationAddress
+    } = this.props;
+
+    let cssClass = 'progress-item';
```
wbobeirne: Just another vote for `classnames`

common/containers/Tabs/Swap/components/SwapProgress.jsx
```diff
@@ -0,0 +1,141 @@
...
+          } else {
+            link = link = bityConfig.btcExplorer.replace(
+              '[[txHash]]',
+              destinationAddress
+            );
+            linkElement = `<a href="${link}" target='_blank' rel='noopener'> View your transaction </a>`;
+          }
+
+          this.setState({ hasShownViewTx: true }, () => {
+            showNotification('success', linkElement);
```
wbobeirne: I'm not sure how showNotification works, but is it possible to use JSX instead of strings? Would be nice to nix any uses of setInnerHTMLDangerously since that's ideally only used for server-side HTML.

wbobeirne: ^ This is possibly off topic for your PR, so feel free to kick that can

common/containers/Tabs/Swap/components/PartThree.jsx
```diff
@@ -0,0 +1,81 @@
...
+  restartSwap: () => RestartSwapAction,
+  stopLoadBityRatesSwap: () => StopLoadBityRatesSwapAction,
+  startOrderTimerSwap: Function,
+  startPollBityOrderStatus: Function,
+  stopOrderTimerSwap: Function,
+  stopPollBityOrderStatus: Function,
+  showNotification: Function
+};
+
+export default class PartThree extends Component {
```
wbobeirne: Is there a more descriptive name for what this does? It might be awkward if we add another step before it, and it's a little tough to tell what it covers without cracking open the code yourself.

webpack_config/webpack.base.js
```diff
@@ -41,10 +41,6 @@ module.exports = {
         exclude: [/node_modules\/(?!ethereum-blockies|idna-uts46)/]
       },
       {
-        test: /\.(ico|webp|eot|otf|ttf|woff|woff2)(\?.*)?$/,
-        loader: 'file-loader?limit=100000'
-      },
-      {
```
wbobeirne: Any reason this was taken out? Do we have no custom fonts?

common/sagas/swap/orders.js
```diff
@@ -0,0 +1,166 @@
...
+            if (!hasShownNotification) {
+              hasShownNotification = true;
+              yield put(showNotification('danger', BITY_TIMEOUT_MESSAGE, 0)); // 0 is forever
+            }
+            break;
+          case 'RCVE':
+            yield put(orderTimeSwap(0));
+            if (!hasShownNotification) {
+              hasShownNotification = true;
+              yield put(showNotification('warning', BITY_TIMEOUT_MESSAGE, 0)); // 0 is forever
```
wbobeirne: Given the explanation added to all 3, maybe we should change the showNotification api to take in `infinity` to be a bit more "readable"

common/sagas/swap/orders.js
```diff
@@ -0,0 +1,166 @@
...
+    const order = yield call(
+      postOrder,
+      payload.amount,
+      payload.destAddress,
+      payload.mode,
+      payload.pair
+    );
+    if (order.error) {
+      // TODO - handle better / like existing site?
+      yield put(showNotification('danger', `Bity Error: ${order.msg}`, 10000));
```
wbobeirne: Might as well use TEN_SECONDS, since ya made it

common/containers/Tabs/Swap/index.jsx
```diff
@@ -43,20 +43,23 @@ type ReduxActionProps = {
...
+  loadBityRatesRequestedSwap: () => LoadBityRatesRequestedSwapAction,
   destinationAddressSwap: (value: ?string) => DestinationAddressSwapAction,
   restartSwap: () => RestartSwapAction,
   stopLoadBityRatesSwap: () => StopLoadBityRatesSwapAction,
   // PART 3 (IGNORE FOR NOW)
-  referenceNumberSwap: typeof swapActions.referenceNumberSwap
+  orderCreateRequestedSwap: any,
+  startPollBityOrderStatus: any,
+  stopOrderTimerSwap: any,
+  stopPollBityOrderStatus: any
```
wbobeirne: Looks like these puppies could use some types

common/components/ui/SimpleButton.jsx
dternyak: That would be great -- thanks.

common/containers/Tabs/Swap/components/CurrencySwap.jsx
dternyak: Huh, wonder why flow never complained.

common/containers/Tabs/Swap/components/PartThree.jsx
dternyak: Not really :(. I tried to think of one. The existing site calls it "Stage-3". I could change it to that?

webpack_config/webpack.base.js
dternyak: It was a duplicate

common/sagas/swap/orders.js
dternyak: heh yeah.

common/sagas/swap/orders.js
dternyak: true

common/containers/Tabs/Swap/index.jsx
dternyak: correct! missed that.

common/actions/swap.js
```diff
@@ -1,10 +1,22 @@
...
+  OrderCreateRequestedSwapAction,
+  OrderStatusSucceededSwapAction,
+  ChangeStepSwapAction,
+  Pairs,
+  RestartSwapAction,
+  LoadBityRatesRequestedSwapAction,
+  StopLoadBityRatesSwapAction,
+  BityOrderResponse,
+  BityOrderPostResponse
+} from './swapTypes';
```
dternyak: It was changed before you made the big redux/flow refactor, but I also like it a bit better if they're separate when there's like 15+ action creators just due to size.

common/api/bity.js
```diff
@@ -1,36 +1,38 @@
...
 /**
  * Gives you multiple rates from Bitys API without making multiple API calls
- * @param arrayOfOriginAndDestinationDicts - [{origin: 'BTC', destination: 'ETH'}, {origin: 'BTC', destination: 'REP}]
+ * @param transactionPairs - [{origin: 'BTC', destination: 'ETH'}, {origin: 'BTC', destination: 'REP}]
  */
-function getMultipleRates(arrayOfOriginAndDestinationDicts) {
+export type TransactionPair = {
+  originKind: string,
+  destinationKind: string
+};
```
dternyak: Will fix. Dunno how Flow isn't picking this stuff up, but then again the function is going unused for now.

common/containers/Tabs/Swap/components/PartThree.jsx
wbobeirne: Not much better. Oh well, as long as it's no more ambiguous than it was before!

common/containers/Tabs/Swap/components/SwapProgress.jsx
dternyak: @crptm ^

Issue #85, opened 2017-07-25
Confirmation Modal


dternyak:
## Send TX Confirmation Modal should display:

- [ ] Warning: You are about to....
- [ ] [ Address ] + [ AddressIcon from ] -> [ Value ] [ CoinType ] -> [ Address ] + [ AddressIcon ] to
- [ ] The [ ETH ] node you are sending through is provided by [ MyEtherWallet ].
- [ ] Address From:
- [ ] Address To:
- [ ] Value:
- [ ] CoinType:
- [ ] Gas Limit:
- [ ] Gas Price:
- [ ] Nonce:
- [ ] Data:
- [ ] Chain ID:
- [ ] [Unsigned TX Textarea]
- [ ] [Signed TX Textarea]
- [ ] Cancel Button
- [ ] Yes, I'm sure Button

### Future:

Current issues with current conf modals

1. People dont read them (had someone the other day literally tell me we need a confirmation for the confirmation....)

3. Too easy to miss the coin you are sending (e.g. ETH vs SNT)

5. Too easy to miss the chain you are on (ETC vs ETH), although the border I added has helped reduce those

7. Not the information the user wants / needs in an easily parsable / skimmible thing (table format will help)

#### Misc notes from Taylor:
"Separately, I hate how modals sometimes close all by themselves sometimes if you click twice or click outside. I do it on Twitter all the time and I think it's worse with bootstrap v4. I lose tweets ALL THE TIME when I'm trying to highlight something.

It shouldn't close unless you hit the X or the CANCEL button or hit the esc button (optional)

--

Other than that, I really like the success / error state inside the modal itself, but we'll have to decide how to handle alerts and things. While some successes and errors come from front-end validation, some are message returned directly from the node (e.g. a TX with that hash has already been imported). It would be safer to handle all alerts via the alerts handler, avoiding a potential situation where you sometimes show alert, sometimes put alert in modal, and if you aren't careful, accidentally don't show the alert at all. Making things fail-proof for our future selves is very important so we can build quickly with less worry. 😉"

wbobeirne:
This was taken care of with #108.

PR #108, opened 2017-08-07
Transaction confirmation modal

wbobeirne:
## What This Is

This addresses #85, and adds a confirmation modal to the send tab. There's still a few missing pieces, and this PR comes with a lot of `// TODO: ...`, but it's at least a start. You can also see a video of it in action [over here](https://gfycat.com/RemoteSelfassuredDove).

## What I Did

* Created and styled a modal for confirming transactions
* Made the modal use the actual transaction object, _not_ what's in the form fields
* Reworked some of the copy and added some additional data (Translation issues?)
* Added a 5 second timer that hopefully gets people to stop and read
* Refactored some unit conversion code, wrote unit tests for it
* Fixed up / added some flow typing

## What I Didn't Do

* Implement _all_ of the bullets in #85. Much of that doesn't exist in the current modal, so I wasn't sure how much we were adding.
* Handle any non-Ethereum transactions (Showing altcoin instead of ETH.) I may need some assistance on how this is handled elsewhere right now.
* Nothing currently generates transactions (See #100) so that's not a part of this. There's no way to actually get this modal to open naturally
* AFAIK there's no infrastructure in the codebase right now for broadcasting a transaction, so the confirm button doesn't work

## Comparison

Old:
old

New: 
new


common/components/ui/Modal.jsx
```diff
@@ -18,6 +17,8 @@ type Props = {
       | 'warning'
       | 'danger'
       | 'link',
+    disabled?: boolean,
+    // $FlowFixMe - Why the fuck doesn't this like onClick?
```
wbobeirne: If anyone could help me out with this, it doesn't like the onClicks I provided in ConfirmationModal.jsx and the error is really opaque, just saying my buttons don't match the button object.

common/containers/Tabs/SendTransaction/components/ConfirmationModal.jsx
```diff
@@ -0,0 +1,153 @@
...
+import translate from 'translations';
+import Big from 'bignumber.js';
+import BaseWallet from 'libs/wallet/base';
+import { toUnit } from 'libs/units';
+import type { NodeConfig } from 'config/data';
+
+import Modal from 'components/ui/Modal';
+import Identicon from 'components/ui/Identicon';
+
+// TODO: Replace me with a shared type defined somewhere
```
wbobeirne: Likely as a part of #100.

common/containers/Tabs/SendTransaction/components/ConfirmationModal.jsx
```diff
@@ -0,0 +1,153 @@
...
+
+export default class ConfirmationModal extends React.Component {
+  props: Props;
+  state: State;
+
+  constructor(props: Props) {
+    super(props);
+    const { value, gasPrice, nonce } = props.rawTransaction;
+
+    // TODO: Fix with wallet address promise
```
wbobeirne: As a part of #105.

common/containers/Tabs/SendTransaction/components/ConfirmationModal.jsx
```diff
@@ -0,0 +1,153 @@
...
+            </li>
+            <li className="ConfModal-details-detail">
+              You are transmitting on the <strong>{node.network}</strong>{' '}
+              network provided by <strong>{node.service}</strong>
+            </li>
+            <li className="ConfModal-details-detail">
+              {rawTransaction.data
+                ? <span>
+                    You are sending the following data:{' '}
+                    <code>{rawTransaction.data}</code>
```
wbobeirne: How big does data get? Should I find a better display for this?

common/containers/Tabs/SendTransaction/components/ConfirmationModal.scss
```diff
@@ -0,0 +1,47 @@
...
+    &-amount {
+      flex-grow: 1;
+      text-align: center;
+      line-height: $summary-height / 2;
+
+      &-arrow {
+        display: inline-block;
+        height: $summary-height / 3;
+        width: 100%;
+        background-image: url('../../../../assets/images/icon-dot-arrow.svg');
```
wbobeirne: Remind me to fix this to not use relative paths.

tayvano:
The only thing I would request is adding the from address. I figured people would recognize their own sending address icon. I was wrong. Here is what the current site looks like now—I'm sure there is a better way. :P 
screen shot 2017-08-07 at 6 05 20 pm

wbobeirne:
Did this just get added? Wasn't in my screenshot of the before, and I took it like minutes before submitting the PR!

But yeah, I'd be happy to add that in.

common/containers/Tabs/SendTransaction/components/ConfirmationModal.jsx
```diff
@@ -0,0 +1,153 @@
...
+          <ul className="ConfModal-details">
+            <li className="ConfModal-details-detail">
+              You are sending to <code>{rawTransaction.to}</code>
+            </li>
+            <li className="ConfModal-details-detail">
+              You are sending <strong>{value} ETH</strong> with a gas price of{' '}
+              <strong>{gasPrice} gwei</strong>
+            </li>
+            <li className="ConfModal-details-detail">
+              You are transmitting on the <strong>{node.network}</strong>{' '}
```
wbobeirne: Not sure where I came up with this copy, original is `interacting with the` which sounds way less nerdy. I'll change it back.

tayvano: @wbobeirne https://appear.in/myetherwallet

wbobeirne:
Added in `from` to the bullet list. Integrated the generate transaction code from #100, so now this actually pops up. I'm going to take `WIP` off the title, I think this is good to go.

crptm:
This should take signed tx

wbobeirne:
I definitely agree in the long term, what would you think about merging this in as-is and making a new issue for that? It's already somewhat better than v3 which just read form inputs, and doesn't change the user experience.

Not downplaying the value of reading from signed though, I really want to make that happen. It would probably just take me a little while to figure out how to convert from that to the raw data, and I want to get in #118 because a lot of work towards handling that is done in there.

crptm:
We should have started with isolated modal that takes rawtx string in the first place. Ethereumjs-tx can parse signed txes.

crptm:
We can merge this as is bc send token PR is on top of this branch.

wbobeirne:
#118 has been merged in, and the modal takes signed transactions now (Along with all the other token handling.) Are there any other outstanding issues?

crptm:
If we only support EIP-155 chains is it possible to check that we broadcast tx on correct network? This could wait for other PR, just curious if this is possible. 

PR should be merged, no reason to make it any bigger.

wbobeirne:
I don't know enough about the alt chains to know that, but presumably we could make that kind of confirmation optional iff they provide chainid right?

Can somebody throw me a 👍 so that I can merge this?

tayvano:
Can you confirm that it shows the sending from address somewhere (underneath the addressicon? This has become an issue of late. Hold on EIP-155 stuff until I get everything sorted with v3 issues and we can circle back. I dont think it applies here regardless.

PR #110, opened 2017-08-08
Fix flow errors

skubakdj:
Addresses the flow type errors introduced by the previous wallet decrypt PR. Essentially moves component-based async calls to `componentDidMount` and sets `address` property in state. 

Stays clear of `SendTransaction/index.jsx` to avoid conflicts with @wbobeirne. 

As always, let me know if anything could be improved! :smile:

### DIFF
--- a/common/components/PaperWallet/index.jsx
+++ b/common/components/PaperWallet/index.jsx
@@ -101,15 +101,9 @@ export default class PaperWallet extends React.Component {
 
   componentDidMount() {
     if (!this.props.wallet) return;
-    this.props.wallet
-      .getAddress()
-      .then(addr => {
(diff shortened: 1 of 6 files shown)


common/components/PaperWallet/index.jsx
```diff
@@ -97,8 +97,16 @@ type Props = {
 
 export default class PaperWallet extends React.Component {
   props: Props;
+  state = { address: '' };
+
+  componentDidMount() {
+    if (!this.props.wallet) return;
+    this.props.wallet.getAddress().then(addr => {
+      this.setState({ address: addr });
+    });
```
skubakdj: Decided against catching anything here since HW wallets can't be used.

Issue #103, opened 2017-08-01
Wallet Class Structure


skubakdj:
Hey everyone!

Let's use this issue to determine our class structure for the different wallet types we support. I've piggybacked on what's currently in our codebase and divised a simple structure that I hope will be a good starting point for discussion: 

```
BaseWallet
|   getAddress()
|   getNakedAddress()   
|
|__ PrivkeyWallet( privkey )
|   |   privKey
|   |   pubKey
|   |   address
|   |   generate()
|   |   getPrivateKey()
|   |   getAddress()
|   |   toKeystoreFile()
|   |   signRawTransaction()
|   |   signMessage()
|   |        
|   |__ EncryptedPrivkeyWallet( key, pass )
|   |       super( decryptPrivateKey( key, pass ) )
|   |
|   |__ MnemonicWallet( phrase, pass )
|   |       super( decryptMnemonicPhraseToPrivkey( phrase, pass ) )
|   |
|   |__ KeystoreWallet( keystore, pass )
|   |       super( decryptKeystoreToPrivkey( keystore, pass ) )
|   |
|   |__ PresaleWallet( presale, pass )
|           super( decryptPresaleToPrivkey( presale, pass ) ) 
|
|__ LedgerWallet
|       address
|       signRawTransaction()
|       signMessage()
|
|__ TrezorWallet
        address
        signRawTransaction()
        signMessage()
```

To summarize, all wallets inherit from `BaseWallet` and any wallet dependent on deriving a private key (`EncryptedPrivkeyWallet`, `KeystoreWallet`, `PresaleWallet`, `MnemonicWallet`) is essentially a wrapper for `PrivkeyWallet`. All wallets should come equipped with methods to sign transactions and messages (`signRawTransaction()` & `signMessage()`, respectively). 

Additional work will probably have to go into `LedgerWallet` and `TrezorWallet`, specifically dealing with HD key paths & verifying addresses exist. 

Please let me know what y'all think :smile: 

@dternyak  @crptm @wbobeirne

wbobeirne:
Great diagram!

Seems to me that all wallets will have a `signRawTransaction()` as well, but that's not shared in any way that I see here. But this is a great idea to standardize, to get rid of [shit like this](https://github.com/kvhnuke/etherwallet/blob/mercury/app/scripts/uiFuncs.js#L114).

One thing I see a lot of in v3 is separate conditionals for each hardware wallet to do their unlocking behavior. I think it would be great to either:
[...]

PR #116, opened 2017-08-14
Keystore & Private Key Wallet Decrypts

skubakdj:
### What this PR does

Continues to implement #103. Completes functionality for the Keystore/Private Key selections on wallet decrypt. 

MEWv3 can handle a few more keystore options than I previously thought. In total, this PR adds support for:

 - `Encrypted private keys`
 - `Presale wallets`
 - `MEWv1 encrypted wallets`
 - `MEWv1 unencrypted wallets`
 - `MEWv2 unencrypted wallets`
 - `UTC wallets` (found in MEWv2 & MEWv3)

Using previous releases, I've been able to manually test all of the above except `Presale wallets`. An unused presale file (or at least the scheme to generate one) would be required to verify this functionality.

I left references to the v3 code some of the functions have been adapted from. Can be removed if desired :smile:


common/components/WalletDecrypt/Keystore.jsx
```diff
@@ -1,63 +1,115 @@
...
+  onPasswordChange = (e: SyntheticInputEvent) => {
+    const valid = this.props.value.file.length && e.target.value.length;
+    this.props.onChange({
+      ...this.props.value,
+      password: e.target.value,
+      valid
+    });
+  };
+
+  handleFileSelection = (e: SyntheticInputEvent) => {
```
kangarang: I believe so!

https://github.com/facebook/flow/issues/218#issuecomment-259780131

https://github.com/facebook/flow/blob/04cc411b6ad34c94b793a6ebf47b1f0a2b8b1957/lib/react.js#L344

skubakdj: Awesome! Thanks for the links :+1:

common/components/WalletDecrypt/Keystore.jsx
```diff
@@ -1,63 +1,115 @@
...
     );
   }
+
+  onKeyDown = (e: SyntheticKeyboardEvent) => {
+    if (e.keyCode === 13) {
+      e.preventDefault();
+      e.stopPropagation();
+      this.props.onUnlock();
+    }
+  };
```
wbobeirne: Much simpler than this is to listen to and preventDefault a `<form onSubmit>` event, surrounding the input. It's better for accessibility, and makes it easier to have a submit button in the future (One event binding instead of onClick and onKeyDown.)

I notice this around a fair amount in the app though, so feel free to stick with it for consistency now. If I ever feel strongly enough about it, I'd just do a PR to clean it up everywhere.

common/components/WalletDecrypt/PrivateKey.jsx
```diff
@@ -80,7 +84,9 @@ export default class PrivateKeyDecrypt extends Component {
 
   onPasswordChange = (e: SyntheticInputEvent) => {
     const fixedPkey = fixPkey(this.props.value.key);
-    const isValid = isValidPrivKey(fixedPkey.length);
+    const isValid =
+      isValidPrivKey(fixedPkey.length) ||
+      isValidEncryptedPrivKey(fixedPkey.length);
```
wbobeirne: This fix, valid, length check series of calls should be a function. It happens 3 times in here.

common/components/WalletDecrypt/PrivateKey.jsx
```diff
@@ -26,7 +26,9 @@ export default class PrivateKeyDecrypt extends Component {
   render() {
     const { key, password } = this.props.value;
     const fixedPkey = fixPkey(key);
-    const isValid = isValidPrivKey(fixedPkey.length);
+    const isValid =
+      isValidPrivKey(fixedPkey.length) ||
+      isValidEncryptedPrivKey(fixedPkey.length);
```
wbobeirne: I see that `isValidPrivKey` was implemented like this before you, but it's kinda weird that we send the length, but the function seems to indicate it would check the whole private key. And maybe in the future we add some additional checks to it (Like to reject non-alphanumeric characters, or something.)

I think we should change these two to take in the key and do `.length` themselves.

common/libs/wallet/privkey.js
```diff
@@ -15,6 +15,9 @@ export default class PrivKeyWallet extends BaseWallet {
   pubKey: Buffer;
   address: Buffer;
   constructor(privkey: Buffer) {
+    if (privkey.length !== 32) {
```
wbobeirne: `isValidPrivKey` instead

common/components/WalletDecrypt/Keystore.jsx
```diff
@@ -1,63 +1,115 @@
...
+    this.props.onChange({
+      ...this.props.value,
+      password: e.target.value,
+      valid
+    });
+  };
+
+  handleFileSelection = (e: SyntheticInputEvent) => {
+    const fileReader = new FileReader();
+    const inputFile = event.target.files[0];
```
wbobeirne: The argument is `e`, but you use `event` here. I'm pretty amazed neither the linter, nor flow didn't catch this.

common/libs/decrypt.js
```diff
@@ -0,0 +1,68 @@
...
+  let tmp = Buffer.concat(ret);
+  return {
+    key: tmp.slice(0, keysize),
+    iv: tmp.slice(keysize, keysize + ivsize)
+  };
+}
+
+export function decipherBuffer(decipher: Object, data: Buffer): Buffer {
+  return Buffer.concat([decipher.update(data), decipher.final()]);
+}
```
wbobeirne: Unit tests for the stuff in here would be 👌 , even though I know they're all coming from v3 anyway.

common/components/WalletDecrypt/Keystore.jsx
wbobeirne: Ah crap, `event` is a global variable on window: https://developer.mozilla.org/en-US/docs/Web/API/Window/event

This can be resolved with http://eslint.org/docs/rules/no-restricted-globals, I'll make a PR for that now.

wbobeirne: #117

skubakdj:
Awesome feedback @wbobeirne!

I have indeed tested the unlocks with one of each wallet type (except `Presale wallets`), but then managed to introduce that `event` bug with my final commit to add a flow type. I'll address your comments and test the wallets again for good measure :+1:

common/components/WalletDecrypt/Keystore.jsx
```diff
@@ -1,63 +1,115 @@
...
+    fileReader.onload = () => {
+      const keystore = fileReader.result;
+      let isPassRequired = false;
+
+      try {
+        isPassRequired = isKeystorePassRequired(keystore);
+      } catch (e) {
+        //TODO: communicate invalid file to user
+      }
+      this.setState({ isPassRequired: isPassRequired });
```
crptm: Do not store derived state in state

common/components/WalletDecrypt/Keystore.jsx
dternyak: Yikes -- good catch @wbobeirne

common/libs/keystore.js
```diff
@@ -5,10 +5,97 @@ import {
...
+}
+
+//adapted from https://github.com/kvhnuke/etherwallet/blob/de536ffebb4f2d1af892a32697e89d1a0d906b01/app/scripts/myetherwallet.js#L218
+export function decryptPresaleToPrivKey(
+  file: string,
+  password: string
+): Buffer {
+  let json = JSON.parse(file);
+  let encseed = new Buffer(json.encseed, 'hex');
+  let derivedKey = pbkdf2Sync(
```
crptm: We prolly want to make decrypts async

dternyak: Why?

crptm: Mist keystores are (were?) too slow to open, i.e. https://github.com/kvhnuke/etherwallet/issues/43

dternyak: @skubakdj ^

skubakdj: Should we switch to the `scrypt-async` lib mentioned in kvhnuke/etherwallet#43?

common/components/WalletDecrypt/PrivateKey.jsx
skubakdj: Created function `validatePkeyAndPass` to DRY things up :smile: how does that look @wbobeirne?

common/libs/keystore.js
dternyak: @skubakdj Yes, let's go with that and see how it turns out.

dternyak: If we don't like it, we can just continue to keep it sync for v4.

skubakdj: Sounds like we might have to figure out a different the wallet class structure. As described in #103, this PR has `super(decryptUtcKeystoreToPkey(keystore, password));` for UTC keystores. I don't think we can have async calls in class constructors?

dternyak: Let's just keep it sync for now then.

common/libs/decrypt.js
skubakdj: Added tests for `decrypt.js` and for the major decrypt functions in `keystore.js`.

common/components/WalletDecrypt/Keystore.jsx
skubakdj: Derived from props instead :+1:

common/components/WalletDecrypt/Keystore.jsx
skubakdj: Definitely agree the `<form onSubmit>` event would be simpler, going to leave this alone for now though.

skubakdj:
@wbobeirne, @crptm I think I addressed your comments with these last few commits. Anything else that needs to be touched up?

PR #60, opened 2017-07-18
WIP: send part3

crptm:

PR #118, opened 2017-08-15
Send token

wbobeirne:
## What This Does

Finishes off the token support @crptm started in #60. Allows you to actually generate a transaction for an ERC20 token, and view it in the confirmation modal. Still no broadcasting.

## Why's This PR So Damn Big?

Well you might look at v3's implementation and think to yourself, doesn't look like much. But that's because v3's implementation is very prone to error. Currently the confirmation modal it uses for token transactions simply echoes what's in the value box before it. This means that if anything goes wrong in the conversion of value -> transaction data, or transaction data -> raw transaction, a user wouldn't see that in the modal. Likewise, the address it shows faced the same issues.

The implementation here actually decodes the `data` param of the raw transaction for display, so that it's going straight to the horse's mouth for the confirmation modal value and receiver address, rather than relying on input values.

In order to implement that, there's a whole lot of scaffolding for the contract and ERC20 interface. I converted it to use `ethereumjs-abi`, and wrote a bunch of unit tests too, just to make sure that it all does as advertised.

## There'd Better Be a Screenshot at the End of This

Yep.

screen shot 2017-08-15 at 4 33 45 am


common/containers/Tabs/SendTransaction/components/AmountField.jsx
```diff
@@ -65,7 +65,7 @@ export default class AmountField extends React.Component {
 
   onSendEverything = () => {
     if (this.props.onChange) {
-      this.props.onChange("everything", this.props.unit);
+      this.props.onChange('everything', this.props.unit);
```
crptm: This has been moved inside component in #60

common/containers/Tabs/SendTransaction/components/ConfirmationModal.jsx
```diff
@@ -16,12 +18,14 @@ type Props = {
   rawTransaction: RawTransaction,
   wallet: BaseWallet,
   node: NodeConfig,
+  token: ?Token,
   onConfirm: RawTransaction => void,
   onCancel: () => void
 };
 
 type State = {
```
crptm: If it doesnt change, it shouldnt be in state

common/containers/Tabs/SendTransaction/components/ConfirmationModal.jsx
```diff
@@ -56,24 +72,51 @@ export default class ConfirmationModal extends React.Component {
...
     clearTimeout(this.readTimer);
   }
 
+  _getStateFromProps(props: Props) {
+    const { rawTransaction, token } = props;
+    const { value, gasPrice, nonce, to } = rawTransaction;
+    let fixedValue;
+    let toAddress;
+
+    if (token) {
```
crptm: Cheating

common/containers/Tabs/SendTransaction/components/ConfirmationModal.jsx
```diff
@@ -56,24 +72,51 @@ export default class ConfirmationModal extends React.Component {
...
+      const tokenData = ERC20.decodeTransfer(rawTransaction.data);
```
crptm: lets use call, $call, for encode/decode names?

common/libs/contract.js
```diff
@@ -46,17 +43,24 @@ export default class Contract {
     return method;
   }
 
-  call(name: string, args: any[]): string {
-    const method = this.getMethodAbi(name);
+  getMethodTypes(method: ABIMethod) {
+    return method.inputs.map(i => i.type);
+  }
+
+  getMethodSelector(method: ABIMethod) {
```
crptm: exists in ethereumjs-abi

common/libs/erc20.js
```diff
@@ -51,12 +57,20 @@ class ERC20 extends Contract {
...
   }
 
-  balanceOf(address: string) {
+  balanceOf(address: string): string {
     return this.call('balanceOf', [address]);
   }
 
-  transfer(to: string, value: Big) {
-    return this.call('transfer', [to, value]);
+  transfer(to: string, value: string | number | Big): string {
```
crptm: prefer casting at call site

common/libs/nodes/rpc/index.js
```diff
@@ -71,13 +63,19 @@ export default class RpcNode extends BaseNode {
 
   async generateTransaction(
```
crptm: Doesnt belong to node

common/libs/nodes/rpc/index.js
```diff
@@ -104,32 +102,51 @@ export default class RpcNode extends BaseNode {
...
+
+      if (token) {
+        // $FlowFixMe - We reject above if tx has no data for token
+        const transferData = ERC20.decodeTransfer(tx.data);
+        value = new Big(transferData.value);
+      } else {
+        value = new Big(tx.value);
+      }
+
+      console.log(value.toString());
```
crptm: logs2 15113336_std

common/libs/transaction.js
```diff
@@ -1,5 +1,6 @@
 // @flow
 
+// TODO: Enforce more bigs, or find better way to avoid ether vs wei for value
```
crptm: lets switch to BN and use wei everywhere

common/containers/Tabs/SendTransaction/components/ConfirmationModal.jsx
wbobeirne: What's `$call`?

common/containers/Tabs/SendTransaction/components/ConfirmationModal.jsx
wbobeirne: Can you elaborate? Are you suggesting I figure this out from to address / transaction data?

common/libs/contract.js
wbobeirne: Is that `methodID()`? Wasn't sure if they were identical or not.

common/libs/nodes/rpc/index.js
wbobeirne: Yeah, this got a lot bigger than I initially intended. What's our current pattern for doing rpc calls in separate libraries? Do you pass the node into each function, or have a class with a constructor that takes it in?

common/containers/Tabs/SendTransaction/components/ConfirmationModal.jsx
crptm: Yep

common/containers/Tabs/SendTransaction/components/ConfirmationModal.jsx
crptm: convention for decodeFoo => $foo

common/libs/contract.js
crptm: yep

common/libs/nodes/rpc/index.js
crptm: Thats the hard part we have to figure something out

common/containers/Tabs/SendTransaction/components/ConfirmationModal.jsx
wbobeirne: Hypothetically it could, since it's calculated from props. I wouldn't want someone to introduce a bug where they reuse the same modal for multiple transactions, only changing the props, and not have it update accordingly.

Unless you're saying I should recalculate these every render, but that seems a little wasteful to me.

common/containers/Tabs/SendTransaction/components/ConfirmationModal.jsx
wbobeirne: Alright, I could take it out of props, loop over all of the tokens, check if the to address matches any of them, and derive the token from that.

common/containers/Tabs/SendTransaction/components/ConfirmationModal.jsx
wbobeirne: Is this something used commonly? `ethereumjs-abi` and most native JS functions use `decode` in function names, and I've not seen that $ prefix for functions before. I would be pretty uncertain of what a function was doing without knowing that.

common/libs/transaction.js
wbobeirne: https://github.com/MyEtherWallet/MyEtherWallet/issues/64#issuecomment-322709914 👍

common/containers/Tabs/SendTransaction/components/ConfirmationModal.jsx
crptm: You def should recalculate on every render. It is never too slow and when it is shouldComponentUpdate can ensure you only rerender when props change.

common/containers/Tabs/SendTransaction/components/ConfirmationModal.jsx
crptm: Not a common pattern, easier to write/read till we get proper contract call interface.  Not insisting tho.

common/containers/Tabs/SendTransaction/components/AmountField.jsx
wbobeirne: Oh yeah I like this component being more self sufficient. There's a lot of cleanup I want to do to send to make it less monolithic (For example, move the form into its own component so there's less `if (this.props.wallet)` going on.) Would you mind if I just cherry picked the commit in from #60 as a part of another cleanup PR?

common/containers/Tabs/SendTransaction/components/ConfirmationModal.jsx
wbobeirne: SGTM 👍

common/libs/nodes/rpc/index.js
wbobeirne: I'll just make it a function in `libs/transaction.js` for now that takes in a node, but leave a note saying we might want to figure out a better way. I'm not sure I see the bigger picture of transactions in MEW beyond Send yet, and I don't want to make a large architectural decision on incomplete information.

common/containers/Tabs/SendTransaction/components/ConfirmationModal.jsx
wbobeirne: Moved most of this out of state, and into a `_decodeTransaction` function:  a8c579b4b7de6136d9102aa3f4887182e6c3d6d8

common/libs/contract.js
wbobeirne: Replaced with 3285f529834a41cfbe20e5861cb7499eed6966e6, but still leaving `getMethodSelector` so that it can take in an ABIMethod object.

wbobeirne:
I think I've addressed everything you mentioned. There are still a few TODOs hanging out, and some improvements and cleanup to be done, but I think we should focus on getting this and #108 in while they're still manageably sized, and make smaller issues for the outstanding tasks.

Let me know if there's anything else you want to see changed!

crptm:
Cleartimeout should be clearinterval

common/libs/contract.js
```diff
@@ -65,25 +65,16 @@ export default class Contract {
...
-      default:
-        throw new Error(`Dont know how to handle abi type ${input.type}`);
-    }
+  decodeArgs(method: ABIMethod, argData: string) {
+    // Remove method selector from data, if present
+    argData = argData.replace(`0x${this.getMethodSelector(method)}`, '');
+    // Convert argdata to a hex buffer for ethereumjs-abi
+    const argBuffer = new Buffer(argData, 'hex');
+    // Decode!
+    return abi.rawDecode(this.getMethodTypes(method), argBuffer);
```
crptm: I think this wants to have whole data. including method selector

common/libs/transaction.js
```diff
@@ -29,3 +39,95 @@ export type BroadcastTransaction = {|
...
+  if (token) {
+    // $FlowFixMe - We reject above if tx has no data for token
+    value = new Big(ERC20.decodeTransfer(tx.data).value);
+    balance = toTokenUnit(await node.getTokenBalance(tx.from, token), token);
+  } else {
+    value = new Big(tx.value);
+    balance = await node.getBalance(tx.from);
+  }
+
+  console.log(value.toString());
```
crptm: logs

common/libs/transaction.js
```diff
@@ -29,3 +39,95 @@ export type BroadcastTransaction = {|
   rawTx: string,
   signedTx: string
 |};
+
+export async function generateTransaction(
+  node: BaseNode,
+  tx: Transaction,
+  wallet: BaseWallet,
+  token: ?Token
```
crptm: If you want to move token along everywhere it should be inside transaction object

common/libs/transaction.js
```diff
@@ -29,3 +39,95 @@ export type BroadcastTransaction = {|
...
+): Promise<BroadcastTransaction> {
+  // Reject bad addresses
+  if (!isValidETHAddress(tx.to)) {
+    return Promise.reject(new Error(translate('ERROR_5')));
```
crptm: You can throw in async function

common/libs/transaction.js
```diff
@@ -29,3 +39,95 @@ export type BroadcastTransaction = {|
...
+  if (gasPriceBig.greaterThan(new Big('1000000000000'))) {
+    return Promise.reject(
+      new Error(
+        'Gas price too high. Please contact support if this was not a mistake.'
+      )
+    );
+  }
+
+  // Ensure their balance exceeds the amount they're sending
+  // TODO: Include gas price too, tokens should probably check ETH too
```
crptm: Eth branch should account for gas prices. Maybe token branch as well?

common/containers/Tabs/SendTransaction/components/ConfirmationModal.jsx
```diff
@@ -56,24 +61,52 @@ export default class ConfirmationModal extends React.Component {
...
-    });
+    this._setWalletAddress(this.props.wallet);
   }
 
   componentWillUnmount() {
     clearTimeout(this.readTimer);
   }
 
+  _setWalletAddress(wallet: BaseWallet) {
+    wallet.getAddress().then(fromAddress => {
```
crptm: No ideas who did this, but this would complicate a lot of our code. If hw wallets need async address resolution they should do it on unlock cc @skubakdj

common/containers/Tabs/SendTransaction/components/ConfirmationModal.jsx
```diff
@@ -3,48 +3,53 @@ import './ConfirmationModal.scss';
...
 
+  componentWillReceiveProps(newProps: Props) {
+    // Reload address if the wallet changes
+    if (newProps.wallet !== this.props.wallet) {
+      this._setWalletAddress(this.props.wallet);
+    }
+  }
+
   // Count down 5 seconds before allowing them to confirm
   readTimer = null;
```
wbobeirne: This is the setInterval reference, the current time left is in state.

wbobeirne: Ah, apparently `clearTimeout` takes `any` but `clearInterval` specifically takes `number`. What a weird quirk. Either way, this is becoming 0.

common/containers/Tabs/SendTransaction/components/ConfirmationModal.jsx
```diff
@@ -3,48 +3,53 @@ import './ConfirmationModal.scss';
...
     super(props);
-    const { value, gasPrice, nonce } = props.rawTransaction;
 
     this.state = {
-      address: '',
-      value: toUnit(new Big(value, 16), 'wei', 'ether').toString(),
-      gasPrice: toUnit(new Big(gasPrice, 16), 'wei', 'gwei').toString(),
-      nonce: new Big(nonce, 16).toString(),
+      fromAddress: '',
       timeToRead: 5
```
wbobeirne: It's async. I saw your note elsewhere about not making it async, and having worked on the Trezor wallet a bit, I totally see what you mean. All of that should probably get resolved before wallet creation.

But right now, I have to keep it in state because async.

common/containers/Tabs/SendTransaction/components/ConfirmationModal.jsx
wbobeirne: I'll take some blame for this, since I 👍 the proposal to make it this way. But I agree, after having worked with wallets a bit more, it seems like all of this should get resolved during unlock. Apologies for my misguided earlier approval.

dternyak: I think I actually made this suggestion @wbobeirne, so no need to take the blame :)

However, I do agree that it makes more sense to construct the wallet instance with the address since we only need to ask once at unlock, and never again (unlike sign).

common/libs/transaction.js
wbobeirne: Yeah, definitely. I wasn't sure if hand-rolling this was the way to go though, or if `ethereumjs-tx` might be able to lend a hand. I was also going to look up a reference implementation later to see exactly what and how I should be checking.

I also would like to learn more about how much the actual broadcast portion of this catches, and how nice those errors look. I would be more inclined to lighten the load of this function if more will be caught at send time. That'll also make offline transactions less of a pain to handle, if we do less talking to the node. v3's implementation seems quite light wrt pre-send error checking.

common/libs/contract.js
wbobeirne: The `erc20` tests will show that this works with this implementation. However, I should definitely add some tests for this function directly, not just the `erc20` one.

`ethereumjs-abi`'s docs are really bad because they show functions that _don't exist yet_. The only ones that exist for this are `rawEncode` and `rawDecode` which cannot handle full encodes, only params. The `encode/decode` and `simpleEncode/simpleDecode` functions they have in the docs are just proposals, not actually working.

If you have advice on a better way to use the library, or a different library that handles this more gracefully, I'm all ears.

common/libs/transaction.js
wbobeirne: Yep, as I've been working my way down, I'm realizing this wants to go damn near everywhere. I'll see if it makes sense to sock it in `Transaction` as an optional key.

common/containers/Tabs/SendTransaction/components/ConfirmationModal.jsx
```diff
@@ -90,6 +123,8 @@ export default class ConfirmationModal extends React.Component {
       }
     ];
 
+    const symbol = token ? token.symbol : 'ETH';
```
wbobeirne: Can you give me a little more guidance on what this means? Or is this something I can drop in as a TODO for the future when we build out the related feature?

common/containers/Tabs/SendTransaction/components/ConfirmationModal.jsx
crptm: If it is interval null is fine i guess

common/containers/Tabs/SendTransaction/components/ConfirmationModal.jsx
crptm: 'ETH' should be 'ETC' on etc network, 'RSK' on rootstock, etc

common/libs/contract.js
crptm: Can you test it w/o stripping selector? I think it worked for me and then code becomes simpler.

common/libs/transaction.js
crptm: I think we should validate as much as we can. We have no guarantees about future error message from not yet written node software :) This should also skip some of those when creating tx offline. Maybe we should lift/duplicate some checks to confirm/broadcast.

common/libs/contract.js
wbobeirne:
```
 FAIL  spec/libs/erc20.spec.js
  ● ERC20 › decodeTransfer › should return the correct transaction given some data

    expect(received).toBe(expected)

    Expected value to be (using ===):
      "0x7cB57B5A97eAbe94205C07890BE4c1aD31E486A8"
    Received:
      "0x0"
```

When I comment the line out. Going to add some more tests just to make sure this isn't something wrong with the ERC20 implementation, but I'm pretty sure it's from the method selector screwing it up.

wbobeirne: Added more tests to $call directly, instead of relying on these erc20 tests. They confirmed that we need to strip the selector.

wbobeirne:
I think I was able to crack using signed transactions with the help of `ethereumjs-tx`. It had some quirks, but I cleaned up interacting with it by adding a `getTransactionFields(tx: EthTx)` function to libs/transactions which makes it much easier to read from one of those objects. Let me know if you have any thoughts on the implementation.

common/libs/contract.js
crptm: Ok i found our problem :) You decoding encoded call and i was thinking it decodes responses (bc thats what #60 does)

wbobeirne: Huh, I thought I had cherry picked most everything out of #60 but I think I had left behind some of the ENS related commits which must have been where the decoding stuff happened. Sorry about that.

wbobeirne:
Created issues that cover outstanding comments in #130 and #131. Let me know if I missed anything critical. Merging this as-is after consulting with @dternyak and @crptm.

PR #124, opened 2017-08-19
Prefer to reference build tools in node_modules

willpiers:
After running `npm install && npm run build` I was getting an error where `lint-staged` was not available in my `PATH`.

This PR changes all npm scripts to reference dependencies that are already in `node_modules` so that all scripts should work after an `npm install`

### DIFF
--- a/package.json
+++ b/package.json
@@ -91,18 +91,18 @@
     "webpack-hot-middleware": "^2.18.2"
   },
   "scripts": {
-    "db": "nodemon ./db",
-    "build": "webpack --config webpack_config/webpack.prod.js",
-    "prebuild": "check-node-version --package",
-    "build:demo": "BUILD_GH_PAGES=true webpack --config webpack_config/webpack.prod.js",
-    "prebuild:demo": "check-node-version --package",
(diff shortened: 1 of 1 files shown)


crptm:
npm scripts should already add node_modules/.bin to PATH before execution, not sure why its not working for you

willpiers:
`lint-staged` and `check-node-version` appear not to conform to the common pattern of putting executables in their `bin` directories - maybe that has something to do with it.

wbobeirne:
Those are both in my `node_modules/.bin` directory. What OS / node / npm version are you on?

dternyak:
@willpiers Have you gotten a chance to examine your dev env?

willpiers:
running OS X 10.12.0
node: node 8.4.0
npm: 5.3.0

I change node versions fairly often though so it could have been different when I created this PR. I tried removing some global node modules and reproducing this issue, to no avail - I should have taken a screenshot when I originally saw the issue.

dternyak:
No worries. I'll close for now, but thanks for letting us now about your issue.

PR #133, opened 2017-08-23
Add reducer tests

dternyak:
## What?
This PR adds "low-hanging fruit" tests for:
- config
- generateWallet
- wallet

Additionally, swap contains significant coverage. 

### Why?
While coverage for all of these reducers is still not optimal, I am opening a pull request now so as to keep the size smaller.

### DIFF
--- a/common/reducers/swap.js
+++ b/common/reducers/swap.js
@@ -1,7 +1,11 @@
 // @flow
-import { combineAndUpper } from 'utils/formatters';
 import type { SwapAction } from 'actions/swapTypes';
 import without from 'lodash/without';
+import {
+  buildDestinationAmount,
+  buildDestinationKind,
+  buildOriginKind
+} from './swapHelpers';
 export const ALL_CRYPTO_KIND_OPTIONS = ['BTC', 'ETH', 'REP'];
 const DEFAULT_ORIGIN_KIND = 'BTC';
 const DEFAULT_DESTINATION_KIND = 'ETH';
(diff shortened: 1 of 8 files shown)


common/reducers/swap.js
```diff
@@ -92,8 +74,10 @@ export function swap(state: State = INITIAL_STATE, action: SwapAction) {
       };
     }
     case 'SWAP_DESTINATION_KIND': {
+      const newOriginKind = buildOriginKind(state.originKind, action.value);
```
dternyak: Previously, we were assuming that there would never be an action created in which the destination kind was the same as the origin kind.

common/reducers/swapHelpers.js
```diff
@@ -0,0 +1,36 @@
...
+export const buildOriginKind = (
+  originKind: string,
+  destinationKind: string
+): string => {
+  if (originKind === destinationKind) {
+    return without(ALL_CRYPTO_KIND_OPTIONS, destinationKind)[0];
+  } else {
+    return originKind;
+  }
+};
```
wbobeirne: What would you think about making these kinds of sub-files in a folder, and making the reducer index.js? I think it would keep the directory cleaner.

spec/reducers/swap.spec.js
```diff
@@ -0,0 +1,184 @@
...
+    expect(
+      swap(undefined, swapActions.destinationAddressSwap(destinationAddress))
+    ).toEqual({
+      ...INITIAL_STATE,
+      destinationAddress
+    });
+  });
+
+  it('should handle SWAP_RESTART', () => {
+    // TODO - should reducer throw if address is not in correct format?
```
wbobeirne: IMO reducers shouldn't throw, that's on the action creator / saga to deal with. I guess just the saga, since we don't have redux-thunk or anything like that.

common/reducers/swapHelpers.js
dternyak: Good idea. I'll make that change now.

spec/reducers/swap.spec.js
dternyak: I'm actually not sure how that TODO got in there, but yeah, agreed.

Issue #132, opened 2017-08-23
Create Warning Banner for Alpha V4


dternyak:
We need a red banner fixed at the top of V4 to notify users that V4 is in alpha and should not be used beyond experimentation.

dternyak:
Fixed by https://github.com/MyEtherWallet/MyEtherWallet/pull/135

Issue #92, opened 2017-07-25
View with Address Only


dternyak:
- [x] Show Address
- [x] Show Balances
- [ ] Print Paper Wallet

Github Ref

1. Feature: kvhnuke/etherwallet: Issue #507 Bookmarkable / refreshable page
2. Feature: Printable mnemonics if they unlock via mnemonic kvhnuke/etherwallet: Issue #493

wbobeirne:
From discussion on weekly call, I think the implementation of this is going to be:

1) Remove "View Wallet Info" tab
2) Rename "Send Ether & Tokens" tab to "View Wallet & Send" (Feel free to suggest something better)
3) Make this tab have subtabs (ala contracts) for sending and viewing wallet info. The latter is where we'll add redownload paper wallet / redownload wallet file / convert to password wallet etc.

This will remove the need for a second tab for this, and remove the need for unlocking your wallet twice to do these things. Let me know if any of that doesn't sound good.

dternyak:
R/e renaming the send tab -- I'm in favor of "Account"

wbobeirne:
Started off on this but unfortunately the changeover to `ethereumjs-wallet` is making this a little tough to fit in. Their [wallet constructors](https://github.com/ethereumjs/ethereumjs-wallet#wallet-api) only have a `fromPublicKey`, not a `fromAddress`. As far as I know, there's no way to derive the public key from the wallet address. So I think my current plan is to stub out the wallet as best as I can. Hopefully this doesn't result in strange behavior where we fail to account for the read-only view.

wbobeirne:
Disregard previous comment, first PR up that works with this :)

WRT "Account", my only concern would be people assuming they have some kind of server-side MEW account. But maybe I'm putting too much stake into that word. I'll start with that for the next PR.

wbobeirne:
I'm hesitant to try and rework the send tab too much while #456 is on the horizon, so I think I'm going to limit my progress this sprint to #386 and #387. I'll circle back around to creating a proper wallet info tab once the refactor is over and merged.

wbobeirne:
This is merged into `transaction-refactor`, so I'm marking this closed, though it's not in develop yet.

PR #135, opened 2017-08-24
Visual Alpha Prep

wbobeirne:
This addresses #132 and a few other things. The complete list is:

* Add an alpha warning banner
    * Used warning color instead of danger to differentiate from v3 more
    * Text and color configurable in `config/data.js`
* Set package.json semver to `4.0.0-alpha.1`
* Set the header version to `4.0.0`
* Disable links that haven't been implemented yet
    * Considered "View Wallet Info" disabled because it's basically a stub (#92)
* Disable wallet options that haven't been implemented yet

This screenshot showcases all of the changes:
screen shot 2017-08-24 at 4 04 29 pm

### DIFF
--- a/common/components/Header/components/NavigationLink.jsx
+++ b/common/components/Header/components/NavigationLink.jsx
@@ -1,0 +1,50 @@
+// @flow
+import React from 'react';
+import classnames from 'classnames';
+import translate from 'translations';
+import { Link } from 'react-router';
+import './NavigationLink.scss';
+
+type Props = {
+  link: {
+    name: string,
+    to?: string,
+    external?: boolean
+  },
+  location: Object
+};
+
+export default class NavigationLink extends React.Component {
+  props: Props;
+
+  render() {
+    const { link, location } = this.props;
+    const linkClasses = classnames({
+      'NavigationLink-link': true,
+      'is-disabled': !link.to,
+      'is-active':
+        location.pathname === link.to ||
+        location.pathname.substring(1) === link.to
+    });
+    const linkLabel = `nav item: ${translate(link.name)}`;
+
+    const linkEl = link.external
+      ? <a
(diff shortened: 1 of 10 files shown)


wbobeirne:
Possible additions: Have tooltips on hover that explain that these aren't implemented yet? But I think the banner should be sufficient.

common/components/Header/components/NavigationLink.jsx
```diff
@@ -0,0 +1,50 @@
...
+  },
+  location: Object
+};
+
+export default class NavigationLink extends React.Component {
+  props: Props;
+
+  render() {
+    const { link, location } = this.props;
+    const linkClasses = classnames({
```
dternyak: This is new to me, but I believe you recommended it in the past. Probably time for me to look into it properly.

wbobeirne: Cool, I threw it in there since we'd talked about it a little before. It's nothing too fancy, just really reduces the amount of:

```
const someConditionalClass = condition ? 'has-class' : '';
const otherConditionalClass = otherCond ? 'has-other-class' : '';

<div className={`MyClass ${someConditionalClass} ${otherConditionalClass`}>
```

to

```
const classes = classnames({
  'MyClass': true,
  'has-class': !!condition,
  'has-other-class': !!otherCondition
});

<div className={classes}>
```

common/components/Header/components/NavigationLink.jsx
```diff
@@ -0,0 +1,50 @@
...
+type Props = {
+  link: {
+    name: string,
+    to?: string,
+    external?: boolean
+  },
+  location: Object
+};
+
+export default class NavigationLink extends React.Component {
```
dternyak: Nice refactor.

Issue #129, opened 2017-08-21
Hardware / Mnemonic wallet modal proposal


wbobeirne:
As someone relatively new to crypto, I was a little confused about the current hardware wallet modal. Specifically the concept of derivation paths. The current modal, for reference:

<img width="819" alt="screen shot 2017-08-21 at 2 56 54 pm" src="https://user-images.githubusercontent.com/649992/29533716-2c287932-8681-11e7-823a-f256fb992991.png">

I had the following questions when I first saw this:

* What is a derivation path?
* Do I need to change this?
* The one I selected mentions Jaxx, Metamask etc. Is that OK?
* What happens when I click "More addresses"? How many are on here? Can I go back?

After working on Trezor for a bit, I've realized that a lot of this stems from the re-use of the same modal for mneumonic wallets. I think it would be smart to keep that re-use, but tailor the modal for each hardware wallet via component props and make this a little simpler.

I put together some wireframes (These are NOT designs, just wireframes) of what I think would keep it simpler, without sacrificing the abilities offered from the current one. I'm not sure I captured _every_ state needed, but this should be enough to go off of when we implement this.

[select-address.pdf](https://github.com/MyEtherWallet/MyEtherWallet/files/1240030/select-address.pdf)

I'm still a little hazy on the derivation paths and BIP44, so please let me know if I lost anything in this simplification. CC @dternyak @crptm @skubakdj @tayvano for feedback.

crptm:
👍  not sure if all hw wallets support passing derivation path.

tayvano:
Yeah its a total shitshow. The problem is if you don't give them the ability to do derivation paths, you will have endless amounts of users unable to access their ETH they sent to an ETC path and vice versa.  There is no doubt a better way to handle it and we should handle it better. 

Right now the most common email regarding HD paths, surprisingly enough, is **not** confusion regarding which path—most just leave it unchanged—which is why the modal itself has remained the hot mess it is *(if it ain't broke....)*

The most common email is: "I sent my ETC to my Jaxx ETH wallet and now it's lost and they told me to contact you" to which we tell them how to find their ETC and they are very happy campers and now use MEW. 🤷 

I would propose:

- Making sure every path is best path depending on mnemonic phrase / hardware wallet / network. 

- Giving users the ability to select a different one via a dropdown

- Placing a warning if they try to change it that they should be careful and email us if they need assistance or clarification. 

[...]

PR #137, opened 2017-08-27
Trezor Unlock + Deterministic Wallet Groundwork

wbobeirne:
## What This Does

This is the first pass at hardware wallet support, starting with Trezor, since that's the one I own. It allows for the unlocking of, and signing transactions for Trezor wallets. It also lays down a lot of the groundwork for future deterministic path generated wallets (Mnemonic, ledger, etc.)

Don't let the line count fool you, this PR isn't _that_ big. The bulk of it was because of `common/vendor/trezor-connect.js`. It's not on NPM, so it has to be loaded into the project directly. Currently in talks with them about the future of Trezor Connect.

The wallet modal was modeled after #129, not v3 directly, so there are some noteworthy differences.

## What This Doesn't Do

* Other hardware wallets
* Nearly enough unit testing, would love to figure out a way to do this better
* Message signing
    * There's no base method for this on BaseWallet, and I wasn't sure how solid the private key implementation was, so I figured I'd hold off on this until it came up
* Guaranteed accurate transaction signing
    * I implemented it as best I could from v3, but without broadcasting, it's hard to be sure I'm doing it right. Would love some help in auditing this.

## Basic Technical Overview

This is how Trezor does everything, and presumably the other HW wallets will have a similar flow:

* All interfacing with Trezor happens at the component layer in `components/WalletDecrypt/Trezor.jsx` via TrezorConnect
* Callbacks from TrezorConnect update props sent to `components/WalletDecrypt/DeterministicWalletsModal.jsx`
* The modal component is connected, and fires off actions to the `deterministicWallets` reducer / saga
* Changes that happen in the modal (changing path, selecting address) are bubbled back up to `Trezor.jsx`
* Once a wallet is selected, `Trezor.jsx` calls `onUnlock` and creates a new `TrezorWallet` instance
* `TrezorWallet` extends the `DeterministicWallet`, and only differs for `signRawTransaction` (And presumably `signMessage` in the future)

## Trezor in Action

_GIFs were recorded at reduced sizes, and addresses / amounts blurred, so don't worry if things look a little odd_

### Connecting

trezor-connect

---

### Token Values

trezor-tokens

---

### Paging to More Addresses

trezor-paging

---

### Path Selection

trezor-path

---

### Unlocking

trezor-unlocking


common/components/WalletDecrypt/Trezor.jsx
```diff
@@ -1,26 +1,139 @@
+// @flow
+import './Trezor.scss';
 import React, { Component } from 'react';
 import translate from 'translations';
+import TrezorConnect from 'vendor/trezor-connect';
+import DeterministicWalletsModal from './DeterministicWalletsModal';
+import TrezorWallet from 'libs/wallet/trezor';
+
+/* eslint-disable quotes */
+const TREZOR_PATHS = [
```
dternyak: consider moving out of component (and into config)

common/libs/transaction.js
```diff
@@ -47,8 +48,8 @@ export function getTransactionFields(tx: EthTx) {
   const [nonce, gasPrice, gasLimit, to, value, data, v, r, s] = tx.toJSON();
 
   return {
-    // No value comes back as '0x', but most things expect '0x0'
-    value: value === '0x' ? '0x0' : value,
+    // No value comes back as '0x', but most things expect '0x00'
```
dternyak: any link to this? seems like we're changing it once already, so would be good to know where this info is coming from.

common/libs/transaction.js
```diff
@@ -121,13 +122,14 @@ export async function generateTransaction(
 
   // Generate the raw transaction
   const txCount = await node.getTransactionCount(tx.from);
+  const cleanHex = hex => addHexPrefix(padToEven(stripHex(hex)));
```
dternyak: What kind of padding are we doing here? Maybe some comments to explain?

common/libs/validators.js
```diff
@@ -139,3 +139,7 @@ export function isValidRawTx(rawTx: RawTransaction): boolean {
 
   return true;
 }
+
+export function isValidPath(dPath: string) {
+  return dPath.split('\'/').length === 4;
```
dternyak: maybe a comment here explaining what's going on. I get that the deterministic path is being split, but its not extremely clear why 4 is important

common/libs/transaction.js
wbobeirne: Yeah, would definitely be good to figure out the "correct" behavior here. I just used these values based on complaining libraries. `0x0` from `0x` came from the node, `0x00` came from Trezor wanting padded values.

common/components/WalletDecrypt/Trezor.jsx
wbobeirne: 👍

common/libs/transaction.js
wbobeirne: This was taken from v3's `sanitizeHex` behavior, Trezor wanted %2 data. I can definitely add some comments to v3, but I don't truly understand the reasoning.

common/libs/validators.js
wbobeirne: https://github.com/bitcoin/bips/blob/master/bip-0032.mediawiki has more information, and test cases, I'll try to add a useful comment and implement some of its test cases.

wbobeirne:
Added clarifying comments, didn't end up implementing the bip32 test cases because they were more about `hdk.derive` and `publicToAddress` from `hdkey` and `ethereumjs-util` respectively.

Ready to have this merged whenever!

jamesob:
:tada:

PR #144, opened 2017-08-30
Create a basic saga pattern to handle potential network errors.

bfillmer:
Basic implementation of error handling pattern. `common/sagas/rates.js` is probably the easiest to grok. Some of the other implementations of our async behavior need more work, wanted to get a small PR in before tackling some of the more in-depth refactors.


common/sagas/rates.js
```diff
@@ -2,18 +2,28 @@
...
-function fetchRates(symbols) {
-  return fetch(
+const fetchRates = symbols =>
+  fetch(
     `https://min-api.cryptocompare.com/data/price?fsym=ETH&tsyms=${symbols.join(
       ','
     )}`
-  ).then(r => r.json());
-}
+  ).then(response => {
```
dternyak: Would prefer adding this to `api/utils/checkHttpStatus.js` and possibly renaming to `ensureOKResponse`. 

Perhaps we can standardize on always chaining `ensureOKResponse` once that is done.

Issue #87, opened 2017-07-25
[BLOCKED] Full ENS


dternyak:
- [ ] Start Auction
- [ ] Place Bid
- [ ] Start & Place Bid
- [ ] Reveal Bid
- [ ] Finalize Auction
- [ ] Resolve using Public Resolver

Available
- [ ] ETH
- [ ] ETC (future)
- [ ] Ropsten
- [ ] Rinkeby

Future (Low Priority)

- [ ] Is there a way to know if I have already revealed my bid for a certain name?
- [ ] Allow people to reveal via a JSON from the official client

Github Ref

1. Issue: Testnet ENS addresses kvhnuke/etherwallet#509
2. Issue: non-ascii characters? kvhnuke/etherwallet#480
3. Feature: Calendar kvhnuke/etherwallet#456
4. Show Address they are sending to: kvhnuke/etherwallet#454
5. Issue: Time kvhnuke/etherwallet#453

Azarielle:
[Added to Asana](https://app.asana.com/0/763966770926912/779983315837145)

PR #147, opened 2017-08-31
ENS: Tab + General Information Panel (Scaffolding)

HenryNguyen5:
Related to #87 
Adds an active ENS tab to the site. 
Inside the tab there's the general information pane that contains the accordions that display information on ENS to the user


common/containers/Tabs/ENS/components/generalInfoPanel/helperStatelessFunctionals.jsx
```diff
@@ -0,0 +1,10 @@
+import * as React from 'react';
+import type { ABlankNooperProps } from './types';
+
+export const ABlankNooper = ({
+  content, //eslint-disable-line
+  href //eslint-disable-line
```
wbobeirne: What's eslint beefing about here?

common/containers/Tabs/ENS/components/generalInfoPanel/helperStatelessFunctionals.jsx
```diff
@@ -0,0 +1,10 @@
...
+}): ABlankNooperProps =>
+  <a target="_blank" rel="noopener" href={href}>
+    {content}
+  </a>;
```
wbobeirne: I really like this component, and would like to see it in the shared components folder. My only suggestion is to use the `children` props so it behaves more like a traditional link. And maybe a more intent-based name, like "NewTabLink" or something.

common/containers/Tabs/ENS/components/generalInfoPanel/index.jsx
```diff
@@ -0,0 +1,135 @@
+// @flow
+import * as React from 'react';
+import GeneralInfoNode from './generalInfoNode';
+import { ABlankNooper } from './helperStatelessFunctionals';
+import type { InfoNode } from './types';
+const generalInfoNodes: Array<InfoNode> = [
```
wbobeirne: Purely stylistic, but we favor `InfoNode[]` over `Array<InfoNode>`

wbobeirne:
I think all the code is good here, but if we merge into develop, we should add a note that we're still building it out, so that people aren't confused about having all the info, but no way to kick it off.

dternyak:
@HenryNguyen5 Great start to ENS!

It looks like you have 8 failing flow checks. After those are fixed up, we're good to merge 👍 

Let me know if you need any help debugging them! :)

common/containers/Tabs/ENS/components/generalInfoPanel/helperStatelessFunctionals.jsx
HenryNguyen5: Right, mainly used content for a short-hand way of setting the HTML inside for one-off links. I can make it so it can accept `children || content`. Also, originally it would accept `{children, ...rest}` as props and then you would be able to over-ride any defaults/add any properties to `<a/>` but flow was throwing a fit  on how `rest` did not contain certain properties.

common/containers/Tabs/ENS/components/generalInfoPanel/helperStatelessFunctionals.jsx
HenryNguyen5: Throwing an error that I dont have prop-type validation on it, it's related to #122. Once it's finished I won't need those disabled lines

common/containers/Tabs/ENS/components/generalInfoPanel/helperStatelessFunctionals.jsx
dternyak: Yes, the shared components is in `/common/components`

PR #150, opened 2017-08-31
Swap Part 5

dternyak:
### What is this?
- QR code when sending Bitcoin
screen shot 2017-08-30 at 10 37 29 pm

- Loading Spinner for Bity Rates
screen shot 2017-08-30 at 10 32 55 pm

- Refactors (whoopee, lots of deletions)
- saga logic fixes


common/containers/Tabs/Swap/components/CurrentRates.jsx
```diff
@@ -25,9 +26,29 @@ export default class CurrentRates extends Component {
     });
   };
 
-  render() {
-    const { ETHBTC, ETHREP, BTCETH, BTCREP } = this.props;
+  buildPairRate = (origin, destination) => {
+    const pair = origin + destination;
+    const statePair = this.state[pair + 'Amount'];
+    const propsPair = this.props[pair];
+    return !propsPair
```
dternyak: Since there is no loading state provided to the component, this implementation is a bit "dirty".

However, because rates are constantly refreshed (every 5 seconds), I thought it better to simply spin when there is no rate data instead of when a rate request is being resolved.

common/containers/Tabs/Swap/components/CurrentRates.jsx
```diff
@@ -25,9 +26,29 @@ export default class CurrentRates extends Component {
     });
   };
 
-  render() {
-    const { ETHBTC, ETHREP, BTCETH, BTCREP } = this.props;
+  buildPairRate = (origin: string, destination: string) => {
+    const pair = origin + destination;
+    const statePair = this.state[pair + 'Amount'];
+    const propsPair = this.props[pair];
+    return !propsPair
```
dternyak: Since there is no loading state provided to the component, this implementation is a bit "dirty".

However, because rates are constantly refreshed (every 5 seconds), I thought it better to simply spin when there is no rate data instead of when a rate request is being resolved.

common/containers/Tabs/Swap/components/BitcoinQR.jsx
```diff
@@ -0,0 +1,39 @@
...
+              rel="noopener"
+            >
+              Please use the recommended TX fees seen here.
+            </a>
+          </p>
+        </section>
+      </div>
+    );
+  }
+}
```
wbobeirne: Was there an issue with the QRCode component?

common/containers/Tabs/Swap/components/CurrentRates.jsx
wbobeirne: Yeah, I think this is preferable. Though it might be good in the future to add a subtle visual cue for when rates are refreshing.

common/containers/Tabs/Swap/components/SwapProgress.jsx
```diff
@@ -40,22 +40,15 @@ export default class SwapProgress extends Component {
...
+          linkElement = `<a href="${link}" target='_blank' rel='noopener'>${notificationMessage}</a>`;
           // BTC uses a different explorer
         } else {
           link = bityConfig.btcExplorer.replace('[[txHash]]', outputTx);
-          linkElement = (
-            <a href={link} target="_blank" rel="noopener">
-              {' '}View your transaction{' '}
-            </a>
-          );
+          linkElement = `<a href="${link}" target='_blank' rel='noopener'>${notificationMessage}</a>`;
```
wbobeirne: Is this going to be alright? Did we switch back to setInnerHtmlDangerously?

dternyak: No, there's no setInnerHtmlDangerously used here.

wbobeirne: The user will see `<a href...` on their screen when it's rendered here: https://github.com/MyEtherWallet/MyEtherWallet/blob/c7146a3d728684a9b77bc3f6b161dcb27dad9b75/common/containers/App/Notifications.jsx#L38

https://facebook.github.io/react/docs/dom-elements.html#dangerouslysetinnerhtml for more information

PR #156, opened 2017-09-01
MEW Alpha 0.01

dternyak:


wbobeirne:
We want to merge this into master, no? And then have some CI thing `npm run build` and copy `dist/` into `gh-pages`?

tayvano:
Not a bad plan. This is the command I currently use to push to `gh-pages`:

`git subtree push --prefix dist origin gh-pages`

PR #164, opened 2017-09-05
Refactor Send

dternyak:
Partially addresses https://github.com/MyEtherWallet/MyEtherWallet/issues/142

* simplifies generateTx and moves significant business logic out of component
* fixes various todos
* simplifies tx types

While this PR is larger than I would have hoped and some of the naming changes are only subjectively better, I believe it is a good start to reducing some of the complexity that has built up around Send.


common/containers/Tabs/SendTransaction/index.jsx
```diff
@@ -153,19 +157,18 @@ export class SendTransaction extends React.Component {
     if (this.state.generateDisabled !== !this.isValid()) {
```
HenryNguyen5: could do without the double negation here

common/containers/Tabs/SendTransaction/index.jsx
```diff
@@ -153,19 +157,18 @@ export class SendTransaction extends React.Component {
...
         this.props.transactions,
         componentStateTransaction.signedTx
       );
       // if there is a matching tx in redux state
-      if (currentTxAsBroadcastTransaction) {
+      if (currentTxAsSignedTransaction) {
         // if the broad-casted transaction attempt is successful, clear the form
-        if (currentTxAsBroadcastTransaction.successfullyBroadcast) {
-          this.resetTransaction();
+        if (currentTxAsSignedTransaction.successfullyBroadcast) {
```
HenryNguyen5: Can combine these into a single if- statement to avoid nesting

common/containers/Tabs/SendTransaction/index.jsx
```diff
@@ -334,44 +338,33 @@ export class SendTransaction extends React.Component {
...
+      unit,
+      value,
+      to,
+      data
+    };
+    return await formatTxInput(wallet, transactionInput);
   }
 
   async estimateGas() {
     if (!isNaN(parseInt(this.state.value))) {
```
HenryNguyen5: Better if you check the negation and break out of the function early if so, saves you a nested block

common/containers/Tabs/SendTransaction/index.jsx
```diff
@@ -334,44 +338,33 @@ export class SendTransaction extends React.Component {
...
       try {
-        const transaction = await this.getTransactionInfoFromState();
+        const transactionWithoutGas = await this.getFormattedTxFromState();
         // Grab a reference to state. If it has changed by the time the estimateGas
         // call comes back, we don't want to replace the gasLimit in state.
         const state = this.state;
```
HenryNguyen5: Confusing variable name, better off naming it `cachedState` or `prevState`

common/containers/Tabs/SendTransaction/index.jsx
```diff
@@ -439,50 +426,49 @@ export class SendTransaction extends React.Component {
...
     } catch (err) {
       this.props.showNotification('danger', err.message, 5000);
     }
   };
 
   openTxModal = () => {
-    if (this.state.transaction) {
-      this.setState({ showTxConfirm: true });
-    }
+    this.setState({ showTxConfirm: true });
```
HenryNguyen5: More concise if you combined `openTxModal/hideConfirmTx` into `toggleTxModal`

common/containers/Tabs/SendTransaction/index.jsx
```diff
@@ -65,7 +73,7 @@ type State = {
   gasLimit: string,
   data: string,
   gasChanged: boolean,
-  transaction: ?BroadcastTransaction,
+  transaction: ?CompleteTransaction,
```
wbobeirne: Much better name :+1:

common/containers/Tabs/SendTransaction/index.jsx
wbobeirne: I don't disagree, but the boilerplate of having some kind of force boolean argument, and not being able to use it directly from event bindings (Because first arg would be `ev`) make for a lot of annoyances.

common/libs/messages.js
```diff
@@ -0,0 +1,7 @@
+// TODO - move this out of transaction; it's only for estimating gas costs
+export type TransactionWithoutGas = {|
+  to: string,
+  value: string | number,
+  data: string,
+  from: string
+|};
```
wbobeirne: Nothing suitable from `libs/transaction` for this?

common/libs/nodes/rpc/index.js
```diff
@@ -39,8 +39,9 @@ export default class RpcNode extends BaseNode {
   async getTokenBalance(address: string, token: Token): Promise<Big> {
     return this.client.call(getTokenBalance(address, token)).then(response => {
       if (response.error) {
+        console.log(response.error);
```
wbobeirne: 🌲

common/containers/Tabs/SendTransaction/index.jsx
dternyak: I'm probably missing something, but I don't think I can in this case, as checking for truthiness of `componentStateTransaction` is necessary before `currentTxAsSignedTransaction` can be initialized.

common/libs/messages.js
dternyak: Nope :/

PR #166, opened 2017-09-05
Refactor Bity

dternyak:
Completely Addresses https://github.com/MyEtherWallet/MyEtherWallet/issues/165

### DIFF
--- a/common/config/data.js
+++ b/common/config/data.js
@@ -15,13 +15,20 @@ export const ANNOUNCEMENT_MESSAGE = `
   If you're interested in recieving updates about the MyEtherWallet V4 Alpha, you can subscribe via <a href="http://myetherwallet.us16.list-manage.com/subscribe?u=afced8afb6eb2968ba407a144&id=15a7c74eab">mailchimp</a> :)
 `;
 
-export const DONATION_ADDRESSES_MAP = {
+const etherScan = 'https://etherscan.io';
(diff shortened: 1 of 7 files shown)


common/config/data.js
```diff
@@ -15,14 +15,21 @@ export const ANNOUNCEMENT_MESSAGE = `
...
 
-export const DONATION_ADDRESSES_MAP = {
+const etherScan = 'https://etherscan.io';
+const blockChainInfo = 'https://blockchain.info';
+const ethPlorer = 'https://ethplorer.io';
+
+export const ETHTxExplorer = `${etherScan}/tx/[[txHash]]`;
+export const BTCTxExplorer = `${blockChainInfo}/tx/[[txHash]]`;
+export const ETHAddressExplorer = `${etherScan}/address/[[address]]`;
+export const ETHTokenExplorer = `${ethPlorer}/address/[[address]]`;
```
HenryNguyen5: Would probably be less error-prone if you instead made these functions that return a string that flow could validate. Ex.
```js
const ETHTxExplorer = (txHash: string) => `${etherScan}/tx/${txHash}`;
```

dternyak: Yeah agreed, this is legacy leftover from V3. I'll adjust it.

dternyak:
@HenryNguyen5 I've addressed your comments :)

PR #167, opened 2017-09-05
Refactor BaseNode to be an interface INode

HenryNguyen5:
Refactored the existing BaseNode class to instead be an interface under INode. Didn't refactor BaseWallet yet as its not 1-1 interface compatible (has one implemented method). Should I convert it to an interface and just implement a base-class that inherits from the interface and has that one method (`getNakedAddress`)? Or have `getNakedAddress` manually implemented into the classes that need it?


dternyak:
@HenryNguyen5 I'm not sure why we need `getNakedAddress` at all, so I would lean towards implementing it for each class that needs it. I'm sure having a base class would be fine as well. Up to you!

PR #168, opened 2017-09-05
Text change - Node Switcher/Nodes to Network Switcher/Network

PositiveVibe:
Changed the english translations to remove the word "node" and replace it with the word "network". https://github.com/MyEtherWallet/MyEtherWallet/issues/106


dternyak:
@PositiveVibe Thanks for your contribution!

Unfortunately, a straight-up find-and-replace will not be sufficient here. I believe we only need to change the text to reflect what is currently on V3 (image attached)
screen shot 2017-09-05 at 6 21 17 pm

Additionally, it appears you have checked in some hidden files that should probably not be part of the PR.

PositiveVibe:
I will remove those two extra files, not sure how they got in there. As far as replacing the text itself, I was under the impression these words all needed to be changed to "node" as well:

image

I would assume the error messages also need to be changed to show the correct words, which is why I changed those. I can remove the error messages text if you would like that. 

My only other question is that for the drop down, I see the "Add Custom Node" is hard coded. Seems to be in two different places in v3 and v4. Which leads me to my next question: should I make the changes for v3 or v4 or both?

tayvano:
Unsure on my feelings on this, so someone should just make a decision. I don't think it matters long  term either way. Basically, it used to be that was a way to change the node between MEW and Etherscan. With the addition of ETC, RSK, UBQ, EXP, and others the word `network` seems more appropriate, but didn't take the time (obviously) to make sure we used consistent language. If we feel `network` is the best describer of this feature, let's make sure to update knowledge base language (George or Weitze can handle that) as well. 

No need to update in v3. Let's just focus on making best choices in v4. 

Thanks!

dternyak:
Hey @PositiveVibe! 

Thanks again for your contribution. We're currently re-thinking some UX, and this issue is no longer relevant. Sorry we weren't able to get your PR in.

Please feel free to message me on slack @dternyak. We love contributors and would love to see you contribute in another way!

Best,
Daniel

PR #170, opened 2017-09-06
Fix Send Regressions

dternyak:
As I was playing with the latest develop, I noticed we had encountered several regressions (listed here https://github.com/MyEtherWallet/MyEtherWallet/issues/169).

This PR fixes both of these regressions, while also simplifying the unit dropdown by making use of the `SimpleDropdown` component.

It also clears out a javascript warning in the console for using the `translate` prop incorrectly.

### DIFF
--- a/common/containers/Tabs/SendTransaction/components/UnitDropdown.jsx
+++ b/common/containers/Tabs/SendTransaction/components/UnitDropdown.jsx
@@ -1,34 +1,16 @@
 // @flow
 import React from 'react';
+import SimpleDropDown from 'components/ui/SimpleDropdown';
 
-class Option extends React.Component {
-  props: {
-    value: string,
-    active: boolean,
-    onChange: (value: string) => void
-  };
-  render() {
-    const { value, active } = this.props;
-    return (
-      <li>
-        <a className={active ? 'active' : ''} onClick={this.onChange}>
-          {value}
-        </a>
-      </li>
-    );
-  }
-
-  onChange = () => {
-    this.props.onChange(this.props.value);
-  };
-}
+type UnitDropdownProps = {
+  value: string,
+  options: string[],
+  onChange?: (value: string) => void
+};
 
 export default class UnitDropdown extends React.Component {
-  props: {
-    value: string,
-    options: string[],
-    onChange?: (value: string) => void
-  };
+  props: UnitDropdownProps;
+
   state: {
     expanded: boolean
   } = {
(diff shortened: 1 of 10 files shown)


common/containers/Tabs/SendTransaction/components/UnitDropdown.jsx
```diff
@@ -1,87 +1,40 @@
...
-    const { value, options, onChange } = this.props;
-    const isReadonly = !onChange;
+    const { value, options } = this.props;
 
     return (
       <div className="input-group-btn">
-        <a
-          style={{ minWidth: 170 }}
+        <SimpleDropDown
           className="btn btn-default dropdown-toggle"
```
wbobeirne: Not sure how I feel about this, we should probably have a separate component for this. The dropdown component that Bootstrap provides doesn't behave like this, it needs completely different markup than a select list.

wbobeirne: Or just be able to toggle the styling, if that's all we care about.

common/containers/Tabs/SendTransaction/components/UnitDropdown.jsx
```diff
@@ -1,87 +1,40 @@
...
-
-  onChange = (value: string) => {
-    this.setState({
-      expanded: false
-    });
-
+  onChange = (e: SyntheticInputEvent) => {
     if (this.props.onChange) {
-      this.props.onChange(value);
+      this.props.onChange(e.target.value);
```
wbobeirne: Perhaps we should keep this interface the same (Just passing back value) rather than having this read in the event. In case we don't end up doing a select list, but instead make it a bootstrap dropdown.

common/containers/Tabs/SendTransaction/components/UnitDropdown.jsx
dternyak: Yeah agreed, the styling needs to be different. I meant it sort of as a placeholder, but we can remove it altogether.

Alternatively, we could create a `NonSimpleDropDown` that uses the fancy `<ul>`s and pretends to be a select menu while also offering more styling.

common/containers/Tabs/SendTransaction/components/UnitDropdown.jsx
dternyak: Would that really happen at the `UnitDropdown` level though, or wouldn't that be a different component.

I guess we need to define what `UnitDropdown` is and what `SimpleDropDown` is, and how many other dropdowns we need.

wbobeirne: I would expect it to happen in SimpleDropDown, so that it calls `this.props.onChange(ev.target.value)`

common/containers/Tabs/SendTransaction/components/UnitDropdown.jsx
wbobeirne: Maybe `SimpleSelect` and `SimpleDropdown`?

common/containers/Tabs/SendTransaction/components/UnitDropdown.jsx
```diff
@@ -1,87 +1,40 @@
...
-  }
-
-  onChange = () => {
-    this.props.onChange(this.props.value);
-  };
-}
+type UnitDropdownProps = {
+  value: string,
+  options: string[],
+  onChange?: (value: string) => void
```
HenryNguyen5: Should this be a type of synthetic event?

dternyak: @wbobeirne Mentioned he would prefer that it is not.

common/sass/styles/overrides/dropdowns.scss
```diff
@@ -0,0 +1,33 @@
...
+
+		> a {
+			font-weight: 300;
+
+			&.active {
+				color: $link-color;
+			}
+		}
+  }
+}
```
wbobeirne: Damn it, prettier didn't catch these, I'm fixing.

common/containers/Tabs/SendTransaction/components/UnitDropdown.jsx
wbobeirne: ^ Basically, dropdowns just receive click events, they're not input changes, so an event doesn't have a `value` attribute. In general, I think components should abstract out their markup as much as possible.

PR #174, opened 2017-09-07
Ether Unit Types and Send UX Improvements

dternyak:
Fixes https://github.com/MyEtherWallet/MyEtherWallet/issues/171

- Create and use Wei, GWei, or Ether unit types instead of `string` / `number` (and have difficulty telling what unit the amount is in)
- various refactors (e.g. breaking out `everything` amount change logic into `handleEverythingAmountChange`).

Also contains a few UX improvements:

1. clear previously generated rawTX/signedTx when changes to transaction inputs are made.
2. reset generated rawTx/signedTx while new generateTx is loading


common/containers/Tabs/SendTransaction/index.jsx
```diff
@@ -404,36 +405,65 @@ export class SendTransaction extends React.Component {
...
+        },
+        () => {
+          resolve();
+        }
+      );
     });
   };
 
   generateTxFromState = async () => {
+    await this.resetJustTx();
```
dternyak: It would probably be okay to not block here and allow setState to happen whenever, but it doesn't hurt anything and allows us to make assumptions about `state.transaction` being null further down.

common/containers/Tabs/SendTransaction/index.jsx
```diff
@@ -404,36 +405,65 @@ export class SendTransaction extends React.Component {
...
       value,
       unit,
-      token
+      token,
+      transaction,
+      generateDisabled
+    });
+  };
+
+  resetJustTx = async (): Promise<*> => {
```
HenryNguyen5: This could be cleaned up using arrow functions implict returns 
```js
resetJustTx = (): Promise<*> =>
  new Promise(resolve =>
    this.setState(
      {
        transaction: null
      },
      resolve
    )
  );
```

dternyak: 👍 I suppose so! Kinda funky though, I haven't really seen this pattern before (but I suppose why not just have resolve be the callback if all you do is call it in the callback anyway).

common/libs/transaction.js
```diff
@@ -174,48 +168,44 @@ export async function formatTxInput(
...
   chainId: number,
   transactionInput: TransactionInput
 ): Promise<CompleteTransaction> {
   const { token } = transactionInput;
-
   const formattedTx = await formatTxInput(wallet, transactionInput);
-
-  const from = await wallet.getAddress();
-
   const transaction: ExtendedRawTransaction = {
```
HenryNguyen5: Could destructure  `transaction` after line 199 to make the return object cleaner

dternyak: 👍

common/containers/Tabs/SendTransaction/index.jsx
HenryNguyen5: If we end up using async set state a lot, a HOF like this would come handy. (Pulled from a react-semantic-ui component which is why you see value as the second param)
```js 
		handleChange = itemToSet => (event, { value }) =>
			new Promise(resolve =>
				this.setState(prevState => {
					console.log('Previous state:', prevState, '\nForm value set:', {
						[itemToSet]: value
					})
					return { [itemToSet]: value }
				}, resolve)
			)
```

PR #175, opened 2017-09-08
Contract Refactor

HenryNguyen5:
Currently have the proposed contract-parser in its own folder at `common/libs/contracts` 

# What this does:
Takes in a contract interface ABI and returns an object with keys equivalent to the ABI function names that each have `.encodeInput`,  `.decodeInput`, `decodeOutput`, `.call`, `.send` methods. 

`.encodeInput`: Accepts key-value parameters and checks them at run-time against the given ABI function to see that sufficient parameters have been given, also allows for run-time parsing of supplied parameters for things such as BN, checking for sufficient address length, making sure numbers are converted into BN, etc...

`.decodeInput`: accepts an encoded argument string and decodes it, returning an object with key-value pairs with a 1-1 mapping to the specific function in the ABI. 

`.decodeOutput`: accepts a encoded output string (such as those returned from `eth_call`), returning an object with key-value pairs with a 1-1 mapping to the specific function in the ABI. **Note that some outputs have no name, so it will default to just using index as a key-value. 

`.call`: Equivalent to encoding the input, sending it to a a node with `eth_call` and then decoding the output

`.send`: Equivalent to encoding the input, making a complete transaction with it, and then returning the signed and raw tx back to the callee

## Why?
- Allow for much stricter checking of parameters before costly JSON-RPC calls
- Reduces error potential by accepting everything as an object with 1-1 ABI mapping instead of guessing at parameters
- Much better error messages when missing parameters or using the wrong type of parameter.
- Stream line our contract usage to one standard


common/libs/contract/ABIFunction.js
```diff
@@ -0,0 +1,158 @@
...
+    return valueMapping[type]
+      ? valueMapping[type](value)
+      : _isBigNumber(value) ? value.toString() : value;
+  };
+
+  _parsePreEncodedValue = (type, value) => {
+    const { _isBigNumber } = this;
+    return _isBigNumber(value) ? value.toString() : value;
+  };
+  _isBigNumber(object) {
```
dternyak: Would move outside of class as a helper utility.

common/libs/contract/ABIFunction.js
```diff
@@ -0,0 +1,160 @@
...
+    this.outputNames = outputs.map(
+      ({ name }, i) => outputMappings[i] || name || i
+    );
+
+    this.methodSelector = abi
+      .methodID(this.name, this.inputTypes)
+      .toString('hex');
+  }
+
+  encodeInput = (suppliedInputs = {}) => {
```
wbobeirne: I know flow typing is TBD which should clear this up, but I don't think we should use default args here, I'd want this to fail loud if the wrong thing is sent.

common/libs/contract/ABIFunction.js
HenryNguyen5: agreed

common/libs/contract/ABIFunction.js
HenryNguyen5: Reason why default arg is an object here is because `_processSuppliedArgs` does type/error checking per object key, which allows for more specific errors on what the supplied input is missing or what specific key is malformed.

HenryNguyen5:
https://github.com/facebook/flow/issues/1323
Seems like this flow issue makes it hard to dynamically assign properties to a class

HenryNguyen5:
Im thinking of switching to using ES6 proxies instead to intercept method calls to a contract, avoiding the mutation of `this` at run-time. Thoughts? @wbobeirne  @dternyak @bfillmer

common/libs/contracts/ABIFunction.ts
```diff
@@ -0,0 +1,207 @@
...
+    const data = this.encodeInput(input);
+
+    const returnedData = await node
+      .sendCallRequest({
+        to,
+        data
+      })
+      .catch(e => {
+        console.error(e);
+        //TODO: Put this in its own handler
```
dternyak: console.log - perhaps address todo at the same time?

PR #176, opened 2017-09-08
[WIP] Add eth call to node

HenryNguyen5:
#175 is related to this.
Adds `eth_call` generic functionality so user supplied contracts or things like the ENS library can form their own `eth_call` requests without cluttering  `rpc nodes` methods


common/libs/contract/index.js
```diff
@@ -0,0 +1,104 @@
...
+    abi.forEach(currentABIMethod => {
+      if (currentABIMethod.type === 'function') {
+        //only grab the functions we need
+        const { encodeInput, decodeInput, decodeOutput } = new ABIFunction(
+          currentABIMethod
+        );
+        const funcToAssign = {
+          [currentABIMethod.name]: { encodeInput, decodeInput, decodeOutput }
+        };
+        Object.assign(this, funcToAssign);
```
wbobeirne: This is going to be a nightmare for flow.

HenryNguyen5:
Already addressed in contracts-refactor

Issue #80, opened 2017-07-25
Wallet Decrypt


dternyak:
### Requirements

- [x] 1. Keystore
- [x] 1b. Presale JSON
- [x] 2. Raw private key
- [x] 2b. Deprecated Encrypted Private Key
- [x] 3. Bip39 Mnemonic with custom path / any number of words / password
- [ ] 5. Ledger
- [x] 6. Trezor
- [ ] 7. Address only (will be used on view wallet and send page for offline?)

##### Future (Low Priority)

"Is having the wallet info disappear on each tab-change a feature? Would it be better if it persisted between tabs? (The only implication I can think of is how would a user "sign out" of their wallet so they could choose another one.)"

- [ ] 5 minutes -> show warning that their wallet is unlocked in the top tabby-guy (meta title tag)
- [ ] 30 minutes of no interaction -> lock wallet, force user to unlock again
- [ ] If user wants to use a different wallet, tell them to refresh the freaking page?
- [ ] I'll make a little wallet icon with a cross through it or something that they can click to clear the current wallet from the current state and show the decrypt panel again
- [ ] Keepkey Hardware Wallet
- [ ] Improve UX of decryption of keys with high n values kvhnuke/etherwallet: Issue #486
- [ ] Hitting enter should unlock wallet

Github Ref

1. Make sure to strip beginning 0x from raw private keys: kvhnuke/etherwallet#400
3. Allow Metamask users to use metamask + mew? kvhnuke/etherwallet#238
4. High n values in keystores -> frozen browser kvhnuke/etherwallet#43

crptm:
I think we can have #2 anyway just treat it as address only?

dternyak:
Took it out

wbobeirne:
Trezor is up via #137, lays groundwork for Ledger and Mnemonic.

wbobeirne:
Pulled myself off of this as Trezor is in. I think we should consider breaking this monolithic task out into many more smaller granular tasks. I see 4-5 possible PRs contained in this, and many of these are questions about how we should implement future stuff, not well defined certain tasks.

dternyak:
Agreed. If you're interested, it would be great to have you break this up into more well defined and smaller issues (especially since you should have a well developed mental model based on your trezor work).

wbobeirne:
I have split this task out into #214, #215, #216, #217, #218, #219, and #220 for the remaining outstanding issues. Closing this one out and following up on those instead.

PR #180, opened 2017-09-08
Wallet Decrypt - Mnemonic

skubakdj:
### **What This Does**
Broadly speaking, two things are accomplished: 
- Adds mnemonic wallet decrypt as described in #103 & referenced by #80 
- Adds mnemonic support to the `DeterministicWalletsModal` & accompanying sagas

### **What Still Needs to Happen**
A few things to get done before this is merged:
- ~~Need to write unit tests for `decryptMnemonicToPrivKey`~~
- ~~Need to do more manual testing~~ 
- Need to test against Trezor to make sure I didn't break anything (may need some help for this one)


common/libs/validators.js
```diff
@@ -143,5 +143,7 @@ export function isValidRawTx(rawTx: RawTransaction): boolean {
 // Full length deterministic wallet paths from BIP32
 // https://github.com/bitcoin/bips/blob/master/bip-0032.mediawiki
 export function isValidPath(dPath: string) {
-  return dPath.split("'/").length === 4;
+  //TODO: use a regex to detect proper paths
+  const len = dPath.split("'/").length;
+  return len === 3 || len === 4;
```
skubakdj: Ledger uses `m/44'/60'/0'` for some reason:

https://ethereum.stackexchange.com/questions/19055/what-is-the-difference-between-m-44-60-0-0-and-m-44-60-0/19061

common/sagas/deterministicWallets.js
```diff
@@ -62,16 +72,22 @@ function* getDeterministicWallets(
...
-
-  for (let i = 0; i < wallets.length; i++) {
-    yield put(
-      updateDeterministicWallet({
-        ...wallets[i],
-        value: balances[i]
-      })
-    );
+
+  try {
```
skubakdj: `try/catch` introduced because derived addresses wouldn't populate on network/api failure

common/components/WalletDecrypt/Mnemonic.jsx
```diff
@@ -1,27 +1,129 @@
...
+          </h4>
+          <div className="form-group">
+            <textarea
+              id="aria-private-key"
+              className={`form-control ${isValidMnemonic
+                ? 'is-valid'
+                : 'is-invalid'}`}
+              value={phrase}
+              onChange={this.onMnemonicChange}
+              placeholder={translateRaw('x_Mnemonic')}
```
dternyak: You can save the `translateRaw` import by passing `true` as the argument for the `textOnly` parameter:

`translate('x_Mnemonic', true)`

dternyak: Makes me question if `translateRaw` should even be exported.

common/config/dpaths.json
```diff
@@ -12,5 +12,31 @@
...
+      "value": "m/44'/61'/0'/0"
+    },
+    {
+      "label": "Testnet",
+      "value": "m/44'/1'/0'/0"
+    },
+    {
+      "label": "Expanse",
+      "value": "m/44'/40'/0'/0"
+    }
```
wbobeirne: Seeing multiple now, might be worth making this a JS file so that we can assign these repeated ones to variables and reuse them.

common/libs/validators.js
wbobeirne: Oh wow, BIP44 is way more readable than BIP32. Would have saved me a lot of headaches: https://github.com/bitcoin/bips/blob/master/bip-0044.mediawiki

Would you mind changing the comment to reflect that BIP 44 is really where we're determining that 4 is the amount, but that ledger is the exception for 3?

common/components/WalletDecrypt/DeterministicWalletsModal.jsx
```diff
@@ -40,13 +40,15 @@ type Props = {
   dPaths: { label: string, value: string }[],
   publicKey: string,
   chainCode: string,
+  seed: string,
```
wbobeirne: This is a required prop, but the DecryptWallet/Trezor.jsx wasn't altered to send it (And probably shouldn't, I think this should be optional.) Any idea why flow didn't catch that?

common/sagas/deterministicWallets.js
```diff
@@ -62,16 +72,22 @@ function* getDeterministicWallets(
...
+      yield put(
+        updateDeterministicWallet({
+          ...wallets[i],
+          value: balances[i]
+        })
+      );
+    }
+  } catch (err) {
+    console.log(err);
+    //TODO: maybe communicate API call error to user?
```
wbobeirne: Definitely, I think you can just get away with a `yield put(showNotification(...`

common/sagas/deterministicWallets.js
```diff
@@ -86,21 +102,27 @@ function* updateWalletTokenValues(): Generator<Yield, Return, Next> {
...
+          tokenValues: {
+            ...wallets[i].tokenValues,
+            [desiredToken]: tokenBalances[i]
+          }
+        })
+      );
+    }
+  } catch (err) {
+    console.log(err);
+    //TODO: maybe communicate API call error to user?
```
wbobeirne: Same as above, showNotification

common/libs/validators.js
skubakdj: yep, will do :+1:

common/components/WalletDecrypt/DeterministicWalletsModal.jsx
skubakdj: Agree, I'll make it optional. Not sure why flow didn't catch that, it definitely should have...

common/components/WalletDecrypt/Mnemonic.jsx
```diff
@@ -1,27 +1,129 @@
...
+
+  onPasswordChange = (e: SyntheticInputEvent) => {
+    this.setState({ pass: e.target.value });
+  };
+
+  onMnemonicChange = (e: SyntheticInputEvent) => {
+    this.setState({ phrase: e.target.value });
+  };
+
+  onChooseAddress = (e: SyntheticInputEvent) => {
```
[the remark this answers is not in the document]
skubakdj: Good call, I'll switch it to `onDWModalOpen`, unless you can suggest something else.

wbobeirne:
Tested the branch on my Trezor, everything seems to work fine. Still not sure how it's not complaining more about the lack of seed arg, but looks like no regressions here.

dternyak:
@wbobeirne Interesting. I'm currently unable to send via Trezor on the latest develop. You're able to send via Trezor on this branch? Notice: The signed tx generates, but when attempting to broadcast it the node always reponds with 
```
Error: Insufficient funds. The account you tried to send transaction from does not have enough funds. Required XXX000000000000 and got: 0.
```

dternyak:
I'm going to attempt to send via Trezor on this branch -- will update with my results.

wbobeirne:
Sorry, allow me to reiterate, I only tested unlocking. This branch is a bit behind develop, and I didn't test sending until it was up to date.

common/components/WalletDecrypt/index.jsx
```diff
@@ -165,7 +171,7 @@ export class WalletDecrypt extends Component {
 
   onUnlock = (payload: any) => {
     this.props.dispatch(
-      WALLETS[this.state.selectedWalletKey].unlock(this.state.value || payload)
+      WALLETS[this.state.selectedWalletKey].unlock(payload || this.state.value)
```
wbobeirne: Payload is a SyntheticEvent in some cases. We'll either need to check if payload isn't one, or change how onUnlock is called so that it's never sent an event.

dternyak: I believe I have fixed this in one of my recent PRs https://github.com/MyEtherWallet/MyEtherWallet/pull/183/files#diff-c2a1968700ed6452c7c0257d4dde4cd4R168

wbobeirne: Sweet, in that case we just need to bring `develop` back into this bad boy.

skubakdj:
@dternyak, @wbobeirne I think I've addressed all the comments with these latest commits. Also added some tests, rebased this branch with `develop`, and did some manual testing. Any additional feedback on the changes?

Also, still haven't been able to test using Trezor, but I can get to it tonight if need be.

common/config/dpaths.json
skubakdj: Converted `dpaths.json` to `dpaths.js`.

dternyak:
@wboberine is working on a fix to Trezor, and is planning on commiting that
directly to this branch.

After that we should be good to merge! :)

PR #181, opened 2017-09-08
Hot module reload fixes

wbobeirne:
Fixes https://github.com/MyEtherWallet/MyEtherWallet/issues/46

This PR fixes hot module reloading by accepting all HMR changes, and moving routes into the root component.

Reloading was behaving weirdly with them as a function, and I wasn't entirely sure the reasoning behind that in the first place, so I simply moved them into the root component. I don't think that should be an issue.

Likewise, I'm not sure if there are any other modules we should have handle their reloading better, but as far as I can tell everything was behaving correctly with the blanket accept.


dternyak:
Looks great! I'm excited for the productivity improvements.

While hot-module replacement seemed to work fine, I received an error message while testing HMR: 

```
"You cannot change <Router routes>; it will be ignored"
```

I fixed it by implementing a solution similar to the one recommended here: https://github.com/ReactTraining/react-router/issues/2704#issuecomment-211352123

## Before
screen shot 2017-09-08 at 11 39 39 am

## After
screen shot 2017-09-08 at 11 41 42 am

PR #185, opened 2017-09-12
Alerting and UX Improvements.

dternyak:
### Why?
While I was forced to be offline during a plane ride, I noticed several UX issues when internet connectivity is unavailable, as well as misc alerting errors. 

## What does this include?
- Refactor rates actions and action creators to use standard network request state pattern (REQUESTED / SUCCEEDED / FAILED)

- Display 'No rates were loaded.' in EquivalentValues when rates are null, instead of providing the user with no feedback.

- Only Request Rates once AccountInfo Component has mounted, instead of upon saga instantiation (unneeded overhead). This allows also us to issue subsequent fiat rates requests to update the "equivalent values" should the users session persist.

 - Show '???' as account balance when balance is null
     - Wallet initial state with balance as null instead of 0. We don't actually know what the balance is, and we shouldn't have 0 as a default as this may confuse users and doesn't accurately reflect their balance.

- Fix error messaging (show error message instead of error Object)

- Open external links (to knowledge-base) in new tab with `target="_blank"`.


dternyak:
@wbobeirne Gah, agreed. I have the hardest time naming actions, I swear. 

I tried to follow the existing convention r/e network requests:

```
export function loadBityRatesRequestedSwap(): LoadBityRatesRequestedSwapAction {
  return {
    type: 'SWAP_LOAD_BITY_RATES_REQUESTED'
  };
}
```
Seems like this is a harder problem than it needs to be (at least for me).

PR #188, opened 2017-09-12
Improve Build Speed

wbobeirne:
## What This Does

I was able to improve build speed by doing the following things:

* Removing eslint-loader. I think this should be a pre-commit hook to enforce (ala prettier) and should be handled by the user's IDE, rather than this slow-ish loader.
* Changing the development devtool. This was the biggest gain, and only results in some minor usability changes with sourcemaps.

| Config                                           | Avg build speed | Avg rebuild speed |
|--------------------------------------------------|-----------------|-------------------|
| Eslint Loader / source-map devtool (current)     | 33.81s          | 3.81s             |
| No Eslint / source-map devtool                   | 28.76s          | 3.51s             |
| No Eslint / cheap-module-eval-source-map devtool | 25.13s          | 1.99s             |

## Future Improvements

* Implement the CommonChunkPlugin, ala [this guide](https://webpack.js.org/guides/build-performance/#minimal-entry-chunk) or otherwise.
* Remove all references to `[hash]` or `[chunkhash]` in dev build

### DIFF
--- a/webpack_config/webpack.base.js
+++ b/webpack_config/webpack.base.js
@@ -31,12 +31,6 @@ module.exports = {
     loaders: [
       {
         test: /\.(js|jsx)$/,
-        enforce: 'pre',
-        loaders: ['eslint-loader'],
-        exclude: [/node_modules/]
-      },
-      {
-        test: /\.(js|jsx)$/,
         loaders: ['babel-loader'],
         exclude: [/node_modules\/(?!ethereum-blockies|idna-uts46)/]
       },
--- a/webpack_config/webpack.dev.js
+++ b/webpack_config/webpack.dev.js
@@ -5,7 +5,7 @@ const webpack = require('webpack');
 const base = require('./webpack.base');
 const FriendlyErrors = require('friendly-errors-webpack-plugin');
 
-base.devtool = 'source-map';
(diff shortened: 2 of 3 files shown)


HenryNguyen5:
This would break debugging in VSCode for me: https://github.com/Microsoft/vscode-chrome-debug-core/issues/83

wbobeirne:
Good to know; perhaps we can make this an environment variable or otherwise configurable?

HenryNguyen5:
Makes sense to me, probably default to slower `source-map` unless a environment variable like`BUILD_SPEED=FAST` is present

HenryNguyen5:
I found that by adding `query: { cacheDirectory: true }` to `query: { cacheDirectory: true }` giving 
```js 
      {
        test: /\.(js|jsx)$/,
        loaders: [
          {
            loader: 'babel-loader',
            query: { cacheDirectory: true }
          }
        ],
        exclude: [/node_modules\/(?!ethereum-blockies|idna-uts46)/]
      },
```
in `./webpack.base.js` gave another 72% build time improvement

wbobeirne:
👏 Hot damn. Maybe open up another PR with that?

HenryNguyen5:
Added DLL plugin to webpack too, im at 6 second build times with 600ms rebuild times

PR #190, opened 2017-09-12
Remove Un-needed isBroadcasting Wallet State

dternyak:
Remove `isBroadcasting` from top-level wallet state. `isBroadcasting` is kept at the BroadcastTransactionStatus level, with a different state for each transaction.

### DIFF
--- a/common/reducers/wallet.js
+++ b/common/reducers/wallet.js
@@ -25,7 +25,6 @@ export const INITIAL_STATE: State = {
   inst: null,
   balance: null,
   tokens: {},
-  isBroadcasting: false,
   transactions: []
 };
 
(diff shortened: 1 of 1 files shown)


common/reducers/wallet.js
```diff
@@ -25,7 +25,6 @@ export const INITIAL_STATE: State = {
   inst: null,
   balance: null,
   tokens: {},
-  isBroadcasting: false,
```
HenryNguyen5: Maybe its better to keep this, and set it to null. Just to keep the state consistent

dternyak: I'm not sure how that's better -- there is no `isBroadcasting` state for a wallet, only for a specific transaction. 

This is just leftover from when I was figuring out the state shape and forgot to remove it.

PR #195, opened 2017-09-14
Build speed v2

HenryNguyen5:
Significantly decrease build-times.

< 6 second cold-build times
< 700ms re-build times

Usage: `webpack --config=./webpack_config/webpack.dll.js` => build the vendor DLL, only need to rebuild if the vendor packages get updated

Then do your typical `npm run dev`

TODO: Need some guidance on how we want to automate the DLL building step, unless you're all cool with manually building it once and then running `npm run dev` normally.


dternyak:
I'm okay with the manually doing the DLL building step, just need to include it in the readme as well.

HenryNguyen5:
It could later be automated if we read from `package.json`, and then only recompile if the hash of `package.json` changes

dternyak:
Oh, that's really not a bad idea. Up to you on implementing that now or later.

wbobeirne:
Adding a `"postinstall"` script for this would simplify that.

README.md
```diff
@@ -3,6 +3,7 @@
 #### Run:
 
 ```bash
+webpack --config=./webpack_config/webpack.dll.js # Only do this if you havent built the .dll.js yet, or if vendor dependencies have changed/updated
```
wbobeirne: Add a "postinstall" script to package.json to force and simplify this

PR #196, opened 2017-09-14
Trezor DPath Fix

wbobeirne:
This fixes Trezor (And any deterministic wallet that extends that class) to use the index of the wallet as a part of the dpath. This was causing Trezor sends to fail.

In the process of debugging, I also trued up a few spots in the send process to better match v3. I'll drop notes linking to where they exist in v3.

### DIFF
--- a/common/libs/transaction.js
+++ b/common/libs/transaction.js
@@ -137,7 +137,7 @@ export async function generateCompleteTransactionFromRawTransaction(
     nonce: cleanHex(nonce),
     gasPrice: cleanHex(gasPrice.toString(16)),
     gasLimit: cleanHex(gasLimit.toString(16)),
-    to: cleanHex(to),
+    to: toChecksumAddress(cleanHex(to)),
     value: token ? '0x00' : cleanHex(value.toString(16)),
     data: data ? cleanHex(data) : '',
     chainId: chainId || 1
--- a/common/libs/transaction.js
+++ b/common/libs/transaction.js
@@ -4,7 +4,7 @@ import translate from 'translations';
 import { padToEven, addHexPrefix, toChecksumAddress } from 'ethereumjs-util';
 import { isValidETHAddress } from 'libs/validators';
 import ERC20 from 'libs/erc20';
-import { stripHex, valueToHex } from 'libs/values';
+import { stripHexPrefixAndLower, valueToHex } from 'libs/values';
 import { Wei, Ether, toTokenUnit } from 'libs/units';
 import { RPCNode } from 'libs/nodes';
 import { TransactionWithoutGas } from 'libs/messages';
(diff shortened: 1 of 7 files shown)


common/libs/transaction.js
```diff
@@ -137,7 +137,7 @@ export async function generateCompleteTransactionFromRawTransaction(
     nonce: cleanHex(nonce),
     gasPrice: cleanHex(gasPrice.toString(16)),
     gasLimit: cleanHex(gasLimit.toString(16)),
-    to: cleanHex(to),
+    to: toChecksumAddress(cleanHex(to)),
```
wbobeirne: I couldn't find _where_ in code this happens, but all of the transactions I generated showed the checksum'd version in v3, whereas v4 was allowing all lower / all caps versions into the raw tx.

common/libs/wallet/trezor.js
```diff
@@ -5,21 +5,25 @@ import Big from 'bignumber.js';
...
-        stripHex(tx.gasLimit.toString()),
-        stripHex(tx.to),
-        stripHex(tx.value),
-        stripHex(tx.data),
+        stripAndLower(tx.nonce),
+        stripAndLower(tx.gasPrice.toString()),
+        stripAndLower(tx.gasLimit.toString()),
+        stripAndLower(tx.to),
+        stripAndLower(tx.value),
+        stripAndLower(tx.data),
```
wbobeirne: https://github.com/kvhnuke/etherwallet/blob/b5e1b75beb7c393e06e5ffd7e540c6bccd9eface/app/scripts/uiFuncs.js#L46

common/libs/wallet/deterministic.js
```diff
@@ -4,17 +4,19 @@ import type { IWallet } from './IWallet';
 export default class DeterministicWallet implements IWallet {
   address: string;
   dPath: string;
+  index: number;
 
-  constructor(address: string, dPath: string) {
+  constructor(address: string, dPath: string, index: number) {
```
wbobeirne: I chose to make this a separate param, rather than sending it all in as the dPath, to make sure people don't make the same mistake I do!

common/libs/wallet/trezor.js
```diff
@@ -5,21 +5,25 @@ import Big from 'bignumber.js';
 import { addHexPrefix } from 'ethereumjs-util';
 import DeterministicWallet from './deterministic';
 import { stripHex } from 'libs/values';
-
 import type { RawTransaction } from 'libs/transaction';
 
+// Identical to ethFuncs.getNakedAddress
+function stripAndLower(str) {
```
dternyak: Any reason not to move into `libs/values`? Seems like this would be useful outside of just trezor.

dternyak: PrivKeyWallet actually implements it on the class:

```
export default class PrivKeyWallet implements IWallet {
...

  getNakedAddress(): Promise<string> {
    return new Promise(resolve => {
      this.getAddress().then(address => {
        resolve(stripHex(address));
      });
    });
  }
```

dternyak: We should probably figure out why that discrepancy exists (not that it needs to be fixed as a part of this PR)

common/libs/wallet/deterministic.js
dternyak: I think thats fine, whoever next works on this should be able to figure out dPath is really dPath - index in that case.

common/libs/wallet/trezor.js
HenryNguyen5: stripHex already lowercases the input 
```js
export function stripHex(address: string): string {
  return address.replace('0x', '').toLowerCase();
}
```

HenryNguyen5: If anything, the `stripHex` in `/libs/values.js` should be renamed to `stripHexAndLower` since thats what it actually does

PR #199, opened 2017-09-15
Add Pre-footer

dternyak:
Solves https://github.com/MyEtherWallet/MyEtherWallet/issues/200

screen shot 2017-09-14 at 5 54 46 pm

- create PreFooter components and styles
- add PreFooter to main Footer Component and adjust Footer styles

### DIFF
--- a/common/components/Footer/PreFooter.jsx
+++ b/common/components/Footer/PreFooter.jsx
@@ -1,0 +1,19 @@
+import React from 'react';
+import './PreFooter.scss';
+
+const PreFooter = () => {
+  return (
+    <section className="pre-footer">
+      <div className="container">
+        <p>
+          MyEtherWallet.com does not hold your keys for you. We cannot access
+          accounts, recover keys, reset passwords, nor reverse transactions.
+          Protect your keys &amp; always check that you are on correct URL.
+          <a href="#"> You are responsible for your security.</a>
+        </p>
+      </div>
+    </section>
+  );
+};
+
+export default PreFooter;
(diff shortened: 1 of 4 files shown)


common/components/Footer/index.jsx
```diff
@@ -94,102 +95,105 @@ const LINKS_SOCIAL = [
...
-            <span>
-              {translate('FOOTER_1')}
-            </span>
-            <span>
-              {translate('FOOTER_1b')}
-            </span>
-          </p>
+      <div>
+        <PreFooter />
+        <footer className="Footer" role="contentinfo" aria-label="footer">
```
dternyak: Most of this is noise that was introduced by an additional layer of nesting created by the top-level div.

common/components/Footer/PreFooter.jsx
```diff
@@ -0,0 +1,19 @@
...
+
+const PreFooter = () => {
+  return (
+    <section className="pre-footer">
+      <div className="container">
+        <p>
+          MyEtherWallet.com does not hold your keys for you. We cannot access
+          accounts, recover keys, reset passwords, nor reverse transactions.
+          Protect your keys &amp; always check that you are on correct URL.
+          <a href="#"> You are responsible for your security.</a>
```
dternyak: We don't yet have the disclaimer modal, but when we do, it will pop up when this is clicked.

On a separate note, I'm working on a action/side-effect based modal instead of component state based modal popups today.

wbobeirne: So it would be like notifications? A saga would fire some kind of modal action? Wouldn't the modal need sufficiently complicated args for content / buttons?

dternyak: Yes, very similar to notifications.

Initially this would be a similar confirmation dialog with only a title, body, cancel and accept buttons to minimize complexity.

PR #202, opened 2017-09-16
All v3 Nodes & Networks

wbobeirne:
## What This Does

* Adds all of v3's supported networks to the NETWORKS object
* Adds all of v3's nodes to the NODES object
* Implements Infura and Etherscan node libs that extend the RPC lib
* Colors the navigation border based on the current active network

I don't have any stake in, and am not _too_ familiar with alternative networks, so anyone who knows more about this and can test it better would be much appreciated. I just used v3 as a baseline, and made sure the requests didn't completely fail.

## What This Doesn't Do

* Custom nodes
* Rainbow dropdown selector
* ~Correctly update all components / redux states when you switch node / network~ <-- State is reset via page reload.

## Screenshot

screen shot 2017-09-16 at 9 44 31 am


common/libs/nodes/infura/index.js
```diff
@@ -0,0 +1,11 @@
+import RPCNode from '../rpc';
+import InfuraClient from './client';
+
+export default class InfuraNode extends RPCNode {
+  client: InfuraClient;
+
+  constructor(endpoint: string) {
+    super(endpoint);
+    this.client = new InfuraClient(endpoint);
```
HenryNguyen5: Shouldn't it work fine if you remove the constructor completely? Default constructor should take care of this I believe

wbobeirne: Default constructor loads the RPC client, I need the InfuraClient here.

HenryNguyen5: Ah you're right. My bad

common/libs/nodes/rpc/client.js
```diff
@@ -1,102 +1,33 @@
...
-}
+import type { RPCRequest, JsonRpcResponse } from './types';
 
 export default class RPCClient {
   endpoint: string;
   constructor(endpoint: string) {
     this.endpoint = endpoint;
   }
 
+  // FIXME is it safe to generate that much entropy?
```
wbobeirne: Considering removing this comment as this matches v3's implementation, and I don't know what this was intended for.

common/config/data.js
```diff
@@ -172,22 +176,90 @@ export type NodeConfig = {
   estimateGas: ?boolean
 };
 
+function makeEtherscanExplorer(url): BlockExplorerConfig {
```
dternyak: `makeEtherscanExplorer` seems like a nismoner. Perhaps `makeETHExplorer`?

wbobeirne: Well, it would only work with sites that have the same URL structure as Etherscan, so I didn't want anyone to throw in a totally different type of site and expect it to work. Though you're right that I do use it for gastracker.io, explorer.rsk.co, and gander.tec. They just all fortunately follow the same path spec.

Perhaps I'll rename, but add a comment to clarify.

dternyak: Oh, yeah, I was thinking this as well. It has to be a compatible url element path. A clarifying comment should be good 👍

dternyak:
@wbobeirne Confirmation always links to Ethereum main net.
screen shot 2017-09-19 at 11 27 32 am

PR #212, opened 2017-09-19
Command for HTTPS dev server

wbobeirne:
## What This Does

Ledger and Digital Bitbox use u2f-api, a library that only works on HTTPS servers. This PR adds the ability to run the dev server as HTTPS instead, by running `npm run dev:https`. Instructions for how to set that up are in the readme. The only caveat is I haven't been able to get Chrome to trust the certificate, but I don't think that matters if you confirm past the security warning.

screen shot 2017-09-19 at 2 24 10 pm

cc @skubakdj you'll need this for Ledger stuff


webpack_config/server.js
```diff
@@ -2,11 +2,14 @@
...
 const express = require('express');
 const webpack = require('webpack');
+const https = require('https');
+const fs = require('fs');
 const webpackConfig = require('./webpack.dev');
 const config = require('./config');
 const LogPlugin = require('./log-plugin');
 
 const app = express();
+let httpsApp;
```
wbobeirne: Not sure how this snuck in, but I'll remove it.

package.json
```diff
@@ -108,6 +108,7 @@
     "test": "jest --config=jest_config/jest.config.json --coverage",
     "pretest": "check-node-version --package",
     "dev": "node webpack_config/server.js",
+    "dev:https": "HTTPS=true node webpack_config/server.js",
```
skubakdj: I was able to launch the HTTPS server with an outdated version of node. Does a command like `predev:https` need to be added?

package.json
```diff
@@ -108,6 +108,7 @@
...
     "predev": "check-node-version --package",
```
skubakdj: What I'm referring to in the previous comment, for reference.

skubakdj: What I'm referring to in the previous comment, for reference.

package.json
wbobeirne: Good call, will do

PR #223, opened 2017-09-21
Wallet Decrypt - Ledger

skubakdj:
I know we're moving to TypeScript so this won't get merged as-is, but I thought I'd open a PR anyway to get some early feedback. Modeled heavily off of @wbobeirne's previous work with Trezor.

Also, don't let the addition count fool you. The bulk of that number comes from the static JS library files required for Ledger.

### What This Does
 - Adds Ledger support to Wallet Decrypt
 - Adds new `LedgerWallet` wallet class
 - Adds Ledger transaction and message signing
 - Includes minor bugfix for the Trezor component related to `dPath` selection

### What Still Need to Happen
 - Refactor to TypeScript
 - A bit more manual testing

### DIFF
--- a/common/components/WalletDecrypt/LedgerNano.jsx
+++ b/common/components/WalletDecrypt/LedgerNano.jsx
@@ -1,26 +1,144 @@
+// @flow
+import './LedgerNano.scss';
 import React, { Component } from 'react';
 import translate from 'translations';
+import DeterministicWalletsModal from './DeterministicWalletsModal';
+import LedgerWallet from 'libs/wallet/ledger';
+import Ledger3 from 'vendor/ledger3';
+import LedgerEth from 'vendor/ledger-eth';
+import u2f from 'vendor/u2f-api';
+import DPATHS from 'config/dpaths.js';
+
+const DEFAULT_PATH = DPATHS.LEDGER[0].value;
+
+type State = {
+  publicKey: string,
+  chainCode: string,
+  dPath: string,
+  error: ?string,
+  isLoading: boolean
+};
 
 export default class LedgerNanoSDecrypt extends Component {
+  props: { onUnlock: any => void };
+  state: State = {
+    publicKey: '',
+    chainCode: '',
+    dPath: DEFAULT_PATH,
+    error: null,
+    isLoading: false
+  };
+
+  _handlePathChange = (dPath: string) => {
+    this._handleConnect(dPath);
+  };
+
+  _handleConnect = (dPath: string = this.state.dPath) => {
+    this.setState({
+      isLoading: true,
+      error: null
+    });
+
+    const ledger = new Ledger3('w0w');
+    const ethApp = new LedgerEth(ledger);
+
+    ethApp.getAddress(
+      dPath,
+      (res, err) => {
+        if (err) {
+          err = err.errorCode ? u2f.getErrorByCode(err.errorCode) : err;
+        }
+
+        if (res) {
+          this.setState({
+            publicKey: res.publicKey,
+            chainCode: res.chainCode,
+            isLoading: false
+          });
+        } else {
+          this.setState({
+            error: err,
+            isLoading: false
+          });
+        }
+      },
+      false,
+      true
+    );
+  };
+
+  _handleCancel = () => {
+    this.setState({
+      publicKey: '',
+      chainCode: '',
+      dPath: DEFAULT_PATH
+    });
+  };
+
+  _handleUnlock = (address: string, index: number) => {
+    this.props.onUnlock(new LedgerWallet(address, this.state.dPath, index));
+  };
+
   render() {
+    const { dPath, publicKey, chainCode, error, isLoading } = this.state;
+    const showErr = error ? 'is-showing' : '';
+
     return (
-      <section className="col-md-4 col-sm-6">
-        <div id="selectedUploadKey">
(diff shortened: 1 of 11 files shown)


common/components/WalletDecrypt/LedgerNano.scss
```diff
@@ -0,0 +1,26 @@
+.LedgerDecrypt {
```
skubakdj: Styles ported from `Trezor.scss`. We could probably refactor to one shared file if that would make more sense.

common/index.jsx
```diff
@@ -11,6 +11,10 @@ import { Root } from 'components';
 import { Routing, history } from './routing';
 import { store } from './store';
 
+import u2f from './vendor/u2f-api';
+
+window.u2f = u2f;
```
skubakdj: Included because `ledger3.js` expects `u2f` to be available globally. We could import it directly into the library file, though `ledger3.js` is licensed under `Apache 2.0`. Not sure what the deal is there.

common/components/WalletDecrypt/LedgerNano.jsx
```diff
@@ -1,26 +1,145 @@
...
+    });
+
+    const ledger = new Ledger3('w0w');
+    const ethApp = new LedgerEth(ledger);
+
+    ethApp.getAddress(
+      dPath,
+      (res, err) => {
+        if (err) {
+          err = err.errorCode ? u2f.getErrorByCode(err.errorCode) : err;
```
skubakdj: May through this piece of logic into a library function. It's used a few times throughout the PR.

common/index.jsx
wbobeirne: tldr legal says it's kosher to edit: https://tldrlegal.com/license/apache-license-2.0-(apache-2.0)

I did the same to digital bitbox, https://github.com/MyEtherWallet/MyEtherWallet/blob/digital-bitbox/common/vendor/digital-bitbox-usb.js#L11

common/components/WalletDecrypt/LedgerNano.jsx
wbobeirne: Yep, digital bitbox looked similar, though it also had a DigitalBitboxEth.getError function. I added a static function to the DigitalBitbox class to the tune of decodeError or something like that.

common/components/WalletDecrypt/LedgerNano.scss
wbobeirne: Could be, Digital Bitbox also looks _pretty_ similar, though there are differences. Might just be best keeping them custom so that we don't have to test all 3 wallets every time we make a change.

common/libs/wallet/ledger.js
```diff
@@ -0,0 +1,93 @@
...
+
+          resolve(addHexPrefix(serializedTx));
+        }
+      );
+    });
+  }
+
+  // modeled after
+  // https://github.com/kvhnuke/etherwallet/blob/3f7ff809e5d02d7ea47db559adaca1c930025e24/app/scripts/controllers/signMsgCtrl.js#L53
+  signMessage(msg: string): Promise<string> {
```
skubakdj: Modeled after the Ledger message signing from MEWv3, though I noticed the input/output differ slightly from the private key implementation. We should probably standardize around one.

common/libs/wallet/ledger.js
```diff
@@ -0,0 +1,93 @@
...
+
+  // modeled after
+  // https://github.com/kvhnuke/etherwallet/blob/3f7ff809e5d02d7ea47db559adaca1c930025e24/app/scripts/uiFuncs.js#L58
+  signRawTransaction(rawTx: RawTransaction): Promise<string> {
+    return new Promise((resolve, reject) => {
+      const eTx = new EthTx(rawTx);
+
+      eTx.raw[6] = Buffer.from([rawTx.chainId]);
+      eTx.raw[7] = 0;
+      eTx.raw[8] = 0;
```
wbobeirne: I see this is the same as uiFuncs, but I think a better way to do this is:
```
const eTx = new EthTx({
  ...rawTx,
  v: Buffer.from([rawTx.chainId]),
  r: 0,
  s: 0
});
```
to make 6, 7, and 8 less mysterious.

common/vendor/ledger3.js
```diff
@@ -0,0 +1,82 @@
...
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ ********************************************************************************/
+
+'use strict';
+
```
wbobeirne: As discussed previously, drop a `var u2f = require('u2f');` in here.

common/libs/wallet/ledger.js
wbobeirne: Agreed, it's not even on the wallet interface.

common/libs/wallet/ledger.js
skubakdj: Yep, I like this a ton better! It's much more clear.

common/index.jsx
skubakdj: Well that's a super helpful website! :smile:  

Good to know the global can be avoided, I'll make the change.

PR #224, opened 2017-09-23
Typescript Refactor

HenryNguyen5:
Full code-base port to typescript.

In things such as redux-form, and RPC nodes, many `any` types were used until either unused functionality is stripped out in `redux-form`'s case, and proper inheritance in the node case.

CI still needs to be done for TSLint + Typescript compiler 

As code base matures, more strict ruling for TSLint will be enabled 

Things such as deep prop passing into components should later be using shared interfaces to make use of deduplication across the codebase.

I suggest pulling it down, doing manual testing. 

No development enviroment changes needed beyond wiping `node_modules` and running `npm install && npm run dev`.

Recommended editors:
VSCode + TSLint plugin
Webstorm + TSLint plugin enabled 

As far I know, Atom isn't very suitible considering VSCodes first class support for typescript projects. Definitely check VSCode out on the new codebase!


wbobeirne:
FWIW atom support is no slouch, https://atom.io/packages/atom-typescript. But I've been meaning to investigate vscode anyhow, so I'll be sure to check it out.

wbobeirne:
I tested generate, send, some of swap, language / node change, and messing with gas, and all seemed to work without a hitch. Couldn't find any (obvious) regressions, which in a 9k line diff, is very impressive!

I'll be honest, I'm a little concerned about the short term productivity loss that we'll probably have, but it's very clear to me that this is the responsible and future-proof way forward. Thanks a ton you two for hunkering down on this.

I'm not going to drop an "approve" until other folks have gotten a chance to vet this, and we'll probably want to figure out what to do with Travis here.

wbobeirne:
I'll be working on getting travis & derivation-checker back into shape.

PR #227, opened 2017-09-23
Typescript - Travis / Dev command fixes

wbobeirne:
A bunch of fixes for the `typescript` branch related to dev workflow stuff:

* Fix derivation-checker by compiling it with webpack, should be future-proof for any other compilation changes
* Remove the `flow` command, replace with `tslint`
* Tell travis to use aforementioned `tslint` command
* Add `tslint-react` package so that we can lint `tsx` files
* Run precommit on `ts,tsx` files, instead of `js,jsx` (Prettier command remains unchanged)

I left `tslint-react` on its defaults, so we may want to configure it a bit more to our liking.


wbobeirne:
^ It's worth noting Travis is failing because there are a bunch of tslint errors in the project currently. Check the logs for more details. We could either lighten up on these rules, or try to fix them all. But it's a lot.

wbobeirne:
Just pushed a fix for `npm run jest`. All but one test passes now, but I'm stumped on how to fix it. It's getting an error in the transpiled code, `__generator is not a function`.

PR #228, opened 2017-09-24
Update tr.js

WhiteSymmetry:


dternyak:
@WhiteSymmetry Thanks so much for your contribution. Do you mind re-submitting to https://github.com/kvhnuke/etherwallet? This is the V4 repo, and is not yet in production, so your translation updates unfortunately wouldn't be used for a while.

Thanks again!

PR #238, opened 2017-09-26
Wallet Decrypt - Ledger: TypeScript Edition

skubakdj:
I've converted my [previous Ledger PR](https://github.com/MyEtherWallet/MyEtherWallet/pull/223) to TypeScript using a fresh branch to avoid handling merge conflicts. I believe I've also addressed all of @wbobeirne's feedback on the original PR. 

Things look functional from my end, though I'd like to do some more testing before it's merged. 

### What This Does
 - Adds Ledger support to Wallet Decrypt
 - Adds new `LedgerWallet` wallet class
 - Adds Ledger transaction and message signing
 - Includes minor bugfix for the Trezor component related to `dPath` selection

#### To Test
`npm run dev:https`


skubakdj:
Hmm looks like there's some Travis errors to figure out. Also just merged the latest from `develop`, so another round of testing is definitely required. Feedback on the existing code is welcome, however.

common/components/WalletDecrypt/LedgerNano.tsx
```diff
@@ -1,29 +1,150 @@
...
+        }
+
+        if (res) {
+          this.setState({
+            publicKey: res.publicKey,
+            chainCode: res.chainCode,
+            isLoading: false
+          });
+        } else {
+          this.setState({
```
dternyak: could simplify multiple setState's

dternyak:
Looks good overall! Unfortunately I don't have a ledger to test with. 

I'll wait for an additional review before merging, and I'll also make a note to ask Ledger to provide typings for their vendor libs.

common/libs/wallet/deterministic.ts
```diff
@@ -1,7 +1,7 @@
 import { IWallet } from './IWallet';
```
wbobeirne: Are we using this at all? And if we're not, should we?

common/libs/wallet/deterministic.ts
```diff
@@ -1,7 +1,7 @@
...
 
 export default class DeterministicWallet {
-  private address: string;
+  public address: string;
```
wbobeirne: Address should remain private / protected, since we have getAddress.

common/vendor/ledger-eth.js
```diff
@@ -0,0 +1,284 @@
+/* prettier-ignore */
+/* eslint-ignore */
```
wbobeirne: We should try to get these in our configs if possible.

common/vendors.js
```diff
@@ -25,3 +25,6 @@ require('qrcode.react');
 require('bignumber.js');
 require('classnames');
 require('./vendor/trezor-connect');
+require('./vendor/u2f-api');
+require('./vendor/ledger3');
+require('./vendor/ledger-eth');
```
wbobeirne: I know these are called vendor, but should we be adding these here? They're in version control, so even if they change, we won't generate a new DLL. I think they should remain in the main bundle.

wbobeirne:
I was unable to get this to work running locally on my machine. I was getting error code 2, `BAD_REQUEST`:

screen shot 2017-09-29 at 2 24 53 pm

However, myetherwallet.com seems to be working fine. Any ideas of where I may have gone wrong?

wbobeirne:
Whoops, I totally forgot (Though you may want to add this to your notes) that to test this, **you must run `npm run dev:https`, not `npm start`**. Worked like a charm after that!

We probably want to improve those error messages to help out with that.

common/libs/wallet/deterministic.ts
skubakdj: Whoops, my bad. Will revert to `private` and use `getAddress()` instead.

common/vendor/ledger-eth.js
skubakdj: Linter config was already thrown into `package.json`, so this line is outdated and will be removed. Not yet sure about Prettier though...

common/libs/wallet/deterministic.ts
skubakdj: We're not using it in this file, so I'll remove the import. I don't think it's necessary here because it gets used further down in the inheritance chain (e.g. `LedgerWallet extends DeterministicWallet implements IWallet`).

common/vendors.js
skubakdj: Not sure, I threw them in because I saw `trezor-connect` was included, though I think you're right. @HenryNguyen5, could you confirm?

dternyak:
Tested, it works 👍 

I'll wait for Henry's comment on how to handle vendor files, and we should be good to merge after tsc errors are resolved.

common/vendors.js
HenryNguyen5: Yeah, `./vendor/*` should be removed for now, my bad.

skubakdj:
Just sat down to work through the new compile errors and looks like I was beaten to it. Thanks @HenryNguyen5 :smile:!

PR #243, opened 2017-09-29
Add disclaimer modal to footer

james-prado:
Closes https://github.com/MyEtherWallet/MyEtherWallet/issues/201

* Add Modal component to footer
* Use Modal to display disclaimer info
* Fix Modal css max-width creating skinny unreadable modals on mobile


common/components/Footer/index.tsx
```diff
@@ -92,7 +93,21 @@ const LINKS_SOCIAL = [
...
 export default class Footer extends Component {
+  constructor(props) {
+    super(props);
+
+    this.state = { isOpen: false };
+  }
+
+  public toggleModal = () => {
+    this.setState({ isOpen: !this.state.isOpen });
+    document.body.classList.toggle('no-scroll');
```
wbobeirne: Not sure if I'm mistaking the intent here, but this should already be happening inside of the modal component: https://github.com/MyEtherWallet/MyEtherWallet/blob/develop/common/components/ui/Modal.tsx#L38

common/components/ui/Modal.scss
```diff
@@ -1,132 +1,130 @@
...
-  100% {
-    opacity: 1;
-  }
+	0% {
+		opacity: 0;
+	}
+	70%,
+	100% {
+		opacity: 1;
+	}
```
wbobeirne: Looks like your spaces became tabs in here, mind fixing your editor to keep them the same?

common/components/ui/Modal.scss
```diff
@@ -1,132 +1,130 @@
...
-  overflow: hidden;
-  display: none;
-  flex-direction: column;
-  animation: modal-open $m-anim-speed ease 1;
-  text-align: left;
+	position: fixed;
+	top: 30px;
+	left: 50%;
+	width: 95%;
+	max-width: 850px;
```
wbobeirne: If you want to target `max-width: 850px`, that should be done for the content of your particular modal. The idea here is that a modal is an unopinionated container, and should be able to be wider than that if someone needs it to be.

common/sass/styles/overrides/forms.scss
```diff
@@ -1,86 +1,87 @@
...
 }
 
 .form-control {
-  margin-top: $space-sm;
-  margin-bottom: $space-sm;
-  transition: $transition;
+	margin-top: $space-sm;
+	margin-bottom: $space-sm;
+	transition: $transition;
+	z-index: auto !important;
```
wbobeirne: I'd like to avoid as many `!important`s as possible, what's this about?

common/components/Footer/index.tsx
```diff
@@ -92,7 +93,21 @@ const LINKS_SOCIAL = [
...
+  };
+
   public render() {
+    const buttons: IButton[] = [
```
wbobeirne: Is it necessary to type these? Or will typescript just know that the shape is 👌

common/sass/styles/overrides/forms.scss
james-prado: At some point when I was adding the modal, the z-index of the the `<input type='password'/>` was making it appear on top of the disclaimer. I'm not quite sure why this was occurring but thought it was a bit odd to have a z-index of 2 on all form inputs.

I just took a peek again and your right, its totally unnecessary. The modal has a z-index of 1050 anyway, not quite sure how this slipped past me as I was making the pr. I'll make sure it gets changed back.

common/components/ui/Modal.scss
james-prado: I'll fix it up so its particular to the disclaimer modal.

common/components/Footer/index.tsx
james-prado: Nope not necessary, still getting familiar with TypeScript, I'll update this.

common/components/Footer/index.tsx
james-prado: Yup, `document.body.classList.toggle('no-scroll');` definitely is already happening inside the modal component and can be removed. The other half of `toggleModal()` (line 103) is using the state of the footer component to store a boolean to determine if the modal is open or closed, similar to how the [MnemonicDecrypt](https://github.com/MyEtherWallet/MyEtherWallet/blob/develop/common/components/WalletDecrypt/Mnemonic.tsx#L72) uses `seed` as a boolean to determine if the modal should be open or closed.

common/sass/styles/overrides/forms.scss
james-prado: It was probably when I began writing a modal component before I realized one already existed in the project

common/components/Footer/index.scss
```diff
@@ -79,8 +79,26 @@
...
+      margin-right: -0.5rem;
     }
   }
 }
+
+.Modal {
+  color: #000;
+}
+
+.modal_open_button {
```
wbobeirne: Class names should follow the naming convention as outlined in the readme: https://github.com/MyEtherWallet/MyEtherWallet#styling

common/components/ui/Modal.scss
```diff
@@ -66,6 +65,11 @@ $m-anim-speed: 400ms;
   animation: modal-open $m-anim-speed ease 1;
   text-align: left;
 
+  &.is-small {
+    width: calc(100% - 40px);
+    max-width: 850px;
+  }
```
wbobeirne: This still isn't what I meant for how the modal works. The idea from the original implementation is that the modal's _content_ is what defines the width, and it just shrinks and grows to that. So the idea is that the Footer's modal content would be wrapped in some element that is `width: 850px;`. Since the content is `display: flex`, it'll shrink it when it gets below that.

It was an unfamiliar way of doing things to me when it was first proposed, but it's actually pretty elegant since you don't have to add things like this.

common/components/ui/Modal.scss
```diff
@@ -117,7 +121,6 @@ $m-anim-speed: 400ms;
   }
 
   &-footer {
-    height: $m-footer-height;
```
wbobeirne: It's possible this was here for a reason, was it causing you any problems?

common/components/ui/Modal.tsx
```diff
@@ -21,6 +21,7 @@ interface Props {
   disableButtons?: boolean;
   children: any;
   buttons: IButton[];
+  className?: string;
```
wbobeirne: Prefer explicit params (i.e. `size`) over free-form class names. These get unmaintainable really quick.

common/components/ui/Modal.scss
james-prado: nope, seems I just removed it by mistake.

common/components/Footer/index.scss
james-prado: I'll change it so it complies to suit css

common/components/ui/Modal.scss
james-prado: I'm more familiar with having set sizes for modals so they remain consistent across the site and adhere to design specs no matter the content. I'm fine with having the content define the modal, but main problem this css is trying to solve is responsiveness on mobile. With the previous implementation, it meant that you would get a skinny modal, about 50% the size of the screen (viewport width).

What about if I added a media query for mobile screens to adjust the width. Something like this

```
@media(max-width: 600px) {
    .modal {
          width: calc(100% - 40px);
     }
}
```

common/components/ui/Modal.tsx
james-prado: good point

common/components/ui/Modal.scss
james-prado: nvm, I double checked and it was causing a problem. Setting the height manually defeats the purpose of having standardized buttons that define the height of the modal footer. Additionally, setting the height this way reduces the height of the modal footer, making the padding mis-aligned

https://imgur.com/a/VdiHS

common/components/Footer/index.tsx
james-prado: Yes it's necessary. TD needs the shape of buttons defined for the modal.

common/components/ui/Modal.scss
wbobeirne: 👍

common/components/ui/Modal.scss
wbobeirne: That makes sense to me! Go ahead and add it. Just as an aside, you can put media queries inside of selectors to reduce repeated selectoring, like so:

```
.modal {
  width: auto;

  @media(max-width: 600px) {
    width: calc(100% - 40px);
  }
}
```

PR #244, opened 2017-09-30
Refactor Dropdowns, Rainbow Node Selector

wbobeirne:
## What This Does

* Provides a `DropdownShell` component for anything that uses dropdown-like behavior (Click button to open / close) (`Dropdown`, `GasPriceDropdown`, `ColorDropdown`)
* Provides a `ColorDropdown` component for the node selector that inserts colored bars and dividers for you.
* Adds customizability to dropdown size and color instead of CSS overrides
* Adds click-off to close dropdowns. No more multiple dropdowns, and clicking anything _but_ the dropdown closes it.

## Things I Don't Like / Want Opinions On

Making color dropdowns a separate component seemed like an obvious choice since most dropdowns don't use that at all. However, extending the existing `Dropdown` component seemed to complicate that, but making a new component ended up with duplicating a decent bit of code. Not sure how I feel about this.

What I really don't like is that, since I needed to add color metadata, options have to be an object of a particular shape in `ColorDropdown`. I actually _really_ like this, but I don't like the inconsistency between `Dropdown` and `ColorDropdown`. Before I go ahead and refactor, what do y'all think of sending along to `Dropdown`:
```
{
  name: 'String representation of object.',
  value: /* literally anything, probably still using the passed `T` for typescript */
}
```
This would get rid of a need for a formatTitle function prop (Since `name` would be pre-calculated) and would allow for future metadata on options (like `divider: true`, `active: true`, or `disabled: true` etc.)

## Gif

dropdowns


dternyak:
I'm okay with having 2 separate components, `ColorDropdown` and `Dropdown`. We can circle back at a later date if we sense that we are having too many similar components.

PR #241, opened 2017-09-27
Update React-Router(-DOM) to 4.x and Simplify Routing

james-prado:
Closes https://github.com/MyEtherWallet/MyEtherWallet/issues/231

* Replace react-router v3 -> React-router-dom v4
* Simplify Root component by merging it with the Routing component
* Add [history](https://github.com/ReactTraining/history) for creating react-router v4 compatible history objects

[react-router-dom](https://github.com/ReactTraining/react-router/tree/master/packages/react-router-dom) installs [react-router](https://github.com/ReactTraining/react-router/tree/master/packages/react-router) as a dependency, so only react-router-dom is required in package.json

[Migration Guide](https://github.com/ReactTraining/react-router/blob/master/packages/react-router/docs/guides/migrating.md)

PR #247, opened 2017-09-30
Fix regressions introduced in #241

james-prado:
Closes issue https://github.com/MyEtherWallet/MyEtherWallet/issues/246

* Update SendTransaction container props (no longer require location.query)
* Manually parse location.search string for queries, and convert to objects

Todo:
* pass location props too `<App/>` & rename it to `<TabSection/>`


common/containers/Tabs/SendTransaction/index.tsx
```diff
@@ -80,7 +80,7 @@ interface State {
 }
 
 interface Props {
-  location; //TODO: update type
+  location; //TODO: fixtype
```
james-prado: I've added types for [history](https://www.npmjs.com/package/@types/history), but I'm a bit confused how I'm actually supposed to use them here. Any help would be much appreciated!

common/containers/Tabs/SendTransaction/index.tsx
```diff
@@ -292,7 +288,20 @@ export class SendTransaction extends React.Component<Props, State> {
...
+            search
+              .substring(1)
+              .replace(/&/g, '","')
+              .replace(/=/g, '":"') +
+            '"}',
+          (key, val) => {
+            return key === '' ? val : decodeURIComponent(val);
+          }
+        )
+      : {};
```
wbobeirne: I get what this is doing, but it feels like black magic that's prone to throwing depending on what's in the search query, and doesn't allow for typing. What happened to URLSearchParams?

james-prado: Your right, it's a bit hackey. Since there will likely be other situations where parsing a query string in a url is useful, I'm writing a few util functions to fix that. I'll change this to WIP till i get that sorted.

common/components/Header/components/NavigationLink.tsx
```diff
@@ -10,12 +10,19 @@ interface Props {
...
   };
-  location: any;
 }
 
-export default class NavigationLink extends React.Component<Props, {}> {
+interface InjectedLocation extends Props {
+  location: { pathname: string };
+}
+
+class NavigationLink extends React.Component<Props, {}> {
```
dternyak: Couldn't we use `InjectedLocation` instead of Props here, and avoid creating the `injected` method?

common/containers/TabSection/index.tsx
```diff
@@ -27,9 +24,10 @@ interface Props {
   changeNode: TChangeNode;
   changeGasPrice: TChangeGasPrice;
 }
-class App extends Component<Props, {}> {
+class TabSection extends Component<Props, {}> {
   public render() {
     const {
+      location,
```
dternyak: Gonna pull this down to confirm, but I don't still don't see how TabSection is getting location passed to it

common/components/Header/components/NavigationLink.tsx
james-prado: yes we can, ill fix that

PR #263, opened 2017-10-04
Travis TypeScript Compiler Checking

dternyak:
fixes https://github.com/MyEtherWallet/MyEtherWallet/issues/258

- add typescript compiler check to travis
- fix existing typescript compiler errors


HenryNguyen5:
Nice, we should have much more confidence in static analysis now

Issue #79, opened 2017-07-25
Handle Javascript Disabled


dternyak:
Right now the site just renders its broken state without Javascript. We should show a message about enabling it.

Likewise, we'll probably want to make a page that tells people to upgrade. We should settle on what features a browser NEEDS to run MEW.

wbobeirne:
Should be straightforward with a `<noscript>` tag.

As for required features, I think the only hard requirement is local storage (Which damn near every browser has,) and a soft requirement of flexbox (A lot of messed up layouts otherwise.) But regardless, we should probably have a bottom line somewhere, if only to improve user experience / awareness of superior browsers and the potential security implications.

dternyak:
@wbobeirne I think this issue ties in with https://github.com/MyEtherWallet/MyEtherWallet/issues/210, in that we need a way to provide messaging to users who are on platforms we do not support.

wbobeirne:
Great, I'll add that to the bad browser check. I have a branch ready to go, just waiting to do some checking once we get browserstack sorted.

PR #274, opened 2017-10-08
No Javascript / Old Browser Blockers

wbobeirne:
## What This Does

Closes out #79:

* Adds a page blocker for if your browser is missing any of the following features:
  * const / let
  * Local storage
  * Object.assign
  * Flexbox support (Otherwise the site will look broken)
* Adds a page blocker if you have Javascript disabled via a noscript tag

The copy for either of these is totally up or editing, I just threw something together. Please advise @tayvano.

## How to Test

1. Run `npm run build`
2. Use `http-server` or something to serve up the `dist/` folder
3. Use out browserstack credentials (Found in slack) and load the page in an old browser to see the bad browser message
4. Load the page in Chrome, and disable javascript to see the no js message like so:

screen shot 2017-10-08 at 12 10 06 pm

## Janky Implementations

Because a few of the third party modules make it into our vendor bundle with a `const` here and a `let` there, I decided that to keep this as fail-proof as possible, I would do the bad browser check outside of the main code bundle. So you'll find it, along with all of the markup for these, at the bottom of `index.html`. This wasn't ideal, but after having the common or vendor bundles fail for a myriad of reasons, felt like a real gordian knot chop.

Also, because they're not served up via React, I couldn't make a nice common component for these kinds of page blockers. So instead I made a sass mixin, `@mixin cover-message` which applies the styles, but allows for customization and overrides.

Requires #273 to work due to needing to be tested with the `npm run build` version of this.

## Screenshots

screen shot 2017-10-08 at 11 58 51 am

screen shot 2017-10-08 at 12 01 05 pm


common/index.html
```diff
@@ -2,25 +2,131 @@
...
+        </a>
+      </div>
+    </div>
+  </div>
+
+  <script>
+    (function() {
+      var badBrowser = false;
+
+      var badBrowser = false;
```
dternyak: duplicate?

dternyak:
Windows 7 with IE 8 doesn't properly show the old browser text, but other than that I had no issues 👍 
screen shot 2017-10-10 at 2 45 38 pm

PR #286, opened 2017-10-12
added verification and editted the main signing functions a tad

VoR0220:
Signed-off-by: RJ Catalano <richard.catalano@itpeoplecorp.com>

### DIFF
--- a/common/libs/signing.ts
+++ b/common/libs/signing.ts
@@ -1,7 +1,14 @@
 import EthTx from 'ethereumjs-tx';
-import { ecsign, sha3 } from 'ethereumjs-util';
+import { ecsign, ecrecover, publicToAddress } from 'ethereumjs-util';
+import sha3 from 'solidity-sha3';
 import { RawTransaction } from 'libs/transaction';
 import { isValidRawTx } from 'libs/validators';
+
+export interface SignatureArgs {
+  r: Buffer;
+  s: Buffer;
+  v: number;
+}
 
 export function signRawTxWithPrivKey(
   privKey: Buffer,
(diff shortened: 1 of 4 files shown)


common/libs/signing.ts
```diff
@@ -1,8 +1,15 @@
...
-import { ecsign, sha3 } from 'ethereumjs-util';
+import { ecsign, ecrecover, publicToAddress } from 'ethereumjs-util';
+import sha3 from 'solidity-sha3';
 import { RawTransaction } from 'libs/transaction';
 import { isValidRawTx } from 'libs/validators';
 
+export interface SignatureArgs {
+  r: Buffer;
+  s: Buffer;
+  v: number;
```
HenryNguyen5: This should also be a buffer

common/libs/signing.ts
```diff
@@ -26,16 +33,23 @@ export function signMessageWithPrivKey(
   const fullMessage = msg + spacer + date;
   const hash = sha3(fullMessage);
   const signed = ecsign(hash, privKey);
-  const combined = Buffer.concat([
```
HenryNguyen5: Using https://github.com/MetaMask/eth-simple-keyring/blob/master/index.js#L61 as a reference, they also concentenate `r,s,v` and return the hex string and `0x` prefixed result. Why did this need to be changed?

common/libs/signing.ts
VoR0220: Should it? It appears that in ethereum-js it is returned as a number? Or am i wrong?

common/libs/signing.ts
VoR0220: Easier to handle as an object then having to reparse it from a hex string. Seemed redundant to do so.

dternyak:
We ended up implementing local / offline signing independently as part of our offline-send functionality.

Sorry we weren't able to get this in.

VoR0220:
Thats fair enough. Keep doing what you're doing. I use your common lib from alpha in a ton of my projects...y'all should really turn that into a development boilerplate of some kind...its really useful.

PR #292, opened 2017-10-14
Fix Missing Address in Paper Wallet

dternyak:
Closes https://github.com/MyEtherWallet/MyEtherWallet/issues/225

Fixes intermittent paper wallet bug by passing address explicitly instead of having to asynchronously request it at the component level.

While the `getAddress` method can be argued to be changed to synchronous, this was a less invasive fix. Additionally, I can still see some time in the future where requested an address may have to be done asynchronously due to external devices requiring users to approve requests for such information.


common/components/PrintableWallet/index.tsx
```diff
@@ -8,13 +8,33 @@ interface Props {
...
+
+const initialState = {
+  address: null,
+  privateKey: null
+};
+
 export default class PrintableWallet extends Component<Props, {}> {
+  public state: State = initialState;
+
+  public async componentDidMount() {
```
wbobeirne: I didn't know you could make componentDidMount async!

PR #300, opened 2017-10-17
Add SingularDTV derivation to mnemonic paths.

danbaragan:
SingularDTV used ConsenSys/eth-lightwallet's default derivation path of `m/0'/0'/0'` since spring of 2016, before `m/44'/60'/0'` getting a lot of traction. Many users have created SingularDTV lightwallets over the past 18 months, and it would help a lot if they could access their wallets directly from MEW without the need to enter the derivation path manually.


tayvano:
PS: this was already updated on v3 a few days ago

wbobeirne:
Code looks fine to me, but from a user perspective, is `SingularDTV` a label that everyone will get? Do we need to add ConsenSys or eth-lightwallet to the label to help other people?

wbobeirne:
@dternyak perhaps this should make it into #849 finally?

dternyak:
@wbobeirne Yes, thanks for the reminder. I'll make a note to include this in Part 2 of that PR.

dternyak:
Fixed in https://github.com/MyEtherWallet/MyEtherWallet/pull/849/commits/5c813528153359d57108cdd684dcd184d3233654

@danbaragan I hope you won't mind that I didn't include your commit - it was easier for me to manually re-use your dPath value.

PR #303, opened 2017-10-18
Wallet Decrypt - Web3 (MetaMask / Mist)

### LINKED ISSUE #220: Support MetaMask / Mist / other web3 wallets
More context for this can be found at https://github.com/kvhnuke/etherwallet/issues/238, but I think the gist of it is to check for `window.web3` and see if we can use that for a wallet.

Unlike that post though, I think we should provide that as an option amongst our other list, rather than use it immediately. I'm sure many folks (such as myself) have a metamask wallet _and_ a hardware or paper wallet that they want to access.

skubakdj:
### What

Implements `web3` as both a wallet and node provider. Closes #220. 

Uses a different transaction model than our other wallets since Web3 providers don't allow for transactions to be signed directly. Instead, a raw transaction is sent to `web3` via the `web3.eth.sendTransaction` method which confirms with the user, submits directly to the network, and returns a TX hash (so there's no signed TX for us to receive here). 

Here's some new web3-specific behaviors to check out: 

 - MetaMask/Mist is hidden as a node option from the user. 
 - On wallet unlock, node should switch to MetaMask & its current network selection. The node dropdown should become disabled. 
 - After MetaMask unlock, if user unlocks different wallet type, node dropdown should become enabled and switch to same network type as Web3 node.
 - Upon submitting a transaction, user should be warned if MetaMask network type has changed since wallet unlock. Transaction will be blocked until this is resolved. 
 - MetaMask/Mist wallet type is disabled in offline mode of `Send Ether & Tokens` tab.

### What Not
Transaction functionality has been added to the `Send Ether & Tokens` tab only. I believe similar changes would be required for the `ENS` and `Contracts` tabs. I can submit them in a separate PR. 

I have tested this implementation with MetaMask, but not Mist (though behavior should be the same). More testing also needs to be done on the `web3` node methods.


common/containers/Tabs/SendTransaction/index.tsx
```diff
@@ -433,6 +442,50 @@ export class SendTransaction extends React.Component<Props, State> {
...
+        bigGasLimit,
+        chainId,
+        transactionInput
+      );
+
+      this.props.showNotification(
+        'success',
+        <TransactionSucceeded
+          txHash={txHash}
+          blockExplorer={network.blockExplorer}
```
skubakdj: The props for this component error in VSCode, though Webpack compiles without throwing and things seem to run fine in the browser. AFAIK, it's the same setup in [the wallet saga](https://github.com/MyEtherWallet/MyEtherWallet/blob/5d4b36d453631a77513b6b55f5bc2732eeded97f/common/sagas/wallet.tsx#L165). Any idea what I'm missing, @HenryNguyen5 ?

common/containers/Tabs/SendTransaction/index.tsx
```diff
@@ -268,125 +272,144 @@ export class SendTransaction extends React.Component<Props, State> {
...
-                  />
-                  {(offline || forceOffline) && (
-                      <div>
-                        <NonceField
-                          value={nonce}
-                          onChange={this.onNonceChange}
-                          placeholder={'0'}
-                        />
+            {unlocked &&
+              !(offline || (forceOffline && isWeb3Wallet)) && (
```
skubakdj: I'm going to assume Prettier reformatted this JSX block. This bit of logic was the only thing I added.

common/actions/wallet/actionTypes.ts
```diff
@@ -17,6 +17,10 @@ export interface UnlockMnemonicAction {
   payload: MnemonicUnlockParams;
 }
 
+export interface UnlockWeb3Action {
+  type: 'WALLET_UNLOCK_WEB3';
```
HenryNguyen5: Not sure how it slipped but all the `type` values in this file should be using the `enum` from `constants.ts`

common/components/WalletDecrypt/Web3.tsx
```diff
@@ -0,0 +1,35 @@
...
+          <button
+            className="Web3Decrypt btn btn-primary btn-lg"
+            onClick={this.props.onUnlock}
+          >
+            {translate('ADD_MetaMask')}
+          </button>
+        </div>
+
+        <div>
+          <a
```
HenryNguyen5: Use `NewTabLink` component in `ui` instead of this anchor

common/config/data.ts
```diff
@@ -242,3 +244,44 @@ export const NODES: { [key: string]: NodeConfig } = {
     estimateGas: true
   }
 };
+
+export function initWeb3Node(): Promise<void> {
+  return new Promise((resolve, reject) => {
+    const { web3 } = window as any;
```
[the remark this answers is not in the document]
skubakdj: Unfortunately the typings are incomplete. `web3.eth.call` and `web3.eth.getTransactionCount` don't accept block parameters, and neither `web3.createBatch` nor `web3.eth.call.request` exist (there's likely more, but those were the only errors being thrown). 

Should we keep `web3` as `any` for this PR or can you recommend some TypeScript magic to make this work?

common/actions/wallet/actionTypes.ts
skubakdj: Converted `constants.ts` to the `enum` format and switched the wallet action files over.

common/config/data.ts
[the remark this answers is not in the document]
HenryNguyen5: Hm, we can either do module augmentation https://www.typescriptlang.org/docs/handbook/declaration-merging.html
or we could help upstream merge in typings.  If either of those prove to be too costly, I guess keeping it as `any` would be fine. Or, a super minimalistic web3 typings file can be hand made with only the methods we're currently using.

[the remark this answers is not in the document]
skubakdj: Went down the declaration merging route, but @HenryNguyen5 and I weren't able to get it working due to some nuances of the library. Will be implementing minimal typings for `web3` via #332, but keeping it as `any` for this PR.

common/libs/wallet/non-deterministic/web3.ts
```diff
@@ -0,0 +1,92 @@
...
+
+  public signRawTransaction(): Promise<string> {
+    return Promise.reject(
+      new Error('Web3 wallets cannot sign raw transactions.')
+    );
+  }
+
+  public signMessage(msg: string): Promise<string> {
+    return new Promise((resolve, reject) => {
+      const msgHex = Buffer.from(msg).toString('hex');
```
HenryNguyen5: you can use the method `bufferToHex` from ethereumjs-utils. It'll convert it to hex and add the prefix for you in one call

PR #304, opened 2017-10-18
Broadcast Tx

dternyak:
closes https://github.com/MyEtherWallet/MyEtherWallet/issues/293

### What?
Broadcast Transaction Tab

- refactor and simplify ConfirmationModal
- create BroadcastTx container and add to tab options.

##### Note: `/pushTx` used as path to mirror etherscan.io


common/Root.tsx
```diff
@@ -31,6 +32,7 @@ export default class Root extends Component<Props, {}> {
             <Route path="/send-transaction" component={SendTransaction} />
             <Route path="/contracts" component={Contracts} />
             <Route path="/ens" component={ENS} />
+            <Route path="/pushTx" component={BroadcastTx} />
```
HenryNguyen5: is there a reason for naming this `pushTx` instead of the format of `broadcast-tx` like the `send-transaction` route?

common/containers/Tabs/BroadcastTx/index.tsx
```diff
@@ -0,0 +1,135 @@
...
+};
+
+class BroadcastTx extends Component<Props, State> {
+  public state = initialState;
+
+  public ensureValidSignedTxInputOnUpdate() {
+    try {
+      const tx = new EthTx(this.state.signedTx);
+      getTransactionFields(tx);
+      if (this.state.disabled) {
```
HenryNguyen5: Redundant `if` check

common/containers/Tabs/BroadcastTx/index.tsx
```diff
@@ -0,0 +1,135 @@
...
+        this.setState({ disabled: false });
+      }
+    } catch (e) {
+      if (!this.state.disabled) {
```
HenryNguyen5: ditto

common/containers/Tabs/BroadcastTx/index.tsx
```diff
@@ -0,0 +1,135 @@
...
+  public state = initialState;
+
+  public ensureValidSignedTxInputOnUpdate() {
+    try {
+      const tx = new EthTx(this.state.signedTx);
+      getTransactionFields(tx);
+      if (this.state.disabled) {
+        this.setState({ disabled: false });
+      }
+    } catch (e) {
```
HenryNguyen5: TS lets you do `} catch { ..rest of code here` if you don't do anything with the error param

common/containers/Tabs/BroadcastTx/index.tsx
```diff
@@ -0,0 +1,135 @@
...
+    this.ensureValidSignedTxInputOnUpdate();
+  }
+
+  public render() {
+    const { signedTx, disabled } = this.state;
+    return (
+      <TabSection>
+        <div className="text-center">
+          <div className="Tab-content-pane row block">
+            <div className="col-md-6">
```
HenryNguyen5: Can these `div`'s be compressed into one `div`?

common/containers/Tabs/BroadcastTx/index.tsx
```diff
@@ -0,0 +1,135 @@
...
+            <div className="col-md-6" style={{ marginTop: '70px' }}>
+              <div
+                className="qr-code text-center"
+                style={{
+                  maxWidth: '15rem',
+                  margin: '1rem auto',
+                  width: '100%'
+                }}
+              >
+                {this.state.signedTx && <QRCode data={this.state.signedTx} />}
```
HenryNguyen5: Use the destructured `signedTx` from the beginning of the render block

common/containers/Tabs/BroadcastTx/index.tsx
```diff
@@ -0,0 +1,135 @@
...
+              </div>
+            </div>
+          </div>
+        </div>
+        {this.state.showConfirmationModal && (
+          <ConfirmationModal
+            signedTx={this.state.signedTx}
```
HenryNguyen5: ditto

common/containers/Tabs/BroadcastTx/index.tsx
```diff
@@ -0,0 +1,135 @@
...
+              <div className="col-md-12 BroadcastTx-title">
+                <h2>Broadcast Signed Transaction</h2>
+              </div>
+              <p>
+                Paste a signed transaction and press the "SEND TRANSACTION"
+                button.
+              </p>
+              <label>{translateRaw('SEND_signed')}</label>
+              <textarea
+                className={`form-control
```
HenryNguyen5: Using the `classnames` lib we have would clean this up

common/libs/values.ts
```diff
@@ -14,3 +14,15 @@ export function valueToHex(value: Ether): string {
   // Finally, hex it up!
   return `0x${wei.toString(16)}`;
 }
+
+export function padLeftEven(hex) {
```
HenryNguyen5: Add typing here

common/libs/values.ts
```diff
@@ -14,3 +14,15 @@ export function valueToHex(value: Ether): string {
...
+  return hex.length % 2 !== 0 ? '0' + hex : hex;
```
HenryNguyen5: Use string template literals to eliminate interpolation

common/libs/values.ts
```diff
@@ -14,3 +14,15 @@ export function valueToHex(value: Ether): string {
...
+}
+
+export function sanitizeHex(hex) {
```
HenryNguyen5: Add typing

common/libs/values.ts
```diff
@@ -14,3 +14,15 @@ export function valueToHex(value: Ether): string {
...
+  hex = hex.substring(0, 2) === '0x' ? hex.substring(2) : hex;
```
HenryNguyen5: Should probably re-name this and make it a constant, mutating parameters isnt a good idea

common/libs/values.ts
```diff
@@ -14,3 +14,15 @@ export function valueToHex(value: Ether): string {
...
+  if (hex === '') {
+    return '';
+  }
+  return '0x' + padLeftEven(hex);
```
HenryNguyen5: Use string literal templates here to eliminate interpolation

common/Root.tsx
dternyak: yes, to match etherscan.

common/libs/values.ts
dternyak: This is all lifted from V3, I but I agree with you. I'll make these changes and note that we're diverging from V3.

dternyak: On second thought, I'm going to leave a TODO and circle back to this once we have a more comprehensive test suite.

PR #305, opened 2017-10-19
Animate transaction notifications & fix styles

james-prado:
Fixes https://github.com/myetherwallet/myetherwallet/issues/270

* Add [react-transition-group](https://github.com/reactjs/react-transition-group) for animating components before dismounting them
* Add id's based on txHash to successful transaction notifications obj's for use as [keys](https://reactjs.org/docs/lists-and-keys.html#keys) when animating notifications in and out.
* Fix white text when verifying transaction


package.json
```diff
@@ -20,6 +20,7 @@
     "font-awesome": "^4.7.0",
     "hdkey": "^0.7.1",
     "idna-uts46": "^1.1.0",
+    "lodash": "^4.17.4",
```
james-prado: Lodash was removed by mistake, this commit reverts that.

spec/containers/utils/shallowWithStore.ts
```diff
@@ -0,0 +1,10 @@
+import { shallow } from 'enzyme';
```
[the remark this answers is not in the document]
james-prado: Yup, must have forgotten about it in a git stash and ended up in the commit. I'll remove this.

common/sass/styles/overrides/buttons.scss
```diff
@@ -1,6 +1,6 @@
 // Button overrides
-@import "common/sass/variables";
```
dternyak: Most of these changes are formatting noise.

common/containers/Tabs/SendTransaction/index.tsx
```diff
@@ -307,14 +307,14 @@ export class SendTransaction extends React.Component<Props, State> {
...
                   {(offline || forceOffline) && (
-                    <div>
-                      <NonceField
-                        value={nonce}
-                        onChange={this.onNonceChange}
-                        placeholder={'0'}
-                      />
-                    </div>
-                  )}
+                      <div>
```
dternyak: I see no changes here except for formatting.

james-prado: I just added a new commit to clean up the rest of the formatting noise, the one thing I couldn't change was the indentation you highlighted, it gets formatted that way by the `precommit` script in `package.json`.

common/containers/TabSection/Notifications.scss
```diff
@@ -142,3 +142,23 @@
...
+.animation-exit {
+  opacity: 1;
+  transform: none;
+}
+
+.animation-exit.animation-exit-active {
+  opacity: 0.01;
+  transform: translateY(100%);
+  transition: opacity 500ms, transform 500ms;
+}
```
wbobeirne: Aaah, the ever familiar class naming of `react-transition-group`. Typically with these, I'm a big fan of:

```
.transitionname {
  &-enter {
    // ...
    &-active {
      // ...
    }
  }

  &-exit {
    // ...
    &-active {
      // ...
    }
  }
}
```

but this is completely stylistic

common/containers/TabSection/Notifications.tsx
```diff
@@ -5,28 +5,38 @@ import {
...
 
 interface Props {
   notifications: Notification[];
   closeNotification: TCloseNotification;
 }
+
+const Transition = props => (
+  <CSSTransition
+    {...props}
+    classNames="animation"
```
wbobeirne: I would pick a more descriptive class than `animation`, this is likely to collide with some other usage of this elsewhere. Maybe `Notification` or something specific like that?

common/containers/TabSection/Notifications.tsx
```diff
@@ -5,28 +5,38 @@ import {
...
+          return (
+            <Transition key={n.id}>
+              <NotificationRow
+                notification={n}
+                onClose={this.props.closeNotification}
+              />
+            </Transition>
+          );
+        })}
+      </TransitionGroup>
```
wbobeirne: I haven't looked at the latest `react-transition-group` in a while, but I'm pretty sure having both of these transition components is overkill. I think all you want is `CSSTransition`. Their [example of CSSTransition](https://codesandbox.io/s/yv645oq21x) doesn't use `TransitionGroup` at all.

common/sagas/wallet.tsx
```diff
@@ -168,7 +168,8 @@ function* broadcastTx(action: BroadcastTxRequestedAction): SagaIterator {
           txHash={txHash}
           blockExplorer={network.blockExplorer}
         />,
-        0
+        0,
+        txHash
```
wbobeirne: Rather than refactor all notifications to have manually specified IDs, what if we generated random IDs for each of them? You could use `Date.now()` or `Math.random()`

common/containers/TabSection/Notifications.scss
james-prado: I think your right here, it is stylistic but its good to maintain consistence. Still getting used to writing in suitcss. I'll update the css.

common/containers/TabSection/Notifications.tsx
james-prado: [TransitionGroup](https://reactcommunity.org/react-transition-group/#TransitionGroup) is used to manage multiple `CSSTransition` or `Transition` elements. It's useful when animating items in a list, like a list of notifications. For animating individual elements, like in their [example of CSSTransition](https://codesandbox.io/s/yv645oq21x), `TransitionGroup` isn't needed, but when you want to animate a list of transitions it is.

common/sagas/wallet.tsx
james-prado: Using `Math.random()` would be better for unique Id's like react keys, I used txHash because it was unique and already a property on the object.

common/containers/TabSection/Notifications.scss
james-prado: Actually, looking at the style guide again, the css should be fine as is. I'm not styling children of the component so this wouldn't be correct
```
.transitionname {
  &-enter {
    // ...
    &-active {
      // ...
    }
  }
...
```
react-transition-group uses classes like `.transitionname-enter`& `.transitionname-enter-active`, to represent the [state of the component](https://github.com/suitcss/suit/blob/master/doc/naming-conventions.md#componentnameis-stateofcomponent). React-transition-group doesn't adhere to the suitcss `.is-stateofcomponent` naming scheme, so sticking with the [react-transition-group naming scheme](https://reactcommunity.org/react-transition-group/#CSSTransition-prop-classNames) seems to be the only option 
```
.transitionname-enter {
     // ...
}

.transitionname-enter.transitionname-enter-active {
     // ...
}
```

common/sagas/wallet.tsx
```diff
@@ -168,7 +168,8 @@ function* broadcastTx(action: BroadcastTxRequestedAction): SagaIterator {
           txHash={txHash}
           blockExplorer={network.blockExplorer}
         />,
-        0
+        0,
+        Math.random()
```
wbobeirne: My aim here wasn't to replace the id here with `Math.random()`, it was to assign the id to `Math.random()` in the `showNotification` function. Otherwise, _every_ call to `showNotification` will have to remember to add a 4th arg of `Math.random()`.

PR #309, opened 2017-10-21
EthereumJS-Wallet Prep (Definition and Package)

HenryNguyen5:
Adds typescript typings for ethereumjs-wallet, ethereumjs-wallet/hdkey and ethereumjs-wallet/provider-engine. This is to be done before the usage of the module to replace our existing `keystore` and `decrypt` libraries.

Could use a thorough review on `hdkey` module, especially `derivePath` and `deriveChild` and on the `V1/V3/EtherWallet` interfaces

Input types e.g. `Buffer` or `string` were derived from the source code of the library itself and its dependencies, so they should match well.

The `provider-engine` submodule is missing its `opts` typings since that would lead to typing up the entirety of the option hierarchy of `MetaMask/provider-engine`


danfinlay:
Hi Henry, replying here b/c you asked for my input.

The ethereumjs-wallet/provider-engine is perfect for when you want your ethereumjs-wallet to auto-sign transactions sent to `sendTransaction()`. If you want to implement custom authorization logic (like a confirmation modal, like MetaMask does), you'll probably want to look at  using the [hooked wallet subprovider](https://github.com/MetaMask/provider-engine/blob/master/subproviders/hooked-wallet.js) with [provider-engine, as seen in its readme](https://github.com/MetaMask/provider-engine), although if you want to get ahead of the curve, you can forego provider-engine entirely (we are currently in the process), and use [json-rpc-engine](https://github.com/kumavis/json-rpc-engine), which is a more general purpose rpc middleware, that allowed @kumavis to factor out some of the ethereum-state-dependent parts of the code (like block tracking).

Issue #299, opened 2017-10-17
Change crypto libs to use ethereumjs-wallet


HenryNguyen5:
https://github.com/ethereumjs/ethereumjs-wallet

The `keystore.ts` and `decrypt.ts` internal libraries can be mostly if not entirely replaced using `ethereumjs-wallet`. This would greatly improve maintainability aswell as security since `ethereumjs-wallet` already has its own full suite of tests and is being also used by Metamask.

PR #316, opened 2017-10-23
EthereumJS-Wallet (Part 3)

HenryNguyen5:
This PR builds on top of #310.
Addresses #299 

What this PR does: 

- Removes `uuid` and `scryptsy` dependencies
- Removes `keystore.ts` test as file does not exist anymore
- Removes `keystore.ts` lib
- Strip out wallet selection based on file type from `sagas/wallet` and put it in `lib/wallet`
- Partition wallets in `lib/wallet` into `deterministic` and `non-deterministic`
- Uses a wrapper/injection function to add `signRawTransaction` and `signMessage` functionality to `ethereumjs-wallet` wallets.
- All non-determinstic based wallets now use a wrapped version of `ethereumjs-wallet` wallets (with the exception of `EncryptedPrivateKeyWallet` and `MnemonicWallet`
- Change IWallet `getAddress` to `getAddressString` to conform to `ethereumjs-wallet` method (`getAddress` returns a buffer in `ethereumjs-wallet` so you need to use the string post-fixed version)
- Change the deterministic wallet class method `getAddress` to `getAddressString`

Testing:
Private key based testing was done for sending transactions and broadcasting them too. I don't have a hardware wallet to test with.

Up for discussion:
Should we remove support for `EncryptedPrivateKeyWallet`? The `decrypt.ts` lib only exists for its functionality now.


common/libs/keystore.ts
```diff
@@ -1,234 +0,0 @@
...
-
-// adapted from https://github.com/kvhnuke/etherwallet/blob/de536ffebb4f2d1af892a32697e89d1a0d906b01/app/scripts/myetherwallet.js#L342
-export function determineKeystoreType(file: string): string {
-  const parsed = JSON.parse(file);
-
-  if (parsed.encseed) {
-    return 'presale';
-  } else if (parsed.Crypto || parsed.crypto) {
-    return 'v2-v3-utc';
-  } else if (parsed.hash && parsed.locked === true) {
```
skubakdj: The piece that identified an encrypted MEWv1 encrypted keystore file, for reference.

common/libs/wallet/IWallet.ts
```diff
@@ -1,6 +1,6 @@
 import { RawTransaction } from 'libs/transaction';
 
 export interface IWallet {
-  getAddress(): Promise<string>;
-  signRawTransaction(tx: RawTransaction): Promise<string>;
+  getAddressString(): Promise<string> | string;
```
wbobeirne: Any reason for the name change? Is address ever expected to be a non-string?

Also, how is the ambiguity of promise or string handled?

common/libs/keystore.ts
```diff
@@ -1,234 +0,0 @@
...
-    derivedKey.slice(0, 16),
-    new Buffer(kstore.crypto.cipherparams.iv, 'hex')
-  );
-  let seed = decipherBuffer(decipher, ciphertext);
-  while (seed.length < 32) {
-    const nullBuff = new Buffer([0x00]);
-    seed = Buffer.concat([nullBuff, seed]);
-  }
-  return seed;
-}
```
wbobeirne: Love to see this stuff going in favor of a common use module 👌

common/libs/wallet/deterministic/mnemonic.ts
```diff
@@ -0,0 +1,9 @@
+import { decryptMnemonicToPrivKey } from 'libs/decrypt';
+import { fromPrivateKey } from 'ethereumjs-wallet';
+
+export const MnemonicWallet = (
+  phrase: string,
+  pass: string,
+  path: string,
+  address: string
+) => fromPrivateKey(decryptMnemonicToPrivKey(phrase, pass, path, address));
```
wbobeirne: 🙌 love how simple this stuff gets with ethereumjs-wallet.

common/libs/wallet/IWallet.ts
wbobeirne: Other than reducing the amount of rewriting in one pr, is there any reason we need to keep these as promiseable anyway?

wbobeirne:
Sorry, the above review was supposed to be "Comment", not "Approve". I didn't get a chance to run this locally yet.

common/libs/keystore.ts
HenryNguyen5: @skubakdj 
scryptsy: Could you try again with a fresh `npm install`? I'm not getting any build errors.
mnemonic wallet: That would be because i forgot to wrap this class :D 
mew wallet: ill have to check this out

common/libs/wallet/IWallet.ts
HenryNguyen5: getAddress (in the ethereumjs-wallet package API) returns a buffer in ethereumjs-wallet so you need to use the string post-fixed version)

HenryNguyen5: We keep them as promisifiable because deterministic wallets are actually async vs the non-deterministic ones

common/libs/keystore.ts
HenryNguyen5: MewV1 wallet should be fixed now too. Was importing from wrong module.

skubakdj: The scryptsy, mnemonic, and MEWv1 encrypted wallet bugs look to be addressed :+1: 

Though I just noticed that unencrypted MEWv1 keystores will prompt for a password in the keystore decrypt component. Seems to be an issue with the `isKeystorePassRequired` function.

common/libs/wallet/non-deterministic/helpers.ts
```diff
@@ -0,0 +1,90 @@
...
+  } else if (parsed.hash && parsed.locked === false) {
+    return KeystoreTypes.v1Unencrypted;
+  } else if (parsed.publisher === 'MyEtherWallet') {
+    return KeystoreTypes.v2Unencrypted;
+  } else {
+    throw new Error('Invalid keystore');
+  }
+}
+
+const isKeystorePassRequired = (file: string): boolean =>
```
skubakdj: This will always return true.

common/libs/wallet/non-deterministic/helpers.ts
```diff
@@ -0,0 +1,90 @@
...
+  determineKeystoreType(file) === KeystoreTypes.presale ||
+  KeystoreTypes.v1Encrypted ||
```
skubakdj: Won't this always return true?

HenryNguyen5: It should return false on any unencrypted wallet

HenryNguyen5: Ah you're right, I should be checking it against the return value of `determineKeyStoreType` on each union

common/libs/keystore.ts
HenryNguyen5: This should be addressed now too

skubakdj: Yep, looks good!

PR #319, opened 2017-10-25
Replace bignumber.js with bn.js

HenryNguyen5:
## Replaces `bignumber.js` with `bn.js`

Addresses #324 

- [x] Handle decimals properly as `bn.js` does not handle decimals, probably do this via react component
- [x] Test functionality that has `bn.js` as a dependency

- Adds some additional typing to DeterministicWallets actions / sagas
- Introduces two semantic units, `Wei` and `TokenValue` that are an alias to BN
- Slight refactor on SendTransaction tab -- Amount Field interaction so a single handler is not responsible for 3 different tasks (sending all, ether, tokens)
- SendTransaction amount field can now be swapped token types on the fly without clearing the input box. 
- Use React components for the display and conversion of units. More below.

## UnitDisplay Base unit --> Display Unit
- Takes in parameters for converting a base unit (ether or token) to a display unit via decimals or ethereum unit type
- Also allows for formatting of numbers 

## UnitConverter Display Unit (User input) --> Base Unit
- Is a render callback that takes a function with a `onUserInput` and `convertedUnit` object variable
- Takes an `onChange` handler as a prop, along with decimals for the conversion to Display -> Base

This way whatever the UnitConverter encloses as children keep their display state (Ex. Ether), and the onChange handler bound to the internal code will be called with the base unit (Ex. Wei). This way we can seperate our base units from display units, with React components handling all of the display part.
Thus, internally, we only need to reason about base units. 

## Possible discussion points 
- Do we want to later on keep everything internally in base 16 also? It would be helpful to know that in the codebase only one base is being used. I've looked through the `ethereueumjs` based libraries, and they all handle `BN` or `string (base 10 / 16 depending if hex prefix exists)`.
- If the above is agreed upon, the only thing that would need to be casted to a string internally would be `Trezor` and possibly `Ledger` wallet files, as I know that `Trezor` requires a prefixless hex string send to its library as it does not handle `bn.js`


common/libs/units.ts
```diff
@@ -1,105 +1,28 @@
...
-  public toPrecision(precision?: number) {
-    return this.amount.toPrecision(precision);
-  }
-
-  public toWei(): Wei {
-    return new Wei(toWei(this.amount, this.unit));
-  }
+export type Wei = BN;
+export const Wei = (input: number | string | BN, base: number = 10): Wei =>
+  new BN(input, base);
```
wbobeirne: There will very likely be spots where decimals are coming in here. I think it'll be hard to eyeball, so we should probably pop an alert in here if it's a number or string with a `.` in it, and try to cover as much ground as we can.

spec/libs/units.spec.ts
```diff
@@ -1,49 +1,97 @@
...
+import {
+  fromWei,
+  Wei,
+  toWei,
+  toTokenBase,
+  fromTokenBase,
+  getDecimal,
+  TokenValue
+} from 'libs/units';
+const Units = {
```
dternyak: newline after imports

common/components/BalanceSidebar/AccountInfo.tsx
```diff
@@ -115,4 +115,9 @@ export default class AccountInfo extends React.Component<Props, State> {
       </div>
     );
   }
+  private getEthereumValue = (value: Wei, etherUnit: UnitKey) => (
+    formatNumber?: boolean
+  ) => (
+    <UnitDisplay value={value} unit={etherUnit} formatNumber={formatNumber} />
+  );
```
wbobeirne: This seems like it would be simpler if you just bassed `showLongBalance` as the third arg rather than returned a function that you then pass it into.

HenryNguyen5: Error was uncaught before because the function parameter was untyped, maybe we should look into setting `--strict` or at least `--noImplictAny` sometime soon

dternyak:
I'm in favor of disallowing implicit any's.

common/components/BalanceSidebar/AccountInfo.tsx
HenryNguyen5: Should be fixed now

common/utils/formatters.ts
```diff
@@ -8,9 +8,38 @@ export function combineAndUpper(...args: string[]) {
   return args.reduce((acc, item) => acc.concat(item.toUpperCase()), '');
 }
 
+const toFixed = (num: string, digits: number = 3) => {
```
dternyak: Could use a test.

HenryNguyen5: It's tested in `formatNumber` unless you wanted a seperate test for it

wbobeirne:
Figured out the above, it's because I was using Etherscan's node that sends back base 10 wei:
```
{"status":"1","message":"OK","result":"38225415000000000000"}
```
Switching to MEW's node causes it to send back base 16 wei:
```
{"jsonrpc":"2.0","id":"7d2a10f669cd14f761fa3f6ba1836b00","result":"0x2127bf46a62e37000"}
```
so it looks like we'll need some special handling for that

common/components/BalanceSidebar/AccountInfo.tsx
```diff
@@ -80,38 +79,40 @@ export default class AccountInfo extends React.Component<Props, State> {
...
-                  ? balance ? balance.toString() : '???'
-                  : balance ? formatNumber(balance.amount) : '???'}
+                <UnitDisplay
+                  value={balance}
+                  unit={'ether'}
+                  displayShortBalance={!showLongBalance}
+                />
               </span>
-              {` ${network.name}`}
+              {network.name}
```
skubakdj: Might want to have a spacer before this so balance reads as `??? ETH` instead of `???ETH`.

wbobeirne:
Also looks to me like everything's fixed up with the latest commits. Thanks Henry!

Issue #89, opened 2017-07-25
Node Switcher


dternyak:
- [x] ETH Nodes
- [x] ETC Nodes
- [x] Ropsten
- [x] Kovan
- [x] Rinkeby
- [x] Custom Node
- [x] Current Node Saved to Local Storage
- [x] Custom Nodes Saved to Local Storage
- [x] Basic Auth user/pw?
- [x] Node determines Chain ID

Github Ref

1. Issue: Kovan / Rinkeby Custom Nodes kvhnuke/etherwallet: Issue #495
2. Initial Feature: kvhnuke/etherwallet: Issue #61

PR #322, opened 2017-10-26
Custom Nodes

wbobeirne:
This addresses the last 3 pieces of #89. I'm not completely thrilled with my implementation so far, so I wanted to put this PR up while I'm still dotting some i's and crossing some t's. I'm looking for feedback on two things primarily:

1) Is the way that I'm putting the node config in the redux store at all confusing / cumbersome, or is it an improvement over `NODES[store.config.nodeSelection]`?
2) Is the separation of custom node configs and hard-coded node configs too annoying, and is there a better way to handle it?

## What It Does

* Makes the "Add Custom Node" button functional
* Provides a modal where users can specify a name, url, port, network type, and http auth for a custom node
* Saves and loads custom nodes in local storage
* Implements latest block in footer as a smoke test for the node
  * If the node fails to fetch the latest block, it reverts to the previous node
* Shows a loading state in the dropdown while switching between nodes

## What It Doesn't Do

* Custom networks. It was already a pain to fix up all of the code that was selecting the node from the `NODES` object, the same would have to be done for `NETWORKS`. This seemed like a lower-priority use case, for now. Should be very doable.
* Handling custom node collisions. The "id" of the custom node is url:port, so if you tried to add two of the same, you'd have a problem. I'll be sure to handle this.
* Unit tests. At least a few will make it in before I try to merge this.
* Manual testing of common custom networks. I need to do more of this before this is ready to merge.

## Screenshots

screen shot 2017-10-26 at 11 42 04 am

screen shot 2017-10-26 at 11 42 13 am


common/store.ts
```diff
@@ -56,10 +59,30 @@ const configureStore = () => {
...
+      savedConfigState.customNodes,
+    );
+    // If we couldn't find it, revert to defaults
+    if (savedNode) {
+      savedConfigState.node = savedNode;
+    }
+    else {
+      savedConfigState.nodeSelection = configInitialState.nodeSelection;
+    }
+  }
```
wbobeirne: This function's getting to be a bit of a doozy, maybe we could formalize loading from local storage into each reducer.

common/components/ui/Modal.tsx
```diff
@@ -3,7 +3,7 @@ import React, { Component } from 'react';
 import './Modal.scss';
 
 export interface IButton {
-  text: string;
+  text: any;
```
dternyak: "Any" (😆) specific reason for this change?

wbobeirne: In order to take in `translate` output, it needed to take in React elements. Wasn't sure what the standard was, but I saw modal's `children` was `any` as well. Buttons should be able to have bold, italics, icons etc. so I figured this was better than forcing `translateRaw` everywhere.

HenryNguyen5: Does ```ts
string | React.ReactElement<string>
```
work?

File: common/components/Header/components/CustomNodeModal.tsx [hunk 3]
```diff
@@ -0,0 +1,260 @@
...
+    username: '',
+    password: '',
+  };
+
+  public render() {
+    const { handleClose } = this.props;
+    const isHttps = window.location.protocol.includes('https');
+    const invalids = this.getInvalids();
+
+    const createInput = (name: string, placeholder?: string, type?: string) => {
```
HenryNguyen5: Could this take an object of `name` `placeholder?` and `type?` instead? Then you can just spread the object inside `input`

HenryNguyen5: Also you can use an implict return here

HenryNguyen5: And it might be worth extracting this function into a method so its not being reinstantiated on each re-render

common/components/Header/components/CustomNodeModal.tsx
```diff
@@ -0,0 +1,260 @@
...
+    // Numeric chain id, if provided
+    if (network === "custom" && customEip155 && !parseInt(customChainId, 10)) {
+      invalids.customChainId = true;
+    }
+
+    return invalids;
+  }
+
+  private handleChange = (ev: React.SyntheticEvent<
+    HTMLInputElement | HTMLSelectElement
```
HenryNguyen5: Instead, you can do `React.FormEvent<HTMLInputElement | HTMLSelectElement>`,
Then when you access `ev.currentTarget.value` you dont need to typecast anything

common/components/Header/components/CustomNodeModal.tsx
```diff
@@ -0,0 +1,260 @@
...
+  >) => {
+    // TODO: typescript bug: https://github.com/Microsoft/TypeScript/issues/13948
+    const name: any = (ev.target as HTMLInputElement).name;
+    const value = (ev.target as HTMLInputElement).value;
+    this.setState({ [name]: value });
+  };
+
+  private handleCheckbox = (ev: React.SyntheticEvent<HTMLInputElement>) => {
```
HenryNguyen5: ^ same here for typing the event

common/components/Header/index.tsx
```diff
@@ -1,55 +1,93 @@
...
+    const { isAddingCustomNode } = this.state;
     const selectedLanguage = languageSelection;
-    const selectedNode = NODES[nodeSelection];
-    const selectedNetwork = NETWORKS[selectedNode.network];
+    const selectedNetwork = NETWORKS[node.network];
     const LanguageDropDown = Dropdown as new () => Dropdown<
       typeof selectedLanguage
     >;
+
+    console.log(node);
```
HenryNguyen5: console logss

common/components/Header/components/CustomNodeModal.tsx
```diff
@@ -0,0 +1,230 @@
...
+  private handleChange = (ev: React.FormEvent<
+    HTMLInputElement | HTMLSelectElement
+  >) => {
+    const { name, value } = ev.currentTarget;
+    this.setState({ [name]: value } as any);
+  };
+
+  private handleCheckbox = (ev: React.FormEvent<HTMLInputElement>) => {
+    const { name } = ev.currentTarget;
+    this.setState({ [name]: !this.state[name] } as any);
```
wbobeirne: @HenryNguyen5 can you check these two `as any`s out?

HenryNguyen5: Checked these out, looks like a legit bug.
One thing you can do is move `as any` inline so it looks like `this.setState({ [name as any]: !this.state[name] }) so you're only setting the computed property name as `any` and not the entire object itself.

wbobeirne:
Was able to test this with a QuikNode, works like a charm.

wbobeirne:
I've done manual testing with Quiknode, and just added unit tests for all of the redux changes and utils functions. I think the remaining issue of custom networks can be made into a new issue, and this can be merged as-is.

common/sagas/config.ts
```diff
@@ -54,19 +64,68 @@ function* reload(): SagaIterator {
...
+  }
+
+  // Grab latest block from the node, before switching, to confirm it's online
+  // Give it 5 seconds before we call it offline
+  let latestBlock;
+  let timeout;
+  try {
+    const { lb, to } = yield race({
+      lb: call(actionConfig.lib.getCurrentBlock.bind(actionConfig.lib)),
+      to: call(delay, 5000)
```
dternyak: 👍

common/libs/nodes/rpc/requests.ts
```diff
@@ -63,4 +64,10 @@ export default class RPCRequests {
       ]
     };
   }
+
+  public getCurrentBlock(): GetCurrentBlockRequest | any {
```
dternyak: | any?

common/components/ui/Modal.tsx
```diff
@@ -17,7 +17,7 @@ export interface IButton {
 }
 interface Props {
   isOpen?: boolean;
-  title: string;
+  title: any;
```
dternyak: What are the expected types here?

common/libs/nodes/rpc/requests.ts
dternyak: World of pain here -- going to merge as is and we can circle back.

PR #329, opened 2017-10-28
Warn / Prevent Outdated Browsers from Wallet Generation

### LINKED ISSUE #210: Don't let clients with insufficient entropy generate wallets
Old browsers that do not support [Window.crypto](https://developer.mozilla.org/en-US/docs/Web/API/Window/crypto) should be provided with a message explaining this, and be prevent from generating a wallet.

wbobeirne:
# Warning for generate wallet if window.crypto is missing

Closes #210. Unfortunately, despite looking very similar to the bad browser block (#274) it can't share a whole lot of code, because that feature exists largely outside of the codebase, inside of `index.html`. Fortunately it's not a whole lot of code.

One thing that I did not do, but might be nice, is to default to a different tab (Send? View wallet?) I'd imagine at some point we'll have a more proper "homepage" so I didn't invest too much thought into this, but let me know if you think that's the way to go. Personally, I think it's good to prompt the upgrade anyhow.

Here's a screenshot:

screen shot 2017-10-27 at 6 17 26 pm

### DIFF
--- a/common/containers/Tabs/GenerateWallet/index.tsx
+++ b/common/containers/Tabs/GenerateWallet/index.tsx
@@ -13,6 +13,7 @@ import { AppState } from 'reducers';
 import DownloadWallet from './components/DownloadWallet';
 import EnterPassword from './components/EnterPassword';
 import PaperWallet from './components/PaperWallet';
+import CryptoWarning from './components/CryptoWarning';
 import TabSection from 'containers/TabSection';
 
 interface Props {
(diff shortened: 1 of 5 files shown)


common/containers/Tabs/GenerateWallet/index.tsx
```diff
@@ -38,38 +39,43 @@ class GenerateWallet extends Component<Props, {}> {
...
-          <AnyEnterPassword
-            walletPasswordForm={this.props.walletPasswordForm}
-            generateNewWallet={this.props.generateNewWallet}
-          />
-        );
-        break;
-
-      case 'download':
-        if (wallet) {
+    if (window.crypto) {
```
dternyak: Has this been tested with browserstack? 

I would be inclined to believe calling window.crypto would throw an exception instead of returning falsy. In this case we would want try/catch instead.

PR #334, opened 2017-10-31
Wallet Loading States & Spinner Update

james-prado:
Closes issue https://github.com/MyEtherWallet/MyEtherWallet/issues/143

* Update loading spinner
    * Make spinner svg instead of png (much better for performance)
    * Add styles for svg
* Expose loading states for wallet balance
    * since [redux-promise-middleware](https://github.com/pburtchaell/redux-promise-middleware/blob/master/docs/introduction.md#use) is applied to the store, I decided to stick to their naming scheme for describing network status, i.e. pending, fulfilled, and rejected instead of introducing a new one.
    * add actions for network status & update wallet reducer
    * call network status actions asynchronously from within the wallet saga
* Display spinners for wallet balance based the on new `state.wallet.balance` property, `isPending: boolean`
* Update initial state of `state.wallet.balance` from `null` to `{ isPending: false, unit: null, amount: null }`


common/actions/wallet/actionCreators.ts
```diff
@@ -42,11 +42,27 @@ export function setWallet(value: IWallet): types.SetWalletAction {
...
+  return {
+    type: constants.WALLET_SET_BALANCE_FULFILLED,
+    payload: { value, status: { isPending: false } }
+  };
+}
+
+export function SetBalanceRejected(): types.SetBalanceRejectedAction {
+  return {
+    type: constants.WALLET_SET_BALANCE_REJECTED,
+    payload: { status: { isPending: false } }
```
wbobeirne: Having a payload is unnecessary for rejected and pending, and isPending is unneeded for fulfilled. Redux's docs recommend:

> It's a good idea to pass as little data in each action as possible.

Because how the action affects the redux store should be handled in the reducer. Rather than passing a bunch of stuff, and doing a bunch of spreads, you should just set isPending to true or false inside of the reducer cases. This makes it easier to trace where that assignment actually happens, and allows changes to be simpler since we only need to change the reducer, not the actions.

Actions are more of your user's intent, while the reducer is how the code responds to that intent.

common/components/BalanceSidebar/AccountInfo.tsx
```diff
@@ -80,9 +80,17 @@ export default class AccountInfo extends React.Component<Props, State> {
...
-                  ? balance ? balance.toString() : '???'
-                  : balance ? formatNumber(balance.amount) : '???'}
+                {this.state.showLongBalance ? (
+                  balance.isPending ? (
+                    <Spinner size="x1" />
+                  ) : (
+                    balance.amount.toString()
+                  )
+                ) : balance.isPending ? (
+                  <Spinner size="x1" />
```
wbobeirne: Spinner shows up twice here, might be simpler to make that the opening ternary check, ala:

```
balance.isPending ? (spinner) : (other checks)
```

common/components/BalanceSidebar/EquivalentValues.tsx
```diff
@@ -33,9 +34,13 @@ export default class EquivalentValues extends React.Component<Props, {}> {
...
-                      {balance
-                        ? formatNumber(balance.amount.times(rates[key]))
-                        : '???'}
+                      {balance ? (
+                        balance.isPending ? (
+                          <Spinner size="x1" />
+                        ) : (
+                          formatNumber(balance.amount.times(rates[key]))
+                        )
+                      ) : null}
```
wbobeirne: Can be simplified to `{balance && onlyRenderMeIfBalanceIsTruthy}`. The only caveat with that method is that if `balance` were an integer, you would do `!!balance` instead.

common/components/ui/Spinner.tsx
```diff
@@ -1,13 +1,25 @@
...
+      <circle
+        className="path"
+        cx="25"
+        cy="25"
+        r="20"
+        fill="none"
+        strokeWidth="5"
+      />
+    </svg>
+  );
```
wbobeirne: 👍

common/reducers/wallet.ts
```diff
@@ -1,19 +1,22 @@
 import {
-  SetBalanceAction,
+  SetBalancePendingAction,
+  SetBalanceFullfilledAction,
+  SetBalanceRejectedAction
+} from './../actions/wallet/actionTypes';
```
wbobeirne: Prefer absolute path rather than relative.

common/reducers/wallet.ts
```diff
@@ -22,18 +25,42 @@ export interface State {
...
+    ...state,
+    balance: { ...action.payload.value.toEther(), ...action.payload.status }
+  };
+}
+
+function SetBalanceRejected(
+  state: State,
+  action: SetBalanceRejectedAction
+): State {
+  return { ...state, balance: { ...state.balance, ...action.payload.status } };
```
wbobeirne: As noted before, the simpler thing to do here is pass no payload and just do:
```
return {
  ...state,
  balance: {
    ...state.balance,
    isPending: false,
  }
};
```

common/reducers/wallet.ts
```diff
@@ -22,18 +25,42 @@ export interface State {
 
 export const INITIAL_STATE: State = {
   inst: null,
-  balance: null,
+  balance: { isPending: false, unit: null, amount: null },
```
wbobeirne: Does the snapshot need to be updated for this?

common/components/BalanceSidebar/AccountInfo.tsx
```diff
@@ -1,15 +1,15 @@
...
+import { IWallet, Balance } from 'libs/wallet';
 import React from 'react';
 import translate from 'translations';
 import { formatNumber } from 'utils/formatters';
+import Spinner from 'components/ui/Spinner';
 import './AccountInfo.scss';
 
 interface Props {
-  balance: Ether;
+  balance: Balance;
```
wbobeirne: `balance` is still `Ether` in BalanceSidebar/index.tsx, and potentially other places. Check the CI logs or run `npm run tslint` to get the full log.

common/actions/wallet/actionCreators.ts
james-prado: I'll tidy that up. Passing as little data as possible in the payload is definitely preferable.

common/reducers/wallet.ts
james-prado: The [test](https://github.com/MyEtherWallet/MyEtherWallet/pull/334/files#diff-ca3d0ee2c7f40670c0116b0b69a2ea4dR19) does, the snapshots don't.

james-prado:
Unlocking the wallet and getting the balance from it are two separate actions. To me, it makes more sense to have an `isPending` key associated with each chunk of data being fetched (`state.wallet.balance`), instead of placing the key at the root of the wallet (`state.wallet`), and keeping a mental note of what data that `isPending` key represents (i.e. is the balance pending, or are the tokens pending?). Also no new nesting is being introduced in `state.wallet.balance`, I just changed the default from `null` to match the shape of the data it expects
```
balance: { 
    isPending: false, 
    unit: null, 
    amount: null 
}
```

dternyak:
The updated spinners can be hard to see against a dark background:

screen shot 2017-11-14 at 12 44 37 pm

I suggest passing an optional prop `dark` that when passed false, will provide a light version of the same spinner.

common/components/ui/UnitDisplay.tsx
```diff
@@ -43,7 +44,7 @@ const UnitDisplay: React.SFC<EthProps | TokenProps> = params => {
   const { value, symbol, displayShortBalance } = params;
 
   if (!value) {
-    return <span>???</span>;
+    return <Spinner size="x1" />;
```
dternyak: I'm not sure if we should have spinners when a user is purposefully using MEW from an offline machine.

dternyak: I would prefer explicit text saying that "Balances are not available when offline".

The same applies for equivalent values (and perhaps across the entire app?)

common/typescript/bn.d.ts
```diff
@@ -14,17 +14,17 @@ declare module 'bn.js' {
     redSqr(): RedBN;
     redISqr(): RedBN;
     /**
-       * @description square root modulo reduction context's prime
-       */
+     * @description square root modulo reduction context's prime
```
dternyak: only formatting changes in this file

Issue #311, opened 2017-10-23
Unencrypted PK -> Encrypted Keystore File


tayvano:
See https://github.com/kvhnuke/etherwallet/issues/1163

Users often want to go from their raw private key to an encrypted keystore file. Right now, we have them use our Chrome Extension of via geth command line. As the use of the Chrome Extension will change moving forward, we need to have this functionality in MEW somewhere. 

The user would need to provide their unencrypted private key and a password and out would pop a keystore file. 

Not sure if this is already done or being worked on, but wanted to get it in writing somewhere so it doesn't get missed.

Not a v4 blocker, just needs to be done before we deprecate Chrome Extension functions.

💖

HenryNguyen5:
Internally the code is already there to handle this, would just need the UI portion!

HenryNguyen5:
@eddiewang 
Code for doing this lives in `ethereumjs-wallet` itself, you'll need to make an instance of a wallet via `fromPrivateKey` then call `.toV3()` on the wallet instance.

eddiewang:
#336

PR #336, opened 2017-11-01
Add Private key to V3 keystore functionality

eddiewang:
Closes issue #311 

A new tab is added to allow a user to regenerate their keystore file with a new password, as long as the user has a valid hex representation of the 256-bit key.

Basic form validation that ensures the password length is at least 9 characters and that the key is 64 valid hex characters.

Shoutout to @HenryNguyen5 for helping me setup my VSCode environment and debugging some TypeScript issues I had.

Here's a screenshot:
image


common/containers/Tabs/RestoreKeystore/components/KeystoreDetails.tsx
```diff
@@ -0,0 +1,156 @@
...
+    const {
+      secretKey,
+      isPasswordVisible,
+      isPrivateKeyVisible,
+      password,
+      wallet,
+      fileName
+    } = this.state;
+    const passwordValid = password.length > 9;
+    const privateKeyValid =
```
HenryNguyen5: ethereumjs-util has a validator for this

common/containers/Tabs/RestoreKeystore/components/KeystoreDetails.tsx
```diff
@@ -0,0 +1,156 @@
...
+      secretKey.length === 64 && /^[a-fA-F0-9]+$/g.test(secretKey);
+    const content = (
+      <div className="KeystoreDetails">
+        <div>
+          <label className="KeystoreDetails-key">
+            <h4 className="KeystoreDetails-label">Private Key</h4>
+            <div className="input-group">
+              <input
+                className={`form-control is-${privateKeyValid
```
HenryNguyen5: You should use the `classnames` package for conditional class names

common/containers/Tabs/RestoreKeystore/components/KeystoreDetails.tsx
```diff
@@ -0,0 +1,156 @@
...
+              />
+            </div>
+          </label>
+        </div>
+        <div>
+          <label className="KeystoreDetails-password">
+            <h4 className="KeystoreDetails-label">Password</h4>
+            <div className="input-group">
+              <input
+                className={`form-control is-${passwordValid
```
HenryNguyen5: Same here regarding `classnames` package

common/containers/Tabs/RestoreKeystore/components/KeystoreDetails.tsx
```diff
@@ -0,0 +1,156 @@
...
+          >
+            Generate Keystore
+          </button>
+        ) : (
+          <a
+            onClick={this.resetState}
+            href={this.getBlob()}
+            className="KeystoreDetails-download btn btn-primary btn-block"
+            aria-label="Download Keystore File (UTC / JSON · Recommended · Encrypted)"
+            aria-describedby="x_KeystoreDesc"
```
HenryNguyen5: You'll need to import the `translate` function from `import translate from 'translations';` to get the value of `x_keystoreDesc` I believe, looks like generateWallet doesn't do that either, would be a good time to fix both

common/containers/Tabs/RestoreKeystore/index.tsx
```diff
@@ -0,0 +1,15 @@
+import React, { Component } from 'react';
+import TabSection from 'containers/TabSection';
+import KeystoreDetails from './components/KeystoreDetails';
+
+class RestoreKeystore extends Component<{}, {}> {
```
HenryNguyen5: This can be a `React.SFC<{}>` instead of a component

common/containers/Tabs/RestoreKeystore/components/Template.tsx
```diff
@@ -0,0 +1,22 @@
+import React, { Component } from 'react';
+
+interface Props {
+  content: React.ReactElement<any>;
+  title: string;
+}
+
+class RestoreKeystoreTemplate extends Component<Props, {}> {
```
HenryNguyen5: This can be a `React.SFC<Props>` instead of a component

common/containers/Tabs/RestoreKeystore/components/KeystoreDetails.tsx
eddiewang: isValidPrivate in ethereumjs-util expects a buffer - which means we would have to convert the hex string to a buffer to do the check, which in the case that the hex string is invalid, that would throw an Error on the conversion.

common/libs/validators.ts
```diff
@@ -86,7 +86,13 @@ function validateEtherAddress(address: string): boolean {
...
-    return privkey.length === 64;
+    const initialCheck =
+      privkey.length === 64 && /^[a-fA-F0-9]+$/g.test(privkey);
+    if (initialCheck) {
+      const keyBuffer = Buffer.from(privkey, 'hex');
+      return isValidPrivate(keyBuffer);
+    }
+    return false;
   } else if (privkey instanceof Buffer) {
     return privkey.length === 32;
```
HenryNguyen5: Could you add the `isValidPrivate` check here too?
Should be able to do the check just once after the string is converted to a buffer.

common/containers/Tabs/RestoreKeystore/components/KeystoreDetails.tsx
```diff
@@ -37,18 +41,19 @@ class KeystoreDetails extends Component<{}, State> {
       fileName
     } = this.state;
     const passwordValid = password.length > 9;
```
HenryNguyen5: Wallets can have password length of >= 0

HenryNguyen5:
Password length can be < 9 characters. E.g. Parity lets you make a password of any length including 0

eddiewang:
@HenryNguyen5 for EnterPassword.tsx in the wallet generation tab we ensure that it's at least 9 characters. Should we stay consistent?

dternyak:
Other wallets may not have this requirement, so no.

wbobeirne:
Thanks for doing this! I just have some notes about the implementation, not the code. This is a pretty niche one-time behavior, so I think dedicating a whole tab to it, and one on the left side at that is a bit much. I think we could either tuck this away somewhere, or present the user with a dialog or button on the send transaction / view wallet pages. Beyond that, some of the copy around this doesn't seem very intuitive to me:

* The tab says "Redownload", but the title is "Regenerate"
* The user may have never downloaded or generated a keystore before, so I think **re** might make users who are looking to convert theirs a little confused
* There's no explanation here of what a keystore is, why it's a good idea to do this, and how they'll unlock their wallet after doing this

I think this page could use a bit of thought on how it will fit into MEW, and how to convince users to use this. No negative reflection on the PR though, this is a great starting point. CC @tayvano if you have any thoughts on where this should live.

eddiewang:
@wbobeirne thank you for the feedback. your comment make sense, the feature may be too specific to dedicate a whole tab to it. What do you think about a modal that pops up from the wallet generation tab sidebar (shown here):

image

common/containers/Tabs/RestoreKeystore/components/KeystoreDetails.tsx
```diff
@@ -0,0 +1,151 @@
...
+            aria-describedby={translate('x_KeystoreDesc')}
+            download={fileName}
+          >
+            Download Keystore
+          </a>
+        )}
+      </div>
+    );
+    return (
+      <div>
```
HenryNguyen5: Is this div needed?

common/libs/validators.ts
```diff
@@ -86,9 +86,15 @@ function validateEtherAddress(address: string): boolean {
 
 export function isValidPrivKey(privkey: string | Buffer): boolean {
```
dternyak: EthereumJS lib that handles this? Not sure how much I like having this custom logic as part of our app code, and definitely not additional regex.

common/libs/validators.ts
```diff
@@ -84,11 +84,24 @@ function validateEtherAddress(address: string): boolean {
...
+    return key.slice(2);
+  }
+  return key;
+}
+
 export function isValidPrivKey(privkey: string | Buffer): boolean {
   if (typeof privkey === 'string') {
-    return privkey.length === 64;
+    const initialCheck =
+      privkey.length === 64 && /^[a-fA-F0-9]+$/g.test(privkey);
```
dternyak: What is this regex checking in simple terms?

Is there an ethereumjs-library method we can use instead?

eddiewang: this regex simply checks if the string is a valid hex.

common/components/WalletDecrypt/PrivateKey.tsx
```diff
@@ -23,7 +17,7 @@ interface Validated {
 }
 
 function validatePkeyAndPass(pkey: string, pass: string): Validated {
-  const fixedPkey = fixPkey(pkey);
+  const fixedPkey = stripHexPrefix(pkey);
```
HenryNguyen5: Probably better to name this to strippedPrivateKey or something

common/containers/Tabs/RestoreKeystore/components/KeystoreInput.tsx
```diff
@@ -0,0 +1,43 @@
...
+import classnames from 'classnames';
+
+interface Props {
+  isValid: boolean;
+  isVisible: boolean;
+  handleInput: any;
+  name: string;
+  value: string;
+  placeholder: string;
+  handleToggle: any;
```
HenryNguyen5: This should be typed

HenryNguyen5:
On a second note, to make sure users have more confidence in the conversion we should implement a run-time validation check.

After converting the private key into a keystore, the component should then call a function that converts the keystore back into a private key and does an assertion to see if they're the same.

UI could be added to this part too if needed to visually tell the user that we've done this check

common/containers/Tabs/RestoreKeystore/components/KeystoreInput.tsx
eddiewang: ack

common/components/WalletDecrypt/PrivateKey.tsx
eddiewang: ack

eddiewang:
@HenryNguyen5 thank you for the review. I've added the changes your comments. One minor thing, I don't know how I feel about `strippedPrivateKey` as we use it more as a universal function that removes hex prefixes from hex strings. That said, most of our usage with that function revolves around dealing with private keys. As an example, `strippedPrivateKey` doesn't make as much sense here:

```
  public getCurrentBlock(): Promise<string> {
    return this.client
      .call(this.requests.getCurrentBlock())
      .then(isValidCurrentBlock)
      .then(({ result }) => new BN(strippedPrivateKey(result)).toString());
  }
```

eddiewang:
Also, I've added a runtime check, that will be helpful if something terrible happens to the ethereum wallet libraries. However, if the user types in a valid private key that is wrong, it would still allow the user to download the keystore as intended.

common/components/WalletDecrypt/PrivateKey.tsx
wbobeirne: From your comment below, I believe henry meant `fixedPkey -> strippedPkey` rather than `stripHexPrefix -> strippedPkey`.

dternyak: Yes, I also he meant to change the variable name. It's minor however, so I'm merging as is.

eddiewang: 🤦‍♂️ sorry

PR #337, opened 2017-11-02
[WIP] ENS name search

HenryNguyen5:
Depends on #319 for BN replacing big number

Adds ENS name search bar, along with different displays depending on the search result.

Todo:
 - [ ] CSS styling @eddiewang  @wbobeirne  @james-prado  could one of you help me with this?
 - [ ] Add tests for ENS lib

Side-note: CI is throwing a weird missing module error, any ideas?


common/components/renderCbs/index.ts
```diff
@@ -0,0 +1 @@
+export * from './UnitConverter';
```
wbobeirne: This is a pretty interesting pattern, but took me a while to grok. Any reason you'd want something like this over inheritance & state, or HOCs?

common/components/ui/Aux.tsx
```diff
@@ -0,0 +1,3 @@
+const Aux = ({ children }) => children;
+
+export default Aux;
```
wbobeirne: I don't see this used anywhere, and it looks like a noop.

common/components/ui/UnitDisplay.tsx
```diff
@@ -0,0 +1,54 @@
...
+
+  const noDecimal = decimal === null || decimal === undefined;
+  if (noDecimal && !unit) {
+    throw Error(
+      `Not enough parameters to convert base unit to display unit. Params given: ${JSON.stringify(
+        params,
+        null,
+        1
+      )}`
+    );
```
wbobeirne: This can be avoided by making props a union of various objects that have the correct combination of required fields.

common/components/ui/UnitDisplay.tsx
```diff
@@ -0,0 +1,54 @@
...
+
+  return (
+    <span>
+      {formattedValue}
+      {symbol ? ` ${symbol}` : ''}
+    </span>
+  );
+};
+
+export default UnitDisplay;
```
wbobeirne: This is a great idea, something we should have done a while ago 👍

common/containers/Tabs/ENS/components/NameResolve/components/CountDown.tsx
```diff
@@ -0,0 +1,69 @@
...
+import moment from 'moment';
+
+class CountDown extends Component<
+  {
+    initialTime: number;
+  },
+  {
+    currentTime: number;
+  }
+> {
```
wbobeirne: Would prefer keeping the habit of defining a `Props` type rather than inline definitions.

common/containers/Tabs/ENS/components/NameResolve/components/CountDown.tsx
```diff
@@ -0,0 +1,69 @@
...
+    const hours = floorTime(hour);
+    timeRemaining -= hours * hour;
+    const minutes = floorTime(minute);
+    timeRemaining -= minutes * minute;
+    const seconds = floorTime(second);
+
+    return `${pad(days)} Days ${pad(hours)} Hours ${pad(minutes)} Minutes ${pad(
+      seconds
+    )} Seconds `;
+  };
```
wbobeirne: If we end up with much more time displays, I'd much prefer a library for this kind of stuff (Like Moment) rather than custom implementations.

wbobeirne: Oh wait, we already have Moment. Can we just use that here?

common/containers/Tabs/SendTransaction/components/AmountField.tsx
```diff
@@ -1,45 +1,54 @@
...
+                  disabled={isReadOnly}
+                  onChange={onUserInput}
+                />
+              )}
+            </UnitConverter>
             <UnitDropdown
               value={unit}
               options={['ether'].concat(this.props.tokens)}
-              onChange={isReadonly ? void 0 : this.onUnitChange}
+              onChange={isReadOnly ? void 0 : this.onUnitChange}
```
wbobeirne: You didn't write this, but `isReadOnly && this.onUnitChange` would suffice here

common/containers/Tabs/ENS/components/NameResolve/components/CountDown.tsx
HenryNguyen5: I couldn't get the custom formatting I wanted with moment, looked like a needed a countdown plugin for it too.

common/components/renderCbs/index.ts
HenryNguyen5: Anything Unit related is part of the BN refactor, it'll be a lot more clear there :) reason why I used a render callback is so it's easy to use within nested jsx, as you want to be using it as the direct parent of a input element

common/components/ui/UnitDisplay.tsx
HenryNguyen5: Good catch, I'll address this in `bn-refactor`

common/containers/Tabs/SendTransaction/components/AmountField.tsx
HenryNguyen5: This is also part of `bn-refactor` I'll address it there. My bad on not mentioning that this PR depends on it

common/containers/Tabs/ENS/components/NameResolve/components/CountDown.tsx
HenryNguyen5: I was contemplating between the two formats for this, I'll switch it back

common/components/ui/Aux.tsx
HenryNguyen5: I was initially using it to return groups of elements without needlessly wrapping them in a div, guess I got rid of its usage later on

HenryNguyen5:
Closing this due to stale branch, going to re-open when appropriate

james-prado:
Closing to open new PR on a different branch with smaller diff

PR #338, opened 2017-11-04
CI Improvements

wbobeirne:
This PR aims to improve continuous integration and the development process in a few ways:

1. Run the lint command pre-commit, since it's fast enough. Should reduce number of failed CIs.
2. Add `tsc --noEmit` to the lint command, since CI runs it as well.
3. Run tests as a part of `prepush`.
4. Reduce noisy output of CI by suppressing the install tree and reducing some of the warnings on tests

Not all of this is complete yet, but I needed to open a PR to get CI to run and show reports. Will update when it's all complete.


package.json
```diff
@@ -126,10 +126,11 @@
     "dev:https": "HTTPS=true node webpack_config/server.js",
     "predev:https": "check-node-version --package",
     "derivation-checker": "webpack --config=./webpack_config/webpack.derivation-checker.js && node ./dist/derivation-checker.js",
-    "tslint": "tslint --project . --exclude common/vendor/**/*",
+    "tslint": "tsc --noEmit && tslint --project . --exclude common/vendor/**/*",
```
HenryNguyen5: I think this should be a semicolon instead of a `&&` that way we get both tsc errors and tslint errors, if tsc fails

wbobeirne: The problem with that is that it would only fail on the right-side command. If the first one fails, and the second one passes, I don't think the build would fail. Maybe I should make these 2 separate commands.

wbobeirne:
Something got messed up with the git history, I'm going to rework it to remove the merge commit.

PR #340, opened 2017-11-06
[WIP] ENS bidding for open names

HenryNguyen5:
Depends on #337 

Allows the user to bid on open names.

Uses re-usable components for bidding on ENS names, so auctioning on ENS names after placing an initial bid should re-use almost all the same components


common/containers/Tabs/ENS/components/NameResolve/components/BidModal.scss
```diff
@@ -0,0 +1,75 @@
...
+      float: none;
+      width: 100%;
+      margin: 0 0 5px;
+
+      &:last-child {
+        margin: 0;
+      }
+    }
+  }
+}
```
wbobeirne: Given that this is the second time we've needed this, I think we need a better solution. Don't worry about it here, just a note for the future.

HenryNguyen5:
Closing this due to stale branch, going to re-open when appropriate

PR #366, opened 2017-11-09
Equivalent Values for Tokens

### LINKED ISSUE #90: Balance Component
- [x] ETH Balance
- [x] Token Balances
- [x] User's address + addressicon
- [x] Donation Blurb on Send Page
- [x] Equivalent values in USD CHF BTC EUR GBP
- [x] Equivalent values for Tokens
- [ ] Equivalent values for all balances (ether + tokens)

Github Ref
1. Better ethplorer integration? kvhnuke/etherwallet: Issue #431
2. GBP kvhnuke/etherwallet: Issue #440
3. Equiv of tokens kvhnuke/etherwallet: Issue #435

wbobeirne:
Closes #90 (or, at least the last task in #90.) Adds a dropdown to the equivalent values sidebar block that shows token values in other currencies. It also shows the currency's value in ETH (When the currency itself is not ETH.)

This was a pretty straightforward change, the main difference is that the rates state shape is now:
```
{
  rates: {
    ETH: { ... },
    BAT: { ... }
  }
}
```
instead of just a flat `rates` object with currencies. It was only being used for `EquivalentValues.tsx` though, so there was little refactoring involved.

The main caveat I see here is that cryptocompare goes entirely off of token symbol. So if a user adds a token with the wrong symbol, or at some point there's a mismatch, the user may see an incorrect equivalent value. I think this is pretty low-impact though, and isn't really too worrisome.

Here's a gif to see it in action


dternyak:
Tested -- works great. No comments on the code either.

I don't want to move the goalposts, but while this implementation is certainly useful, the original feature request was for to show the cumulative value of all ether plus token balances. I've adjusted the issue to reflect that given that the original spec was vague. Perhaps an "all" option would suffice for that. 

This is certainly a useful feature as well, and can certainly be merged immediately and without further alteration. However, it shouldn't be difficult to add a default "all" option with the cumulative value to fulfill the additional requirement.

wbobeirne:
Aaah, that actually makes a lot of sense. I'll go ahead and add that option, but I may not be able to default to it; It takes a while to come back with all of the token values so I wouldn't know which ones to ask for from CryptoCompare.

wbobeirne:
I'm actually surprised at how annoying this is, because the CryptoCompare API only accepts one symbol to compare to at a time. I'll definitely try to get this done, but it may require a bit of overhauling.

Issue #372, opened 2017-11-10
Etherscan estimate gas returning an error


HenryNguyen5:
Found during the bn.js refactor, on develop branch etherscan `eth_estimateGas` also throws an api error that we dont really handle.

PR #384, opened 2017-11-12
RPC Error Handling

eddiewang:
This PR is for #472 and #372

It works in two main parts.

The first is an integration test that covers all RpcNode and children endpoints. We write json schemas to validate against, which essentially acts as a whitelist of valid results. When we run our tests, we will be notified if any of the nodes stop working.

The second works similarly, but runs in the main app. We write method validations for each RpcNode method, using the same json schema to validate against. If it's not in the whitelisted schemas, and error will be thrown the error is displayed through the UI component.


common/libs/validators.ts
```diff
@@ -177,3 +179,28 @@ export const isValidByteCode = (byteCode: string) =>
 
 export const isValidAbiJson = (abiJson: string) =>
   abiJson && abiJson.startsWith('[') && abiJson.endsWith(']');
+
+// JSONSchema Validations for Rpc responses
+const v = new Validator();
+
+export const schema = {
```
dternyak: This approach looks solid to me 👍

common/libs/nodes/rpc/index.ts
```diff
@@ -5,6 +5,22 @@ import { Wei } from 'libs/units';
 import { INode, TxObj } from '../INode';
 import RPCClient from './client';
 import RPCRequests from './requests';
+import { isValidGetBalance } from '../../validators';
+
+// TODO - understand response values
+// "RPC" requests will sometimes resolve with 200, but contain a payload informing there was an error.
+function ensureOkResponse(response: any) {
```
dternyak: May be redundant since validation should catch an error response (it *should* have a different shape as the success response.

package-lock.json
```diff
@@ -23,9 +23,9 @@
       "dev": true
```
dternyak: package-lock.json can be removed entirely.

common/components/Footer/index.tsx
```diff
@@ -94,7 +94,7 @@ const LINKS_SOCIAL = [
 
 interface Props {
   latestBlock: string;
-};
+}
```
HenryNguyen5: Looks like this file got formatted without semis

eddiewang: @HenryNguyen5 Is that supposed to have semis? I have semis set to true, but it's not adding the semi there. The semi is added on line 96. confusing.

eddiewang:
PR should be ready to review. Before merging, I will need to take out the 2 console.logs. I left them in there so reviewers can test the response requests / responses easily, and ensure the api validation is working. Also, the console.log will display all the api calls when the jest tests are run.

This PR has tests covering all RpcNodes, InfuraNodes, and EthscanNodes.

It also contains fixes for InfuraNode and EthscanNode class, that caused the Etherscan api to fail on estimateGas and other methods #372 . The fix ensures that when we run through the headers, only non-empty values are added. It also uses the Header object properly by adding a createHeader helper class method.

The InfuraNode did not work previously, as we were sending an invalid body object. This fix changes the way the id is generated, as Infura expects a number rather than the string.

common/utils/node.ts
```diff
@@ -6,13 +6,15 @@ export function makeCustomNodeId(config: CustomNodeConfig): string {
 }
 
 export function getCustomNodeConfigFromId(
```
dternyak: only formatting changes in this file -- revert?

common/reducers/config.ts
```diff
@@ -88,27 +84,24 @@ function forceOffline(state: State): State {
 function addCustomNode(state: State, action: AddCustomNodeAction): State {
   return {
     ...state,
-    customNodes: [
-      ...state.customNodes,
-      action.payload,
-    ],
+    customNodes: [...state.customNodes, action.payload]
```
dternyak: only formatting changes here as well

common/libs/nodes/rpc/requests.ts
```diff
@@ -7,7 +7,7 @@ import {
   GetTokenBalanceRequest,
   GetTransactionCountRequest,
   SendRawTxRequest,
-  GetCurrentBlockRequest,
+  GetCurrentBlockRequest
```
dternyak: only formatting changes here

common/libs/nodes/rpc/index.ts
```diff
@@ -63,39 +62,29 @@ export default class RpcNode implements INode {
...
-        return response.map(item => {
-          // FIXME wrap in maybe-like
-          if (item.error) {
-            return TokenValue('0');
-          }
-          return TokenValue(item.result);
-        });
-      });
-    // TODO - Error handling
+      .then(response => response.map(item => TokenValue(item.result)));
```
dternyak: Error handling TODO is removed, but its not clear to me that there is in-fact error handling here. What happens when only one token request fails, but all others succeeds?

common/libs/nodes/rpc/index.ts
```diff
@@ -6,13 +6,15 @@ import { stripHexPrefix } from 'libs/values';
...
-}
+import {
+  isValidGetBalance,
+  isValidEstimateGas,
+  isValidCallRequest,
+  isValidTokenBalance,
+  isValidTransactionCount,
+  isValidCurrentBlock,
+  isValidRawTxApi
+} from '../../validators';
```
dternyak: prefer absolute imports (can we make a tsconfig flag for this?)

common/libs/nodes/infura/client.ts
```diff
@@ -2,7 +2,7 @@ import { randomBytes } from 'crypto';
 import RPCClient from '../rpc/client';
 
 export default class InfuraClient extends RPCClient {
-  public id(): string {
-    return `0x${randomBytes(5).toString('hex')}`;
+  public id(): number {
+    return parseInt(randomBytes(5).toString('hex'), 16);
```
dternyak: 👍

common/containers/TabSection/index.tsx
```diff
@@ -8,7 +8,7 @@ import {
   TChangeLanguage,
   TChangeNodeIntent,
   TAddCustomNode,
-  TRemoveCustomNode,
+  TRemoveCustomNode
```
dternyak: only formatting changes in this file

common/components/Header/components/CustomNodeModal.tsx
```diff
@@ -35,23 +35,26 @@ export default class CustomNodeModal extends React.Component<Props, State> {
     network: NETWORK_KEYS[0],
```
dternyak: only formatting changes in this entire file

spec/config/data.spec.ts
```diff
@@ -0,0 +1,91 @@
...
+    describe(`RPC (${service}) should work`, () => {
+      it(`RPC: ${testType} ${service}`, () => {
+        return RPCTests[testType](node).then(d => expect(d.valid).toBeTruthy());
+      });
+    });
+  });
+}
+
+const mapNodeEndpoints = (nodes: { [key: string]: NodeConfig }) => {
+  const RpcNodeList = ['eth_mew', 'etc_epool', 'etc_epool', 'rop_mew'];
```
dternyak: could we import from some config instead of re-hardcoding here?

spec/config/data.spec.ts
```diff
@@ -0,0 +1,91 @@
...
+
+  const EthscanNodeList = ['eth_ethscan', 'kov_ethscan', 'rin_ethscan'];
```
dternyak: ditto ^

spec/config/data.spec.ts
```diff
@@ -0,0 +1,91 @@
...
+
+  const InfuraNodeList = ['eth_infura', 'rop_infura', 'rin_infura'];
```
dternyak: ditto ^

package-lock.json
```diff
@@ -0,0 +1,15014 @@
+{
```
dternyak: remove package-lock.json

common/libs/nodes/rpc/index.ts
eddiewang: My thinking here is that rather than softly resolving the error by hardcoding a TokenValue('0'), we would throw and error so we don't confuse the user. But I guess if a single call fails, then the rest of the calls would subsequently stop as well. What I can do is add an optional overloaded callback that serves as a custom resolver for isValidEthCall.

common/libs/nodes/infura/client.ts
HenryNguyen5: So infura takes a number for ID but RPC node takes a string?

common/libs/nodes/rpc/client.ts
```diff
@@ -22,22 +22,30 @@ export default class RPCClient {
...
         'Content-Type': 'application/json',
-        ...this.headers,
-      },
+        ...this.headers
+      }),
       body: JSON.stringify(requests.map(this.decorateRequest))
     }).then(r => r.json());
   };
+
+  private createHeaders = headerObject => {
```
HenryNguyen5: Should type up `headerObject` with the appropriate header keys

common/libs/nodes/rpc/index.ts
```diff
@@ -24,78 +26,73 @@ export default class RpcNode implements INode {
...
-    return this.client.call(this.requests.ethCall(txObj)).then(r => {
-      if (r.error) {
-        throw Error(r.error.message);
-      }
-      return r.result;
-    });
+    return this.client
+      .call(this.requests.ethCall(txObj))
+      .then(isValidCallRequest)
+      .then(({ result }) => {
```
HenryNguyen5: Could use implict return here (and in the function itself) to avoid brace / return keyword noise

common/reducers/wallet.ts
```diff
@@ -15,7 +15,7 @@ export interface State {
   // in ETH
   balance: Balance | { wei: null };
   tokens: {
-    [key: string]: TokenValue;
+    [key: string]: TokenValue | string;
```
HenryNguyen5: Why is there a string being added here? Everything should be a BN

common/selectors/wallet.ts
```diff
@@ -11,7 +11,7 @@ export function getWalletInst(state: AppState): IWallet | null | undefined {
 
 export interface TokenBalance {
   symbol: string;
-  balance: TokenValue;
+  balance: any;
```
HenryNguyen5: This should be a BN too.

spec/config/RpcNodeTestConfig.js
```diff
@@ -0,0 +1,5 @@
+module.exports = {
```
HenryNguyen5: Prefer a default export instead of a module export here

spec/config/data.spec.ts
```diff
@@ -0,0 +1,88 @@
...
+  },
+  token: {
+    address: '0x4156d3342d5c385a87d264f90653733592000581',
+    symbol: 'SALT',
+    decimal: 8
+  }
+};
+
+const testGetBalance = async (n: RPCNode) => {
+  const data = await n.client.call(
```
HenryNguyen5: These functions would probably be cleaner if you used `.then` instead of async since its only a single call.

common/libs/nodes/infura/client.ts
dternyak: Yes, oddly

eddiewang: Yeah, RpcNodes for our apis take a hex string, but the Infura nodes need a integer or it replies with a Golang error about parsing hex.

common/libs/nodes/rpc/index.ts
eddiewang: ack

common/reducers/wallet.ts
eddiewang: Instead of defaulting to a 0 value token on error, we could have a "Failed" string instead of trying to soft-fail. We can show the token value as "failed" in the UI to avoid confusion.

spec/config/RpcNodeTestConfig.js
eddiewang: same. for some reason that doesn't work with jest. probably es6 related stuff.

common/reducers/wallet.ts
dternyak: I don't think we need to keep "Failed" in redux state -- we can simply display that in the UI based on some computation of other state

common/libs/nodes/rpc/client.ts
eddiewang: @HenryNguyen5 The headerObject been console.logging as blank, aside from the content-type, but previously the code was written to loop around the Object keys, so I wrote it this way for compatibility reasons.

common/selectors/wallet.ts
eddiewang: ack

package.json
```diff
@@ -102,8 +103,9 @@
     "tslint-config-prettier": "1.6.0",
     "tslint-react": "3.2.0",
     "types-rlp": "0.0.1",
-    "typescript": "2.5.2",
+    "typescript": "2.6.2",
```
dternyak: remove change

package.json
```diff
@@ -102,8 +103,9 @@
...
     "url-loader": "0.6.2",
+    "url-search-params-polyfill": "2.0.1",
```
dternyak: separate PR (also, why do we need this?)

spec/config/data.spec.ts
```diff
@@ -0,0 +1,87 @@
+import { NODES, NodeConfig } from '../../common/config/data';
+import { RPCNode } from '../../common/libs/nodes';
+import { Validator } from 'jsonschema';
+import { schema } from '../../common/libs/validators';
+import 'url-search-params-polyfill';
```
dternyak: unclear why we need this

spec/reducers/wallet.spec.ts
```diff
@@ -60,7 +60,12 @@ describe('wallet reducer', () => {
   });
 
   it('should handle WALLET_SET_TOKEN_BALANCES', () => {
-    const tokenBalances = { OMG: TokenValue('20') };
+    const tokenBalances = {
```
dternyak: would be great to see a test for the error case

spec/config/data.spec.ts
eddiewang: @dternyak it injects search params so that the Etherscan Nodes work as seen here:
```
 public encodeRequest(request: EtherscanRequest): string {
    const encoded = new URLSearchParams();
    Object.keys(request).forEach(key => {
      if (request[key]) {
        encoded.set(key, request[key]);
      }
    });
    return encoded.toString();
  }
```

spec/reducers/wallet.spec.ts
eddiewang: ack

PR #386, opened 2017-11-13
Read-Only Address Wallet (Pt. 1 - The Basics)

wbobeirne:
This is the very least possible done for a read-only address wallet ala #92. You simply punch in an address, and can view any tab as though you were unlocked. However, it throws if you try to do anything that requires signing or private keys.

## What This Does

* Creates 2 new types to replace the previous 1 type-to-rule-them-all for wallets, `IFullWallet` and `IReadOnlyWallet`. `IWallet` is now an or of these two.
* Implements an `AddressOnlyWallet` class that extends the aforementioned `IReadOnlyWallet`.
* Changes most transaction and signing functions to require an `IFullWallet` rather than an `IWallet`
* Adds a boolean to wallet classes, `isReadOnly`, to hide unwanted behavior.

## What This Doesn't Do... <sup>and what I plan to do later</sup>

* Add a prop to the decrypt component to prevent address only from showing up on tabs like sign and deploy contract, and have the typing ensure that it will only return `IFullWallet`s when that prop is provided to avoid over-checking.
* Alter the send page so that it doesn't show the send form to read-only wallets.
* Create a view wallet info tab on send, and turn it into an "Account" tab.

What I'm looking for now is feedback on the organization and implementation of the wallet class. As noted above, there are still quite a few UX holes to plug here. However, I think this wouldn't be problematic to merge as-is if people like it, since the feature pretty much works. There's no broken UI, it always throws and displays messages properly if you try to do something you shouldn't.


common/containers/Tabs/Contracts/components/withTx.tsx
```diff
@@ -4,12 +4,12 @@ import { GWei, Wei, Ether } from 'libs/units';
...
 import { showNotification, TShowNotification } from 'actions/notifications';
 import { broadcastTx, TBroadcastTx } from 'actions/wallet';
-import { IWallet } from 'libs/wallet/IWallet';
+import { IFullWallet } from 'libs/wallet/IWallet';
 import { RPCNode } from 'libs/nodes';
 import { NodeConfig, NetworkConfig } from 'config/data';
 
 export interface IWithTx {
-  wallet: IWallet;
+  wallet: IFullWallet;
```
wbobeirne: @HenryNguyen5 Despite saying that this should only receive an `IFullWallet`, it's still able to receive an `IReadOnlyWallet` (Since that's what the redux store injects, any `IWallet`.) I'm quite confused about how this component works, and why it's not getting caught by the typechecker. Got any thoughts?

common/containers/Tabs/SignAndVerifyMessage/components/SignMessage/index.tsx
```diff
@@ -10,7 +10,7 @@ import { AppState } from 'reducers';
 import './index.scss';
 
 interface Props {
-  wallet: IWallet;
+  wallet: IFullWallet;
```
wbobeirne: @HenryNguyen5 Another instance of specifying `IFullWallet` but typescript allowing any `IWallet`.

wbobeirne: Come to think of it, the reducer doesn't even say it's a sure thing; it's `wallet?: IWallet | null`. Kinda scares me that it's letting this thru.

common/containers/Tabs/Contracts/components/withTx.tsx
HenryNguyen5: Hm, looks like we've been typing these up wrong.
It should be 
```ts
  wallet: AppState['wallet']['inst'];
```

HenryNguyen5: This will throw the correct error on the files since you're grabbing AppState interface and specifying the exact type it'll be

wbobeirne: I think what I would prefer here is to keep `IFullWallet` as the type, and just enforce that this component only renders with an `IFullWallet`, rather than having every wallet component have some conditional in it like
```
if (!this.props.wallet || this.props.wallet.isReadOnly) {
  return null;
}
```
along with every callback function that uses full wallet only functions. Maybe this could be done with a higher order component?

HenryNguyen5: Yep, HOC or Render callback would be perfect here

common/components/WalletDecrypt/ViewOnly.tsx
```diff
@@ -1,29 +1,60 @@
...
   }
+
+  private changeAddress = (ev: React.FormEvent<HTMLInputElement>) => {
+    this.setState({ address: ev.currentTarget.value });
+  };
+
+  private openWallet = (ev: React.SyntheticEvent<HTMLFormElement>) => {
+    const { address } = this.state;
+    ev.preventDefault();
+    if (isValidETHAddress(address)) {
```
HenryNguyen5: Would it be better here to disable the button outright if its not a valid address? Then you dont need to check validity before calling the unlock prop

common/containers/Tabs/SendTransaction/index.tsx
```diff
@@ -465,6 +465,10 @@ export class SendTransaction extends React.Component<Props, State> {
 
   public async getFormattedTxFromState(): Promise<TransactionWithoutGas> {
     const { wallet } = this.props;
+    if (wallet.isReadOnly) {
```
HenryNguyen5: Instead of throwing when the user tries to generate a tx, instead the fields can just all be disabled. 
`transaction-refactor` completely re-writes this component so perhaps i can help implement it there

common/components/WalletDecrypt/ViewOnly.tsx
```diff
@@ -1,29 +1,60 @@
...
+      const wallet = new AddressOnlyWallet(address);
+      console.log(wallet);
```
wbobeirne: looooog

common/components/WalletDecrypt/ViewOnly.tsx
wbobeirne: I'll leave the check here, since it's pretty inexpensive, but I'll definitely add disabling the button.

common/containers/Tabs/SendTransaction/index.tsx
wbobeirne: In a future PR, I want to not even show this form. I skipped over that now for exactly that `transaction-refactor` branch.

common/components/WalletDecrypt/ViewOnly.tsx
wbobeirne: This is done via 1d7c7b89a2affd0653c092ed6bfd3edd574cba2f

common/containers/Tabs/Contracts/components/withTx.tsx
wbobeirne: Implemented this in #387 over here: https://github.com/MyEtherWallet/MyEtherWallet/pull/387/files#diff-931b59e1826e8a1147efbbd7c1384306

PR #388, opened 2017-11-13
Strip hex prefix from raw private key displayed during wallet generation

eddiewang:
The new ethereumjs-lib we're using adds a hex prefix to the private key, which is undesirable behaviour, as it doesn't fit our current V3 Keystore format. The printed wallet also prints out the private key with a 0x.

This PR strips the hex prefix from a raw private key from the wallet lib, and displays it to the user as a private hex without the 0x prefix.

Also, we discovered that there are many different functions all doing minor string manipulations, and created an issue committed to tracking that in a future PR: #474


common/containers/Tabs/GenerateWallet/components/DownloadWallet.tsx
```diff
@@ -37,7 +37,11 @@ export default class DownloadWallet extends Component<Props, State> {
 
   public render() {
     const { hasDownloadedWallet } = this.state;
-    const filename = this.props.wallet.getV3Filename();
+    const getFilename = this.props.wallet.getV3Filename().split('--');
+    const checksumAddress = toChecksumAddress(getFilename[2]);
+    const filename = [getFilename[0], getFilename[1], checksumAddress].join(
```
dternyak: Would prefer to move this logic out of the component.

spec/libs/wallet/non-deterministic/helpers.spec.ts
```diff
@@ -0,0 +1,26 @@
+import { checksumKeystoreFilename } from '../../../../common/libs/wallet/non-deterministic/helpers';
```
dternyak: Prefer absolute import instead of relative.

eddiewang:
To keep everything consistent with MEW V3, we no longer checksum the public address in the filename, or add the 0x prefix to the private key. This also applies to the Paper Wallet that is printed out.

common/components/PrintableWallet/index.tsx
```diff
@@ -27,7 +28,7 @@ const print = (address: string, privateKey: string) => () =>
 
 const PrintableWallet: React.SFC<{ wallet: IFullWallet }> = ({ wallet }) => {
   const address = wallet.getAddressString();
-  const privateKey = wallet.getPrivateKeyString();
+  const privateKey = stripHexPrefix(wallet.getPrivateKeyString());
```
dternyak: Prefer stripping hex prefer at wallet method

HenryNguyen5: We'd have to wrap the library itself if we did this

dternyak:
Agreed. I don't expect breaking changes here, so I don't think we should
have any trouble wrapping.

common/components/PrintableWallet/index.tsx
dternyak: Agreed. I don't expect breaking changes here, so I don't think we should
have any trouble wrapping.

@HenryNguyen5 Perhaps we should introduce a separate PR for this, as I want to ensure that we design the wrapped lib correctly / with an independent design process from this PR.

eddiewang: We have a bunch of utility functions dealing with `stripHex` `sanitizeHex` `stripHexPrefixAndLower` etc. I think having a consistent set of methods would be ideal, unless we really need all these different types of sanitization/mutation methods.

PR #411, opened 2017-11-16
CI & Commit Hook Improvements

wbobeirne:
This is identical to #338, I just needed to clean up the git history which required a force push

> This PR aims to improve continuous integration and the development process in a few ways:
>
> 1. Run the lint command pre-commit, since it's fast enough. Should reduce number of failed CIs.
> 2. Add `tsc --noEmit` to the lint command, since CI runs it as well.
> 3. Run tests as a part of `prepush`.
> 4. Reduce noisy output of CI by suppressing the install tree and reducing some of the warnings on tests


dternyak:
The pre-commit seems (subjectively) a bit too long for my taste.

Takes me out of "flow state" to wait 5-10s.

The rest of the changes look fine to me!

wbobeirne:
Maybe we should move `npm run tscheck` to prepush, and remove `npm run test` altogether?

dternyak:
👍 That should be reasonably fast.

wbobeirne:
Sorry for the delay; I moved `tslint` and `tscheck` to prepush to keep committing fast. Precommit should be the same as it was.

dternyak:
After some thought, I'm not sure we want to exit and prevent pushing when those fail - it can be sometimes useful to push commits that fail (e.g. https://github.com/MyEtherWallet/MyEtherWallet/commit/ecddd7a76b20ef562e87353ba08e378c157ccf05).

Perhaps remove the pre-push entirely as well?

wbobeirne:
You can still push without hook if you're certain your change is good with `git [command] --no-verify`. Personally I would find this change to be a huge time saver in my workflow, since I often push a branch, leave it, and then have to come back to it 5-10 minutes later because I had missed something. But I won't push for it, I can always add it to my own pre-commit / pre-push hooks if I really want it.

At the very least though, we should get the silent install on CI stuff up. Going through those logs is a pain in the ass every time.

dternyak:
True, `--no-verify` is a decent stop-gap if you want to push anyway.

If it's a huge time-saver for you and not a significant drawback for others (as long as they know about `--no-verify`), let's merge!

dternyak:
@HenryNguyen5 @james-prado @eddiewang @skubakdj Ping ^ in case you weren't familiar with `--no-verify` and don't want to wait on pre-push.

PR #420, opened 2017-11-17
Equivalent values for all tokens (ETH + ERC20s)

### LINKED ISSUE #90: Balance Component

wbobeirne:
This is a continuation of #366 and finally closes #90. This adds an option (and makes it the default) to show equivalent values of your ETH balance as well as your tokens.

To reduce roundtrips, this calculates token values by going from token -> eth -> usd / gbp etc., instead of using the API to go directly from token to other. This may result in some _very_ minor differences, but it should be pretty small. Perhaps we should add a note here saying these are rough values and should not be taken as facts.

Here's what it looks like:

screen shot 2017-11-17 at 1 13 27 pm


common/components/BalanceSidebar/EquivalentValues.tsx
```diff
@@ -63,23 +68,19 @@ export default class EquivalentValues extends React.Component<Props, CmpState> {
...
-              {balance.isPending ? (
-                <Spinner />
-              ) : (
-                <UnitDisplay
-                  unit={'ether'}
-                  value={balance ? balance.muln(rates[currency][key]) : null}
-                  displayShortBalance={2}
-                />
-              )}
+              <UnitDisplay
```
dternyak: Why get rid of the spinner?

wbobeirne: The spinner should never happen; It was replaced with the one large spinner on line 83.

common/actions/rates/actionPayloads.ts
```diff
@@ -1,31 +1,52 @@
...
+const CCRates = (symbols: string[]) => {
+  const tsyms = rateSymbols.concat(symbols).join(',');
+  return `${CCApi}/data/price?fsym=ETH&tsyms=${tsyms}`;
 };
 
 export interface CCResponse {
-  symbol: string;
-  rates: {
-    BTC: number;
+  [symbol: string]: {
```
skubakdj: This should break the rates reducer test recently merged into `develop`.

common/components/BalanceSidebar/EquivalentValues.tsx
```diff
@@ -63,23 +68,19 @@ export default class EquivalentValues extends React.Component<Props, CmpState> {
...
+                unit={'ether'}
+                value={values[key]}
+                displayShortBalance={3}
```
skubakdj: Should we up this to `5`? Low-value accounts (~$1) currently get rounded out to `0 BTC` in the equiv rates display.

common/actions/rates/actionPayloads.ts
wbobeirne: Good catch, fixed!

common/components/BalanceSidebar/EquivalentValues.tsx
wbobeirne: I was talking to henry about this, I think we need to have UnitDisplay show `< 0.001` when the value is below the threshold, but above 0.

common/components/BalanceSidebar/EquivalentValues.tsx
wbobeirne: Un-redacted, this is fixed.

PR #435, opened 2017-11-21
Remove Bity Rates from Local Storage

james-prado:
Closes https://github.com/MyEtherWallet/MyEtherWallet/issues/392
* Replace `swap.bityRates` with initial value when storing in local storage
* Change initial value of `state.bityRates` from `{}` to `null`

bityRates no longer persist in localstorage and are updated on each page load.

### DIFF
--- a/common/store.ts
+++ b/common/store.ts
@@ -111,12 +111,13 @@ const configureStore = () => {
           languageSelection: state.config.languageSelection,
           customNodes: state.config.customNodes
         },
-        swap: state.swap,
+        swap: { ...state.swap, bityRates: {} },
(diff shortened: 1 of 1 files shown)


common/store.ts
```diff
@@ -111,12 +111,13 @@ const configureStore = () => {
           languageSelection: state.config.languageSelection,
           customNodes: state.config.customNodes
         },
-        swap: state.swap,
+        swap: { ...state.swap, bityRates: swapInitialState.bityRates },
```
dternyak: This still persists state.

`bityRates: null` is what you're looking for.

PR #448, opened 2017-11-22
Modified File Blacklist for Travis

### LINKED ISSUE #347: Integrate Modifed-File Blacklist Checking into CI
After audit, changes to crypto-specific files should be automatically failed by CI.

A CI service (perhaps similar to the site-diff tool @skubakdj has been working on) should be created that will allow for configuration of specific files that should not be allowed to have changes without explicit override (in practice, just fail the CI status check).

skubakdj:
## What

Have Travis fail if a PR build contains changes to files that have been previously marked as `frozen` (closes #347). 

## How

This PR introduces two new NPM scripts: `freezer` and `freezer:validate`. 

`freezer` will perform a diff between the current branch and `develop`, enumerate any changed files, and throw if any of these files match those specified in the `frozen` property of the `package.json` from the `develop` branch. 

`freezer:validate` will loop through the `frozen` property of the current branch's `package.json`, check that the specified files and folders exist, and throw if any invalid configuration is found. This should prevent us from committing any bad config. 

Files are specified using the project's relative path. Folders are the same, but with `/*` appended. Here's an example `package.json` snippet:

```
...
  "frozen": [
    "common/libs/decrypt.ts",
    "common/libs/wallet/*"
  ],
...
```

I've forked the repo for ease of testing. [You can see these commands in action here](https://github.com/skubakdj/MyEtherWallet/pull/1/). Note: these checks will be skipped on Travis `push` builds and only run during `pr` builds. 

I have some more testing I'd like to do before this is merged, but any feedback would be appreciated!


common/freezer.ts
```diff
@@ -0,0 +1,176 @@
...
+    }
+
+    const newFiles = getNewFiles(diff);
+    const frozenFiles = getFrozenFiles(frozen);
+    const frozenFolders = getFrozenFolders(frozen);
+
+    ensureNewFilesAreNotFrozen(newFiles, frozenFiles, frozenFolders);
+  } catch (err) {
+    console.log(err.message);
+    exit();
```
wbobeirne: This should be an `exit(1)` to ensure that the build fails if something's wrong with freezer.

webpack_config/webpack.freezer.js
```diff
@@ -0,0 +1,22 @@
...
+  },
+
+  // Point at freezer, make sure it's setup to run in node
+  target: 'node',
+  entry: {
+    'freezer': './common/freezer'
+  }
+});
+
+module.exports = derivationConfig;
```
wbobeirne: Are there features that webpack offers that node doesn't have that are valuable enough for the added build time & complexity of this? I feel like node alone should be good enough.

common/freezer.ts
skubakdj: The `exit` here is a constant that calls `process.exit(1)`, ensuring a failing build on error.

webpack_config/webpack.freezer.js
skubakdj: Webpack is used here to compile TS to JS, so it's necessary with the current setup. It would be pretty easy though to move to pure JS if that would be preferable.

skubakdj:
Good point on the potential dismantling. Not exactly sure either on how to completely solve for that, but I'll give it some more thought. Working this into an NPM module sounds like a good first step.

I'm not sure the two commands should be merged since the run result of `freezer` depends on config coming from `develop`, whereas `freezer:validate` checks against your working branch. You could have invalid config that fails `freezer:validate` but has no impact on `freezer`.

PR #454, opened 2017-11-24
Show confirm transaction copy during hardware wallet signing

james-prado:
Closes https://github.com/MyEtherWallet/MyEtherWallet/issues/270

* Prompt user to confirm transaction when using hardware wallet


HenryNguyen5:
This is a great idea, but i think it'd be better to be based on `transaction-refactor`
I could merge it in there.

james-prado:
@HenryNguyen5 Sure, I don't see why not

dternyak:
Merging into develop after discussion (`transaction-refactor` will be in development for a long enough period of time that it is worth merging as-is).

common/containers/Tabs/SendTransaction/index.tsx
```diff
@@ -265,6 +271,8 @@ export class SendTransaction extends React.Component<Props, State> {
         ? getDecimal('ether')
         : (this.state.token && this.state.token.decimal) || 0;
     const isWeb3Wallet = this.props.wallet instanceof Web3Wallet;
+    const isLedgerWallet = this.props.wallet instanceof LedgerWallet;
+    const isTrezorWallet = this.props.wallet instanceof TrezorWallet;
```
dternyak: a utility function could be helpful as additional hardware wallets become supported, but its not worth making holding up the PR.

Issue #379, opened 2017-11-11
Offline polling unreliable


HenryNguyen5:
Currently we use `navigator.onLine` to check if the browser is in an offline state, but that might or might not mean we can access node RPC calls.

According to: https://developer.mozilla.org/en-US/docs/Web/API/NavigatorOnLine/onLine
For chrome / safari browsers, even if the browser does not have an internet connection but is just on a LAN it counts as being online. In this case any RPC calls made would fail anyway.

Another case is that the user has no internet connection, but has a local node running that the user is connected to. In this case they would still be able to broadcast a transaction despite no internet.

IMO, the best way to check if they're online would probably to send a `ping` type request to the currently connected node to see if MEW is able to send RPC calls.

PR #478, opened 2017-11-28
Better Offline Detection / Handling

### LINKED ISSUE #421: Flashing "Balance isn't available offline"
I'm not sure what determines this, but I see "Balance isn't available offline" flashing while balances are loading in the mnemonic wallet. We should probably have a better way of determining if the user is offline than what's currently in place.

screen shot 2017-11-17 at 4 34 56 pm

wbobeirne:
This closes #421 and #379. It does the following:

1. Implements a `NodeLib.ping()` function for nodes, and pings that whenever navigator.onLine changes. Coming back with anything except a properly formed JSON response will result in flipping you into an offline state.
    * This should account for HTML login redirects or blocked traffic at hotels & airports etc.
    * Going into forced offline ceases the check.
2. Notifications for when the user is flipped to online or offline states by a failed ping.
3. Connects the `UnitDisplay` component and uses the `state.config.offline` attribute for showing the offline text. If the value is null, show a loader instead, because presumably it's still fetching the value.
    * An alternative to this would be to allow for sending either a `Balance` object, or the `isPending` attribute on it for this. Let me know if that's preferable.
4. A minor alteration of notifications, so that if you want duration to be explicitly infinite, you pass the number `Infinity`, not the string `'infinity'`.

## Example of going offline

<br/>

offline

<br/>

## Example of the UnitDisplay loader

<br/>

balance-loader


wbobeirne:
As per some discussion with @dternyak, the logic has been adjusted such that it only pings the node:

1. Once initially
2. If `navigator.onLine === false` and `state.config.offline === false`, and checks if the node cannot be pinged. If so, toggles `state.config.offline = true`
3. If `navigator.onLine === true` and `state.config.offline === true`, and checks if the node *can* be pinged. If so, toggles `state.config.offline = false`
    * If it cannot be pinged, it will try again every 5 seconds.

Issue #462, opened 2017-11-27
CryptoCompare equivalent values API needs to be batch-requested


wbobeirne:
## Updated issue

Now that the API is more resilient, we'll need to handle the 500+ characters use case if you request an obscene number of tokens. We'll simply need to split the request into multiple groups, and combine all of the results at the end.

## Old issue

If you load a wallet that has tokens that CryptoCompare doesn't know about, the price API call fails:

https://min-api.cryptocompare.com/data/price?fsym=ETH&tsyms=USD,EUR,GBP,BTC,CHF,REP,ETH,ANT,BeerCoin%20%F0%9F%8D%BA%20,CryptoCarbon,DAO,DGD,DGX%201.0,EDG,EMV,GNT,GUP,HMQ,ICN,JET,JetCoins,MCAP,MCO,MDA,MGO,MKR,MLN,NxC,OMG,PAY,PLU,PTOY,QRL,REP,RLC,RLT,ROUND,SGT,SHIT,SNGLS,SNT,STORJ,SWT,TIME,TRST,TaaS,Unicorn%20%F0%9F%A6%84%20,VERI,VRS,VSL,WINGS,XAUR

```
{"Response":"Error","Message":"tsyms param seems to be invalid.","Type":1,"Aggregated":false,"Data":[]}
```

The immediate problem is that the fetch call doesn't catch this (It comes back as a 200 response) The bigger problem is that one bad token kills this feature entirely. I have a few possible solutions:

1) Get a white-list of CryptoCompare tokens. This kinda sucks, and would have to be updated frequently.
2) Make the request, and if it fails, make the request again with _only ETH_. Make EquivalentValues show only ETH, and only allow them to select per token, not all tokens + ETH.
3) Try to work out better handling from CryptoCompare's side. Either an OPTIONS endpoint that describes which tokens it acknowledges, or have it return even with bad tokens, just omit them from the response.

cc @tayvano on number 3, do we have a relationship with them?

vcealicu:
I'll fix it on our side. We'll just omit the tokens that don't trade. That's what we do for the pricemultifull path. I'll implement the same logic on this.

vcealicu:
I found the bug.... it's just a limit that is being hit and I return the wrong error. The actual error should be: the tsyms length is higher than 100 (that's the current limit). 

I'm pushing out a change that will return the correct error and increase the limit to 500 characters, so you can request on average around 80-100 assets per call (depending on the ticker length of each asset). 

[...]

PR #486, opened 2017-11-28
Throw on CryptoCompare API error

wbobeirne:
This is a small bandaid for #462, CryptoCompare price error responses come back with valid json and a 200 status code so they were running through the rest of the code as though it were a success response. This caused an uncaught error to bubble up. This just throws if the API comes back with an error, causing the failure case for this action to trigger.


dternyak:
I'm inclined to wait until vlad updates CC backend as specified in https://github.com/MyEtherWallet/MyEtherWallet/issues/462 and implement paged token requests

wbobeirne:
We'll definitely want to implement that, but this change handles any class of error from their API, not just this one. The current behavior is to very incorrectly end up with rates that are `undefined`, which throws uncaught killing the site in dev and could have a pretty bad user experience in production. I think we should keep this, but not close out #462 and finish fixing that once the change is live.

PR #500, opened 2017-11-30
Equivalent Values Fixes

wbobeirne:
This PR solves two issues:

1. There's a bit of caching going on to prevent re-requesting rates if we're not asking for new data. The initial value of the cache is for no tokens. So if you had no tokens to request (ETH only wallet,) it assumed the rates were up to date, preventing the request. The fix here is to initialize the cache as null, not empty array. That indicates there's no cache yet, so it runs the first time regardless.
2. Non ether-like decimal tokens (where the decimal wasn't 18) were getting their equivalent values compared as though they were. So a decimal 0 token (Like SNGLS) was being shown at value/10^18. This just adjusts the equivalent value to match as though it were decimal 18 for display only.


common/components/BalanceSidebar/EquivalentValues.tsx
```diff
@@ -197,8 +202,15 @@ export default class EquivalentValues extends React.Component<Props, CmpState> {
       return {};
     }
 
+    // Tokens with non-ether like decimals need to be adjusted to match
+    const decimal = this.decimalLookup[currency];
+    const adjustedBalance =
```
HenryNguyen5: is it possible to use functions inside `libs/units` instead of this?

wbobeirne: This now uses `toTokenBase`, however I think that I may not be using it as intended. I'm just using it to shift decimals, not to convert it to a particular tokens decimal. Perhaps the function should be renamed to be more general, or I should make a new function that makes it a little more clear what I'm doing, maybe something like `fromDecimalBaseToDecimalBase(value, decimalOld, decimalNew)`.

wbobeirne: Made a function `convertTokenBase` that does behavior we discussed in Slack. Also wrote unit tests for it. Should be cleaner now!

PR #515, opened 2017-12-03
Fix Prettier Configuration

HenryNguyen5:
This PR adds a config file that prettier will use, keeping everyone's config the same


dternyak:
Tested, confirmed works. No more formatting wars! (after we fix the existing).

Issue #494, opened 2017-11-30
Token Redesign


dternyak:
- Flag will be set for each account address that is initially unlocked (flag will be the hash of the account address in local storage to protect info leakage via local storage).
- When flag is not present, request all default token balances and provide a UI for the user to choose which tokens he would like to monitor in the sidebar (default to all).
- When user enters token-choosing UI post-initial configuration and select additional token options, request balance for each token.
- User can activate token-choosing configurator to re-request all token balances to auto-discover all non-zero balances with helper button.

Notes:
- should significantly reduce backend load due to only requested "selected" tokens stored in LS after user configures which tokens they would like to monitor.

dternyak:
@wbobeirne These mockups look great to me!

wbobeirne:
This is merged into `transaction-refactor`, so I'm marking this closed, though it's not in develop yet.

Issue #341, opened 2017-11-07
Global Account Unlock


dternyak:
At its core, this proposal addresses the issue of treating a user's wallet as an account. Once a wallet is authenticated, it can be accessed across the site. Implementation details aside, this change will pave the way for a dashboard design, and thats a good thing.

Before any wireframes or tough decisions are made, I think some research into how to build a dashboard should be done. Looking through dashboards ([coinbase](https://www.coinbase.com/), [shopify](https://www.shopify.ca/), [stripe](https://stripe.com/ca), [trezor](https://trezor.io/), etc...) other design teams have built, sketching prototypes, and eventually creating wireframes and mockups are all part of that process. Going through this design process is an investment, but the payoff for the users, and the designers / developers down the line will be worth it.

I'm absolutely behind this proposal and I'm open to collaborating on solutions. - @james-prado

dternyak:
UX Proposals #1

Account Sidebar (similar to Trezor UX).
    - Disabled tabs based on if account unlock is required to use.
         - Send Required
         - Swap not required

james-prado:
At its core, this proposal addresses the issue of treating a user's wallet as an account. Once a wallet is authenticated, it can be accessed across the site. Implementation details aside, I think this change will pave the way for a dashboard design. 

Before any wireframes or tough decisions are made, I think some research into how to build a dashboard should be done. Looking through dashboards ([coinbase](https://www.coinbase.com/dashboard), [shopify](https://www.shopify.ca/), [stripe](https://stripe.com/ca), [trezor](https://trezor.io), etc...) other design teams have built, sketching prototypes, and eventually creating wireframes and mockups are all part of that process. Going through this design process is an investment, but the payoff for the users, and the designers / developers down the line will be worth it. 

I'm absolutely behind this proposal and I'm open to collaborating on solutions.

dternyak:
I Didn't mean to delete @james-prado's comment, but it's a good enough summary of the issue that I've moved it into the main issue description.

wbobeirne:
See #434 for some previous discussion & mockups.

james-prado:
Issue moved to https://github.com/MyCryptoHQ/research/issues/3

PR #520, opened 2017-12-04
Local storage wallet config infrastructure

wbobeirne:
The following PR adds some infrastructure around saving configurations per wallet. Rather than making this a part of an upcoming PR that will need it, I wanted to keep it separate so that we could discuss how this could best be implemented. Let me lay out some of the use-case I see for it:

### Short Term

#494 will need to keep track of which tokens each wallet is watching, so it needed some way to store wallet information. I could make some kind of `${address}-tokens` address, but I figured having one unified way of storing information per-wallet would be better, because...

### Medium Term

#341's mockups have all shown some amount of custom configuration per wallet. Keeping track of what wallet type they are, naming them, maybe assigning icons. We'll probably want to have a quick and easy way to allow users more control over their individual wallets. As for the local storage key, I hash and salt the address rather than just save it as the address key because...

### Long Term

I could see us setting up a redis store of configs. In the future, ideally when you unlock a wallet, it carries with it all of your configuration. This will make MEW feel really cohesive across all your devices, whether it's an app, electron, or web instance of it. We probably won't want to associate any possibly identifying information with the configs, so hashing the address will ensure that even if redis is compromised, we aren't leaking user data.

---

Anyway, looking for discussion on my implementation (Especially the address hashing, should we use something better than this?) and the general ideas outlined above. Let me know if you've got an upcoming use-case that would need a wallet config, and if this covers that case.


common/reducers/wallet.ts
```diff
@@ -1,17 +1,13 @@
...
+import { SetTokenBalancesAction, SetWalletAction, WalletAction, TypeKeys } from 'actions/wallet';
 import { TokenValue } from 'libs/units';
 import { BroadcastTransactionStatus } from 'libs/transaction';
-import { IWallet, Balance } from 'libs/wallet';
+import { IWallet, Balance, WalletConfig } from 'libs/wallet';
 import { getTxFromBroadcastTransactionStatus } from 'selectors/wallet';
 
 export interface State {
   inst?: IWallet | null;
+  config?: WalletConfig | null;
```
dternyak: Confused about the shape of WalletConfig: `tokens?: string[] | null;` 

`tokens` is already a property - will that be removed in favor of WalletConfig? 

I would have expected something more along the lines of:
```
interface Accounts {
    [key: string]: {
        [token: string]: TokenObj
    }
}

accounts?: Accounts | null```

Comment (wbobeirne) [reply common/reducers/wallet.ts]: Yeah, now that I look at it, the nomenclature is pretty bad. What is currently on `state.wallet.tokens` is actually just token _balances_ (As well as an error state.) I should probably rename that property `tokenBalances` and just make them balance objects.

The tokens object on the config would just be an array of token symbols, i.e. `["BAT", "GNT", ...]` that will flag which tokens we're currently tracking for that wallet.

I didn't want to store full token data in local storage (address, decimal etc.) in case we ever change it or expand on it (icons, written names, website addresses etc.) so that it just looks up the token object each time.

Comment (wbobeirne) [reply common/reducers/wallet.ts]: And to be more clear about why this is under `WalletConfig`, I think the future of `WalletConfig` will look something like:

```
interface WalletConfig {
  name?: string | null;
  tokens?: string[] | null;
  type?: keyof WalletTypes | null;
  syncConfigToCloud?: boolean;
}
```

that keeps track of user preferences or other UX-centric things to make wallets more persistent.

dternyak: Got it, much more clear now.

wbobeirne:
Closing this issue to roll it into the branch built off of `transaction-refactor` for issue #494

Issue #219, opened 2017-09-20
Improve UX of high n value keys


wbobeirne:
From https://github.com/kvhnuke/etherwallet/issues/486, the browser locks up (especially on slower computers) so let's see if we can async this.

tayvano:
This was sent to me via twitter maybe like 6 months ago as a "this is how it could be handled better but this lib isnt production ready" : https://github.com/ryancdotorg/scrypt-async-js

HenryNguyen5:
Yeah, looks like they're doing the setTimeout trick

PR #548, opened 2017-12-11
[WIP] Attempt empty password for keystore unlock before prompting for pass

aitrean:
Closes https://github.com/MyEtherWallet/MyEtherWallet/issues/423

Hi guys. Since Henry and James were having so much fun with MEW I thought I'd try my hand at a few of your issues - if that's alright. This solution is for issue https://github.com/MyEtherWallet/MyEtherWallet/issues/423 which I tested against some keystore files I generated with ethereumjs-wallet. It's working, however, there is a 1 - 2 second pause when we decrypt an empty-password encoded file. There doesn't seem to be any indicator that the wallet is being generated either, so there is a 1 -2 second period where it doesn't look like the UI is responsive. In V3, it looks like you guys use a notifier to indicate the file has been received, but I'm not seeing that notification in this build. Is that an issue?

### DIFF
--- a/common/Root.tsx
+++ b/common/Root.tsx
@@ -27,7 +27,7 @@ export default class Root extends Component<Props, State> {
     error: null
   };
 
-  public componentDidCatch(error) {
+  public componentDidCatch(error: Error) {
     this.setState({ error });
   }
 
--- a/common/actions/rates/actionPayloads.ts
+++ b/common/actions/rates/actionPayloads.ts
@@ -1,31 +1,39 @@
 import { handleJSONResponse } from 'api/utils';
 
-export const rateSymbols = ['USD', 'EUR', 'GBP', 'BTC', 'CHF', 'REP', 'ETH'];
+export const rateSymbols: Symbols = ['USD', 'EUR', 'GBP', 'BTC', 'CHF', 'REP', 'ETH'];
+
+export type Symbols = (keyof ISymbol)[];
 // TODO - internationalize
 const ERROR_MESSAGE = 'Could not fetch rate data.';
 const CCApi = 'https://min-api.cryptocompare.com';
 
 const CCRates = (symbols: string[]) => {
-  const tsyms = rateSymbols.concat(symbols).join(',');
+  const tsyms = rateSymbols.concat(symbols as any).join(',');
   return `${CCApi}/data/price?fsym=ETH&tsyms=${tsyms}`;
 };
 
 export interface CCResponse {
-  [symbol: string]: {
-    USD: number;
-    EUR: number;
-    GBP: number;
-    BTC: number;
-    CHF: number;
-    REP: number;
-    ETH: number;
-  };
+  [symbol: string]: ISymbol;
+}
+
+interface ISymbol {
+  USD: number;
+  EUR: number;
+  GBP: number;
+  BTC: number;
+  CHF: number;
+  REP: number;
+  ETH: number;
+}
+
+interface IRates extends ISymbol {
+  Response?: 'Error';
 }
 
 export const fetchRates = (symbols: string[] = []): Promise<CCResponse> =>
   fetch(CCRates(symbols))
     .then(response => handleJSONResponse(response, ERROR_MESSAGE))
-    .then(rates => {
+    .then((rates: IRates) => {
       // API errors come as 200s, so check the json for error
       if (rates.Response && rates.Response === 'Error') {
         throw new Error('Failed to fetch rates');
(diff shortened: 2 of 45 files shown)


aitrean:
Looking at it again, the pause between uploading a password-empty file and loading the wallet can be more significant than just 1 - 2 seconds, up to 30 seconds sometimes. It looks like a lot of this time is waiting for the ethereumjs-wallet to generate the wallet with the empty passphrase. I can see similar wait time in your live version when I upload a password-empty encrypted keystore, but here it just looks unresponsive since we immediately try the empty passphrase. Perhaps a UI indicator would help to indicate the wallet is loading?

dternyak:
Thanks for your contribution and interest @aitrean!

Great point about the UI being un-responsive while decrypt is taking place. Unfortunately, the call itself is synchronous, and thus blocks the UI, so we aren't able to show a loading indicator (or at the very least, the loading indicator would be blocked as well).

Let me do some research and decide if there is a way to ensure the UI is responsive while we attempt to decrypt the wallet. 

We definitely don't want to hurt overall UX by trying to fix a relatively un-common use-case.

HenryNguyen5:
You could use `setTimeout` with 0 `ms` to push the sync call onto the event loop, preventing the UI from being blocked. And then place a spinner component while it's waiting

dternyak:
In this case, we'd be showing a spinner to all users (while the decryption runs) who have uploaded a keystore, which is probably not the desired UX.

aitrean:
@HenryNguyen5 I'll look into that

@dternyak Perhaps I'm misunderstanding, but would a spinner on the unblocked UI be detrimental to users with real passwords? Files encrypted with real passwords immediately reject the blank password, so they wouldn't see the spinner. Also, if you only unblock the UI, it will look like the interface didn't acknowledge your file change. If not a spinner, what about those notifications you have in the bottom of your live version when a file gets uploaded? I don't see them in this build.

dternyak:
@aitrean Got it - as long as the existing flow for users with passwords doesn't change (e.g. they don't briefly see a loading indicator and then immediately disappear) that approach should be fine 👍

dternyak:
The issue is that there are encrypted keystores with empty passphrases, so I don't believe you can just check with `isPassRequired` -- you actually have to attempt the decrypt with an empty passphrase, which would block, requiring you to show a loading indicator to all users.

HenryNguyen5:
Another (arguably cleaner and more scalable) solution would be to use webworkers for decryption tasks as it's CPU heavy

aitrean:
@dternyak Gotcha. I like Henry's idea to use a webworker for handling the ethereumjs library decryption behaviour. To address the ultra-brief spinner appearance on password-encrypted files, could we set up a timer to wait for a webworker? If the webworker doesn't signal within x seconds, load the spinner so that the user doesn't think the UI has ignored them.

dternyak:
@aitrean I haven't worked with web workers much before, but I think this should work in theory. @wbobeirne I think you may have had some web worker experience iirc - do you happen any thoughts?

wbobeirne:
It definitely is a good use case for workers, but it's not exactly a very trivial refactor. Workers take in completely different files, so this would mean [adding a new loader to webpack](https://github.com/webpack-contrib/worker-loader), and splitting out our decrypting logic into their own files that then get output and loaded in separately. I think all of this could be reasonably be tucked away into a wrapper async function, so there'd be little change on the component side.

I think this'd be a great way to go, as it'd also close out #219.

common/sagas/wallet.tsx
```diff
@@ -116,7 +116,9 @@ export function* unlockKeystore(action: UnlockKeystoreAction): SagaIterator {
   try {
     wallet = getKeystoreWallet(file, password);
   } catch (e) {
-    yield put(showNotification('danger', translate('ERROR_6')));
+    if (password !== '') {
+      yield put(showNotification('danger', translate('ERROR_6')));
+    }
```
wbobeirne: Even if the password is blank, it could throw for other reasons (Invalid file?) We should ideally check if the password is blank _and_ if it's a certain type of error.

aitrean: Woops! I should have seen that. Thanks for the feedback!

aitrean: Actually, testing this issue again without my changes, I see that there's no error at all for incorrect file uploads. I can upload jpgs to the keystore and it will give me an unlock button. Can't find this issue logged. Shall I open one?

dternyak: Yes, please do 👍

PR #552, opened 2017-12-11
Mnemonic Delimiters

timeffect:
Closes https://github.com/MyEtherWallet/MyEtherWallet/issues/541


common/components/WalletDecrypt/Mnemonic.tsx
```diff
@@ -86,18 +88,21 @@ export default class MnemonicDecrypt extends Component<Props, State> {
   };
 
   public onMnemonicChange = (e: React.SyntheticEvent<HTMLTextAreaElement>) => {
-    this.setState({ phrase: (e.target as HTMLTextAreaElement).value });
+    const phrase = (e.target as HTMLTextAreaElement).value;
+    const formattedPhrase = phrase.replace(/(\r\n|\n|\r|,)/gm," ").trim();
+    this.setState({ phrase });
+    this.setState({ formattedPhrase })
```
dternyak: can combine into one setState

common/components/WalletDecrypt/Mnemonic.tsx
```diff
@@ -19,14 +20,15 @@ interface State {
...
+    formattedPhrase: '',
     pass: '',
     seed: '',
     dPath: DEFAULT_PATH
   };
 
   public render() {
-    const { phrase, seed, dPath, pass } = this.state;
-    const isValidMnemonic = validateMnemonic(phrase);
+    const { phrase, formattedPhrase, seed, dPath, pass } = this.state;
```
dternyak: can remove phrase from de-structure (unless I'm missing something further down)

common/components/WalletDecrypt/Mnemonic.tsx
```diff
@@ -86,18 +88,21 @@ export default class MnemonicDecrypt extends Component<Props, State> {
   };
 
   public onMnemonicChange = (e: React.SyntheticEvent<HTMLTextAreaElement>) => {
-    this.setState({ phrase: (e.target as HTMLTextAreaElement).value });
+    const phrase = (e.target as HTMLTextAreaElement).value;
+    const formattedPhrase = phrase.replace(/(\r\n|\n|\r|,)/gm," ").trim();
```
dternyak: Move regex into utility file. Would be nice to see if lodash has something like this out of the box to avoid rolling our own regex if possible.

dternyak:
Thanks for the PR @timeffect! I left a few comments :)

common/components/WalletDecrypt/Mnemonic.tsx
```diff
@@ -86,19 +88,21 @@ export default class MnemonicDecrypt extends Component<Props, State> {
   };
 
   public onMnemonicChange = (e: React.SyntheticEvent<HTMLTextAreaElement>) => {
```
HenryNguyen5: I know this isnt part of your PR, but it should be `React.FormEvent<HTMLTextAreaElement>`, then you won't need to typecast `e.target` as you can use `e.currentTarget.value` instead

common/components/WalletDecrypt/Mnemonic.tsx
wbobeirne: Unit tests would also be good for the utility function

common/components/WalletDecrypt/Mnemonic.tsx
timeffect: it's used at least here: https://github.com/MyEtherWallet/MyEtherWallet/blob/develop/common/components/WalletDecrypt/Mnemonic.tsx#L41

PR #569, opened 2017-12-13
Minor Visual Touchups

### LINKED ISSUE #521: Improve global link / form element transitions
This has been bugging me for a while, and I think it's worth doing. Right now we're doing a a 500ms transition on _all_ links and _all_ form elements, on _all_ attributes. I think we should do the following:

* Set a `transition-property` to determine what attributes should be animated. `color, background-color, box-shadow, opacity, transform` seems like a good set. Width, height, margin, padding etc should not be animated on for performance and responsive breakpoint reasons.
* Reduce the default timing. 500ms feels really sluggish, in my opinion. I'd say something like 150ms will feel snappier.
* Give buttons the global transitions as well

Let me know if this is an unpopular opinion.

wbobeirne:
Closes #521. In an effort to modernize the site a bit, differentiate it from v3, and generally just make things look better, I've made a quick hit-list of visual improvements to the site. Each is its own commit, so if there's anything that people don't like, it should be easy to revert. Please check out the branch and browse around, it's hard to capture _everything_ in screenshots. But here are the changes with screenshots:

## Give Navigation Background

I've always hated that this was floating in space, so I gave it a background.

nav

## Adjust Button Styles

They're now slightly rounded, have quick animations to transition between states, and the default color is now white. I'm also on a mission to remove as many of the `btn-info`s as I can, I really don't think we need that additional tier of color information. We should ideally only ever be providing two button options, a primary / success / warning / danger button, and a default cancel button.

buttons

## Reduce Shadows on White Blocks

The sprawling shadows on these didn't give them much definition, slightly rounded corners and tighter shadows should make these stand out a little more.

blocks

## Adjust Form Validation Styles

Generally reduced the neon glow of our forms to keep things tighter like everything else.

forms

## Adjust Notification Spacing and animation

Notifications have a little more padding, now stack directly on each other instead of having 20px margins, and animate faster

notifications

## Speed Up Global Transitions

Down to 120ms from 500ms, limited to opacity and colors, and now applied to buttons. Makes most things feel snappier, and reduces resizing jank from widths and heights being animated.

---

Let me know if you disagree with any of these adjustments. Expect more changes to come for specific components that need improvement, this PR is only for global style improvements.

### DIFF
--- a/common/components/Header/components/CustomNodeModal.tsx
+++ b/common/components/Header/components/CustomNodeModal.tsx
@@ -64,6 +64,7 @@ export default class CustomNodeModal extends React.Component<Props, State> {
         disabled: !!Object.keys(invalids).length
       },
       {
+        type: 'default',
         text: translate('x_Cancel'),
         onClick: handleClose
       }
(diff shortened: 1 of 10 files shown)


dternyak:
Solid improvement 👍 

We can iteratively improve with more feedback after the next alpha point release with these updates includes.

PR #587, opened 2017-12-14
Update icon fonts & Fix small css issues

james-prado:
This PR isn't related to any open issues. I cherry-picked these commits off a larger pr I'm working on, they should be easier to manage here

- Rename the social media icon font & expand it to include all icons
- Add svg icons to the icon font, 'eye', 'eye-closed', and 'question-circle'
- Fix misaligned margins on swap dropdowns for small screens
screen shot 2017-12-15 at 4 18 33 pm,
- Replace old `` components with `<i />` & Update css accordingly


common/sass/fonts/icons/_mixins.scss
```diff
@@ -0,0 +1,17 @@
...
+}
+
+@mixin nc-flip($horiz, $vert, $rotation) {
+  filter: progid:DXImageTransform.Microsoft.BasicImage(rotation=#{$rotation});
+  -webkit-transform: scale($horiz, $vert);
+  -moz-transform: scale($horiz, $vert);
+  -ms-transform: scale($horiz, $vert);
+  -o-transform: scale($horiz, $vert);
+  transform: scale($horiz, $vert);
+}
```
wbobeirne: You can ditch these mixins, the browsers that need this level of fallback don't support the JS features needed to run the site anyway. And if we do need fallbacks, we should be using something like [cssnext](http://cssnext.io/).

common/sass/fonts/icons/icons.scss
```diff
@@ -0,0 +1,44 @@
...
+
+.nc-icon.nc-alert-circle-que:before {
+  content: "\ea0f";
+}
+
+.nc-icon.nc-eye-19:before {
+  content: "\ea10";
+}
+
+.nc-icon.nc-eye-ban-20:before {
```
wbobeirne: What are the numbers for here? Can they just be `nc-eye` and `nc-eye-closed` or something?

common/sass/fonts/icons/nucleo.scss
```diff
@@ -0,0 +1,167 @@
...
+    transform: rotate(0deg);
+  }
+  100% {
+    -webkit-transform: rotate(360deg);
+    -moz-transform: rotate(360deg);
+    -ms-transform: rotate(360deg);
+    -o-transform: rotate(360deg);
+    transform: rotate(360deg);
+  }
+}
```
wbobeirne: Same as above, no need to vendor prefix.

james-prado:
I didn't realize we were using font-awesome icons on the site currently. I inspected a few icons and found them to be individual svgs and thought it'd be better to replace them with an icon font and open up a discussion about what icons we should use, and how to use them (icon fonts, individual svgs, etc...)

common/sass/fonts/icons/_mixins.scss
james-prado: yup they aren't necessary, they're just auto-generated mixins I copy pasted [nucleo](https://nucleoapp.com/)

common/sass/fonts/icons/icons.scss
james-prado: They correspond to the icon names in the [nucleo](https://nucleoapp.com/) icon-font, I'll go back to using font-awesome and change the names accordingly

common/containers/Tabs/GenerateWallet/components/EnterPassword.scss
```diff
@@ -1,4 +1,4 @@
-@import "common/sass/variables";
+@import 'common/sass/variables';
```
dternyak: only formatting change

wbobeirne:
+1 on opening the discussion. I definitely don't think we should move forward with multiple ways of doing icons. I'm concerned about relying on Nucleo for our process, since it's a paid app with no option for linux developers.

My vote would be to convert all of our existing icons to https://www.npmjs.com/package/react-fontawesome (Or something like it, maybe https://github.com/gorangajic/react-icons) instead of relying on classnames. Should keep payload size down (only include the icons we need,) make adding new icons simple (just include a new component,) and reduce the guesswork and lookup of classnames.

How would you feel about making that us making that anew issue, to convert existing nucleo icons and fontawesome icons all to react components?

tayvano:
The downside of font icons is you are limited to a single color and a single library. I would recommend sticking with SVGs as moving forward, the icons are going to be more and more mutli-colored.

Font Awesome and font icons in general are extremely bloated. Font Awesome specifically is atrociously bloated and adding another 1-2MB+ when we have maybe 10 single-color icons total seems like waste. This is why you are seeing them move to the [SVG w/ JS](https://fontawesome.com/how-to-use/svg-with-js) system which, while it solves the bloated issue, still has some kinks to work out.

In order to keep the source clean and reduce external requests to 0, inline the SVGs and call it a day. It's worth cleaning up the organization of these icons in general but it's been a long time since I saw any advantage to using a font icon over inlined SVGs. 

I made most of the icons myself or they came from https://www.flaticon.com/. I have a premium account with them and can switch that to be a shared MEW account  so you guys can easily pull any SVGs icons. I use https://www.npmjs.com/package/svgo after downloading the SVG to minify them (there's a sublime text plugin called Minify that makes it super easy, but you could do it in the build process as well)

wbobeirne:
The bloat is a non-issue with react components, you only include what you need (FontAwesome's, for example: https://github.com/FortAwesome/react-fontawesome#explicit-import.) Either way, I don't much care what the underlying graphic is (font, svg asset, inline svg), as long as we tuck away the implementation into easily reusable components and have props for the aspects that are configurable.

james-prado:
I'm open to using other ways to organize icons instead of using icon-fonts, I think having some system in place to organize and maintain icons is important. A good alternative is [SVG symbols](https://css-tricks.com/svg-symbol-good-choice-icons/). Nucleo also supports exporting as SVG symbols.

Icon-fonts are bloated, but using Nucleo we can explicitly export only the icons we need, even if the icons come from different icon sets.

Nucleo does support multi-color icons (even as icon-fonts), and importing multiple icon sets. Individual svg's can also be added and managed like any other icon.

Nucleo does cover many of the implimentation details (creating icon-font files, svg symbol files, etc..), but doesn't go as far as making a react component for icons like the [font-awesome react component](https://github.com/FortAwesome/react-fontawesome). I don't think it would take much effort to make a similar react component to manage the implimentation details of our icons.

Nucleo doesn't offer the same [SVG w/ JS](https://fontawesome.com/how-to-use/svg-with-js) solution font-awesome does, but still provides developers with raw svg's they can quickly copy paste for inline svgs if we want to use inline svg's.

 [Nucleo does cost money](https://nucleoapp.com/pricing), I've had a really great experience using it for years to manage icons. The best free alternative I found is [IcoMoon](https://icomoon.io/app/#/select), related tutorials for exporting [SVG symbols](https://css-tricks.com/svg-symbol-good-choice-icons/) available [here](https://css-tricks.com/svg-sprites-use-better-icon-fonts/#article-header-id-4).

@wbobeirne makes a good point, Nucleo currently does not have a linux app. It does have a webapp available but after taking a look it seems to be missing a few features.

I think Nucleo is great, but really it's more important to decide on a system of organizing icons, Nucleo and IcoMoon just make that process easier.

james-prado:
Issues have been solved by other pr's.

Issue #512, opened 2017-12-03
Mobile Web Support


dternyak:
- [x] Adjust CSS as time permits for better responsiveness on mobile (or at least ensure at parity with V3)
- [ ] Fix iOS 11 Safari Whitescreen

wbobeirne:
Just a reminder that we have a BrowserStack account, feel free to PM me in slack if you need to do some mobile browser testing. Using Chrome's responsive tools is a good start, but it's always good to make sure it's working fine on the real hardware.

wbobeirne:
Unfortunately BrowserStack doesn't have dev tools for iOS 11, but iOS 9 is getting

> SyntaxError: Invalid character '\u8544'

Looks like the js bundle might be getting some kind of malformed characters. Not sure if this is what's causing iOS 11's problem though.

dternyak:
Re-opening - iOS 11 white-screens onload.

PR #599, opened 2017-12-16
Fix iOS Safari with Polyfills

wbobeirne:
Fixes the whitescreen mentioned in #512 by adding polyfills. Unfortunately these balloon up the size a bit, but iOS safari is missing a ton in the last version or two. This adds the standard babel-polyfill, and a fetch polyfill. Not my favorite fix, but I think this is better than rejecting a ton of our mobile users.

While I was in the neighborhood, I also added error messages to the red screen of death. Makes it a ton easier to debug on mobile, and should make our user reports 100x better. Feel free to massage the copy:

screen shot 2017-12-15 at 7 33 27 pm

### DIFF
--- a/common/Root.tsx
+++ b/common/Root.tsx
@@ -21,25 +21,28 @@ interface Props {
 }
 
 interface State {
-  hasError: boolean;
+  error: Error | null;
 }
 
 export default class Root extends Component<Props, State> {
   public state = {
-    hasError: false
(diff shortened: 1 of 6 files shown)


dternyak:
Made a copy update in case the exception somehow reveals private information.

screen shot 2017-12-15 at 5 31 32 pm

PR #651, opened 2017-12-21
Add HashRouter for downloadable version

james-prado:
Closes https://github.com/MyEtherWallet/MyEtherWallet/issues/584

Reproduce Issue:
- Build production site with`npm run build`
- Serve production build via http-server, or access it from local file system (ie. open index.html in `dist`)
- Navigate to any sub route (`/account/info`)
- Refresh page

Changes:
- Replace `<Router />` with `<BrowserRouter />`
	- This creates a history object, so we don't have to manually add one.
- Use `<HashRouter />` for downloaded site
	- Read about some of the differences [here](https://medium.com/@pshrmn/a-simple-react-router-v4-tutorial-7f23ff27adf)
	- This solves the issue of routes breaking when the user refreshes a page if they've downloaded the site
- Add npm script `build:downloadable` to produce a production build for users to download
- Change publicPath for production build from `'/'` to `'./'` in webpack config based on env variable set by `npm run build:downloadable`
- Change path to assets from `'/...'` to `'./...'` in `index.html` by injecting them through webpack plugins (this allows the paths to be prefixed by the publicPath set in the webpack config)
	- These two changes to paths fix failed requests for assets on downloaded versions
- Add dev dependency `hoist-non-react-statics`, required to by react when building
	- see why [here](https://reactjs.org/docs/higher-order-components.html#static-methods-must-be-copied-over)


webpack_config/config.js
```diff
@@ -4,7 +4,9 @@ const path = require('path');
 module.exports = {
   port: process.env.HTTPS ? 3443 : 3000,
   title: 'MEW',
-  publicPath: process.env.BUILD_GH_PAGES ? '/react-semantic.ui-starter/' : '/',
+  publicPath: process.env.BUILD_GH_PAGES
```
dternyak: We may as well get rid of the `/react-semantic.ui-starter/` from the original boilerplate, and fix the nested ternary.

common/Root.tsx
```diff
@@ -32,33 +31,38 @@ export default class Root extends Component<Props, State> {
...
-            </Route>
-            <Route path="/send-transaction" component={SendTransaction} />
-            <Route path="/contracts" component={Contracts} />
-            <Route path="/ens" component={ENS} />
-            <Route path="/sign-and-verify-message" component={SignAndVerifyMessage} />
-            <Route path="/pushTx" component={BroadcastTx} />
-            <LegacyRoutes />
-          </div>
-        </Router>
+        {process.env.HTTPS || process.env.NODE_ENV === 'development' ? (
```
wbobeirne: I'd like to key in on something more declarative than checking the HTTPS environment variable. Maybe something like `process.env.DOWNLOADABLE` for HashRouter. I could easily see us changing some config related to HTTPS and breaking this part of the build.

package.json
```diff
@@ -78,6 +78,7 @@
     "file-loader": "1.1.6",
     "friendly-errors-webpack-plugin": "1.6.1",
     "glob": "7.1.2",
+    "hoist-non-react-statics": "^2.3.1",
```
wbobeirne: Thank you, I was sporadically getting build errors related to this module but had no idea what it was about.

webpack_config/config.js
```diff
@@ -4,7 +4,9 @@ const path = require('path');
...
+    ? '/react-semantic.ui-starter/'
+    : process.env.NODE_ENV === 'production' ? './' : '/',
```
wbobeirne: Would rather key in on a `DOWNLOADABLE` env variable. I think the idea here is that producing the downloadable version should be considered special, not the default.

common/index.html
```diff
@@ -8,14 +8,14 @@
...
-  <link rel="mask-icon" href="favicon/safari-pinned-tab.svg" color="#0e97c0">
-  <meta name="msapplication-config" content="/favicon/browserconfig.xml">
+  <script src="./dll.vendor.js"></script>
+  <link rel="manifest" href="./manifest.json">
+  <link rel="shortcut icon" href="./favicon.ico">
+  <link rel="apple-touch-icon" sizes="180x180" href="./favicon/apple-touch-icon.png">
+  <link rel="icon" type="image/png" href="./favicon/favicon-32x32.png" sizes="32x32">
+  <link rel="icon" type="image/png" href="./favicon/favicon-16x16.png" sizes="16x16">
+  <link rel="mask-icon" href="./favicon/safari-pinned-tab.svg" color="#0e97c0">
+  <meta name="msapplication-config" content="./favicon/browserconfig.xml">
```
wbobeirne: These changes will break the production site build, if you land on `myetherwallet.com/account` these will load `/account/dll.vendor...` etc. We'll need to find a way to link these using publicPath.

common/index.html
```diff
@@ -89,7 +63,10 @@ <h2>Your Browser is Out of Date</h2>
   </div>
 
   <script>
-    (function() {
+    (function () {
+      // check where the file is running
+      __webpack_public_path__ = window.location.protocol.startsWith('file') ? './' : '/'
```
wbobeirne: I don't see this used anywhere else, is this necessary?

HenryNguyen5:
Instead of having environment variables or flags for checking if the app is of `downloadable` version, it now checks itself if it's being served via a server or being opened as a file, and changes `publicPath` and router type appropriately. This makes testing easier too as we don't need to be rebuilding twice to test both versions.

common/index.html
wbobeirne: Ah, for context it looks like this was the intent: https://webpack.js.org/guides/public-path/#on-the-fly. I don't think this works though because this variable is only picked up on at compile time, but this js only runs on page load time.

This has the problem of when a page is run on a webserver (i.e. production) that any page with 2 levels of nesting will not load. I tested this by running the build through superstatic, a simple node webserver. As you can see in the screenshot, it loaded the assets at `/account/*` instead of the root level. Webservers need assets at `/`, where as the downloadable version needs them at `./` or no slash at all.

screen shot 2017-12-24 at 4 26 18 pm

wbobeirne: tl;dr I think the only reasonable way to do this is have a separate build for the downloadable version.

PR #658, opened 2017-12-23
Broadcast UI Cleanup

dternyak:
## What?

Fixes https://github.com/MyEtherWallet/MyEtherWallet/issues/657

Adds `onlyTransactionParameters` prop to SendButtonFactory to only show Transaction Parameters and not the serialized TX. 

Useful in Broadcast Signed Transaction so that the user is not presented with redundant data.

## Before
screen shot 2017-12-22 at 7 13 31 pm

## After
screen shot 2017-12-22 at 7 13 42 pm


wbobeirne:
If anyone would be willing, while we're in the neighborhood, I also think the styles on this one are a bit weird:

* It's left aligned, when it should be centered
* It uses an h2 for the title when it should be an h1
* The title has a huge margin bottom, while the help text is smushed to the content

I can make these changes myself shortly.

wbobeirne:
Fixed the above issues
screen shot 2017-12-23 at 12 23 53 pm

Issue #557, opened 2017-12-12
Discourage Private Keys / Keystore


dternyak:
- Make Send/Account the default view (with additional tweaks/messaging to make it viable as the landing page)
- Show messaging on the Generate Wallet tab to let users know we recommend metamask / hardware wallets instead of keystores. (Optional) Force user to acknowledge they do not want to use MetaMask. 
- Show messaging when selecting keystore / raw private key as a decrypt option about the insecurity of these wallet formats
     - Don't show this when user is offline
- [Controversial] Force user to wait N seconds and read warning when attempting to decrypt a keystore / private key for send
    - Don't show this when user is offline

tayvano:
```We strongly encourage you to use a [hardware wallet](https://myetherwallet.github.io/knowledge-base/hardware-wallets/hardware-wallet-recommendations.html) or [MetaMask](https://myetherwallet.github.io/knowledge-base/migration/moving-from-private-key-to-metamask.html) to interact with your wallet on MyEtherWallet.com. It is safer & easier. Using your private key, mnemonic, or keystore file can result in lost funds if you visit a fake MyEtherWallet.com. [Learn how easy it is to switch today ->](https://myetherwallet.github.io/knowledge-base/migration/)"```

+1 for forcing using to wait 10 seconds when decrypting (while showing a message like the above)—don't show if offline or on localhost / file

If you convert to password field, make sure there is an eye icon. We moved away from the password field because even if the very early days people seemed incapable of debugging when they had some extra characters or 0x at the beginning or something. 

Unlocking existing wallets online and on phishing sites is probably 95% of phishing. It's uncommon someone generates on a phishing site as they typically read guides or tutorials. Not to say we shouldn't encourage doing it offline, just that unlocking is by far the biggest problem.

wbobeirne:
> Convert input type to password (with option to reveal/eye icon) for:

Unfortunately these are both textareas, which do not support `type="password"`. We could make them inputs, or try to come up with a hacky way of replacing characters with dots. Let me know what's preferred, otherwise I'm just going to leave this out.

dternyak:
@wbobeirne Any reason not to go with an input? Seems fairly straightforward.

wbobeirne:
@dternyak after #925 is merged, do you think we should close this? Or do we want to implement any of the other notes from the original post?

dternyak:
[...]

PR #659, opened 2017-12-23
Generate Mnemonic Wallet

### LINKED ISSUE #109: Mnemonic Generation Support
Alongside raw private key and UTC wallet generation, it would be valuable for MEW to provide a mnemonic based wallet (potentially in the same format as MetaMask).

# Misc Notes
- [Metamask implementation](https://github.com/MetaMask/eth-hd-keyring)
- BIP 32 - core
- BIP 39 - wordlist
- BIP 44 - paths

To create a binary seed from the mnemonic, we use the PBKDF2 function with a mnemonic sentence (in UTF-8 NFKD) used as the password and the string "mnemonic" + passphrase (again in UTF-8 NFKD) used as the salt. The iteration count is set to 2048 and HMAC-SHA512 is used as the pseudo-random function. The length of the derived key is 512 bits (= 64 bytes).

So any passphrase you choose, gets concatenated with "mnemonic" and is used as a salt. ie. if your passphrase was "Th1sI5mYp@ssphr@se"... it would be "mnemonicTh1sI5mYp@ssphr@se" being used as the salt and an iteration count of 2048.

Also, as mentioned in the BIP, ANY passphrase provided will generate a valid seed... so an attacker would then need to check addresses for every generated/tested passphrase for coins further slowing down any attack.

-- BIP 39?

wbobeirne:
Closes #109, and addresses a lot of #557. This PR looks a lot bigger than it really is, so don't let the line count fool ya. This reworks the generate flow to give users the option between a keystore file and a mnemonic phrase. You can see #109 for the discussion around all of the changes, or you can just watch the following example videos:

### [Gfycat - Updated keystore wallet generation](https://gfycat.com/PlushAromaticHuman)

<br/>

There were a few additional changes related to this:

* Removed generateWallet redux stuff. Everything is handled in state now.
    * Rather than have two separate flows share the same reducer, or add a new reducer, I decided the extremely simple data related to this could be kept in state. If there ever comes a need to output actions, we can add some info-only actions that get fired.
* Removed sidebar from generate altogether.
    * As @tayvano noted, we may want to add back in the guide links somewhere in the flow. Feedback is welcome on where they could fit!
* Re-organized files and styles to better match other tabs.
    * That's where the bulk of the additions / removals come from.


dternyak:
Tested, works great!

Implementation looks great as well.

I did notice that there is no way to reset or go back once you have selected a keystore / mnemonic based wallet generation. It would be nice to have a restart / back button, but it doesn't need to hold up the merge.

common/containers/Tabs/GenerateWallet/components/FinalSteps.tsx
```diff
@@ -0,0 +1,77 @@
...
+import ProvideKeystoreImage from 'assets/images/unlock-guide/provide-keystore.png';
+import SelectMnemonicImage from 'assets/images/unlock-guide/select-mnemonic.png';
+import ProvideMnemonicImage from 'assets/images/unlock-guide/provide-mnemonic.png';
+import './FinalSteps.scss';
+
+interface Props {
+  walletType: WalletType;
+}
+
+export default class FinalSteps extends React.Component<Props> {
```
HenryNguyen5: This could just be an SFC instead, and then you can have all the logic in a separate function too

common/containers/Tabs/GenerateWallet/components/Mnemonic/Mnemonic.tsx
```diff
@@ -0,0 +1,150 @@
...
+              to reveal it.
+            `
+              : `
+              Write these words down. Do not copy them to your clipboard, or save
+              them anywhere online.
+            `}
+          </p>
+
+          <div className="GenerateMnemonic-words">
+            {[firstHalf, lastHalf].map((ws, i) => (
```
HenryNguyen5: I would consider unrolling the first level array just to reduce complexity, with having a helper function / SFC generate the `Word` components

common/containers/Tabs/GenerateWallet/components/Mnemonic/Mnemonic.tsx
```diff
@@ -0,0 +1,150 @@
...
+
+    return <div className="GenerateMnemonic Tab-content-pane">{content}</div>;
+  }
+
+  private regenerateWordArray = () => {
+    this.setState({ words: generateMnemonic().split(' ') });
+  };
+
+  private handleConfirmChange = (index: number, value: string) => {
+    console.log(index, value);
```
HenryNguyen5: Got some console logs here

common/containers/Tabs/GenerateWallet/components/Mnemonic/Mnemonic.tsx
```diff
@@ -0,0 +1,150 @@
...
+    const confirmValues = [...this.state.confirmValues];
+    confirmValues[index] = value;
+    console.log(confirmValues);
+    this.setState({ confirmValues });
```
HenryNguyen5: This could all be done with a setState cb that'll give you the previous state. Should be more clean.

common/containers/Tabs/GenerateWallet/components/Mnemonic/Word.tsx
```diff
@@ -0,0 +1,64 @@
...
+    const { isShowingWord } = this.state;
+    const readOnly = isReadOnly || isShowingWord;
+
+    return (
+      <div className="MnemonicWord">
+        <span className="MnemonicWord-number">{index + 1}.</span>
+        <div className="MnemonicWord-word input-group">
+          <input
+            className={classnames({
+              'MnemonicWord-word-input': true,
```
HenryNguyen5: This can be shortened to 
```ts
classnames(
               'MnemonicWord-word-input',
               'form-control',
               {'is-valid': word === value}
             )
```

HenryNguyen5: Or using the same style as LN 49 to keep it consistent

common/containers/Tabs/GenerateWallet/components/Template.tsx
```diff
@@ -2,23 +2,11 @@ import React from 'react';
...
-        <div className="GenerateWallet-column col-md-9">
-          <main className="GenerateWallet-column-content Tab-content-pane">{content}</main>
-        </div>
-
-        <div className="GenerateWallet-column col-md-3">
-          <aside className="GenerateWallet-column-help Tab-content-pane">{help}</aside>
-        </div>
-      </div>
-    );
+    return <div className="GenerateWallet Tab-content-pane">{this.props.children}</div>;
```
HenryNguyen5: This could be a SFC instead

common/containers/Tabs/GenerateWallet/components/WalletTypes.tsx
```diff
@@ -0,0 +1,78 @@
...
+import translate from 'translations';
+import { WalletType } from '../GenerateWallet';
+import { NewTabLink } from 'components/ui';
+import './WalletTypes.scss';
+
+interface Props {
+  onSelect(type: WalletType): void;
+}
+
+export default class WalletTypes extends React.Component<Props, {}> {
```
HenryNguyen5: This could be an SFC too

common/containers/Tabs/GenerateWallet/components/Mnemonic/Mnemonic.tsx
wbobeirne: Can you elaborate on
> unrolling the first level array
Not sure what that looks like

common/containers/Tabs/GenerateWallet/components/WalletTypes.tsx
wbobeirne: Would be dope if there was a linter rule or something that caught implementations of React.Component that only had a render method. I'll convert this, but lmk if you think that's out there.

wbobeirne:
Personally I find keeping the markup as together as possible makes reading the structure of the markup a little easier, but I can pretty easily convert some of those maps to helper functions.

wbobeirne:
@dternyak added a UI back button, but also refactored this so it no longer uses state to keep track of which wallet we're generating, but uses the router instead. This makes the back button work as well. Added routes for /generate, /generate/keystore, and /generate/mnemonic.

screen shot 2017-12-28 at 2 06 29 pm

@HenryNguyen5 I think I addressed everything! Let me know if I got any changes wrong.

common/containers/Tabs/GenerateWallet/components/Mnemonic/Mnemonic.tsx
wbobeirne: I think I simplified this by making a makeWord helper function.

wbobeirne:
I alertified the warning text upfront:

screen shot 2017-12-28 at 2 40 39 pm

PR #663, opened 2017-12-23
Miscellaneous Security Improvements

### LINKED ISSUE #558: Miscellaneous Security Improvements
1. Enforce no HTTP external links (from https://github.com/MyEtherWallet/MyEtherWallet/issues/556). Inspired by MM-01-001 from MetaMask Cure53 audit

- [x] Integrate CI Script that fails on finding any non https links throughout codebase

- [x] Additionally, convert all existing HTTP -> HTTPS (if there are any)

2. Prevent Reverse Tabnabbing
More information can be found here: https://github.com/asciidoctor/asciidoctor/issues/2071

- [x] Create script to ensure `rel="noopener noreferrer"` is added to all anchor elements with` target="_blank"`
- [x] Integrate into CI (exit code on finding elements without `rel`)

jclancy93:
Closes #558

- [x] Create custom linting setup via tslint
- [x] Enforce no external http links 
- [x] Enforce all <a> with ```target = _blank``` has ```rel="noopener noreferrer"```
- [x] Turn react anchor black rule on and lock ```tslint-microsoft-contrib``` version


dternyak:
Hey @jclancy93!

Thanks for getting started on this.

I wanted to check-in and see how things were coming along. Are you having any issues continuing with completing the bounty?

jclancy93:
Definitely. Was traveling for the holidays but have gotten a feel for the tslint walker and hoping to finish the no external http links tomorrow.

jclancy93:
Hey @dternyak so I'm working on enforcing all with target = _blank has rel="noopener noreferrer". 

I found this exact rule implemented in a microsoft ts lint extension, so I have borrowed heavily from that. You can find the associated repo [here](https://github.com/Microsoft/tslint-microsoft-contrib/blob/master/src/reactAnchorBlankNoopenerRule.ts)

Let me know what you think

jclancy93:
Hey @dternyak so I'm working on enforcing all with target = _blank has rel="noopener noreferrer". 

I found this exact rule implemented in a microsoft tslint extension, so I have borrowed heavily from that. You can find the associated repo [here](https://github.com/Microsoft/tslint-microsoft-contrib/blob/master/src/reactAnchorBlankNoopenerRule.ts)

Let me know what you think

jclancy93:
No problem. My logic in just copy-pasting the rules over was to avoid adding a bunch of unnecessary rules. Also, since ```tslint-microsoft-contrib``` doesn't contain a rule to enforce no external http links, I will need to extend their ruleset to contain that. I could extend it from the tslint-react library, but the utilities that I included from ```tslint-microsoft-contrib``` allow for easier testing of jsx elements

Let me know if that has any impact on your opinion.

wbobeirne:
Well we can select which rules from that repo we want to add, I agree that we don't want them all.

I see what you mean about the utilities though. There are a couple of ways we could go about that:

1) Fork `tslint-microsoft-contrib` and make the rule ourself in our fork.
2) I think we can import the utilities directly for our custom rules. It's definitely a little dirty, but something like:

```
import { getJsxAttributesFromJsxElement } from 'tslint-microsoft-contrib/src/utils/utils/JsxAttribute';
...
```

wbobeirne:
Since it's purely compile time, there's nothing wrong with having a module but only using part of it. Not like the users have to load it!

If the second works out, I'd go for that. Maintaining a whole new repo seems like a pain.

jclancy93:
Good point. I guess the only negative user implication of including the entire lib is download time/repo size for users that want to use MEW locally. 

Working on this now

jclancy93:
@wbobeirne @dternyak  ok fixed this up. Also added enforce no external http links rule.

I'm still doing research on if/how I can make the auto-fix to ```https://``` work with tslint. I'm aware of the ```--fix``` flag but the way it works under the hood is to replace the node (jsx opening element) with a string, I'm not sure if we can make that work with the specific rule, but gonna look into it more. 

If you guys have any feedback or insight about the fix, let me know. Thanks for the guidance 😄

wbobeirne:
I wouldn't worry about --fix; some websites won't even support HTTPS so I'd rather the developer have to manually change it and confirm the website has HTTPS.

jclancy93:
Ok cool, so this PR should be good to go once i fix the merge conflicts.

Do you want me to rebase the commits?

dternyak:
@jclancy93 No need! I can squash at the time of merge

dternyak:
For onlookers: here's an example of CI failing with a non HTTPS external link: 
screen shot 2018-01-04 at 4 58 55 pm

package.json
```diff
@@ -109,6 +109,7 @@
     "ts-loader": "^3.2.0",
     "tslint": "5.8.0",
     "tslint-config-prettier": "1.6.0",
+    "tslint-microsoft-contrib": "^5.0.1",
```
dternyak: Could you lock down the version?

jclancy93: should be good to go. I also turned on the target blank noopener rule so you should see a lot more failures from running tslint now 😃

jclancy93:
Just one thing to keep in mind. As @wbobeirne noted, not all sites will have https versions. This means that in those cases, whoever is contributing will have to wrap the violating <a> tag in tslint disable comment syntax, like so: 

``` 
/* tslint:disable:no-external-http-link */
<a href="http://somelinkhere.com"></a>
/* tslint:enable:no-external-http-link */
```

It might be possible to have tslint throw a warning instead of an error, but it seems wiser to force the contributor to wrap the tag in comments.

jclancy93:
Just one thing to keep in mind, as @wbobeirne mentioned, not all sites will have https versions. For an offending link, all one has to do is wrap it in tslint disable comments like so:

```
{/* tslint:disable:no-external-http-link */}
<a href="http://somelink.here"></a>
{/* tslint:enable:no-external-http-link */}
```

It might be possible to change the rule to display a warning instead of throw an error, but I think its wiser to force the user to double check the link. 

Sorry about the double notification, was trying to edit my original comment and ended up deleting it 😅

tayvano:
I'm surprisingly okay with all links having https, even if the site doesn't have https. I can't think of an instance where they would not have https AND be linked to from MEW, especially with letsencrypt being so easy.

I would actually opt for choosing to not link to a non-https site over spending more time on this issue. ((Just my two cents!))

jclancy93:
@tayvano both forcing all links to https or the commenting option I mentioned above can both work with the current implementation of the rule 👍

custom_linting_rules/noExternalHttpLinkRule.ts
```diff
@@ -0,0 +1,99 @@
...
+const FAILURE_STRING: string = 'Anchor tags with an external link must use https';
+
+/**
+ * Implementation of the no-external-http-link rule.
+ */
+export class Rule extends Lint.Rules.AbstractRule {
+  public static metadata: ExtendedMetadata = {
+    ruleName: 'tno-external-http-link',
+    type: 'functionality',
+    description: 'Anchor tags with an external link must use https',
```
wbobeirne: Repeat from above FAILURE_STRING

custom_linting_rules/noExternalHttpLinkRule.ts
```diff
@@ -0,0 +1,99 @@
...
+import { ExtendedMetadata } from '../node_modules/tslint-microsoft-contrib/utils/ExtendedMetadata';
+import { Utils } from '../node_modules/tslint-microsoft-contrib/utils/Utils';
+
+import {
+  getJsxAttributesFromJsxElement,
+  getStringLiteral,
+  isEmpty
+} from '../node_modules/tslint-microsoft-contrib/utils/JsxAttribute';
+
+const FAILURE_STRING: string = 'Anchor tags with an external link must use https';
```
wbobeirne: No need for explicit typing when you define a const.

custom_linting_rules/noExternalHttpLinkRule.ts
```diff
@@ -0,0 +1,99 @@
...
+  }
+
+  private validateOpeningElement(openingElement: ts.JsxOpeningLikeElement): void {
+    if (openingElement.tagName.getText() === 'a') {
+      const allAttributes: { [propName: string]: ts.JsxAttribute } = getJsxAttributesFromJsxElement(
+        openingElement
+      );
+      /* tslint:disable:no-string-literal */
+      const href: ts.JsxAttribute = allAttributes['href'];
+      /* tslint:enable:no-string-literal */
```
wbobeirne: I don't think we use this linter rule, but is there any advantage of doing this over `.href`?

custom_linting_rules/noExternalHttpLinkRule.ts
```diff
@@ -0,0 +1,99 @@
...
+      if (href != null && !isSafeHrefAttributeValue(href)) {
```
wbobeirne: Prefer `!==` or explicit casting over implicit casting with `!=`

custom_linting_rules/noExternalHttpLinkRule.ts
```diff
@@ -0,0 +1,99 @@
...
+    description: 'Anchor tags with an external link must use https',
+    options: null,
+    optionsDescription: '',
+    typescriptOnly: true,
+    issueClass: 'SDL',
+    issueType: 'Error',
+    severity: 'Critical',
+    level: 'Mandatory',
+    group: 'Security',
+    commonWeaknessEnumeration: '242,676'
```
wbobeirne: Do we know what this extended metadata is for and what it means? I tried looking this up, but wasn't sure what it meant.

custom_linting_rules/noExternalHttpLinkRule.js
```diff
@@ -0,0 +1,100 @@
...
+    }
+    var stringValue = JsxAttribute_1.getStringLiteral(attribute);
+    if (stringValue === '#') {
+        return true;
+    }
+    else if (stringValue == null || stringValue.length === 0) {
+        return false;
+    }
+    return stringValue.indexOf('https://') >= 0;
+}
```
wbobeirne: It would be ideal if we didn't check in compiled code, but I'm not sure what the linting process would look like in that case. I think this is OK for now.

custom_linting_rules/noExternalHttpLinkRule.ts
jclancy93: Good catch, I was just following the format of the microsoft lib. Don't see any reason why we can't use dot notation

custom_linting_rules/noExternalHttpLinkRule.js
jclancy93: I agree. According to the docs found [here](https://palantir.github.io/tslint/develop/custom-rules/) (final notes section at the end), we can run the rule in typescript. However, I tried running this way and couldn't get it to work

custom_linting_rules/noExternalHttpLinkRule.ts
jclancy93: Again following the formatting of the microsoft tslint lib, but I don't think using extendedMetadata is necessary. 

I think its necessary to define rule metadata, but we could use the standard tslint metadata interface, like [this](https://github.com/palantir/tslint-react/blob/master/src/rules/jsxAlignmentRule.ts#L24)

dternyak:
@jclancy93 Looks good for a merge, but would definitely appreciate fixing up those final comments.

Thanks a lot for your work on this :)

jclancy93:
Should be good to go now.

> Do we know what this extended metadata is for and what it means? I tried looking this up, but wasn't sure what it meant.

@wbobeirne turns out the interface for Abstract Rule requires an ExtendedMeta object, so no way around it for now. 

Glad I could contribute to such an awesome project @dternyak!

dternyak:
@jclancy93 I noticed a false positive on requiring `rel="noopener noreferrer"`. 


ERROR: 
```
/Users/danielternyak2/ActiveDevelopment/MEW/MyEtherWallet/common/components/DataField.tsx[8, 3]: Anchor tags with an external link must use https
```

Anchor Element:
```
const expander = (expandHandler: ExpandHandler) => (
  <a onClick={expandHandler}>
    <p className="strong">{translate('TRANS_advanced')}</p>
  </a>
);
```

As you can see, the anchor element does not have a `target="_blank"` property, but there is still an error when running tslint.

dternyak:
Another example in ```
ERROR: /Users/danielternyak2/ActiveDevelopment/MEW/MyEtherWallet/common/components/GasSlider/GasSlider.tsx[67, 15]: Anchor tags with an external link must use https
```

**Comment by dternyak** [comment]
@jclancy93 Looks good to go after further review :)

**Comment by dternyak** [comment]
@jclancy93 Actually I did end up finding an issue (sorry for crying wolf before):


Error:
```
ERROR: /Users/danielternyak2/ActiveDevelopment/MEW/MyEtherWallet/common/containers/Tabs/SendTransaction/components/Donate.tsx[19, 9]: Anchor tags with an external link must use https
```

Example Snipper:
```
  <a className="btn btn-primary btn-block" onClick={this.onClick}>
          {translate('sidebar_donate')}
        </a>
```

There isn't an external link referenced here at all, but there is a warning.

Using anchor tags this way is probably an anti-pattern, but there is existing stylesheets that would be a pain to re-write.

Do you mind adjusting the linter to not require https on anchor tags that don't specify any URL?

jclancy93:
@dternyak no problem. Not sure why this is happening but I will take a look and fix it.

dternyak:
Replaced by https://github.com/MyEtherWallet/MyEtherWallet/pull/773

PR #665, opened 2017-12-24
Webpack Upgrade

### LINKED ISSUE #206: Remove dll.vendor.js dependency in index.html during production builds
`/dll.vendor.js` currently is not needed for production build, it should be removed from `index.html` if the node.env for production is set

### LINKED ISSUE #268: Re-compile webpack DLL on vendor or package.json change
We should have a script that computes the md5 hash of the package.json and specified vendor directories. If it changes from the previous stored hash, then recompile the webpack DLL.

The script could also auto-add new dependencies from package.json or vendors to the DLL file.

HenryNguyen5:
Fixes #268
Fixes #206 
Adds rebuild and build speed improvements across the board 
Removes depreciated config keys
Adds significant cpu usage reduction when webpack server is watching via turning off polling
Reduces production build size by around 6.5mb


webpack_config/webpack.base.js
```diff
@@ -5,8 +5,58 @@ const HtmlWebpackPlugin = require('html-webpack-plugin');
...
+      'react-transition-group',
+      'redux',
+      'redux-logger',
+      'redux-promise-middleware',
+      'redux-saga',
+      'scryptsy',
+      'store2',
+      'uuid',
+      'wallet-address-validator',
+      'whatwg-fetch'
```
wbobeirne: Can we maintain this array in a shared way between config.js and this?

dternyak:
@HenryNguyen5 FYI I got errors about `html-webpack-include-assets-plugin` when running `npm run dev` as wll.

superstatic.json
```diff
@@ -1,6 +0,0 @@
-{
-	"rewrites": [{
-		"source": "/**/*",
-		"destination": "/index.html"
-	}]
-}
```
wbobeirne: Thanks, that was in fact a mistake.

PR #671, opened 2017-12-27
Request Payment Sub-Tab - EIP 681

skubakdj:
### What This Does

- Implements basic support for [EIP 681](https://github.com/nagydani/EIPs/blob/4eca54f2899e78ea7a91981aa42f0c4c34526946/EIPS/eip-681.md) by introducing a `Request Payment` tab in the account view. Allows users to generate a QR code that can be scanned and processed by wallets that support this standard. Also displays the URI that can be used directly. 
- Makes unit dropdown searchable if list length > 20.
- Makes `AmountField` component accept two new props: `showAllTokens` and `customValidator`. 

This PR should be in a reviewable state, but I still need to: 
 - ~~Fix a ts-lint bug.~~
 - ~~Add additional input validation in the QR generation code.~~
 - Test this implementation against others. Since this is an ongoing effort with other wallets, this is probably something to accomplish in the future.


dternyak:
Hey @ligi! 

I wanted to give you a ping so that you can provide feedback on our implementation of EIP 681. 😄

ligi:
Hey - thanks for the ping - will have a look - but realistically after the 34c3 - which basically means next year.

dternyak:
@ligi Just wanted to give you a ping and let you know this functionality is now live on alpha.myetherwallet.com

Looking forward to your feedback!

Issue #86, opened 2017-07-25
Sign and Verify


dternyak:
- [x] Sign Message
- [x] Verify Message
- [x] Disable decrypt options that don't offer sign functionality.

Future: 
- [x] Sign via Ledger
~- [ ] Sign via Trezor~ (Trezor API not currently available)

dternyak:
Remaining task:

Disable decrypt options that don't offer sign functionality

PR #673, opened 2017-12-27
Disable Trezor in Sign & Verify Message Tab

skubakdj:
Completes the remaining task on #86 by:
 - Making `WalletDecrypt` accept a `disabledWallets` prop that conditionally disables wallet radio buttons.
 - Disabling Trezor in the `Sign & Verify Message` tab's `WalletDecrypt`.

### DIFF
--- a/common/containers/Tabs/SignAndVerifyMessage/components/SignMessage/index.tsx
+++ b/common/containers/Tabs/SignAndVerifyMessage/components/SignMessage/index.tsx
@@ -64,7 +64,7 @@ export class SignMessage extends Component<Props, State> {
               onSignMessage={this.onSignMessage}
             />
           )}
-          <WalletDecrypt hidden={unlocked} />
+          <WalletDecrypt hidden={unlocked} disabledWallets={['trezor']} />
 
           {!!signedMessage && (
             <div>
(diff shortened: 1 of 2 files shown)


common/containers/Tabs/SignAndVerifyMessage/components/SignMessage/index.tsx
```diff
@@ -64,7 +64,7 @@ export class SignMessage extends Component<Props, State> {
               onSignMessage={this.onSignMessage}
             />
           )}
-          <WalletDecrypt hidden={unlocked} />
+          <WalletDecrypt hidden={unlocked} disabledWallets={['trezor']} />
```
wbobeirne: My preference, if possible, would be to either A) maintain a dictionary of wallet disable arrays keyed by reason (e.g. `DISABLED_WALLETS.CANT_SIGN`) or B) make the props booleans for reasons why to disable them `hideWalletsThatCantSign={true}`. Support for these features is subject to change, and it'd be nice to have one canonical place to track this. I also can see a need for wallets that are read-only, wallets that don't support offline etc.

wbobeirne: Ah, didn't see that this was merged. If there comes a time I need to adjust this, I hope you don't mind that I implement this! I'll probably need to rework this a bit for the new unlock UI.

PR #675, opened 2017-12-27
Hardware Wallet Unlock Loading Indicator

james-prado:
Closes https://github.com/MyEtherWallet/MyEtherWallet/issues/578

- Add loading indicator while decrypting hardware wallets
load
- Display a tip prompting the user to check they're using the ethereum app on their hardware wallet when a timeout occurs
screen shot 2017-12-27 at 5 13 18 pm

I don't actually have a Tezor to test this with, the implementation is pretty similar though so it _should_ work, but I can't test to confirm


common/components/WalletDecrypt/LedgerNano.tsx
```diff
@@ -65,9 +86,12 @@ export default class LedgerNanoSDecrypt extends Component<Props, State> {
...
         </div>
-
-        <div className={`LedgerDecrypt-error alert alert-danger ${showErr}`}>{error || '-'}</div>
-
+        {error && (
+          <p className="LedgerDecrypt-error alert alert-danger is-showing">
+            <strong>Error: </strong>
+            {error}
+          </p>
+        )}
```
wbobeirne: Not saying to change this, but just wanted to make sure it was clear that this was always rendered before, and is-showing was added so that the content didn't jump quite as much when an error popped up. It always filled the area it would in the content.

common/components/WalletDecrypt/Trezor.tsx
```diff
@@ -24,21 +26,41 @@ export default class TrezorDecrypt extends Component<Props, State> {
...
-    const { dPath, publicKey, chainCode, error, isLoading } = this.state;
+    const { dPath, publicKey, chainCode, error, isLoading, showTip } = this.state;
     const showErr = error ? 'is-showing' : '';
 
     return (
       <section className="TrezorDecrypt col-md-4 col-sm-6">
+        {showTip && (
+          <p>
+            <strong>Tip: </strong>Make sure you're logged into the ethereum app on your hardware
+            wallet
```
wbobeirne: Trezors don't actually have you confirm signing in on the wallet.

common/components/WalletDecrypt/Trezor.tsx
```diff
@@ -101,6 +124,9 @@ export default class TrezorDecrypt extends Component<Props, State> {
             isLoading: false
           });
         } else {
+          if (res.error === 'TIMEOUT') {
```
wbobeirne: I'm not sure that res.error will ever be `TIMEOUT`. I think this is the exhaustive list of "errors": https://github.com/trezor/connect/blob/master/connect.js#L30

Between this and the comment above, I would probably just remove the showTip functionality on TREZOR and leave it at a loading indicator.

common/components/WalletDecrypt/LedgerNano.tsx
james-prado: I was actually just messing around with it and forgot to remove it before committing. I'll make sure to revert it

common/components/WalletDecrypt/Trezor.tsx
james-prado: Yea I don't actually have a TREZOR to test any of these changes on so I sort of just guessed it would be similar to ledger. I'll revert the TREZOR changes

common/components/WalletDecrypt/Trezor.scss
```diff
@@ -23,4 +23,13 @@
   &-buy {
     margin-top: 10px;
   }
+
+  &-message {
```
dternyak: This isn't used anywhere.

dternyak:
@james-prado I get a TIMEOUT error immediately after unlock when testing with ledger.

PR #680, opened 2017-12-28
Web Worker Decrypt

aitrean:
Closes https://github.com/MyEtherWallet/MyEtherWallet/issues/423, https://github.com/MyEtherWallet/MyEtherWallet/issues/219

This PR includes three solutions. 
1. Attempt an empty password every time a keystore is uploaded.
2. Delegate scrypt decryption (ie ethereumjs-wallet.fromV3) to its own web worker and interface with it through an async typescript function that gets handled in the wallet saga. This keeps the UI unblocked when scrypt takes a long time to decrypt.
3. Add logic to show a spinner x number of milliseconds after file upload so the user will understand when a wallet is being decrypted.


common/components/WalletDecrypt/Keystore.tsx
```diff
@@ -21,12 +22,14 @@ function isPassRequired(file: string): boolean {
...
   public props: {
     value: KeystoreValue;
+    isWalletLoading: boolean;
     onChange(value: KeystoreValue): void;
     onUnlock(): void;
   };
 
   public render() {
     const { file, password } = this.props.value;
+    const isWalletLoading = this.props.isWalletLoading;
```
dternyak: Why not de-structure?

dternyak:
Impressive work! Going to leave it open a bit more and solicit feedback from @HenryNguyen5 and @wbobeirne who participated in design.

dternyak:
Here's an encrypted Keystore with no password for testing purposes: 
[UTC--2017-12-30T16-45-58.340Z--80965bfdbb6c44c8ed652add058bf372e2b9dc86.zip](https://github.com/MyEtherWallet/MyEtherWallet/files/1594575/UTC--2017-12-30T16-45-58.340Z--80965bfdbb6c44c8ed652add058bf372e2b9dc86.zip)

common/libs/web-workers/workers/scrypt-worker.worker.ts
```diff
@@ -0,0 +1,18 @@
+import { fromV3, IFullWallet } from 'ethereumjs-wallet';
+
+declare var postMessage: any;
+interface DecryptionParameters {
+  keystore: string;
+  password: string;
+  nonStrict: boolean;
+}
+
+onmessage = event => {
```
dternyak: event could use typing

package.json
```diff
@@ -110,7 +110,8 @@
     "webpack": "3.10.0",
     "webpack-dev-middleware": "2.0.3",
     "webpack-hot-middleware": "2.21.0",
-    "webpack-sources": "1.0.1"
+    "webpack-sources": "1.0.1",
+    "worker-loader": "^1.1.0"
```
wbobeirne: Lock down version # (We try to avoid `^` versions to keep builds consistent.)

common/libs/web-workers/scrypt-wrapper.ts
```diff
@@ -0,0 +1,22 @@
+import { IFullWallet, fromPrivateKey } from 'ethereumjs-wallet';
+import { toBuffer } from 'ethereumjs-util';
+const Worker = require('worker-loader!./workers/scrypt-worker.worker.ts');
```
wbobeirne: We should setup typing for this, as outlined here: https://github.com/webpack-contrib/worker-loader#integrating-with-typescript. Right now it's an implicit `any`.

common/actions/wallet/actionCreators.ts
```diff
@@ -43,6 +43,13 @@ export function setWallet(value: IWallet): types.SetWalletAction {
   };
 }
 
+export function setWalletLoading(loadingStatus: boolean): types.SetWalletLoadingAction {
```
wbobeirne: Existing nomenclature would be `setWalletPending`

common/components/WalletDecrypt/Keystore.tsx
```diff
@@ -57,6 +60,7 @@ export default class KeystoreDecrypt extends Component {
                 type="password"
               />
             </div>
+            {isWalletLoading ? <Spinner /> : ''}
```
wbobeirne: Minor, but you can simplify these boolean checks to `{isWalletLoading && <Spinner />}`

common/libs/web-workers/workers/scrypt-worker.worker.ts
```diff
@@ -0,0 +1,18 @@
+import { fromV3, IFullWallet } from 'ethereumjs-wallet';
+
+declare var postMessage: any;
```
wbobeirne: From https://github.com/webpack-contrib/worker-loader#integrating-with-typescript:

```
const ctx: Worker = self as any;
ctx.postMessage({ foo: "foo" });
```
for typing

common/sagas/wallet/wallet.ts
```diff
@@ -141,18 +146,43 @@ export function* unlockPrivateKey(action: UnlockPrivateKeyAction): SagaIterator
   yield put(setWallet(wallet));
 }
 
+export function* startLoadingSpinner(): SagaIterator {
+  yield call(delay, 1000);
```
wbobeirne: I'd probably shorten this to 300-500, a full second before having the site respond is a little long.

common/libs/web-workers/scrypt-wrapper.ts
wbobeirne: Still typed as any, but
```
import Worker from 'worker-loader!./workers/scrypt-worker.worker.ts';
```
seems to fix it!

PR #684, opened 2017-12-29
Offline Polling on WalletDecrypt Mount

dternyak:
Fixes https://github.com/MyEtherWallet/MyEtherWallet/issues/661


wbobeirne:
How would you feel about polling online from the start? It'd be nice if when you went to the send view, it didn't suddenly pop in the changes, rather than detecting it before you get there.

dternyak:
Sure, don't see any issue with that. We're not hitting the nodes to poll for offline connectivity anyway, so I don't see any reason why not to do it starting onload.

dternyak:
https://github.com/MyEtherWallet/MyEtherWallet/pull/785 replaces

PR #693, opened 2017-12-30
Ensure CI fails on non-prettier compliant PRs

james-prado:
Closes https://github.com/MyEtherWallet/MyEtherWallet/issues/601

- Add npm scripts to list formatting diffs 
	- [list-difference flag](https://prettier.io/docs/en/cli.html#list-different)
- Check prettier diff in ci

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,6 +21,8 @@ install:
 jobs:
   include:
     - stage: test
+      script: npm run prettier:diff
+    - stage: test
       script: npm run test
     - stage: test
       script: npm run tslint
--- a/package.json
+++ b/package.json
@@ -140,7 +140,10 @@
     "tscheck": "tsc --noEmit",
     "start": "npm run dev",
     "precommit": "lint-staged",
-    "formatAll": "find ./common/ -name '*.ts*' | xargs prettier --write --config ./.prettierrc --config-precedence file-override",
+    "formatAll":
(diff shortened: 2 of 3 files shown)


package.json
```diff
@@ -140,6 +140,11 @@
     "precommit": "lint-staged",
     "formatAll":
       "find ./common/ -name '*.ts*' | xargs prettier --write --config ./.prettierrc --config-precedence file-override",
+    "prettier:diff-ts":
+      "prettier --write --config ./.prettierrc --list-different \"common/**/*.ts\"",
+    "prettier:diff-tsx":
+      "prettier --write --config ./.prettierrc --list-different \"common/**/*.tsx\"",
+    "prettier:diff": "npm run prettier:diff-ts && npm run prettier:diff-tsx",
```
dternyak: We don't need both `prettier:diff-tsx` and `prettier:diff-ts`, a single regex should be able to accommodate both.

.travis.yml
```diff
@@ -20,6 +20,8 @@ install:
 
 jobs:
   include:
+    - stage: Prettier Diff
```
dternyak: Why run it in a different stage?

.travis.yml
```diff
@@ -20,6 +20,8 @@ install:
...
+      script: npm run prettier:diff
```
dternyak: To confirm, `--list-different` returns an exit code if there are any different files?

PR #694, opened 2017-12-31
Fix Help Nav isActive Condition

james-prado:
Add condition to fix help tab appearing as active at `localhost:3000/`


common/components/Header/components/NavigationLink.tsx
```diff
@@ -18,13 +18,14 @@ class NavigationLink extends React.Component<Props, {}> {
...
     // 1) Current path is the same as link
-    // 2) the first path is the same for both links (/account and /account/send)
+    // 2) the first path is the same for both links (/account and /account/send) && is not an external link
     // 3) we're at the root path and this is the "homepage" nav item
     const isActive =
       location.pathname === link.to ||
-      (link.to && location.pathname.split('/')[1] === link.to.split('/')[1]) ||
+      (link.to &&
+        location.pathname.split('/')[1] === link.to.split('/')[1] &&
+        location.pathname.split('/')[0] === link.to.split('/')[0]) ||
```
dternyak: We should variable these conditions, it's no longer readable.

dternyak:
It would be nice if this PR took care of https://github.com/MyEtherWallet/MyEtherWallet/issues/620 as well, closely related.

common/components/Header/components/NavigationLink.tsx
wbobeirne: Splits should def be a variables. The external check could also just be a 4th condition:
```
... ||
location.pathname.split('/')[0] === link.to.split('/')[0]
```
No need to piggyback off of the 2nd condition

PR #701, opened 2018-01-01
Implement `offline-plugin` for Service Workers / App Cache

### LINKED ISSUE #570: Cache assets with service workers
Service workers will reduce subsequent page load times and data usage on subsequent requests, and allow users to visit the webpage without an internet connection. We can probably load up an off-the-shelf solution for high yield / low effort, but might require some digging in to make sure we're busting caches properly.

https://github.com/NekR/offline-plugin looks like a good candidate for this, but chime in if you have experience using something else.

wbobeirne:
Closes #570. This is probably the highest benefit to lowest effort PR I've ever made. Implements the [`offline-plugin`](https://github.com/NekR/offline-plugin) on the production build to cache assets via service workerss, or fallback to app cache. This is really cool for two reasons:

1) Subsequent visits the the site will not re-request assets, and will instead use a cache, resulting in hella fast page loads.
2) You can load myetherwallet.com even if you're offline.

This will be exceptionally good for Android phone users, but will still be great for anyone using Chrome or Firefox. It falls back to AppCache apparently, but I'm not really sure how that behaves.

Here's what a second page load looks like on a 3g connection today:

screen shot 2018-01-01 at 3 25 15 pm

Here's what it looks like with the service worker installed:

screen shot 2018-01-01 at 3 25 35 pm

From 27.2s to 2.22s

---

I think we should put this up on the next alpha release, so that we can do substantial testing on various devices more easily. I definitely want to make sure we're not going to get into a situation where we can't bust a cache.

### DIFF
--- a/common/index.tsx
+++ b/common/index.tsx
@@ -5,13 +5,15 @@ import 'sass/styles.scss';
 import 'babel-polyfill';
 import 'whatwg-fetch';
 import React from 'react';
+import OfflineRuntime from 'offline-plugin/runtime';
 import { render } from 'react-dom';
 import Root from './Root';
 import { configuredStore } from './store';
 import consoleAdvertisement from './utils/consoleAdvertisement';
 
-const appEl = document.getElementById('app');
+OfflineRuntime.install();
 
+const appEl = document.getElementById('app');
 render(<Root store={configuredStore} />, appEl);
 
 if (module.hot) {
--- a/package.json
+++ b/package.json
@@ -93,6 +93,7 @@
     "node-sass": "4.7.2",
     "nodemon": "1.14.7",
     "null-loader": "0.1.1",
+    "offline-plugin": "4.9.0",
     "prettier": "1.9.2",
     "progress": "2.0.0",
     "react-hot-loader": "3.1.3",
--- a/webpack_config/webpack.prod.js
+++ b/webpack_config/webpack.prod.js
@@ -6,7 +6,7 @@ const webpack = require('webpack');
 const ExtractTextPlugin = require('extract-text-webpack-plugin');
 const ProgressPlugin = require('webpack/lib/ProgressPlugin');
 const BabelMinifyPlugin = require('babel-minify-webpack-plugin');
-// const OfflinePlugin = require('offline-plugin')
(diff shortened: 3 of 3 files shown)


dternyak:
I get a few exceptions after building for production and running locally:


```
sw.js:444 DOMException: Entry was not found.
(anonymous) @ sw.js:444
Promise rejected (async)
cacheAssets @ sw.js:443
(anonymous) @ sw.js:452
Promise resolved (async)
cacheChanged @ sw.js:450
(anonymous) @ sw.js:380
sw.js:1 Uncaught (in promise) DOMException: Entry was not found.
Promise rejected (async)
ExtendableEvent.waitUntil @ sw.js:313
(anonymous) @ sw.js:385
```

```
at  
 SyntaxError: Identifier 'a' has already been declared
    at i (http://localhost:8080/vendor.9c6dd251.js:24:159237)
    at x (http://localhost:8080/vendor.9c6dd251.js:24:162288)
    at y (http://localhost:8080/vendor.9c6dd251.js:24:161048)
    at http://localhost:8080/vendor.9c6dd251.js:24:163141
    at Array.forEach (<anonymous>)
    at D (http://localhost:8080/vendor.9c6dd251.js:24:163114)
    at y (http://localhost:8080/vendor.9c6dd251.js:24:160951)
    at a (http://localhost:8080/vendor.9c6dd251.js:24:159881)
    at d (http://localhost:8080/vendor.9c6dd251.js:24:165548)
    at t.a (http://localhost:8080/vendor.9c6dd251.js:24:158234)
```


screen shot 2018-01-01 at 3 28 45 pm

wbobeirne:
Hmm, could possibly this issue in their gh issues: https://github.com/NekR/offline-plugin/issues/195

wbobeirne:
Actually, it's more likely the error above it is causing the install to fail. Not sure how vendor is using the same variable name twice.

dternyak:
After some testing, I was not able to re-create the issue, and then vendor error was determined to already be in `develop`. Changes looks good otherwise, I'll be proceeding with the merge 👍 

We'll need to circle back to fix the vendor error in `develop` at a future time.

PR #705, opened 2018-01-01
Add "404" Page

### LINKED ISSUE #551: 404 Page
Right now if you go to an undefined path, you get a white screen. Ideally we could inject some fun into this, but we at least need something.

eddiewang:
Closes #551 
Would love to have a few different error messages from the community that we randomize (kind of like Discord's status messages), but this is a simple error page for now.

image


common/containers/PageNotFound/index.ts
```diff
@@ -0,0 +1,3 @@
+import PageNotFound from './PageNotFound';
+
+export default PageNotFound;
```
wbobeirne: This is pretty minor, but the `container` nomenclature is for connected components. Since the page not found is just a dumb component, it could live in the components folder.

common/containers/PageNotFound/PageNotFound.tsx
```diff
@@ -0,0 +1,25 @@
+import React, { Component } from 'react';
+import { Link, RouteComponentProps } from 'react-router-dom';
+import TabSection from '../TabSection';
+import './PageNotFound.scss';
+
+export default class PageNotFound extends Component<RouteComponentProps<{}>> {
```
wbobeirne: This could be a static function component

dternyak:
Travis is currently down, merging without waiting for CI to pass. https://www.traviscistatus.com/incidents/yxqx6brg5nl0

PR #708, opened 2018-01-02
Fix deterministic modal token values

### LINKED ISSUE #692: [BUG] DeterministicWallet Modal Issues
screen shot 2017-12-30 at 12 16 03 pm

Steps to reproduce:

1. Open an HD Wallet
2. Select any token balanace

wbobeirne:
Closes #692. Frustratingly, Typescript did not catch a change in types because of the way we type Redux Saga `yield call`s explicitly. There are also tests for this functionality, which I was surprised to see didn't catch this either. I'm not intimately familiar with saga testing, so maybe @skubakdj could shed some light on that bit of weirdness.

Either way, this fixes it up and tightens up some of the typing.

### DIFF
--- a/common/components/WalletDecrypt/components/DeterministicWalletsModal.tsx
+++ b/common/components/WalletDecrypt/components/DeterministicWalletsModal.tsx
@@ -30,8 +30,8 @@ interface Props {
   seed?: string;
 
   // Redux state
-  wallets: DeterministicWalletData[];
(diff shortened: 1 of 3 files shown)


wbobeirne:
Hahaha, now that I fixed it the test is failing. So I guess the test was written while this was bugged. I'll see if I can fix that.

PR #711, opened 2018-01-02
Show warning when no tokens are found after scan

### LINKED ISSUE #613: Show "No Tokens Found" Messaging in Scanner
The messaging related to token scanning may confuse users today, as we simply show no tokens to select when none are found.

Instead, display messaging ("No Tokens Found").

screen shot 2017-12-18 at 11 54 32 am

wbobeirne:
Closes #613. Looks like so:

screen shot 2018-01-02 at 1 12 36 pm


dternyak:
The user isn't able to add custom tokens / re-scan if they scan and receive no tokens.

We should continue to include buttons for re-scanning / adding custom tokens.

wbobeirne:
Great point, I had forgotten about the case where a user has no mainstream tokens, but has some custom ones.

wbobeirne:
Reworked this to function better, updated OP with new screenshot.

PR #765, opened 2018-01-09
Use network unit everywhere, fix network redux state

### LINKED ISSUE #726: Update Send Transaction modal when sending ETC or other network tokens
#### Alpha

2018-01-04 at 8 34 50 am

#### Current (for reference)

2018-01-04 at 8 38 16 am

wbobeirne:
Closes #726.

This originally was only meant to have the confirmation show ETC or whatever instead of ETH when on a different network, but turns out I kind of pulled up the floorboards on `state.config.network` in the process.

The problem was that `selectors/config`'s `getNetworkConfig` was having to fetch which network we were on each time, rather than just using `state.config.network`. If you made the mistake of accessing the latter, it would just be whatever network the site bootstrapped with. Worse yet, the site was always bootstrapping with ETH, even if the starting node was an ETC node or otherwise.

The full list of changes are as follows:

* Confirmation modal now shows correct network token (Unless ERC20 token, then it shows that)
* Send unit dropdown now shows correct network token (Instead of "ether" for all networks)
* Changing node now correctly updates `state.config.network`
* Loading saved node from local storage now correctly sets `state.config.network` at bootstrap

All in all, I think this PR is a good example of doing _more_ in sagas, and _less_ in selectors. Selectors should ideally not be creating new objects or doing overly complex logic, but instead just be selecting things off of the redux store.

## Screenshots

screen shot 2018-01-09 at 12 50 35 pm

screen shot 2018-01-09 at 12 51 17 pm

## Steps to Test

1. Load up the site connected to an ETH node
2. Go to send and unlock
3. Confirm the dropdown says ETH
4. Switch to an ETC node
5. Confirm the dropdown now says ETC
6. Refresh the page with ETC as your network
7. Confirm that your page starts connected to the ETC node
8. Confirm that send shows ETC in the dropdown
9. Generate a transaction, and check that the modal says ETC
10. Switch back to an ETH node
11. Generate a transaction, and check that the modal says ETH
12. Generate a token transaction, and check that the modal says the token name


common/sagas/config.ts
```diff
@@ -119,25 +119,28 @@ export function* reload(): SagaIterator {
...
-  const currentNode = yield select(getNode);
-  const currentConfig = yield select(getNodeConfig);
-  const currentNetwork = currentConfig.network;
+  const currentNode: string = yield select(getNode);
+  const currentConfig: NodeConfig = yield select(getNodeConfig);
+  const customNets: CustomNetworkConfig[] = yield select(getCustomNetworkConfigs);
+  const currentNetwork =
+    getNetworkConfigFromId(currentConfig.network, customNets) || NETWORKS[currentConfig.network];
+
+  function* bailOut(message: string) {
```
dternyak: Any specific reason for the closure? Seems like this could be re-usable in other sagas.

wbobeirne: `changeNode(currentNode, currentConfig, currentNetwork)` relies on variables defined in the function. I could move it out, but the behavior of error message + revert node seemed specific to `handleNodeChangeIntent` anyway. Or I could pass them, but then I'd be copying that 3 times.

dternyak: Yeah, I figured all that as well. I don't think there's extreme value by making it re-usable, we can worry about that later.

PR #768, opened 2018-01-09
Unit tests for token & contract JSON

wbobeirne:
These are just some quick check tests against our contract and token json files to make sure they're formatted correctly. It checks:

* Is token address valid?
* Are token symbols and addresses unique (per network)?
* Does the token have a decimal, and is it between 0 and 18?
* If the contract has an address, is it valid?
* Is the ABI json valid json?

This also makes a change to network contracts, allowing them to not have an address. On v3, the Mist multisig contracts just filled in garbage as the address, and had a message telling the user to change it. v4 did not have that message, and rather than continue doing that, I simply changed it so that contracts don't _have_ to have an address.


spec/config/tokens.spec.ts
```diff
@@ -0,0 +1,40 @@
...
+describe('Tokens JSON', () => {
+  Object.keys(TOKENS).forEach(network => {
+    it(`${network} tokens array properly formatted`, () => {
+      const tokens = TOKENS[network];
+      const addressCollisionMap = {};
+      const symbolCollisionMap = {};
+
+      tokens.forEach(token => {
+        if (!isValidETHAddress(token.address)) {
+          throw Error(`Token ${token.symbol} has invalid contract address '${token.address}'`);
```
dternyak: Any reason for throwing errors instead of using `asserts`? Seems purely stylistic since CI should fail either way, but I'm curious.

wbobeirne: Maybe this is silly, but I wanted to reduce the overall test count (Hoping for some kind of performance gain, not to mention making it clear we don't really have 5000+ tests) which is why I didn't wrap these in `it`s. The throwing is so that you get a better error message than simply having a test called "ETH tokens list is properly formatted" failing, with no understanding of why.

I could convert these to individual `it`s, and use `expect` if that's preferred.

dternyak: I don't think it's a major issue either way, but it does go against convention and raises eyebrows.

Happy to merge either way.

wbobeirne: In that case I would prefer to leave as-is. The test failures you get this have very specific error messages that point to the problem, and should make it easier to find, i.e.

> Token GNT has invalid contract address '0xa74476443119A942dE498590Fe1f2454d7D4aC0doopsimsomeextracharacters'

PR #771, opened 2018-01-09
Rename Aux to AuxComponent for Windows

wbobeirne:
The filename `AUX` (With any extension, any capitalization) is reserved in Windows. When you clone the repo, it omits the file, totally messing up the build. This just renames the Aux.tsx component file to AuxComponent.tsx. No other changes to code.


dternyak:
Interesting, we're getting prettier conflicts here. Any ideas @james-prado?

wbobeirne:
This is almost certainly because I committed this from Windows, and that probably didn't respect the commit hooks somehow.

dternyak:
@wbobeirne I've pushed a commit that reformats those files with pre-commit - hopefully it's all fixed up after.

dternyak:
Overall very interesting issue. Thanks for your time debugging this.

PR #772, opened 2018-01-09
Display Address on Hardware Wallets

james-prado:
Closes https://github.com/MyEtherWallet/MyEtherWallet/issues/624

- Add 'display address' button in the sidebar when logged into a hardware wallet
display-hw-address

- Replace ledger vendor files with [ledger's api](https://github.com/LedgerHQ/ledger-node-js-api), available [here on npm](https://www.npmjs.com/package/ledgerco)
	- I'm looking for some feedback on this change. Using this module over the current ledger files provides async functions, and better error handling. Our current ledger files are also copy pasted from an out of date version of this module, and have no custom functions added, so I don't see any reason we should keep our current files.
	- **Update**: seems the vendor files were a holdover from u2f not being a node module at the start of v4 
- Update Trezor Connect types and vendor files 
- Install libusb-1.0 on ci.
	- This is prevents libusb dependency in the [ledger api](https://www.npmjs.com/package/ledgerco) from breaking during build


common/selectors/wallet.ts
```diff
@@ -106,6 +106,14 @@ export const getWalletType = (state: AppState): IWalletType => {
   return { isWeb3Wallet, isHardwareWallet };
 };
 
+export const getHWType = (state: AppState): string => {
+  const wallet = getWalletInst(state);
+  const isLedgerWallet = wallet instanceof LedgerWallet;
+  const isTrezorWallet = wallet instanceof TrezorWallet;
+  const type = isLedgerWallet ? 'ledger' : isTrezorWallet ? 'tezor' : '';
```
wbobeirne: *Trezor, and I'd use `translateRaw('x_Trezor' | 'x_Ledger')`

common/components/BalanceSidebar/AccountInfo.tsx
```diff
@@ -46,23 +49,53 @@ export default class AccountInfo extends React.Component<Props, State> {
...
+        (result, error) => {
+          if (result) {
+            console.log(result);
+          } else {
+            throw new Error(error);
+          }
+        },
+        undefined
+      );
+    }
```
wbobeirne: I would prefer this logic to live in the associated wallet lib files. My suggestion would be to make a new (optional) function like `wallet.showAddressOnHardware`, and only show the button if the wallet has that function.

common/components/BalanceSidebar/AccountInfo.tsx
```diff
@@ -5,11 +5,14 @@ import React from 'react';
...
 import './AccountInfo.scss';
 import Spinner from 'components/ui/Spinner';
+import ledger from 'ledgerco';
+import TrezorConnect from 'vendor/trezor-connect';
 
 interface Props {
   balance: Balance;
   wallet: IWallet;
   network: NetworkConfig;
+  hwType: string;
```
wbobeirne: Once you add wallet lib function, nix this proptype

common/components/BalanceSidebar/AccountInfo.tsx
```diff
@@ -46,23 +49,53 @@ export default class AccountInfo extends React.Component<Props, State> {
...
+        <div className="AccountInfo-section AccountInfo-address-section">
+          <div className="AccountInfo-address-icon">
+            <Identicon address={address} size="100%" />
+          </div>
+          <div className="AccountInfo-address-wrapper">
             <div className="AccountInfo-address-addr">{address}</div>
           </div>
         </div>
 
+        {hwType && (
```
wbobeirne: Change to check for show address function

common/components/BalanceSidebar/AccountInfo.tsx
```diff
@@ -46,23 +49,53 @@ export default class AccountInfo extends React.Component<Props, State> {
...
+      });
+    } else if (hwType === 'trezor') {
+      TrezorConnect.ethereumGetAddress(
+        dPath,
+        (result, error) => {
+          if (result) {
+            console.log(result);
+          } else {
+            throw new Error(error);
+          }
```
wbobeirne: You can get rid of these two. We can probably just swallow the error if there was one, it's probably because they disconnected their device. Alternatively you could show a notification.

common/components/BalanceSidebar/AccountInfo.tsx
```diff
@@ -46,23 +49,53 @@ export default class AccountInfo extends React.Component<Props, State> {
...
   };
 
+  public displayHWAddress = () => {
+    const { hwType, wallet } = this.props;
+    const { dPath, index } = wallet as any;
+    if (hwType === 'ledger') {
+      ledger.comm_u2f.create_async().then(comm => {
+        new ledger.eth(comm).getAddress_async(dPath + '/' + index, true, false);
+      });
+    } else if (hwType === 'trezor') {
```
wbobeirne: Confirming this works when `hwType === 'tezor'`. Other than the typo, showed up fine on my device.

wbobeirne:
Ah crap, looks like Travis is failing with (what I assume is) the new Ledger modules:
```
prebuild-install WARN install /usr/lib/x86_64-linux-gnu/libstdc++.so.6: version `GLIBCXX_3.4.21' not found (required by /home/travis/build/MyEtherWallet/MyEtherWallet/node_modules/node-hid/build/Release/HID.node)
make: Entering directory `/home/travis/build/MyEtherWallet/MyEtherWallet/node_modules/node-hid/build'
  CC(target) Release/obj.target/hidapi/hidapi/libusb/hid.o
../hidapi/libusb/hid.c:47:20: fatal error: libusb.h: No such file or directory
 #include <libusb.h>
                    ^
```

According to [this issue](https://github.com/node-hid/node-hid/issues/219), looks like we might need to be installing `libusb` on Travis?

common/libs/wallet/deterministic/deterministic.ts
```diff
@@ -1,7 +1,7 @@
 export class DeterministicWallet {
-  private address: string;
-  private dPath: string;
-  private index: number;
```
james-prado: Is there any reason these need to be private? Declaring these as `private` means these variables were would be retrieved from the redux state then passed as arguments back into methods extending `DeterministicWallet`. Declaring them as `public` allows classes like `LedgerWallet` or `TrezorWallet`, which extend `DeterministicWallet`, can use `this.dPath, this.address, ...` in their methods instead of having them passed as arguments

wbobeirne: I believe what you'd want here is the `protected` keyword. Children of the class can access them, but not external code.

I'd prefer we keep as few attributes public as possible, and instead offer up functions wherever possible that do what need doing with the attributes. Just makes refactoring a bit easier.

package.json
```diff
@@ -23,6 +23,7 @@
     "hdkey": "0.7.1",
     "idna-uts46": "1.1.0",
     "jsonschema": "1.2.2",
+    "ledgerco": "^1.2.1",
```
dternyak: pin package version

dternyak: we should setup a linter for this.

dternyak:
@james-prado While not a strict requirement for merge, I noticed during testing that the user is not given any prompt / indicator that they should move their attention to the ledger to confirm the address is correct.

I think replacing the onClick text with a loading indicator would be a small but definite UX improvement.

dternyak:
It may even be worth it to to replace with a loading indicator and text that says "Please check your device to confirm your account address."

PR #776, opened 2018-01-10
[WIP] Fix Todo Issues

james-prado:
Closes more todo issues https://github.com/MyEtherWallet/MyEtherWallet/issues/519

- Add regex check for BIP44 paths
- Fix swap dropdown button height
- Remove `common/config` from ignored paths in jest_config


common/components/WalletDecrypt/components/Keystore.tsx
```diff
@@ -1,34 +1,36 @@
...
+    return isKeystoreValid(file);
+  }
 
   public render() {
-    const { file, password } = this.props.value;
-    const passReq = isPassRequired(file);
-    const unlockDisabled = !file || (passReq && !password);
+    const { file, password, valid } = this.props.value;
+    const passReq = isKeystorePassRequired(file);
+    const unlockDisabled = !file || !valid || (passReq && !password);
```
dternyak: variablize

common/libs/validators.ts
```diff
@@ -116,9 +116,8 @@ export function isPositiveIntegerOrZero(num: number): boolean {
 // https://github.com/bitcoin/bips/blob/master/bip-0044.mediawiki
 // normal path length is 4, ledger is the exception at 3
 export function isValidPath(dPath: string) {
```
dternyak: could use a test. Where is the regex coming from? Handrolled?

common/libs/wallet/non-deterministic/helpers.ts
```diff
@@ -33,22 +33,33 @@ export const signWrapper = (walletToWrap: IFullWallet): WrappedWallet =>
...
+      return KeystoreTypes.v1Encrypted;
+    } else if (parsed.hash && parsed.locked === false) {
+      return KeystoreTypes.v1Unencrypted;
+    } else if (parsed.publisher === 'MyEtherWallet') {
+      return KeystoreTypes.v2Unencrypted;
+    } else {
+      return '';
+    }
+  } catch {
+    return '';
```
dternyak: Do we really want to return an empty string? What's the situation in which that's the desired return value?

common/libs/wallet/non-deterministic/helpers.ts
```diff
@@ -33,22 +33,33 @@ export const signWrapper = (walletToWrap: IFullWallet): WrappedWallet =>
...
   }
 }
 
+const isKeystoreValid = (file: string): boolean => {
+  if (determineKeystoreType(file) === '') {
```
dternyak: Ah, it's here. This seems quite unintuitive though.

common/libs/validators.ts
james-prado: Hand rolled, I'll provide some tests and some comments to clarify

common/libs/wallet/non-deterministic/helpers.ts
james-prado: The function is intended to return a string. Returning an empty string is falsey and cuts down on the amount of `try catch` statements if I just threw an error in the final `else {}` statement

dternyak: @wbobeirne Maybe you have some thoughts on how we could make this more intuitive (or confirm you agree this is the best approach).

common/libs/wallet/non-deterministic/helpers.ts
james-prado: Yea it isn't very intuitive, the error handling still could use some polish. Errors thrown by parsing an invalid file as JSON in `determineKeystoreType` make it difficult to write something easier to use. I spent some time on it the issue but didn't make much progress. I don't think it's a blocking issue, and it is still functional, so I'm going to come back to it later after I've fixed more critical issues.

james-prado:
Stale

Issue #778, opened 2018-01-10
Present non-generated wallets as options on Create New Wallet


wbobeirne:
As I've been working on #557, I realize that while we may brow-beat users with other options for wallet on unlock, we'd be best off trying to guide them towards other wallets _before_ they generate one. Rather than show a bunch of (easily ignored) warnings during generate with links to other wallets, what if we presented MetaMask / Ledger / Trezor as options for creating a new wallet? Quick n dirty mockup, but the idea is there'd be much more emphasis on the more secure options, and we put the insecure generatable ones towards the bottom with warnings.

![create-multi-option](https://user-images.githubusercontent.com/649992/34791854-33beba50-f614-11e7-8f85-67b6c3d5ba29.png)

Let me know if y'all have thoughts on this.

dternyak:
Definitely makes sense to start letting them know what wallet options we recommend earlier in the process rather than later.

👍

dternyak:
The Parity Mobile Signer is a good candidate for this view as well.

wbobeirne:
Closing in favor of #557, to do all together.

PR #780, opened 2018-01-10
Discourage private keys (pt. 1)

wbobeirne:
Checks off a few from #557. Implements the warning text from v3, but cranks it up to 11. This was a pretty quick pass at this, so I'm looking for plenty of feedback on how it looks and reads. I call this part 1 since it doesn't hit all checkboxes, and I'd like to see #778 get resolved as a part 2.

## What This Does

* Makes users read and confirm warnings before unlocking private keys, keystore files, or mnemonic phrases
* Makes the Account view the default homepage.
* Adds a link to the Generate tab on Account's wallet unlock for if you don't have a wallet.

## What This Doesn't Do

* [Passwordify private key / mnemonic](https://github.com/MyEtherWallet/MyEtherWallet/issues/557#issuecomment-356710322)
* Force user to wait. I felt the checkboxes would be enough of a roadblock.
* Homepagify account. Could use some guidance on what this might look like.

### New Homepage

screen shot 2018-01-10 at 2 47 08 pm

### Insecure Unlock Warning

screen shot 2018-01-10 at 2 47 17 pm

### Acknowledging

acknowledge


wbobeirne:
From Taylor on Slack:
> I like it for when we have the apps. let’s circle back in a week after we see what the general feedback on the current changes are from the community
> Im curious to know if we’re being “too scary”
> finding the balance between educate, scary, etc will be good

I'll leave this PR open while we wait for more feedback about the messaging in v3.

wbobeirne:
I'm pretty sure we're moving forward on this one. I've merged the latest develop in, and confirmed that it does not bother you on the Electron or HTML download builds.

PR #785, opened 2018-01-10
Better Offline UX

### LINKED ISSUE #661: Offline status no longer being polled
I'm guessing there was a regression in `transaction-refactor`, but going offline is no longer detected, the saga is no longer polling for it.

wbobeirne:
Closes #661, and implements a bunch of QOL fixes for when the user is offline.

## What This Does

* Starts "polling" user online status right at page load.
     * Only actually polls if navigator.onLine disagrees with our current online status, doesn't just poll every X seconds.
* Doesn't allow attempting unlock of wallets that require internet connection
    * MetaMask, Trezor
* Doesn't make requests while offline to reduce error messages
    * Wallet balance, token balances, swap, equivalent values
    * If offline status changes, will request again without needing to reload page
* Doesn't render tabs that are online-only while offline
    * Swap, Contracts, ENS, Broadcast Tx
* Shows a less severe offline notification if you start the page off in online mode
    * I'm not super wild about the way this looks, so let me know if you have a better idea
* Adds a little indicator to the header of your current status

### Unlock Offline

screen shot 2018-01-10 at 5 44 31 pm

### Send Offline

screen shot 2018-01-10 at 5 48 26 pm

### Online-Only Tab Block

screen shot 2018-01-10 at 5 48 50 pm

<sup>It's got a neat lil' animation too!</sup>

### Status Indicator

screen shot 2018-01-10 at 5 44 45 pm

screen shot 2018-01-10 at 5 45 00 pm

### Info Message When Starting Offline

screen shot 2018-01-10 at 5 58 42 pm


dternyak:
🔥 🔥 🔥 

This implementation is extremely user-friendly. Awesome work.

A few notes:

- [ ] While nonce displays when the user is "truly" offline, it should still be displayed when toggling advanced. This doesn't **need** to be part of this PR, but it's similar enough to the new functionality that it's worth getting in imo.

- [ ] Force offline reducers / selectors / action (and the general concept) is still present. We can delete all of that as a result of this new UX.

- [ ] Tests are failing (minor)

dternyak:
I noticed a small regression in the token card styles:

OLD
screen shot 2018-01-10 at 5 54 56 pm

NEW
screen shot 2018-01-10 at 5 54 53 pm

wbobeirne:
Fixed the style regression, and all of the saga tests I broke with the new offline yields.

dternyak:
Extremely minor, but `Nonce` doesn't curerntly have a placeholder, and it's slightly off-putting. Throwing in a placeholder there would be great.

wbobeirne:
Checked off nonce & forceOffline removal. Hope that didn't add _too_ much PR noise.

screen shot 2018-01-10 at 7 32 45 pm

dternyak:
@wbobeirne Breaking tests - otherwise looks good 👍

PR #789, opened 2018-01-11
Add webpack-subresource-integrity plugin

### LINKED ISSUE #690: Add resource integrity attributes to script tags
As suggested in [this gh issue](https://github.com/jantimon/html-webpack-plugin/issues/491), the [webpack-subresource-integrity](https://www.npmjs.com/package/webpack-subresource-integrity) module will add hashes to script tags that will reject the script if it doesn't match the hash. Should prevent a case in which someone is able to get a hold of whatever cdn we're using.

eddiewang:
This PR closes #690 and adds the integrity to the script that's injected. The script will then be rejected if it doesn't match the hash in the attribute (aka. if our CDN was compromised in some way). I'm not too familiar with this process - I was able to find the integrity attribute in the HTML source after adding the plugin, some eyes to confirm this plugin is working as intended would be helpful.


dternyak:
I don't know what's up here, but CI refuses to behave. 

I'll try re-running again tomorrow, otherwise we'll proceed without CI.

PR #790, opened 2018-01-11
Remove extra fonts

james-prado:
Closes https://github.com/MyEtherWallet/MyEtherWallet/issues/629

- Remove Lato-Bold (it wasn't being used)
- Remove .ttf, .eot, .woff, and .svg files types that aren't being used
- Replace all instances of `font-weight: 200;` with `font-weight: 300`
    - Font weight was rounding up to 'Lato-Light' (300 font-weight), because 'Lato-Ultralight' (200 font-weight) isn't available


common/assets/styles/etherwallet-fonts.less
```diff
@@ -1,24 +1,13 @@
...
 }
-
-@font-face {
-  font-family: 'Lato';
-  src: url('../fonts/Lato-Bold.woff') format('woff'),
-    url('../fonts/Lato-Bold.ttf') format('truetype');
-
-  font-style: normal;
-  font-weight: 700;
-}
```
wbobeirne: It's not? I see that h3's and h4's apparently use it.

common/assets/styles/etherwallet-fonts.less
```diff
@@ -1,24 +1,13 @@
...
+  src: url('../fonts/Lato-Light.woff2') format('woff2');
   font-style: normal;
   font-weight: 300;
 }
 
 @font-face {
   font-family: 'Lato';
-  src: url('../fonts/Lato-Regular.woff') format('woff'),
-    url('../fonts/Lato-Regular.ttf') format('truetype');
+  src: url('../fonts/Lato-Regular.woff2') format('woff2');
```
wbobeirne: [Not all browsers are up to woff2 yet](https://caniuse.com/#feat=woff2), if we're going to narrow down to one I'd stick with woff.

common/assets/styles/etherwallet-fonts.less
wbobeirne: I also see a couple of `font-weight: bold;`s. I think we should keep it.

common/assets/styles/etherwallet-fonts.less
james-prado: Good catch, yea it'd be better to support woff then

PR #793, opened 2018-01-11
Stricter balance checks during swap

### LINKED ISSUE #745: Loading indicator while lite-send balance is resolving
Lite-send will default to "valid" transaction while balances are loaded to determine if the user has enough funds to complete the swap.

In cases where balance takes significant time to retrieve, this will allow the user to complete the transaction before they should be allowed to.

Instead of defaulting to valid, show a loading indicator until the balance state is resolved.

eddiewang:
Closes #745

I believe this is the desired code - but need some clarity on how to reproduce the bug. Essentially I use the getBalance selector to ensure that I get back a non-null balance (we don't care if it's a TokenBalance or Ether balance) before showing the Generate Transaction button. This way, we ensure the use can't accidentally generate a signed tx before the actual valid tx is loaded for LiteSend.


eddiewang:
@dternyak the bug you mentioned seems to only occur with MetaMask - I haven't been able to reproduce with my Ledger after ~20 tests with 5 different tokens with both valid and invalid values. Generally in this interface, what I found was that if there balance was less than the raw value, then the CurrentValue.value is null.

```
interface ICurrentValue {
  raw: string;
  value: TokenValue | Wei | null;
}
```

I also confirmed that on slow networks, lite-send will not let you sign the tx, even if the error takes a little while to show up:



That said - just for sanity, I added a balance check so that if a non-null value is passed into the value prop even when CurrentValue.raw > CurrentValue.value then that should be caught as well.

PR #807, opened 2018-01-13
ENS Resolution for AddressField Component

HenryNguyen5:
## What this does
- Adds ENS resolution to the `AddressField` component 

## Features 
- Caches previously searched names in redux state for successful lookups
- Displays notifications on failed lookups
- Locks address input when ENS is resolving the name w/ spinner and loading text
- Displays resolved address below input box on completion


common/components/AddressFieldFactory/AddressInputFactory.tsx
```diff
@@ -1,29 +1,51 @@
...
+  ensAddress,
+  rawAddress
+}) => {
+  const isENS = isValidENSAddress(ensAddress);
+  const text = 'Loading ENS address...';
+  if (isLoading) {
+    return (
+      <>
+        <Spinner /> {text}
+      </>
```
wbobeirne: Whaaaat is this syntax?!

common/actions/ens/actionTypes/resolveDomain.ts
```diff
@@ -0,0 +1,28 @@
...
+
+export interface ResolveDomainSucceeded {
+  type: TypeKeys.ENS_RESOLVE_DOMAIN_SUCCEEDED;
+  payload: { domain: string; domainData: DomainRequest };
+}
+
+export interface ResolveDomainCached {
+  type: TypeKeys.ENS_RESOLVE_DOMAIN_CACHED;
+  payload: { domain: string };
+}
```
wbobeirne: For simplicity, does this need to be a separate action? Can we just not re-use succeeded?

common/libs/ens/index.ts
```diff
@@ -0,0 +1,86 @@
...
+}
+
+export const modeStrMap = name => [
+  `${name} is available and the auction hasn’t started`,
+  `${name} is available and the auction has been started`,
+  `${name} is taken and currently owned by someone`,
+  `${name} is forbidden`,
+  `${name} is currently in the ‘reveal’ stage of the auction`,
+  `${name} is not yet available due to the ‘soft launch’ of names.`
+];
```
wbobeirne: Just for readability, would be nice to make this an object that uses the enums to make it clear they're 1:1.

common/libs/validators.ts
```diff
@@ -22,6 +22,9 @@ export function isValidETHAddress(address: string): boolean {
   }
 }
 
+export const isGenesisAddress = (address: string): boolean =>
```
dternyak: I'm unfamiliar with genesis addresses.

common/sagas/ens/ens.ts
```diff
@@ -0,0 +1,68 @@
...
+      const shouldResolve = yield call(shouldResolveDomain, domain);
+      if (!shouldResolve) {
+        yield put(resolveDomainCached({ domain }));
+        continue;
+      }
+
+      const node: INode = yield select(getNodeLib);
+      const result = yield race({
+        domainData: call(resolveDomainRequest, domain, node),
+        err: call(delay, 4000)
```
wbobeirne: Seems a little short, that Colombia internet has made me appreciate how rough RTTs can be on slow speeds.

common/components/AddressFieldFactory/AddressInputFactory.tsx
```diff
@@ -1,29 +1,51 @@
...
+    );
+  } else {
+    return isENS ? <>{`Resolved Address: ${rawAddress}`}</> : null;
+  }
```
wbobeirne: The loader shows up even if the address is invalid (i.e. too short) but then you get no message explaining what happened. There should be a message for "invalid ens".

screen shot 2018-01-15 at 1 20 49 am

common/sagas/ens/ens.ts
```diff
@@ -0,0 +1,68 @@
...
+      if (!domainData) {
+        throw Error();
+      }
+      const domainSuccessAction = resolveDomainSucceeded(domain, domainData);
+      yield put(domainSuccessAction);
+      yield;
+    } catch (e) {
+      const domainFailAction = resolveDomainFailed(domain, e);
+      yield put(domainFailAction);
+      yield put(showNotification('danger', e.message || 'Could not resolve ENS address', 5000));
```
wbobeirne: I feel like inline errors were a lot easier to grok as a user than notifications. Would be nice if we could maintain that (but by no means a showstopper.)

common/components/AddressFieldFactory/AddressInputFactory.tsx
HenryNguyen5: Its a `React.Fragment`

dternyak: React Fragment short-hand :)

common/actions/ens/actionTypes/resolveDomain.ts
HenryNguyen5: Would rather have it explicitly different for readability reasons

common/sagas/ens/ens.ts
HenryNguyen5: I think we could do that in a seperate PR, that would give better error messages for all transaction fields. Adding inline error messages to the `AddressField` component means having different error messages being stored in state instead of just checking `null`, which at that point all of the transaction state should support specialized error messages.

PR #808, opened 2018-01-13
Toggle setting gasLimit automatically

james-prado:
Closes https://github.com/MyEtherWallet/MyEtherWallet/issues/796

- Show loading indicator while estimating gas
	- times out after 10 seconds, this should be more than enough, throttling my network to the low end of 3G I still couldn't timeout. Even if I could, I don't think any user would want to sit for 10+, or 30+ seconds for a gasLimit estimate
- Prompt user on timeout to set gasLimit manually in 'Advanced settings'
- Rename 'Advanged: GasPrice, GasLimit, Nonce, Data' to 'Advanced settings'
- Add toggle for auto setting gasLimit in advanced settings
	- new `setGasLimit: boolean` prop in `state.config`
	- this prop is persists through localstorage with the rest of the redux state


dternyak:
While this was not originally discussed in the issue, I've realized that it may be dangerous to simply allow users to input their own gas limits if we're not able to provide gas estimation within a reasonable timeframe via the currently selected node. In general, users will not be willing/able to research the expected gas limit for token transfers, and certainly not for more complex contract interaction via Contracts tab. 

While this is certainly major scope creep, a more robust solution may include implementing https://github.com/MyEtherWallet/MyEtherWallet/issues/802. With https://github.com/MyEtherWallet/MyEtherWallet/issues/802, any JSON RPC request taking over ~10 seconds can be cancelled, and the request can be re-attempted against another healthy node.

common/components/GasSlider/components/AdvancedGas.tsx
```diff
@@ -3,53 +3,78 @@ import classnames from 'classnames';
...
+            value={gasPrice}
             onChange={this.handleGasPriceChange}
           />
         </div>
 
-        <div className="col-md-4 col-sm-6 col-xs-12">
+        <div className="col-md-4 col-sm-6 col-xs-12 gasLimit-input-grp">
           <label>{translate('OFFLINE_Step2_Label_4')}</label>
+          <div className="SimpleGas-flex-spacer" />
+          <CSSTransition in={estimatingGas} timeout={300} classNames="fade">
```
dternyak: duplicate

common/components/GasSlider/components/SimpleGas.tsx
```diff
@@ -3,22 +3,41 @@ import Slider from 'rc-slider';
...
+    const { gasPrice, gasEstimationStatus } = this.props;
+    const estimatingGas = gasEstimationStatus === 'PENDING' ? true : false;
+    const estimatingGasTimeout = gasEstimationStatus === 'TIMEOUT' ? true : false;
     return (
       <div className="SimpleGas row form-group">
-        <div className="col-md-12">
+        <div className="col-md-12 SimpleGas-title">
           <label className="SimpleGas-label">{translate('Transaction Fee')}</label>
+          <div className="SimpleGas-flex-spacer" />
+          <CSSTransition in={estimatingGas} timeout={300} classNames="fade">
```
dternyak: duplicate

common/components/GasSlider/components/SimpleGas.tsx
```diff
@@ -3,22 +3,41 @@ import Slider from 'rc-slider';
...
+            <div className={`SimpleGas-estimating small ${estimatingGas ? 'active' : ''}`}>
+              Setting gas limit
```
wbobeirne: Nitpick: I think "Setting" doesn't make a lot of sense here, since that shouldn't take time. Maybe "Calculating" or "Determining" or "Fetching" or something like that.

common/components/GasSlider/components/SimpleGas.tsx
```diff
@@ -3,22 +3,41 @@ import Slider from 'rc-slider';
...
+              <Spinner />
+            </div>
+          </CSSTransition>
         </div>
+        {estimatingGasTimeout && (
+          <div className="col-md-12 prompt-toggle-gas-limit">
+            <p className="small">
+              Couldn't set gas limit, try setting manually in advanced settings
```
wbobeirne: I'd recommend switching nodes, or telling them to wait before asking a user to set the gas limit manually. Maybe 1% of users will know what to do in this situation. But for 99% of cases, we should probably just be able to set it to 21000 right?

common/components/GasSlider/components/SimpleGas.scss
```diff
@@ -34,3 +50,26 @@
     }
   }
 }
+
+.fade {
```
wbobeirne: We should add a slight transition delay on entry here (~300ms.) I noticed that this loader would pop up _every change I made_, even if there was no gas request made to the node. Either that, or we should make it so it only goes into PENDING state on network request.

common/libs/nodes/rpc/index.ts
```diff
@@ -16,6 +16,15 @@ import {
   isValidRawTxApi
 } from '../../validators';
 
+const timeout = (ms, promise: Promise<any>): any => {
+  return new Promise((resolve, reject) => {
+    setTimeout(() => {
+      reject(new Error('timeout'));
```
wbobeirne: Just in case there are some catches out there that display `error.message`, something like:
> The request timed out, try switching nodes or checking your connection

common/components/GasSlider/components/AdvancedGas.tsx
```diff
@@ -3,53 +3,78 @@ import classnames from 'classnames';
...
       <div className="AdvancedGas row form-group">
+        <div className="col-md-12">
+          <label className="checkbox">
+            <input
+              type="checkbox"
+              className="form-check-input"
+              defaultChecked={setGasLimit}
+              onChange={this.handleToggleSetGasLimit}
+            />
+            <span>Automatically Set Gas Limit</span>
```
wbobeirne: As with previous comments, I'd call this calculate or something. Also, now that we have an explicit checkbox, I think we should make the gasLimit input disabled while this is unchecked.

common/components/GasSlider/components/AdvancedGas.tsx
```diff
@@ -3,53 +3,78 @@ import classnames from 'classnames';
...
             type="number"
             placeholder="e.g. 40"
-            value={vals.gasPrice}
+            value={gasPrice}
             onChange={this.handleGasPriceChange}
           />
         </div>
 
-        <div className="col-md-4 col-sm-6 col-xs-12">
+        <div className="col-md-4 col-sm-6 col-xs-12 gasLimit-input-grp">
```
wbobeirne: Any reason for the nonstandard classname?

common/components/GasSlider/components/SimpleGas.tsx
```diff
@@ -3,22 +3,41 @@ import Slider from 'rc-slider';
...
+          <div className="SimpleGas-flex-spacer" />
+          <CSSTransition in={estimatingGas} timeout={300} classNames="fade">
+            <div className={`SimpleGas-estimating small ${estimatingGas ? 'active' : ''}`}>
+              Setting gas limit
+              <Spinner />
+            </div>
+          </CSSTransition>
         </div>
+        {estimatingGasTimeout && (
+          <div className="col-md-12 prompt-toggle-gas-limit">
```
wbobeirne: Non-standard classname

common/components/GasSlider/components/AdvancedGas.tsx
```diff
@@ -3,53 +3,78 @@ import classnames from 'classnames';
...
+    // Shadow var names on gasLimit
+    const { gasEstimationStatus, setGasLimit, gasPrice, nonce } = this.props;
+    const estimatingGas = gasEstimationStatus === 'PENDING' ? true : false;
     return (
       <div className="AdvancedGas row form-group">
+        <div className="col-md-12">
+          <label className="checkbox">
+            <input
+              type="checkbox"
+              className="form-check-input"
```
wbobeirne: This is a Bootstrap v4 classname, we're still on [3.3](getbootstrap.com/docs/3.3/css/), so it's not doing anything.

common/components/GasSlider/components/SimpleGas.tsx
dternyak: Token transfers are a common usecase, and 21000 would not suffice there.

common/components/GasSlider/components/SimpleGas.scss
james-prado: I don't think delaying it will help, it just pushed back the problem. The conditions for the spinner were wrong when I commited, this has been fixed now so it shouldn't show up on every change

common/components/GasSlider/components/SimpleGas.tsx
wbobeirne: Right, I meant the behavior could be `if (!data) return 21000`, and only bug the user about estimate failures if they have data (from token or otherwise.)

PR #811, opened 2018-01-13
Handle Shapeshift Error Messages

### LINKED ISSUE #806: Swap doesn't gracefully handle ShapeShift's blocked regions
If you try to swap from my lovely state of New York, ShapeShift responds with:

```
{
  "error": {
    "code": "geoRestriction",
    "message": "You have tried to access a protected end point from a blocked region. See http://pleaseprotectconsumers.org for more info",
    "data": {
      "url": "http://pleaseprotectconsumers.org"
    }
  }
}
```

While the error message that pops up of:

> Connection Error. Please check the developer console for more details and/or contact support

Isn't necessarily wrong, it'd be nice if it handled this a little more graciously. Perhaps we should have a dictionary of `error.code` lookups for ShapeShift.

eddiewang:
Closes #806 . I wasn't able to clearly understand what Shapeshift's Error Message structure is like from the Docs (It looks like some are simply string messages while others have the message contained as a property in an object), so this PR checks for both and tries to display a more helpful err message for the user.


common/sagas/swap/orders.ts
```diff
@@ -185,10 +185,13 @@ export function* postShapeshiftOrderCreate(
...
     }
   } catch (e) {
-    const message =
-      'Connection Error. Please check the developer console for more details and/or contact support';
-    yield put(showNotification('danger', message, TEN_SECONDS));
-    yield put(shapeshiftOrderCreateFailedSwap());
+    if (e.message) {
+      yield put(showNotification('danger', e.message, TEN_SECONDS));
+    } else {
+      const message = `Connection Error. ${e}`;
```
wbobeirne: `e` could be a whole host of things (Some of which do not stringify well). I would stick with the original error message, and `console.error` e to help with debugging.

eddiewang:
@wbobeirne added console.errors for the two sagas (bity + shapeshift).

common/sagas/swap/orders.ts
```diff
@@ -185,10 +186,15 @@ export function* postShapeshiftOrderCreate(
       yield put(startPollShapeshiftOrderStatus());
     }
   } catch (e) {
-    const message =
-      'Connection Error. Please check the developer console for more details and/or contact support';
-    yield put(showNotification('danger', message, TEN_SECONDS));
-    yield put(shapeshiftOrderCreateFailedSwap());
+    if (e && e.message) {
+      yield put(showNotification('danger', e.message, TEN_SECONDS));
```
wbobeirne: Just putting this in the review for posterity: doesn't this case also need
```
yield put(shapeshiftOrderCreateFailedSwap());
```

eddiewang: @wbobeirne i'll move it after the conditional but yes - i think you're right we need to call that on both cases.

wbobeirne:
This is what I'm getting when I hit the error:

screen shot 2018-01-14 at 11 26 26 pm

I think we throw a generic error somewhere. I can finish this task off since I can trigger it, if that's OK with you.

dternyak:
👍

wbobeirne:
Unfortunately the failure happens during the CORS preflight, so the response is not available. So I just added a catch-all for TypeErrors (Which AFAIK only happen from CORS rejections.)

screen shot 2018-01-15 at 12 32 55 am

PR #813, opened 2018-01-14
Remove non-Ethereum networks

dternyak:
As of https://github.com/kvhnuke/etherwallet/issues/1400, RSK node is no longer being maintained. 

It has been removed on V3, and we can follow suite on V4.

While we intend to support UBIQ, EXP, and ETC, the current integration is not robust enough for a release candidate (as seen in issues such as https://github.com/MyEtherWallet/MyEtherWallet/issues/814 and unhandled wallet format accessibility issues).

We also make assumptions about Ethereum as the network in the Payment request tab, and potentially other yet unknown areas.

After a more complete integration and testing we can re-enable these networks.

### DIFF
--- a/common/config/data.ts
+++ b/common/config/data.ts
@@ -173,24 +173,6 @@ export const NETWORKS: { [key: string]: NetworkConfig } = {
     tokens: require('./tokens/rinkeby.json'),
     contracts: require('./contracts/rinkeby.json'),
     isTestnet: true
-  },
-  EXP: {
-    name: 'EXP',
-    unit: 'EXP',
-    chainId: 2,
-    color: '#673ab7',
-    blockExplorer: makeExplorer('http://www.gander.tech'),
-    tokens: require('./tokens/exp.json'),
-    contracts: require('./contracts/exp.json')
-  },
-  UBQ: {
-    name: 'UBQ',
-    unit: 'UBQ',
-    chainId: 8,
-    color: '#b37aff',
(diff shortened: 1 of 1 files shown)


wbobeirne:
This seems like a shame to remove, as it's a lot harder to find the areas where we hard-coded ETH if we don't have other networks. Do you know if we get a lot of ETC usage? I'd definitely like to get at least that figured out pre-launch.

dternyak:
We don't have specific metrics, but I'm sure they do get significant usage.

The integration itself has some faults, but I would like to prioritize a
correct integration as well.

The issues I know of today include wallet format incompatibilites (e.g
trezor on exp) and equivalent values. I'm fairly confident that there are
other issues given how little attention these additional networks have had
so far.

ShapeShift integration is also not fully completed - alternative networks
as origin kind are not supported.

Overall, these networks have known issues and should be disabled until we
can cleanly integrate them.

dternyak:
We don't have specific metrics, but I'm sure they do get significant usage.

The integration itself has some faults, but I would like to prioritize a
correct integration as well.

The issues I know of today include:

- wallet format incompatibilities (e.g trezor on exp) 
- equivalent values
- payment requests

I'm fairly confident that there are
other issues given how little attention these additional networks have had
so far.

ShapeShift integration is also not fully completed - alternative networks
as origin kind are not supported.

Overall, these networks have known issues and should be disabled until we
can cleanly integrate them.

PR #815, opened 2018-01-14
Change network without window refresh

dternyak:
Closes https://github.com/MyEtherWallet/MyEtherWallet/issues/642

I believe wallet and transaction state is the only state that needs to get reset when changing networks. I couldn't think of any any other state that would be need to be reset during a network change, but please comment if you think there is.

Branched off of https://github.com/MyEtherWallet/MyEtherWallet/pull/813, will re-base branch after merge.

### DIFF
--- a/common/components/ui/UnlockHeader.tsx
+++ b/common/components/ui/UnlockHeader.tsx
@@ -20,7 +20,7 @@ export class UnlockHeader extends React.Component<Props, State> {
   };
 
   public componentDidUpdate(prevProps: Props) {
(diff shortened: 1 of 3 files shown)


wbobeirne:
Oh but did you mean to merge this into remove-rsk instead of develop?

dternyak:
@wbobeirne I mentioned that I had branched off of it, so was waiting on it to get merged before re-basing against develop.

I can also cherry-pick these commits off of develop instead of `remove-rsk` so we don't have to depend on it, but if it's going to get merged shortly it won't matter.

Issue #828, opened 2018-01-15
Alpha beginner impressions / usability issues


jidanni:
On https://alpha.myetherwallet.com/
```
  <ul class="ColorDropdown dropdown-menu dropdown-menu-right">
    <li class="ColorDropdown-item" style=
    "border-color: rgb(14, 151, 192);">
      <a class="active"><span>ETH
      <small>(MyEtherWallet)</small></span></a>
    </li>
    <li class="ColorDropdown-item" style=
    "border-color: rgb(14, 151, 192);">
      <a class=""><span>ETH
      <small>(Etherscan.io)</small></span></a>
    </li>
    <li class="ColorDropdown-item" style=
    "border-color: rgb(14, 151, 192);">
      <a class=""><span>ETH <small>(infura.io)</small></span></a>
    </li>
    <li class="ColorDropdown-item" style=
    "border-color: rgb(102, 144, 115);">
      <a class=""><span>ETC <small>(Epool.io)</small></span></a>
    </li>
    <li class="ColorDropdown-item" style=
    "border-color: rgb(173, 193, 1);">
      <a class=""><span>Ropsten
      <small>(MyEtherWallet)</small></span></a>
    </li>
    <li class="ColorDropdown-item" style=
    "border-color: rgb(173, 193, 1);">
      <a class=""><span>Ropsten
      <small>(infura.io)</small></span></a>
    </li>...
```
is broken up into subgroups. The first group has three items, the next one, etc.
However they are anonymous sub-groups. Each one should have a header giving them a title.

dternyak:
@jidanni Thanks for all of these accessibility issues!

Would you mind if I asked you to collapse these into a single issue with multiple comments / bullet points? It would certainly help on our end to have a single source of accessibility issues instead of searching against all open issues. 

Let me know if this is something you're okay with doing :)

jidanni:
@dternyak OK. Feel free to retitle this bug "Alpha beginner impressions / usability issues". I'll append the rest here. Thanks.

dternyak:
@jidanni Great. Feel free to close the other issues yourself once they're moved over, or I can do so myself if you let me know.

jidanni:
OK.
Here we see the user has 0.5 Ropstein in his account, but can only select sending Ether.

![nj](https://user-images.githubusercontent.com/1230959/34933471-ebbf6608-fa11-11e7-8a2b-af5132963381.jpg)

jidanni:
[...]

PR #849, opened 2018-01-17
Support Non-Ethereum Networks

dternyak:
Closes https://github.com/MyEtherWallet/MyEtherWallet/issues/843

Adds network support for:
- UBQ
- ETC
- EXP

Adds Legacy path for:
- SingularDTV

Includes
- Refactoring **(please excuse the diff-noise related to module exporting config objects)**
- Groups dPaths by network instead of referencing network by string in object (fragile)
- References correct default in MnemonicModal instead of 1st index in platform array
    - In practice, this means a trezor/ledger unlock while on ETC network will default to ETC Trezor path instead of ETH path
- Improved type safety across network interaction
- Whitelist wallet formats based on network (`isSupportedWalletFormat` selector)

TODO
- [x] Write tests to make sure each network has dPathFormats
    - alternatively fix typing so that runtime network additions are allowed to not have dPathFormats without impacting "compile time" networks.
- [x] Write tests to ensure that each network has a different chainId
- [x] Blacklisting specific features per network (perhaps via sub-tab?).
- [ ] Don't assume ether for equivalent values
- [x] Audit / research other areas where we assume ether

Nice to have
- [ ] Group nodes by network instead of referencing network in object (fragile)

screen shot 2018-01-17 at 2 21 07 am
screen shot 2018-01-17 at 2 21 25 am


common/config/networks.ts
```diff
@@ -0,0 +1,332 @@
...
+
+export interface DPathFormats {
+  trezor: DPath;
+  ledgerNanoS: DPath;
+  mnemonicPhrase: DPath[];
+}
+
+export interface NetworkConfig {
+  // TODO really try not to allow strings due to custom networks
+  name: NetworkKeys | string;
```
dternyak: We could type this as `NetworkKeys` if it wasn't for `makeNetworkConfigFromCustomConfig` - would appreciate feedback on how we could address this.

common/selectors/config.ts
```diff
@@ -4,18 +4,80 @@ import {
...
+    SecureWallets.TREZOR,
+    InsecureWallets.MNEMONIC_PHRASE
+  ] as any; // TODO understand why "as Wallets" is a type error
+  if (CHECK_FORMATS.includes(format)) {
+    return !!getAnyDPath(format as AnyDPathFormat, network);
+  }
+  if (format === SecureWallets.WEB3) {
+    if (network !== 'ETH') {
+      // TODO -- determine if we can select testnets via MetaMask
+      return false;
```
dternyak: MetaMask will be disabled on TestNets until we address this.

common/components/WalletDecrypt/components/DeterministicWalletsModal.tsx
```diff
@@ -118,14 +126,16 @@ class DeterministicWalletsModalClass extends React.Component<Props, State> {
             <select
               className="form-control"
               onChange={this.handleChangePath}
-              value={isCustomPath ? 'custom' : dPath}
+              value={this.state.currentLabel || this.findDPath('value', dPath).value}
             >
               {dPaths.map(dp => (
-                <option key={dp.value} value={dp.value}>
+                <option key={dp.value} value={dp.label}>
```
dternyak: reminder to myself to double check this

common/components/WalletDecrypt/components/Mnemonic.tsx
```diff
@@ -1,11 +1,13 @@
...
 import React, { Component } from 'react';
 import translate, { translateRaw } from 'translations';
 import DeterministicWalletsModal from './DeterministicWalletsModal';
 import { formatMnemonic } from 'utils/formatters';
+import { ETH_DEFAULT } from 'config/dpaths';
+import { getPaths } from 'selectors/config';
+import { InsecureWalletName } from 'config';
 
-const DEFAULT_PATH = DPATHS.MNEMONIC[0].value;
+const DEFAULT_PATH = ETH_DEFAULT.value;
```
dternyak: Need to use selector here as well to get default mnemonic path instead of assuming ether (for other networks)

dternyak:
Re-opens https://github.com/MyEtherWallet/MyEtherWallet/issues/642 (DeterministicWallet component state needs to be flushed)

common/components/WalletDecrypt/WalletDecrypt.tsx
```diff
@@ -171,7 +175,7 @@ export class WalletDecrypt extends Component<Props, State> {
       unlock: this.props.unlockPrivateKey,
       helpLink: `${knowledgeBaseURL}/private-keys-passwords/difference-beween-private-key-and-keystore-file.html`
     },
-    'view-only': {
+    [MiscWalletName.VIEW_ONLY]: {
```
wbobeirne: Purely a question: Is there any way to just do `WalletName.X` instead of secure / insecure / misc? Or is this important for typescript?

common/components/WalletDecrypt/components/DeterministicWalletsModal.tsx
```diff
@@ -113,21 +122,20 @@ class DeterministicWalletsModalClass extends React.Component<Props, State> {
         handleClose={onCancel}
       >
         <div className="DWModal">
+          {/* TODO: replace styles for flexbox with flexbox classes in https://github.com/MyEtherWallet/MyEtherWallet/pull/850/files#diff-2150778b9391533fec7b8afd060c7672 */}
```
wbobeirne: From my comments over there, I think I would prefer flex ui components over classes. Just a thought.

common/components/WalletDecrypt/components/DeterministicWalletsModal.tsx
```diff
@@ -113,21 +122,20 @@ class DeterministicWalletsModalClass extends React.Component<Props, State> {
...
           <form className="DWModal-path form-group-sm" onSubmit={this.handleSubmitCustomPath}>
-            <span className="DWModal-path-label">Addresses for</span>
-            <select
-              className="form-control"
+            <span className="DWModal-path-label">Addresses </span>
```
wbobeirne: Did you have any thoughts from the commentary in #828 on how we could make it more clear what this is? Could be good to get in some usability improvements while we're here, but feel free to punt to a future PR.

common/components/WalletDecrypt/components/DeterministicWalletsModal.tsx
```diff
@@ -113,21 +122,20 @@ class DeterministicWalletsModalClass extends React.Component<Props, State> {
...
+            <Select
+              name="fieldDPath"
+              className=""
+              value={this.state.currentLabel || this.findDPath('value', dPath).value}
```
wbobeirne: I feel like the nomenclature of using "label" for something that's a value is fairly confusing.

common/components/WalletDecrypt/components/LedgerNano.tsx
```diff
@@ -22,11 +28,13 @@ interface State {
...
 
-export class LedgerNanoSDecrypt extends Component<Props, State> {
+type Props = OwnProps & StateProps;
+
+class LedgerNanoSDecryptClass extends Component<Props, State> {
   public state: State = {
     publicKey: '',
     chainCode: '',
-    dPath: DEFAULT_PATH,
+    dPath: this.props.dPath.value,
```
wbobeirne: Can you do this? I thought this had to happen in constructor.

common/components/WalletDecrypt/components/Trezor.tsx
```diff
@@ -133,7 +145,16 @@ export class TrezorDecrypt extends Component<Props, State> {
...
 }
+
+function mapStateToProps(state: AppState): StateProps {
+  const network = getNetworkConfig(state);
+  return {
+    dPath: getSingleDPath(SecureWalletName.TREZOR, network)
+  };
+}
+
+export const TrezorDecrypt = connect(mapStateToProps)(TrezorDecryptClass);
```
wbobeirne: To reduce # of connected components (Which we should strive to do more for performance, imo), I think we should just send each wallet the `network` and they can run this in constructor or whatever. It needs it anyway to disable certain wallets on certain networks.

common/components/WalletDecrypt/components/WalletButton.tsx
```diff
@@ -1,23 +1,33 @@
...
+  walletType: WalletName;
   isSecure?: boolean;
   isReadOnly?: boolean;
   isDisabled?: boolean;
   onClick(walletType: string): void;
 }
 
-export class WalletButton extends React.Component<Props, {}> {
+interface StateProps {
+  isFormatDisabled?: boolean;
```
wbobeirne: Do we need to distinguish this?  Could just do `isDisabled: ownProps.isDisabled || ...` in `mapStateToProps`. But from my comment above about reducing connected components, maybe we should be doing this in WalletDecrypt and just have that keep track of `network`.

common/components/ui/UnlockHeader.tsx
```diff
@@ -1,20 +1,23 @@
...
-  disabledWallets?: string[];
+  disabledWallets?: WalletName[];
 }
+
 interface State {
   isExpanded: boolean;
 }
-export class UnlockHeader extends React.Component<Props, State> {
+
+export class UnlockHeader extends React.PureComponent<Props, State> {
```
wbobeirne: We should def. be doing this more. In some cases it costs performance, but it can provide some nice savings on some more complex components.

common/components/ui/UnlockHeader.tsx
```diff
@@ -57,7 +60,7 @@ export class UnlockHeader extends React.Component<Props, State> {
     );
   }
 
-  public toggleisExpanded = () => {
+  public toggleisExpanded = (_: React.FormEvent<HTMLButtonElement>) => {
```
wbobeirne: If it's unused, is there any reason in specifying?

common/config/networks.ts
wbobeirne: I don't think there's a way around this. The custom configs are a pain for typing in general (See also, custom nodes.) We could make a specific type for custom network keys that is just string, but to be more specific about when we expect it to be either / or.

common/sagas/config.ts
```diff
@@ -171,12 +168,14 @@ export function* handleNodeChangeIntent(action: ChangeNodeIntentAction): SagaIte
   yield put(setLatestBlock(latestBlock));
   yield put(changeNode(action.payload, actionConfig, actionNetwork));
 
-  const currentWallet: IWallet | null = yield select(getWalletInst);
-
+  // TODO - renable once DeterministicWallet state is fixed to flush properly.
+  // TODO DeterministicWallet keeps path related state we need to flush before we can stop reloading
```
wbobeirne: Is there a reason we can't implement a `DW_RESET` action, and fire that on network changes?

common/components/WalletDecrypt/components/DeterministicWalletsModal.tsx
```diff
@@ -8,14 +8,16 @@ import {
...
+import { NetworkConfig } from 'config';
 import { isValidPath } from 'libs/validators';
 import React from 'react';
 import { connect } from 'react-redux';
 import { getNetworkConfig } from 'selectors/config';
 import { getTokens, MergedToken } from 'selectors/wallet';
 import { UnitDisplay } from 'components/ui';
 import './DeterministicWalletsModal.scss';
+import { DPath } from 'config/dpaths';
+import Select from 'react-select';
```
wbobeirne: I'm very down with using a select component. We should probably replace most if not all selects with this.

common/selectors/config.ts
```diff
@@ -4,13 +4,69 @@ import {
...
+  // TODO -- determine if we can select testnets via MetaMask
+  if (format === SecureWalletName.WEB3) {
+    if (network.name !== 'ETH') {
+      return false;
+    }
+  }
+
+  // All other wallet formats are supported
+  return true;
+}
```
wbobeirne: These are all handy, but I'd call these utils, not selectors. Selectors are a redux concept that are only supposed to take in state, and return a subset of the state.

common/components/WalletDecrypt/WalletDecrypt.tsx
dternyak: @HenryNguyen5 ^

common/components/WalletDecrypt/components/DeterministicWalletsModal.tsx
dternyak: Agreed, however these were modifications, not additions. Ideally DeterministicWalletsModal could be re-written to use much less component state, and these types of changes could be made then.

common/components/WalletDecrypt/components/Trezor.tsx
dternyak: Agreed

common/components/WalletDecrypt/components/WalletButton.tsx
dternyak: Agreed, I believe I mentioned somewhere this could use some refactoring.

common/components/ui/UnlockHeader.tsx
dternyak: @HenryNguyen5 ^

common/sagas/config.ts
dternyak: We certainly could, and that's the long-term solution. Probably out of scope here though.

common/selectors/config.ts
dternyak: Agreed, actually debated this with @HenryNguyen5 and we were both uncertain. I can move it to a utility file since you've brought it up as well.

common/components/WalletDecrypt/components/LedgerNano.tsx
dternyak: I'll need to check back, but during testing it works fine. It might however be because some state is being updated immediately after mount, so I can't be certain.

common/components/WalletDecrypt/components/DeterministicWalletsModal.tsx
dternyak: I think it's probably out of scope for now, considering this PR needs a Part 2 on it's own. It'll be good to circle back to address (😉) them though.

common/components/WalletDecrypt/components/DeterministicWalletsModal.tsx
dternyak: @james-prado Do you mind taking a look here?

common/components/WalletDecrypt/WalletDecrypt.tsx
```diff
@@ -132,16 +145,7 @@ export class WalletDecrypt extends Component<Props, State> {
       unlock: this.props.setWallet,
       helpLink: 'https://doc.satoshilabs.com/trezor-apps/mew.html'
     },
-    'digital-bitbox': {
```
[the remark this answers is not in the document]
dternyak: we do :)

common/config/networks.ts
```diff
@@ -0,0 +1,356 @@
...
+  tokens: require('./tokens/exp.json'),
+  contracts: require('./contracts/exp.json'),
+  dPathFormats: {
+    [SecureWalletName.TREZOR]: EXP_DEFAULT,
+    [SecureWalletName.LEDGER_NANO_S]: EXP_DEFAULT,
+    [InsecureWalletName.MNEMONIC_PHRASE]: EXP_DEFAULT
+  }
+};
+
+export type NetworkKeys = keyof typeof NETWORKS;
```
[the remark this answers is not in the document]
dternyak: Thanks for the heads up.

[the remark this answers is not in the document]
HenryNguyen5: not sure if it matters in this case as the types will get stripped out before JS processes it

common/selectors/config.ts
```diff
@@ -4,13 +4,69 @@ import {
...
+
+  const isHDFormat = (f: string): f is DPathFormat => CHECK_FORMATS.includes(f as DPathFormat);
+
+  // Ensure DPath's are found
+  if (isHDFormat(format)) {
+    return !!(network.dPathFormats && network.dPathFormats[format]);
+  }
+
+  // Ensure Web3 is only enabled on ETH
+  // TODO -- determine if we can select testnets via MetaMask
```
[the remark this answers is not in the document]
dternyak: The intent here was to be able to allow users to sign into there desired testnet on metamask if they have selected that network in the node dropdown.

I'll adjust the todo to clarify, agreed it's not very clear.

common/config/networks.ts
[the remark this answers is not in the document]
skubakdj: Good point, it probably doesn't :smile:

common/components/WalletDecrypt/components/WalletButton.tsx
dternyak: Addressed in https://github.com/MyEtherWallet/MyEtherWallet/pull/849/commits/43b5833c54d5a87abc686501b8478014c8455e86

common/components/WalletDecrypt/components/Trezor.tsx
dternyak: Addressed in https://github.com/MyEtherWallet/MyEtherWallet/pull/849/commits/43b5833c54d5a87abc686501b8478014c8455e86

common/components/WalletDecrypt/components/DeterministicWalletsModal.tsx
james-prado: For flexbox styles I prefer using classes. Elements wrapped in a `display: flex` style tend to have a lot of custom styles specific to each element that is hard to replicate with components, and all of this is becomes even harder when trying to provide support for phone and tablet screen widths 
ie. placing spacer components (empty divs with `flex-grow: 1`), wrapping, aligning

james-prado: Also, I'm not big on using many custom classes for flex-box, but a few classes that are regularly used and aren't too specific seem to work really well with the vast amount of configuration options for flexbox

common/components/WalletDecrypt/components/WalletButton.tsx
dternyak: Fixed in 366871c

dternyak:
@skubakdj Thanks for catching the issue with MetaMask. It's since been resolved.

common/components/WalletDecrypt/WalletDecrypt.tsx
HenryNguyen5: We can, but it helps with typing if you group together each wallet type into its own enum. Then when you map over them TS can infer which wallet type you're referring to and return the specific wallet info instead of a union of all of them

common/components/ui/UnlockHeader.tsx
HenryNguyen5: It's just to specify what it's being used on as an event handler.

common/utils/network.ts
```diff
@@ -26,3 +53,65 @@ export function getNetworkConfigFromId(
...
+type PathType = keyof DPathFormats;
+
+type DPathFormat =
+  | SecureWalletName.TREZOR
+  | SecureWalletName.LEDGER_NANO_S
+  | InsecureWalletName.MNEMONIC_PHRASE;
+
+export function getPaths(pathType: PathType): DPath[] {
+  const networkPaths: DPath[] = [];
+  Object.values(NETWORKS).forEach(networkConfig => {
```
HenryNguyen5: specifiying networkConfig as `NetworkConfig` in the function parameter will eliminate the need to do `path as DPath` on ln 69. This stops type coercion

common/utils/network.ts
```diff
@@ -26,3 +53,65 @@ export function getNetworkConfigFromId(
...
+  const dPathFormats = network.dPathFormats;
+  return dPathFormats[format];
+}
+
+export function isNetworkUnit(network: NetworkConfig, unit: string) {
+  const validNetworks = Object.values(NETWORKS).filter((n: NetworkConfig) => n.unit === unit);
+  return validNetworks.includes(network);
+}
+
+export function isSupportedOnNetwork(format: WalletName, network: NetworkConfig): boolean {
```
HenryNguyen5: would prefer more descriptive function names here such as `isWalletSupportedOnNetwork`

common/utils/network.ts
```diff
@@ -26,3 +53,65 @@ export function getNetworkConfigFromId(
...
+  // Ensure Web3 is only enabled on ETH or ETH Testnets (MetaMask does not support other networks)
+  if (format === SecureWalletName.WEB3) {
+    return isNetworkUnit(network, 'ETH');
+  }
+
+  // All other wallet formats are supported
+  return true;
+}
+
+export function allSupportedOnNetwork(network: NetworkConfig): WalletName[] {
```
HenryNguyen5: same here regarding function naming

common/utils/network.ts
```diff
@@ -26,3 +53,65 @@ export function getNetworkConfigFromId(
...
+  return walletNames.filter(walletName => isSupportedOnNetwork(walletName, network));
+}
+
+export function unSupportedOnNetwork(network: NetworkConfig): WalletName[] {
```
HenryNguyen5: ditto

common/utils/network.ts
dternyak: Typescript still didn't like this.

PR #850, opened 2018-01-17
Improve Gas Price UX (Part 2)

james-prado:
Closes https://github.com/MyEtherWallet/MyEtherWallet/issues/571

- [x] Replace Contract Interact call function gas limit with a readonly gas slider component
- [x] Replace Contract Deploy gas limit with a readonly gas slider component
- [x] Remove old gas slider from header
- [x] Put gas $ estimate in the confirmation modal
	- [works on across non-ethereum networks too](https://imgur.com/a/PrLCY)
	- (tested on ETC, transaction fee displays properly)

Also:
- Rename `<GasSlider /`> to `<Gas />`
- Add option to check if valid amount field on `<GenerateTransaction />` button
	-  this is useful for deploying contracts, where the value may or may not be specified
- Add `customLabel` prop to `<GasLimitField />` 
- Add option to `isValidAmount()` selector to determine if transaction field value should be checked
- Add new `getTransactionFee()` selector 
- Add new `options` prop for `<AdvancedGas />` & set default option values
- Set nonce when `<Gas />` mounts
- Reset transaction fields when `<Gas />` mounts
	- prevents stale data in fields when switching tabs (happens most often with subtabs)
	- example: contract interact & deploy tabs on alpha won't reset the data field when you fill them then swap to the other tab

New `<Gas />` component let you:
- set initial state ('simple' or 'advanced')
- disable toggle for basic and advanced settings
- specify options for `<AdvancedGas />`
	- currently only which input fields to display

TODO:
- clean up a few small css issues


dternyak:
A few CSS issues with the Gas Limit field:

screen shot 2018-01-17 at 11 07 33 am

screen shot 2018-01-17 at 11 07 16 am

common/containers/Tabs/Contracts/components/Interact/components/InteractExplorer/components/Fields.tsx
```diff
@@ -12,9 +11,8 @@ export class Fields extends Component<OwnProps> {
   public render() {
     const makeContent = () => (
       <React.Fragment>
-        <GasLimitField />
+        <GasSlider />
```
wbobeirne: We're going to need a `hideData={true}` prop, since on contract interact, the data is determined from the rest of the form. We don't want a user entering data here.

common/containers/Tabs/Contracts/components/Deploy.tsx
```diff
@@ -46,27 +45,12 @@ class DeployClass extends Component<DispatchProps> {
...
-                })}
-              />
-            )}
-          />
-        </label>
         <div className="row form-group">
-          <div className="col-xs-11">
-            <NonceField alwaysDisplay={false} />
+          <div className="col-xs-12 clearfix">
+            <GasSlider />
```
wbobeirne: Same as interact, should have a `hideData={true}` prop since the textarea will be the data.

james-prado:
Since deploy.tsx doesn't set a value field - `state.transaction.fields.value`, The generate transaction button is currently stuck as disabled. 

The issue seems to be coming from [these few lines](https://github.com/MyEtherWallet/MyEtherWallet/blob/develop/common/selectors/transaction/current.ts#L48-L51)

If someone with more knowledge about the `<Deploy />` component could provide some feedback on whether deploy ever should be updating the value field, that'd be super helpful

common/components/ConfirmationModal/components/Details/AmountAndGasPrice/components/TransactionFee.tsx
```diff
@@ -0,0 +1,61 @@
...
+class TransactionFeeClass extends React.Component<Props> {
+  public render() {
+    const { rates, network, isOffline } = this.props;
+    return (
+      <SerializedTransaction
+        withSerializedTransaction={serializedTransaction => {
+          const transactionInstance = makeTransaction(serializedTransaction);
+          const { gasPrice, gasLimit } = getTransactionFields(transactionInstance);
+          const fee = Wei(gasPrice).mul(Wei(gasLimit));
+          const usdBN = network.isTestnet
```
HenryNguyen5: This should be done in a selector instead

common/containers/Tabs/SendTransaction/index.tsx
```diff
@@ -17,6 +17,7 @@ import SubTabs, { Tab } from 'components/SubTabs';
 import { getNetworkConfig } from 'selectors/config';
 import { isNetworkUnit } from 'utils/network';
 import { RouteNotFound } from 'components/RouteNotFound';
+import { getNonceRequested, TGetNonceRequested } from 'actions/transaction';
```
HenryNguyen5: A much better place to put this would be in the `GasSlider` `componentDidMount` component lifecycle method if we're going to be using that for anything gas price / gas limit /nonce/ related

common/components/GasSlider/GasSlider.tsx
```diff
@@ -49,18 +51,26 @@ class GasSlider extends React.Component<Props, State> {
...
   public render() {
-    const { offline, disableAdvanced, gasPrice } = this.props;
+    const { offline, disableAdvanced, gasPrice, readOnly } = this.props;
     const showAdvanced = (this.state.showAdvanced || offline) && !disableAdvanced;
 
     return (
       <div className="GasSlider">
-        {showAdvanced ? (
+        {readOnly ? (
+          <GasLimitField
```
HenryNguyen5: GasSlider should be re-named something else, and have `slider` as a boolean prop instead as having it display an input field sometimes based on the `readOnly` prop breaks semantics

common/components/GenerateTransaction/GenerateTransaction.tsx
```diff
@@ -32,13 +33,14 @@ class GenerateTransactionClass extends Component<StateProps> {
...
-      validGasLimit
+      validGasLimit,
+      checkValidAmount
     } = this.props;
 
     const isButtonDisabled =
       !isFullTransaction ||
       networkRequestPending ||
-      !validAmount ||
+      (checkValidAmount && !validAmount) ||
```
HenryNguyen5: `checkValidAmount` should be a parameter to the `isValidAmount` selector instead, allowing it to be used in more scenarios and easier to test for

HenryNguyen5:
2018-01-22-173948_1618x565_scrot

`Replace Contract Deploy gas limit with a readonly gas slider component`

It's showing as an input field to me instead

common/containers/Tabs/SendTransaction/index.tsx
dternyak: Great idea, gasSlider can handle getting its own nonce.

We may need to make updates to the reducer not to clear nonce since it's no longer really part of the transaction state in this case.

PR #857, opened 2018-01-18
Fix Offline Send

### LINKED ISSUE #838: Offline send no longer works
Amount field was marked invalid, presumably due to some validation that the input was lower than balance.

- [ ] Don't include balance related validation offline for amount field

Steps to reproduce:

1. Turn off wifi / ensure no internet connectivity.
2. Sign in (I used mnemonic)
3. Fill out fields 
4. Notice that transaction generation is disabled, and amount field is red.

skubakdj:
Makes validation of `Amount` field offline-aware. Closes #838.

Also addresses a different bug I came across. When in the collapsed / vertical arrangement, the advanced `Gas Price`, `Gas Limit`, and `Nonce` fields are unclickable. ~~This adjusts their `z-index` values to bring them to the front.~~ This applies the `col-xs-12` bootstrap class to size the elements correctly.


common/components/GasSlider/components/AdvancedGas.tsx
```diff
@@ -52,7 +52,7 @@ class AdvancedGas extends React.Component<Props> {
           />
         </div>
 
-        <div className="col-md-4 col-sm-6 col-xs-12 AdvancedGas-gasLimit">
+        <div className="col-md-4 col-sm-6 col-xs-12 AdvancedGas-boostZIndex AdvancedGas-gasLimit">
```
wbobeirne: I actually realized there's a much simpler fix to this, which is to add `col-xs-12` to the columns that have `col-sm-12` and `col-md-12`. I guess Bootstrap wants you to specify the smallest column size. I'd prefer this over messing with z-index, that always comes back to bite you.

skubakdj: Great suggestion, that's definitely the correct approach! Just pushed the change.

PR #858, opened 2018-01-18
Rotate promos

james-prado:
Closes https://github.com/MyEtherWallet/MyEtherWallet/issues/750

- Seperate promos into individual components
	- creating banners is becoming too difficult with a one-size-fits-all component, separating them makes them easier to write and compose
- Rotate promos every 10 seconds
- Add fade animation when transitioning between promos

### DIFF
--- a/common/components/BalanceSidebar/Promos.scss
+++ b/common/components/BalanceSidebar/Promos.scss
@@ -114,6 +114,6 @@
   transition: opacity 300ms;
   pointer-events: none;
   &.carousel-exit-active {
-    opacity: 0.01;
+    opacity: 0;
   }
 }
--- a/common/components/BalanceSidebar/Promos.scss
+++ b/common/components/BalanceSidebar/Promos.scss
@@ -1,15 +1,34 @@
-@import "common/sass/variables";
-@import "common/sass/mixins";
+@import 'common/sass/variables';
+@import 'common/sass/mixins';
 
 .Promos {
+  overflow: hidden;
+
+  &-promo-wrapper {
+    height: 6rem;
+    overflow: hidden;
+  }
+
+  &-Bity {
+    background-color: #006e79;
+  }
+
+  &-Coinbase {
+    background-color: #2b71b1;
+  }
+
+  &-HardwareWallets {
+    background-color: #6e9a3e;
+  }
+
   &-promo {
     position: relative;
-    height: 6rem;
+    height: inherit;
     display: block;
     color: #fff;
     text-decoration: none;
     text-align: center;
-    transition-duration: 200ms;
+    transition: opacity 200ms;
     @include clearfix;
 
     &:hover,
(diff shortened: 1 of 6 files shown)


common/components/BalanceSidebar/Promos.scss
```diff
@@ -88,3 +107,13 @@
...
   }
 }
+
+.carousel-exit {
+  opacity: 1;
+  transform: translate(0%, -100%);
+  transition: opacity 300ms;
+  pointer-events: none;
+  &.carousel-exit-active {
+    opacity: 0.01;
```
wbobeirne: You can just make it zero no problem.

common/components/BalanceSidebar/Promos.scss
```diff
@@ -1,15 +1,34 @@
...
   &-promo {
     position: relative;
-    height: 6rem;
+    height: inherit;
     display: block;
     color: #fff;
     text-decoration: none;
     text-align: center;
-    transition-duration: 200ms;
+    transition: opacity 200ms;
```
wbobeirne: You can just kill this if it's being handled in `.carousel-exit`

common/components/BalanceSidebar/Promos.tsx
```diff
@@ -94,6 +58,15 @@ export default class Promos extends React.Component<{}, State> {
   }
 
   private navigateToPromo = (idx: number) => () => {
+    // stop rotating when user begins interacting with promos
+    clearInterval(this.timer);
```
wbobeirne: This is a nice touch I was going to suggest, glad you beat me to the punch!

common/components/BalanceSidebar/Promos.tsx
```diff
@@ -94,6 +58,15 @@ export default class Promos extends React.Component<{}, State> {
...
     this.setState({ activePromo: Math.max(0, Math.min(promos.length, idx)) });
   };
+
+  private rotate = () => {
+    const previousIndex = this.state.activePromo;
+    const nextIndex = previousIndex + 1;
+    const nextIndexExists = typeof promos[nextIndex] !== 'undefined' ? true : false;
```
wbobeirne: You can make this code _much_ simpler with modular math:
```
const activePromo = (this.state.activePromo + 1) % promos.length;
this.setState({ activePromo });
```
The % sign is a modulo operator, which causes numbers to wrap.

common/components/BalanceSidebar/Promos.scss
james-prado: the transition here is for `:hover` effect

common/components/BalanceSidebar/Promos.tsx
james-prado: Thx 😄

common/components/BalanceSidebar/Promos.tsx
james-prado: Good use of modulo 👍 , thx for the suggestion

PR #859, opened 2018-01-18
Routing Improvements

james-prado:
Closes https://github.com/MyEtherWallet/MyEtherWallet/issues/652

- Remove any `<Route />` components nested inside a `<Route />`
- Update routing in 'Generate Wallet', 'Account View' & 'Send, Contracts', and 'Sign and Verify Message'
- Add `<CaptureRouteNotFound />` & `<RouteNotFound />` component to provide proper 404 page support
- Update `<SubTabs />` component
- Handle disabled subroutes and paths with `<Redirect />`'s


common/containers/Tabs/GenerateWallet/GenerateWallet.tsx
```diff
@@ -13,25 +14,23 @@ export enum WalletType {
...
+        <TabSection>
+          <section className="Tab-content">
+            {window.crypto ? (
+              <Switch>
+                <Route exact={true} path={currentPath} component={WalletTypes} />
+                <Route path={`${currentPath}/keystore`} component={Keystore} />
+                <Route path={`${currentPath}/Mnemonic`} component={Mnemonic} />
+              </Switch>
+            ) : (
+              <Route component={CryptoWarning} />
```
wbobeirne: Does this need to be a route, or can it just be `<CryptoWarning />`?

common/Root.tsx
```diff
@@ -50,16 +50,10 @@ export default class Root extends Component<Props, State> {
     const routes = (
       <Switch>
         <Route exact={true} path="/" component={GenerateWallet} />
-        <Route path="/generate" component={GenerateWallet}>
-          <Route path="keystore" component={GenerateWallet} />
-          <Route path="mnemonic" component={GenerateWallet} />
-        </Route>
+        <Route path="/generate" component={GenerateWallet} />
```
wbobeirne: Upon testing, I've found that what this means is that this matches _every_ `/generate` route. I think what would be preferrable is to make all of these routes `exact={true}`, and to specify all routes in here:

```
<Route exact={true} path="/generate" component={GenerateWallet} />
<Route exact={true} path="/generate/keystore" component={GenerateWallet} />
<Route exact={true} path="/generate/mnemonic" component={GenerateWallet} />
```

Right now if you go to `/generate/likewhateverman`, you don't get the 404 page, you just get a contentless template page.

common/components/SubTabs/index.tsx
```diff
@@ -1,63 +0,0 @@
...
-        </div>
-
-        <main className={`SubTabs-content col-sm-${columnSize}`} key={tab.path}>
-          {tab.render(this.props.subTabProps || {})}
-        </main>
-        {this.props.sideBar ? this.props.sideBar : null}
-      </div>
-    );
-  }
-}
```
wbobeirne: I don't like the idea of removing the subtabs component, I think we were considering transitioning signing and contract pages to it instead of the oversized header things.

dternyak:
Agreed @wbobeirne would be good to get these changes in first.

common/Root.tsx
wbobeirne: Just another point for being explicit about routes, it would allow us to do something like this: https://www.npmjs.com/package/react-static-webpack-plugin and get rid of HashRouter, by having actual HTML files for each route. Might also simplify our gh-pages configuration as well so that non-index pages aren't 404ing!

common/components/SubTabs/index.tsx
dternyak: Yeah, we definitely want to keep this. I'm going to rely on it for disabling features that are only valid on specific networks as well.

james-prado: I don't think the subtabs component is necessary. Everything that it can do can be done much cleaner by using [NavLink](https://reacttraining.com/react-router/web/api/NavLink)

example using [NavLink](https://reacttraining.com/react-router/web/api/NavLink)
```
const tabs = [
  {
    path: 'valid',
    name: 'foo'
  },
  {
    path: 'invalid',
    name: 'bar'
  }
];

const validTabs = ['foo']

<div className="SubTabs row">
	<div className="SubTabs-tabs col-sm-8">
		{tabs.map((t, i) => (
			<NavLink
				className={`SubTabs-tabs-link ${!validTabs.includes(t.name) ? 'is-disabled' : ''}`}
				activeClassName="is-active"
				to={currentPath + '/' + t.path}
				key={i}
			>
				{t.name}
			</NavLink>
		))}
	</div>
</div>
```

common/components/SubTabs/index.tsx
```diff
@@ -1,62 +1,38 @@
...
-import { Link } from 'react-router-dom';
+import { NavLink, RouteComponentProps } from 'react-router-dom';
 import './SubTabs.scss';
 
-export interface Tab<SubTabProps = null> {
+interface Tabs {
+  name: string | React.ReactElement<any>;
   path: string;
-  name: TranslateType;
-  isDisabled?(props: SubTabProps | {}): boolean;
```
dternyak: Just a heads up, we do still need this prop.

dternyak: Ah, I see you've added `disabled`. You'll have to dig into each usage of the subtab and pass the computed value instead of the function.

Not sure this change is useful / necessary in the scope of this PR.

common/Root.tsx
james-prado: I've added `<RouteWithSubRoutes />` component which should now take care of 404 pages  without having to use `exact={true}` on every path.

Also I don't think we're using any dynamic routes, so something like [react-static-webpack-plugin](npmjs.com/package/react-static-webpack-plugin) should still work

common/components/SubTabs/index.tsx
james-prado: The only subtab using `disabled` is send. I've updated the pr to handle disabled subroutes & tabs

dternyak: Whichever get's merged first can handle the conflict, it's not a big change.

common/Root.tsx
```diff
@@ -46,22 +46,43 @@ export default class Root extends Component<Props, State> {
...
+            route => subRoutes.includes(route) || route === path
+          );
+          return validRoutes.includes(props.location.pathname) ? (
+            <BaseComponent {...props} />
+          ) : (
+            <PageNotFound {...props} />
+          );
+        }}
+      />
+    );
```
wbobeirne: While tedious, I'd rather we use the `exact={true}` prop because all of the other routes (Not just these sub-tabbed ones) have the same issue of rendering anything at `route/*`, and this doesn't fix them (e.g. /ens/blah).

james-prado:
@wbobeirne I had a few issues using exclusively `exact={true}`, but I think I've found the correct solution now. Paths are set to `exact={true}` if they have no subRoutes, which is the case for most routes in mew. If they do have subRoutes, then there is no need to add `exact={true}`. I've also updated the components for handling 404 pages. Together these changes allow only the routes we specify, all other routes will deliver a 404 page

PR #876, opened 2018-01-22
Sweeper Component

oojr:
work in progress, For the next parts, I am thinking of creating a new redux saga` function* estimateBatchGasCost`, I also need to create a new redux key to create a list of transactions

### DIFF
--- a/common/containers/Tabs/SendTransaction/components/SweepRow.tsx
+++ b/common/containers/Tabs/SendTransaction/components/SweepRow.tsx
@@ -1,0 +1,89 @@
+import removeIcon from 'assets/images/icon-remove.svg';
+import React from 'react';
+import { TokenValue } from 'libs/units';
+import { UnitDisplay } from 'components/ui';
+
+type ToggleTrackedFn = (symbol: string) => void;
+
+interface Props {
+  balance: TokenValue;
+  symbol: string;
+  custom?: boolean;
+  decimal: number;
+  tracked: boolean;
+  broadcasted: boolean;
+  toggleTracked: ToggleTrackedFn | false;
+  onRemove(symbol: string): void;
+}
+interface State {
+  showLongBalance: boolean;
+}
+
+export default class TokenRow extends React.Component<Props, State> {
+  public state = {
+    showLongBalance: false
+  };
+
+  public render() {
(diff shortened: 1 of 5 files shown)


dternyak:
Hey @oojr,

I hope things are going well on your side.

We didn't make this as clear as we should have initially, but ideally this work would get done in the next 1-2 weeks.

Just wanted to make sure you're still up for that sort of deadline :)

oojr:
@dternyak I'm up for it, I want to try to have most of the functionality done sooner than a week and have it reviewed

oojr:
@dternyak @HenryNguyen5 

Before scanning for tokens

screen shot 2018-01-29 at 7 46 18 am

The screen after scanning for tokens

screen shot 2018-01-29 at 7 48 14 am

oojr:
@dternyak @HenryNguyen5 the balances component inside the sweeper page is not going to work like that, I feel I need to keep a different list of tokens with different state, similar to this.state.trackedTokens in `Balances.tsx`, I am thinking of adding another component like Balances but it is a a table/checkbox list so the user can know which token has been broadcasted and if that token is mined into a block

dternyak:
@oojr Sure, feel free to create new components as necessary. 

We're getting close to the 1 week mark since we last discussed - are you still planning on having a reviewable PR in the next few days?

oojr:
@dternyak I have a good idea on what I want to do, i'll have something reviewable today or tomorrow if I don't run into any roadblocks

oojr:
@dternyak update on this, I am thinking about how to architect a saga at the moment, looping through with a for..of and adding doing a yield put for each token that needs to be sent with setUnitMeta action, trigger the estimateGas saga, and save the batch with estimates in a new store is my plan

oojr:
@HenryNguyen5 @dternyak  I think I might have to pass this over to someone with more expertise, I need to gather a better understanding of how sending a transaction works, 

the flow I was targeting for was estimating transactions one by one and keeping a list of token balances with information if the transaction was broadcasted or not

wbobeirne:
That's fair @oojr, this was by no means a small task. Thanks for taking a swing at it. I'm going to close this PR out, but the code will be around for whoever picks it up next, and they can decide if they want to build off of what you got started with.

PR #886, opened 2018-01-22
Minimize rerender from gasprice

HenryNguyen5:
This prevents the entire tab from re-rendering on gas price slider changes via removing passed down state props and instead directly connecting the components using the prop.


common/components/Header/components/GasPriceDropdown.tsx
```diff
@@ -5,25 +5,32 @@ import DropdownShell from 'components/ui/DropdownShell';
...
   }
 
   public render() {
-    const { value } = this.props;
     return (
       <DropdownShell
         color="white"
         size="smr"
-        ariaLabel={`adjust gas price. current price is ${value} gwei`}
+        ariaLabel={`adjust gas price. current price is ${this.props.gasPrice.raw} gwei`}
```
dternyak: can we keep the de-structure?

HenryNguyen5: I can revert, but this prevents creating another variable that's only going to be used once anyway.

dternyak: It get's used quite a bit in that component afaict.

dternyak: Nevermind, across multiple methods. No real benefit.

PR #899, opened 2018-01-23
Performance Improvements (Pure components, debounced gas slider)

### LINKED ISSUE #885: Make complex non-connected pure components extend React.PureComponent
Components that do a lot in render, but are purely determined based off of state and props should extend React.PureComponent for wee lil' performance boost.

wbobeirne:
Closes #885. I started noticing some rough performance with the recent gas slider component. The issue is that a ton of components were updating very rapidly, even if they had nothing to do with gas price. My solution was twofold:

1. Identify non-connected pure components that don't implement their own `shouldComponentUpdate`, and have them extend `React.PureComponent` instead of `React.Component`
2. Debounce the gas slider so that it only submits an action with its value after you stop sliding it for 200ms

The results are pretty good for general performance thanks to PureComponent, but especially good for swingin' around the gas slider like you just don't care:

### Before (Dev Build)

screen shot 2018-01-23 at 2 22 36 pm

### After (Dev Build)

screen shot 2018-01-23 at 2 19 39 pm

<sub>Performance across the board goes _way_ up with the production build, so don't worry too much about sub-60fps</sub>


common/components/GasSlider/GasSlider.tsx
```diff
@@ -79,6 +89,15 @@ class GasSlider extends React.Component<Props, State> {
   private toggleAdvanced = () => {
     this.setState({ showAdvanced: !this.state.showAdvanced });
   };
+
+  private handleGasPriceInput = (raw: string) => {
+    const gasBn = new BN(raw);
+    const value = gasBn.mul(new BN(Units.gwei));
+    this.setState({
+      gasPrice: { raw, value }
+    });
```
wbobeirne: @HenryNguyen5 is this OK for fudging the component for 200ms? Or did I miss any other factors that go into gasPrice throughout the sagas?

HenryNguyen5: The only race condition i can think of is if the user changes gas price and then generates the transaction within 200ms. This would cause the transaction to be generated with the old gas price. If you want to prevent this situation from ever happening I recommend debouncing it within a saga with the `throttle` effect. This would also require you to make an action creator / action type for a "pending" gas price input, which would disable the `GenerateTransaction` component, similar to how the `isNetworkRequestPending` selector works.

dternyak: A practical solution could be bringing down the throttle to 25/50ms, as in practice I don't see users being able to generate a TX quicker than that.

The solution @HenryNguyen5 mentioned does seem to be robust, but with the tradeoff of additional engineering time/complexity.

wbobeirne: I like Henry's idea, I'll give that a shake. You can hold off on merging until I implement that.

wbobeirne: @HenryNguyen5 this is done, though I'll admit it ended up being a lot more code than I wanted. I ended up housing the redux state under `network` because 1. adding a field to `state.transaction.fields` that wasn't an object broke some stuff, and 2. it's more inline with tracking network requests than it is tracking field values. If you have suggestions, please let me know.

common/selectors/transaction/helpers.ts
```diff
@@ -23,7 +23,7 @@ export const isFullTx = (
   const partialParamsToCheck = { ...rest };
   const validPartialParams = Object.values(partialParamsToCheck).reduce<boolean>(
     (isValid, v: AppState['transaction']['fields'] & ICurrentTo & ICurrentValue) =>
-      isValid && !!v.value,
+      console.log(v) || (isValid && !!v.value),
```
HenryNguyen5: got a console log here

common/sagas/transaction/fields/fields.ts
```diff
@@ -47,5 +56,6 @@ export const fields = [
   takeEvery(TypeKeys.DATA_FIELD_INPUT, handleDataInput),
   takeEvery(TypeKeys.GAS_LIMIT_INPUT, handleGasLimitInput),
   takeEvery(TypeKeys.GAS_PRICE_INPUT, handleGasPriceInput),
-  takeEvery(TypeKeys.NONCE_INPUT, handleNonceInput)
+  takeEvery(TypeKeys.NONCE_INPUT, handleNonceInput),
+  takeLatest(TypeKeys.GAS_PRICE_INPUT_INTENT, handleGasPriceInputIntent)
```
HenryNguyen5: Instead of `takeLatest` + `call(delay ...` you should use `throttle`. Using `takeLatest` + `delay` wont actually reduce the amount of actions being dispatched, it'll just delay all of the actions by 300 ms. `throttle` implements a sliding buffer + delay with an action channel, the result being that every action that is dispatched while the throttled saga is paused is dropped, except for the latest one (which is exactly what you want)

HenryNguyen5: Totally my bad here, i was reading the code as `takeEvery` even though I mentioned `takeLatest`. This is fine

wbobeirne:
@dternyak conflicts resolved, sorry for the delay

HenryNguyen5:
@wbobeirne  looks like there's still conflicting files here

PR #916, opened 2018-01-24
Fix misc send issues

### LINKED ISSUE #915: Token balance validation off by one

HenryNguyen5:
Fixes 
#915 
#914 
#745

### DIFF
--- a/common/components/AmountFieldFactory/AmountInputFactory.tsx
+++ b/common/components/AmountFieldFactory/AmountInputFactory.tsx
@@ -1,6 +1,6 @@
 import React, { Component } from 'react';
 import { Query } from 'components/renderCbs';
-import { ICurrentValue, getCurrentValue, isValidAmount } from 'selectors/transaction';
+import { ICurrentValue, getCurrentValue, nonStandardTransaction } from 'selectors/transaction';
 import { AppState } from 'reducers';
 import { connect } from 'react-redux';
 import { CallbackProps } from 'components/AmountFieldFactory';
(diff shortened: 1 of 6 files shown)


webpack_config/makeConfig.js
```diff
@@ -295,11 +296,11 @@ module.exports = function(opts = {}) {
     filename: options.isProduction ? '[name].[chunkhash:8].js' : '[name].js',
     publicPath: isDownloadable && options.isProduction ? './' : '/',
     crossOriginLoading: 'anonymous'
-  }
-
+  };
 
   // The final bundle
   return {
+    devtool,
```
wbobeirne: For anyone else reviewing: This is the only non-formatting change.

Can't believe I missed this D:

HenryNguyen5: ah yeah, looks like we dont have prettier setup for things outside of `common`

common/components/AmountFieldFactory/AmountInputFactory.tsx
```diff
@@ -11,23 +11,23 @@ interface OwnProps {
...
+    const { currentValue, onChange, withProps, isNonStandard } = this.props;
 
     return (
       <Query
         params={['readOnly']}
         withQuery={({ readOnly }) =>
           withProps({
             currentValue,
-            isValid: validAmount,
+            isValid: !!currentValue.value || isNonStandard,
```
wbobeirne: It took me a while to figure out that `currentValue.value` is null if it exceeds the balance. This is a little unintuitive IMO. What do we do to get the value field's value in the non-standard transactions?

HenryNguyen5: https://github.com/MyEtherWallet/MyEtherWallet/wiki/Transaction-Validity-Spec
^ For checking for non-standard transactions

HenryNguyen5: And yeah, i hear ya on the null value, https://github.com/MyEtherWallet/MyEtherWallet/issues/901
The initial thought process behind it was to make it impossible to send a transaction if a field was invalid ( as the transaction would have null fields and then be rejected by a node). But this lead to the inability to set custom error messages.

PR #924, opened 2018-01-25
Disable more wallets conditionally + explain why

wbobeirne:
This adds conditional disabling of wallets for the HTML downloadable build, and the Electron app build, and disables read-only on lite send.

This also reworks disabling so that we explain to the user why they can't unlock. The UX was getting pretty bad when you'd go to a page like, say, sign message on Electron, and 4 wallets would be disabled with no reasoning. So now you get a tooltip on hover.

### Electron App

screen shot 2018-01-25 at 4 31 14 pm

### HTML Downloadable

screen shot 2018-01-25 at 4 32 54 pm

### Electron + Sign Message

screen shot 2018-01-25 at 4 31 23 pm


dternyak:
👍 Great touch on including tooltips to explain to users why formats are disabled.

Issue #851, opened 2018-01-17
Update Tx Confirmation Modal


james-prado:
I think we've squeezed out every bit of use from the old design. Unordered lists can only do so much and the modal has a lot of technical details which I don't think are suited for sticking in lists. 

It may not be something that makes it in before audits, but It's a part of the site thats frequently used, and seems to be in need of an update

I played around with some idea's in https://github.com/MyEtherWallet/MyEtherWallet/pull/850 but reverted and pushed them to a separate branch to push only what's required in this issue. 

Currently
- Clarity is an issue, everything is just thrown in a list
- Technical details are mixed with important information
	- Nonce and transaction data should be displayed separately, most important information should come first 
- Instead of just not showing tx data when it isn't there, another `</li>` is added telling you there is no data
- Copy is verbose

Proposal
- Prioritize important information, like 'from', 'to', and 'amount / value being sent'
- Separate technical details
	- add a 'details' area
- Display usd conversions along side amounts & transaction fees
- Update the copy

PR #928, opened 2018-01-26
Confirm TX Modal Upgrades

james-prado:
Closes https://github.com/MyCryptoHQ/MyCrypto/issues/851

- [x] Prioritize important information, like 'from', 'to', and 'amount / value being sent'
- [x] Separate technical details from important information
- [x] Display usd conversions along side amounts & transaction fees
- [x] Update the copy

also: 

- Add details section
- Display which contract tokens are sent through
- Update `<Identicon />` styles
- Animate Modal transition in & out 
	- Right now it's just a simple fade, maybe it should scale instead?
- Add new font [Roboto Mono](https://fonts.google.com/specimen/Roboto+Mono)
	- Provides more font-weights over our current monospaced fonts
	- Pairs really well with [Lato](https://fonts.google.com/specimen/Lato) (our primary font
- Replace `.less` fonts file with `.sass`
- Update generic `<Modal />` styles
- Round all amounts to 6 decimal places
	- the raw value (decimals and all) is displayed in details
- Add white fade on text overflow (when scrollbars appear)
screen shot 2018-02-03 at 12 20 20 am


common/components/ConfirmationModal/ConfirmationModal.tsx
```diff
@@ -89,30 +89,25 @@ class ConfirmationModalClass extends React.Component<Props, State> {
     ];
 
     return (
-      <div className="ConfModalWrap">
+      <React.Fragment>
```
wbobeirne: Do you need the fragment at all? Can probably just return the Modal.

common/components/ConfirmationModal/components/Body/Body.scss
```diff
@@ -0,0 +1,104 @@
...
+.Body {
+  margin: auto;
+  max-width: 35rem;
+  width: 100%;
+  @media screen and (min-width: $screen-sm) {
+    min-width: 32rem;
+  }
+}
+
+.Testnet-warn {
```
wbobeirne: These are styles for separate components, and should be in their own respective .scss files.

common/components/ConfirmationModal/components/Body/Body.scss
```diff
@@ -0,0 +1,104 @@
...
+  text-align: center;
+  margin: 0;
+}
+
+.Address {
```
wbobeirne: I'd be very careful about class name collisions here. Maybe be a little more specific, like `.ConfAddress` or something like that.

common/components/ConfirmationModal/components/Body/Body.scss
```diff
@@ -0,0 +1,104 @@
...
+    > img {
+      height: 3rem;
+      width: 3rem;
+      transform: scale(0.75);
+    }
+  }
+}
+
+.Amount,
+.Address {
```
wbobeirne: Same note as above, be careful with class collisions.

common/components/ConfirmationModal/components/Body/components/Amounts.tsx
```diff
@@ -0,0 +1,107 @@
...
+        <div className="Amount-send-positioning-wrapper">
+          <h5>You'll Send: </h5>
+          <h5>
+            <UnitDisplay
+              decimal={decimal}
+              value={sendValue}
+              symbol={isToken ? unit : networkUnit}
+              checkOffline={false}
+            />
+            {!isTestnet && (
```
wbobeirne: We won't always have rates for all currencies, not just testnets.

common/components/ConfirmationModal/components/Body/components/Amounts.tsx
```diff
@@ -0,0 +1,107 @@
...
+              <span style={{ marginLeft: '8px' }} className="small">
```
wbobeirne: Move styles to CSS

common/components/ConfirmationModal/components/Body/components/Amounts.tsx
```diff
@@ -0,0 +1,107 @@
...
+                  displayShortBalance={2}
+                  displayTrailingZeroes={true}
+                  checkOffline={true}
+                />
+              </span>
+            )}
+          </h5>
+        </div>
+      </div>
+      <div className="Amount-total">
```
wbobeirne: Send, fee, and total all look very similar from a markup perspective. Maybe convert this to a loop or function that returns the markup?

common/components/ConfirmationModal/components/Body/components/Amounts.tsx
wbobeirne: Also if we're offline, we may not have rates either.

common/components/ui/Identicon.tsx
```diff
@@ -12,31 +13,12 @@ export default function Identicon(props: Props) {
...
-          top: 0,
-          bottom: 0,
-          borderRadius: '50%',
-
-          boxShadow: `
-                    inset rgba(255, 255, 255, 0.5) 0 2px 2px,
-                    inset rgba(0, 0, 0, 0.6) 0 -1px 8px
-                  `
-        }}
-      />
```
wbobeirne: This by all accounts should be a good change, but (not your fault for missing this, it's a hack) inline styling was because this component was being used for the printable paper wallet, since the window that pops up for it doesn't use CSS. Honestly I want to convert that thing to use _no components_ and be entirely self-contained because of how goddamn annoying it is as a use-case. I'll probably go through and make a PR for that at another point.

But suffice it to say, these need to stay linine for a bit.

james-prado:
Thanks for the great review @wbobeirne!

Following up on your comments:
- Displaying the amount field twice, once in the visual summary, and again just below that is a bit overkill, I'll address this
- More emphasis should be given to the amount "You'll Send" to make it more obvious that's the amount the destination address will receive. Otherwise, I think the copy indicating exactly what "You'll send" and what the "Transaction Fee" is makes it pretty clear
- Displaying the contract address users are sending tokens through is a great idea, I'll make sure to implement this

common/translations/lang/en.json
```diff
@@ -3,7 +3,8 @@
   "data": {
     "x_CancelReplaceTx": "Cancel or Replace Transaction",
     "x_CancelTx": "Cancel Transaction",
-    "x_PasswordDesc": "This password *encrypts* your private key. This does not act as a seed to generate your keys. **You will need this password + your private key to unlock your wallet.**",
+    "x_PasswordDesc":
```
dternyak: Significant diff noise here. For now, let's use raw text instead of creating new keys in translation files (if there are any new keys)

wbobeirne:
New updates look great, just had one issue which is modals that overflow content have the buttons ride right up next to the content. It would be good to add some padding on top of the button, even better to add a border or box-shadow to the content when it's scrollable to indicate that to the user (Important on systems that don't do scrollbars, like OSX default or phones.)

screen shot 2018-02-01 at 8 21 59 pm

The transaction modal could also use a little love on mobile. Ideally it would keep the same layout (icon on left, from / to on right) but the address would wrap lines.

screen shot 2018-02-01 at 8 33 53 pm

common/components/ConfirmationModalTemplate/components/Body/components/Amounts.scss
```diff
@@ -0,0 +1,47 @@
...
+
+  &-total {
+    &-positioning-wrapper {
+      padding: 1rem 0rem;
+      padding-bottom: 0;
+      background-image: linear-gradient(to right, #c2cfd6 33%, #fff0 0%);
+      background-position: top;
+      background-size: 5px 1px;
+      background-repeat: repeat-x;
+      > h5:nth-child(0n + 2) {
```
wbobeirne: Again, kind of a spooky brittle selector

common/components/ConfirmationModalTemplate/components/Body/components/Details.scss
```diff
@@ -0,0 +1,34 @@
...
+    &:hover {
+      background-color: rgba(153, 153, 153, 0.12);
+    }
+  }
+  &-pair {
+    margin-bottom: 1rem;
+    > h5 {
+      display: inline-block;
+      margin: 0;
+      &:nth-child(0n + 2) {
```
wbobeirne: That same selector again

common/components/SendButtonFactory/OnlineSend.tsx
```diff
@@ -24,7 +24,7 @@ interface DispatchProps {
 }
 
 interface OwnProps {
-  Modal: typeof ConfirmationModal;
+  Modal: typeof ConfirmationModalTemplate;
```
wbobeirne: Is this prop even being used anymore?

common/sass/variables/tables.scss
```diff
@@ -1,8 +1,12 @@
...
-$table-bg-hover: $gray-lightest;
+$table-bg-accent: rgba(0, 0, 0, 0.03);
+$table-bg-hover: rgba(0, 0, 0, 0.06);
 $table-bg-active: $table-bg-hover;
 $table-border-color: transparent;
 $table-cell-padding: 0.75rem 1rem;
+
+tr {
+  transition: background-color 120ms;
+}
```
wbobeirne: Style in a variables file will cause this to be repeated every time someone `@include`'s it. This should be in an overrides or something.

Also, while a transition is fairly innocuous, we should try to hide opinionated element selectors behind Bootstrap's classes for styled things: http://getbootstrap.com/docs/3.3/css/#tables.

common/components/ConfirmationModalTemplate/components/Body/components/Addresses.scss
```diff
@@ -0,0 +1,67 @@
+@import 'common/sass/variables';
+
+.tx-modal-address {
+  display: flex;
+  flex-direction: column;
+  width: fit-content;
```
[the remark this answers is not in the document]
james-prado: The support for this isn't perfect. However the support for just `width: fit-content` is pretty good. The only exception is ie & edge. Assuming we aim to also support IE, together they total 4.34%, IE – 2.84%, Edge – 1.5%.

In this case achieving the same alignment without `width: fit-content;` would be difficult and verbose. I'd suggest instead of scrapping `fit-content` we fallback to another with property. And with the cascading effect of css it's super easy
```
width: 100%;
width: fit-content;
```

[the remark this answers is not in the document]
james-prado: nvm, went back and double checked, I don't actually need it.

common/translations/lang/en.json
james-prado: I'll clean up the diff noise. I think changing the keys until copy is modified would be a good thing in this case. I changed the copy for the modal 'confirm' and 'cancel' buttons, they were really long, and were causing issues even medium sized screens. 

I think it'll make a big difference in the mean time while we're in alpha / beta.

common/utils/rawUnitConversion.ts
```diff
@@ -0,0 +1,24 @@
...
+
+interface EthProps extends Args {
+  unit: UnitKey;
+}
+interface TokenProps extends Args {
+  decimal: number;
+}
+
+// export const rawUnitConversion = (args: Args & (EthProps | TokenProps)) => {
+export const rawUnitConversion = (args: Args & (EthProps | TokenProps)) => {
```
HenryNguyen5: This should probably be in `libs/units` instead, and instead of re-writing the logic that the `UnitDisplay` component uses internally, you could pull out the logic within the component and re-use it in the lib.

common/utils/rawUnitConversion.ts
```diff
@@ -0,0 +1,24 @@
...
+  const { value } = args;
+  const isEthereumUnit = !!(args as EthProps).unit;
+  return value
+    ? isEthereumUnit
+      ? fromTokenBase(value, getDecimalFromEtherUnit((args as EthProps).unit))
```
HenryNguyen5: Should use type guards instead of type coercion as seen in `UnitDisplay` L39 (`IsEthereumUnit`)

common/components/ConfirmationModalTemplate/components/Body/Body.tsx
```diff
@@ -0,0 +1,119 @@
...
+                isTestnet={isTestnet}
+                sendValue={sendValue}
+                fee={transactionFee}
+                networkUnit={networkUnit}
+                decimal={decimal}
+                unit={unit}
+                rates={rates}
+              />
+              <button
+                className={`tx-modal-details-button ${showDetails &&
```
HenryNguyen5: If `showDetails` is false, it'll include the className of `false`

common/components/ConfirmationModalTemplate/components/Body/components/Details.tsx
```diff
@@ -0,0 +1,54 @@
...
+import Code from 'components/ui/Code';
+import './Details.scss';
+import BN from 'bn.js';
+import { rawUnitConversion } from 'utils/rawUnitConversion';
+
+interface Props {
+  sendValue: BN;
+  network: string;
+  provider: string;
+  gasPrice: any;
```
HenryNguyen5: Any types here

common/components/ConfirmationModalTemplate/components/Body/components/Amounts.tsx
```diff
@@ -0,0 +1,104 @@
+import React from 'react';
+import { UnitDisplay } from 'components/ui';
+import BN from 'bn.js';
+import './Amounts.scss';
+import { AppState } from 'reducers';
+
+interface Props {
+  sendValue: BN;
```
HenryNguyen5: Should be `Wei`

common/components/ConfirmationModalTemplate/components/Body/components/Amounts.tsx
```diff
@@ -0,0 +1,104 @@
...
+  fee: BN;
```
HenryNguyen5: ditto

common/components/ConfirmationModalTemplate/components/Body/Body.tsx
```diff
@@ -0,0 +1,119 @@
...
+    const { rates, from, to, unit, node, network, decimal } = this.props;
+    const { showDetails } = this.state;
+    const networkUnit = network.unit;
+    return (
+      <SerializedTransaction
+        withSerializedTransaction={serializedTransaction => {
+          const transactionInstance = makeTransaction(serializedTransaction);
+          const rawTx = getTransactionFields(transactionInstance);
+          const { value, gasPrice, gasLimit, data, nonce, chainId } = rawTx;
+          const isToken = !(unit === 'ether' || unit === 'ETH');
```
HenryNguyen5: This will break on non-ether unit networks. Instead it should be checking if the supplied unit is a non-network unit `!network.unit`.  This is also more suited to be a selector rather than in-component

common/components/ConfirmationModalTemplate/components/Body/Body.tsx
```diff
@@ -0,0 +1,119 @@
...
+          const isTestnet = network.isTestnet;
```
HenryNguyen5: Would be better to have this beside LN44 to keep declaration groupings together.

common/components/ConfirmationModalTemplate/components/Body/Body.tsx
```diff
@@ -0,0 +1,119 @@
...
+import { getTransactionFee } from 'libs/transaction/utils/ether';
+import { Wei, TokenValue } from 'libs/units';
+import ERC20 from 'libs/erc20';
+import './Body.scss';
+
+interface State {
+  showDetails: boolean;
+}
+
+interface Props {
```
HenryNguyen5: Typing up props this way makes them fragile to State/Selector changes. Any changes to the selectors / state will break these typings as they dont reference `AppState`

HenryNguyen5: Also, re-naming it as `StateProps` will help readability to tell where the props are coming from.

common/components/ConfirmationModalTemplate/ConfirmationModalTemplate.tsx
```diff
@@ -114,29 +114,23 @@ class ConfirmationModalTemplateClass extends React.Component<Props, State> {
...
+        disableButtons={transactionBroadcasting}
+        isOpen={isOpen}
+      >
+        {transactionBroadcasting ? (
+          <React.Fragment>
+            <Spinner size="x5" />
+          </React.Fragment>
+        ) : (
+          <React.Fragment>
+            <Body />
```
HenryNguyen5: This breaks #940, a custom body should be able to be supplied for different transaction use-cases (ENS, contract interact, contract deploy could all have different displays to the user that make more sense in that context. Such as if the user is interacting with a contract we should be displaying what function they're calling in the confirmation modal. Or if it's ENS, to display ENS specific data in the confirmation modal.)

common/components/ConfirmationModalTemplate/ConfirmationModalTemplate.tsx
```diff
@@ -42,8 +43,7 @@ export interface ConfirmButtonCBProps {
 }
 
 export interface OwnProps {
-  summary: React.ReactElement<any> | null;
-  details: React.ReactElement<any> | null;
+  isOpen?: boolean;
```
HenryNguyen5: Should accept a `body` parameter as per #940, which can either be this current implementation of `Body`, or a custom implementation

common/components/ConfirmationModalTemplate/ConfirmationModalTemplate.tsx
HenryNguyen5: Ideally, the `Body` component should be placed in a folder where `ConfirmationModal` implements `ConfirmationModalTemplate`, see https://github.com/MyCryptoHQ/MyCrypto/pull/940/files#diff-2bac20a9e428f854eb8c93128d97e2ea

common/components/ConfirmationModalTemplate/components/Body/Body.tsx
```diff
@@ -0,0 +1,119 @@
...
+          const isToken = !(unit === 'ether' || unit === 'ETH');
+          const isTestnet = network.isTestnet;
+          const sendValue = isToken
+            ? TokenValue(ERC20.transfer.decodeInput(data)._value)
+            : Wei(value);
+          const transactionFee = getTransactionFee(gasPrice, gasLimit);
+          return (
+            <div className="tx-modal-body">
+              {isTestnet && <p className="tx-modal-testnet-warn small">Testnet Transaction</p>}
+              <Addresses
```
HenryNguyen5: I think it'd be better to have these as connected components with their own selectors for getting their own slice of state. Then custom implementations of `Body` can re-use the components `Addresses` `Amounts` `Details` without them being dependent on each other

common/components/ConfirmationModalTemplate/components/Body/components/Addresses.tsx
```diff
@@ -0,0 +1,55 @@
+import React from 'react';
+import ERC20 from 'libs/erc20';
+import { Identicon } from 'components/ui';
+import arrow from 'assets/images/tail-triangle-down.svg';
+import './Addresses.scss';
+import { ETHAddressExplorer } from 'config';
+
+interface Props {
```
HenryNguyen5: Typing props this way makes them brittle as described in `Body`

common/components/ConfirmationModalTemplate/components/Body/components/Amounts.tsx
```diff
@@ -0,0 +1,104 @@
+import React from 'react';
+import { UnitDisplay } from 'components/ui';
+import BN from 'bn.js';
+import './Amounts.scss';
+import { AppState } from 'reducers';
+
+interface Props {
```
HenryNguyen5: Typing props this way makes them brittle as described in `Body`

common/components/ConfirmationModalTemplate/components/Body/components/Details.tsx
HenryNguyen5: Typing props this way makes them brittle as described in `Body`

common/components/SendButton.tsx
```diff
@@ -1,15 +1,15 @@
 import React from 'react';
 import { SendButtonFactory } from './SendButtonFactory';
 import translate from 'translations';
-import { ConfirmationModal } from 'components/ConfirmationModal';
+import { ConfirmationModalTemplate } from 'components/ConfirmationModalTemplate';
 
 export const SendButton: React.SFC<{
   onlyTransactionParameters?: boolean;
-  customModal?: typeof ConfirmationModal;
+  customModal?: typeof ConfirmationModalTemplate;
```
HenryNguyen5: This should be `typeof ConfirmationModal` where `ConfirmationModal` implements `ConfirmationModalTemplate`

common/components/SendButtonFactory/OnlineSend.tsx
```diff
@@ -24,7 +24,6 @@ interface DispatchProps {
 }
 
 interface OwnProps {
-  Modal: typeof ConfirmationModal;
```
HenryNguyen5: Removing this prop breaks #940 as a custom modal implementation can not be passed down to `OnlineSend` from `SendButtonFactory`

common/components/SendButtonFactory/SendButtonFactory.tsx
```diff
@@ -19,7 +19,7 @@ interface StateProps {
 }
 interface OwnProps {
   onlyTransactionParameters?: boolean;
-  Modal: typeof ConfirmationModal;
+  Modal: typeof ConfirmationModalTemplate;
```
HenryNguyen5: This should be typeof ConfirmationModal where ConfirmationModal implements ConfirmationModalTemplate

HenryNguyen5: And also is not used anymore, breaking #940

common/components/ConfirmationModalTemplate/components/Body/components/Details.tsx
```diff
@@ -0,0 +1,54 @@
...
+  gasLimit: any;
+  nonce: any;
+  data: any;
+  chainId: any;
+  decimal: number;
+  symbol: string;
+}
+
+export const Details: React.SFC<Props> = ({
```
HenryNguyen5: There could be way less props here if a selector was used instead to derive all the correct data

PR #930, opened 2018-01-26
Address messages, warnings and gas limits

### LINKED ISSUE #900: Warn users if they are sending tokens to a token contract address (address in the list in /ethereum-lists)?
Non-blocking, but would be excellent for UX. Same vein as an "Address not checksummed" warning where it displays but doesn't actually prevent user from sending if they truly want to.

We're having more and more users who send their tokens *to* the token contract address and these token teams can't / won't help them recover their funds, which is the real shame here.

### LINKED ISSUE #907: Send Warnings
V3: https://github.com/kvhnuke/etherwallet/blob/mercury/app/scripts/customGas.js

In V3, users are warned when attempting to send to a known malicious address, expired ICO contract address, and other instances where there is a high likelihood they do not really want to send their funds.

- [ ] Port warn functionality from V3.

wbobeirne:
Closes #900, closes #907, and adds a couple of extras.

### What This Does

* Updates the list of hard-coded messages for addresses from v3 (#907)
    * Moves the message list to config/addressMessages.ts to be somewhat less buried in the code
* Warns users if they're sending to a token contract (#900)
* Uses the hard-coded gas limit provided by addressMessages.ts, if there is one
    * cc @HenryNguyen5, was my implementation on this alright?
* Warns users if they were going to send to themselves, but won't stop them

### How to Test

1. Open the site with the `token-contract-warning` branch
2. Attempt to send to `0x75aa7b0d02532f3833b66c7f0ad35376d373ddf8`, confirm you get a message and your gas limit is set to `300000` with no load time.
3. Attempt to send to `0x0d8775f648430679a709e98d2b0cb6250d2887ef`, confirm you get a message
4. Attempt to send to your own address, confirm you get a message

### Screenshots

screen shot 2018-01-26 at 2 55 54 pm

screen shot 2018-01-26 at 2 56 19 pm

screen shot 2018-01-26 at 2 57 48 pm


common/sagas/transaction/network/gas.ts
```diff
@@ -48,6 +50,18 @@ export function* shouldEstimateGas(): SagaIterator {
       continue;
     }
 
+    // If we have a hard-coded limit, use it immediately and skip estimate
+    const msg = action.type === TypeKeys.TO_FIELD_SET && getAddressMessage(action.payload.raw);
```
HenryNguyen5: This way of overriding gas limit could be unreliable as some actions cause a casade of actions to be fired, which would mean `TypeKeys.TO_FIELD_SET` wouldn't be the action type, but the gas limit would still be re-estimated, wiping out the override.

A better way of implementing this is running this in a seperate saga which watches over the `toField` of the state. Then when the field is changed and matches (e.g. getAddressMessage returns a valid value). 

0) Store the previous state of what the auto gas estimation variable was in the saga
1) turn of auto gas estimation
2) grab the messsage gas limit
3) Dispatch the custom gas limit (now it wont be overriden because auto gas is off)

While steps 0-3 are running, if the to field changes by anything:
1) set auto gas estimation back to its original value
2) Run steps 0-3 again

If the state is cleared / transaction is sent (they should be under `reset` action)
1) set auto gas estimation back to its original value

Also, it may be a good idea to warn the user that you're no longer auto estimating, and instead its a predefined value.

wbobeirne: That's a good point about gas limit estimation possibly being kicked off by other actions. However, I'd like to leave gas estimate on and still give the users the visual that gas limit is uneditable & auto-calculated in these instances. To me, the gas estimation checkbox means "figure it out for me", not literally estimate it via the node. If we know better, I think we should leave the field disabled and just set it for them. And if they want to change it themselves, unchecking the checkbox is still an option (Which would override this code path, since the early exit for offline or estimate off happens before it.)

So I think the only change I want to make here is to open it up to all actions, not just TO_FIELD_SET. Let me know if you think there was anything else important that this misses.

HenryNguyen5: That works too, I'd still split this into another saga though and call it from within this one just for maintainability / testability sake. You'll also need to modify how you get the message (probably by using a selector to get the address instead) so that it works on any gas estimation related actions.

wbobeirne: Made a selector for address messages, busted it out into its own function, and added unit tests. Lemme know if there's anything else!

PR #940, opened 2018-01-28
Generic Confirmation Modal

HenryNguyen5:
This allows for a generic confirmation modal to be used instead of the default one. 

Main use-cases that I can see off the top of my head include custom interactions such as:
1. contract deploy
2. contract interaction
3. ENS

The `ens-bidding` PR actually uses this generic modal structure already to provide ENS specific information in the modal, with custom confirmation button support but still having the same broadcasting / time limit functionality.


wbobeirne:
Just a heads up, this'll probably be a heavy conflict with #928. Not sure what the ideal way to reconcile this would be.

PR #941, opened 2018-01-28
[PRE-ENS-2] Refactor Generate Transaction

HenryNguyen5:
Precursors: #940 

This simplifies the transaction signing process from the react component side, and offloads that to sagas instead. It also splits the generate transaction component into a factory and default implementation to allow further customization and more ways to implement the button.


common/components/GenerateTransaction.tsx
```diff
@@ -0,0 +1,13 @@
+import { GenerateTransactionFactory } from './GenerateTransactionFactory';
+import React from 'react';
+import translate from 'translations';
+
+export const GenerateTransaction: React.SFC<{}> = () => (
+  <GenerateTransactionFactory
+    withProps={({ disabled, isWeb3Wallet, onClick }) => (
+      <button disabled={disabled} className="btn btn-info btn-block" onClick={onClick}>
+        {isWeb3Wallet ? translate('Send to MetaMask / Mist') : translate('DEP_signtx')}
```
wbobeirne: I've been meaning to mention, this copy doesn't actually really make sense since clicking this button doesn't do anything with metamask or mist. It's not until the confirmation modal that we call out to web3 to do the transacting. Not necessarily a part of this PR though.

wbobeirne:
Looks like some unit tests are failing now though, cc @HenryNguyen5

PR #953, opened 2018-01-29
Update Readme, bump nvmrc, move some to wiki

wbobeirne:
I've updated the readme to better reflect our current build commands and folder structure. I also called out which version of Node we're on, and updated the nvmrc to reflect that. For the more detailed developer style guides, I've [moved those to a wiki page](https://github.com/MyEtherWallet/MyEtherWallet/wiki/Style-Guides) so that we can update them more frequently, and to keep the readme to the important basics.


wbobeirne:
Shite, I've included on of Henry's branches in here. Closing and reopening with fixed branch.

PR #963, opened 2018-01-30
[DO NOT MERGE] Nodes-Networks-Redux-PT3

HenryNguyen5:
This is the third PR of Nodes-Networks-Redux refactor.
For a more detailed explanation visit #961 

Part 1: Redux state, selectors, action creators
Part 2: Re-typing repo using new static and custom types
Part 3: Refactor components and sagas to use new selectors to derive data instead
Part 4: Testing and finalisation


shared/types/node.d.ts
```diff
@@ -14,13 +19,18 @@ interface CustomNodeConfig {
 }
 
 interface StaticNodeConfig {
+  isCustom: false;
```
HenryNguyen5: This is used for type guards throughout the codebase so TS knows what node config its dealing with.

PR #981, opened 2018-02-02
MEW-01-004 - Stronger Keystores

### LINKED ISSUE #889: MEW-01-004: Keystore Protection allows weak passwords
- [ ] Block users from generating keystore wallets with weak passwords 
- [ ] Increase N Factor to acceptable level (Mist uses ~10000 rounds)
     - Note: We'll need to benchmark this to make sure UX doesn't completely fall off a cliff. 

Notes: 
https://github.com/dropbox/zxcvbn may be a good resource for estimating password complexity.

wbobeirne:
Closes #889, addresses the confirm issue in #828. This PR makes a lot of changes regarding keystore file generation.

## What This Does

* Implements password checking via [zxcvbn](https://github.com/dropbox/zxcvbn) and ups the minimum characters to 12.
    * Also uses that module's handy feedback to inform the user of their weak password.
* Adds a "Confirm password" field to keystore generation.
* Moves keystore generation off to a web worker, and shows a spinner while it's generating.
* Ups the n-factor of keystores to 8192.

## How to Test

Generate a keystore file, make sure everything's 👌 

### The New Form

screen shot 2018-02-01 at 7 48 45 pm

<br/>

### Performance of Generating (6x CPU slowdown)

screen shot 2018-02-01 at 7 47 51 pm

^ Notice that the graph is mostly purple, which is just redraws of the spinner, not Javascript CPU usage.

## Other Notes

At some point, we should rework the generate flow to require you to unlock before getting at private key / paper wallet. That would ensure no one sends to an address that they haven't unlocked yet, and would also mean we don't let the private key exit the web worker. However, this may not be worth working on at all if we remove generation from web and make it Electron-only.


webpack_config/config.js
```diff
@@ -69,7 +69,6 @@ module.exports = {
     'redux-promise-middleware',
     'redux-saga',
     'scryptsy',
-    'store2',
```
dternyak: Thanks, forgot to remove this here after removing from `package.json`

dternyak:
Copy pasting any string into the confirm field and then revealing will show a `stronger-passwords` string.

wbobeirne:
That's actually at a browser level, you can't copy-paste password fields. `stronger-passwords` was just the last thing in your clipboard (Probably from checking this branch out.)

I'm not really sure what the best way to handle this is, other than maybe just disabling togglability on confirm?

wbobeirne:
That might mess up legit use cases, for instance copying from a password manager / generator (i.e. lastpass.) Maybe we should listen for the copy event on the fields and alert them that passwords can't be copied? I'm not even sure if it fires on password fields.

dternyak:
@wbobeirne Good point. I think I'm doing a bit of both premature optimization and bike-shedding, so maybe we should just step back and let user reports come in if it really is a problem.

PR #990, opened 2018-02-05
Forward query arguments during redirect

### LINKED ISSUE #968: Prefilled transaction is broken
Since the change from `/send-transaction` to `/account/send`, we lose the query parameters upon navigation. We'll need to find a way to carry over the data from the query.

aitrean:
Closes #968 

Copy search arguments on router redirects to preserve prefilled transaction queries.


dternyak:
A more generic solution might be valuable so that this isn't implemented only for the `send-transaction` redirect.

I know that custom networks are at least one other instance where we might want pre-filled fields via query params.

common/Root.tsx
```diff
@@ -95,7 +95,7 @@ export default class Root extends Component<Props, State> {
 
 const LegacyRoutes = withRouter(props => {
   const { history } = props;
-  const { pathname, hash } = props.location;
+  const { pathname, hash, search } = props.location;
 
   if (pathname === '/') {
     switch (hash) {
```
wbobeirne: Search carry-over needs to happen for `#send-transaction` and `#offline-transaction` too.

dternyak:
@wbobeirne Mind re-reviewing when you have a chance?

Issue #809, opened 2018-01-13
chrome password autofill removed unlocking keystore


james-prado:
Chrome autofill's the password (indicated by the yellow), and [this onLoad](https://github.com/MyEtherWallet/MyEtherWallet/blob/develop/common/components/WalletDecrypt/components/Keystore.tsx#L112) function seems to be removing it
<img width="444" alt="screen shot 2018-01-13 at 5 59 19 am" src="https://user-images.githubusercontent.com/13836920/34905461-02ffee7e-f827-11e7-8252-da5eaf6761e2.png">

dternyak:
Is this an issue? I would rather not allow autofill to fill-in the password, but perhaps there's a cleaner way of doing so (e.g. `no-autofill` property).

james-prado:
You can't actually turn off [autofill for passwords](https://developer.mozilla.org/en-US/docs/Web/Security/Securing_your_site/Turning_off_form_autocompletion). I think it would be better to just fix the issue instead of replacing the password with a blank string

PR #993, opened 2018-02-05
Fix keystore password autofill

james-prado:
Closes https://github.com/MyCryptoHQ/MyCrypto/issues/809

- Add `attemptEmptyPass` prop to the `PrivateKeyValue` interface
	- set `attemptEmptyPass` to `true` and attempt unlock on file change
- Revert setting password to an empty string 
- Increase spinner size
- Hide inputs & unlock button while unlocking wallet

This solution works, but sometimes there's a small flicker where the spinner shows for a few milliseconds when attempting to unlock with an empty pass. 

It seems to me like the work & the compromises aren't worth it compared to just allowing the user to enter an empty password on their own. ie. leave pass blank and hit 'unlock.


dternyak:
> It seems to me like the work & the compromises aren't worth it compared to just allowing the user to enter an empty password on their own. ie. leave pass blank and hit 'unlock.

It seems to me that you're commenting on two separate issues:

1. Disable browser autofill of passwords
2. Attempt empty password on unlock of keystores

Additionally, Im not aware of why we can't simply use `autocomplete="off"` as mentioned in https://developer.mozilla.org/en-US/docs/Web/Security/Securing_your_site/Turning_off_form_autocompletion.

However, due to a recent internal decision to simply disable keystores/mnemonics/private keys on the web based MyCrypto entirely, https://github.com/MyCryptoHQ/MyCrypto/issues/809 has begun redundant.

I'm going to close this PR for now, but keep the branch around in case we change our minds at a later point.

james-prado:
`autocomplete="off"` [doesn't work on password inputs ](https://developer.mozilla.org/en-US/docs/Web/Security/Securing_your_site/Turning_off_form_autocompletion#The_autocomplete_attribute_and_login_fields)

PR #994, opened 2018-02-05
[WIP] Nodes-Networks-Redux-PT4

HenryNguyen5:
This PR adds tests and fixes any bugs found for the nodes/networks rework


spec/reducers/config/nodes/staticNodes.spec.ts
```diff
@@ -0,0 +1,128 @@
+import { web3SetNode, web3UnsetNode } from 'actions/config';
+import { staticNodes, INITIAL_STATE } from 'reducers/config/nodes/staticNodes';
+import { EtherscanNode, InfuraNode, RPCNode } from 'libs/nodes';
+import { Web3NodeConfig } from 'types/node';
+import { networkIdToName } from 'libs/values';
+import { Web3Service } from 'libs/nodes/web3';
+
+const expectedInitialState = JSON.stringify({
```
dternyak: Hm, don't really like that all this is duplicated. Is there really no way to import an object from `reducer` that is used to create the initial state?

```
const nodes = {..}

INITIAL_STATE =  {
    nodes,
    blah,

}
```

spec/reducers/config/networks/staticNetworks.spec.ts
```diff
@@ -0,0 +1,148 @@
...
+    isCustom: false,
+    color: '#0e97c0',
+    blockExplorer: makeExplorer('https://etherscan.io'),
+    tokenExplorer: {
+      name: ethPlorer,
+      address: ETHTokenExplorer
+    },
+    tokens: require('config/tokens/eth.json'),
+    contracts: require('config/contracts/eth.json'),
+    dPathFormats: {
```
dternyak: Same here, I think this could be imported.

common/libs/values.ts
```diff
@@ -24,7 +24,9 @@ export function sanitizeHex(hex: string) {
   return hex !== '' ? `0x${padLeftEven(hexStr)}` : '';
 }
 
-export function networkIdToName(networkId: string | number): StaticNetworkIds {
+export function networkIdToName(
+  networkId: 1 | 3 | 4 | 42 | '1' | '3' | '4' | '42'
```
dternyak: Are we hardcoding the expected chainIds here? I don't like this for long-term maintainability reasons (multiple places to configure adding the chainId when adding a new network)

spec/reducers/config/nodes/staticNodes.spec.ts
HenryNguyen5: I hard coded the expected initial state here instead of importing the initial state so that the test checks for equality, or else it doesnt do anything.

HenryNguyen5: We could just get rid of the initial state test too if you think thats too much overhead

spec/reducers/config/networks/staticNetworks.spec.ts
HenryNguyen5: Same as above ^

common/libs/values.ts
HenryNguyen5: The expected chainIds are already hard coded in the switch statement, so i just added type checking so that if you enter an invalid chainId you dont have to wait for the switch statement to throw an error before you find out

dternyak: Why are we hardcoding them in the switch statement? We should be able to traverse the network object and find the matching network.

HenryNguyen5: That's legacy code that's been there for awhile. Not sure why its there

dternyak: This type of approach would allow us to configure additional networks in only one place, which I think is valuable.

dternyak: I think we should refactor this now to make sure it won't bite us later.

PR #997, opened 2018-02-05
Add tooltip to gas slider

wbobeirne:
Because so many people use ETHGasStation to look up what gwei to use, we should communicate this in some way so that less advanced users don't have to go to the advanced tab to set the right gwei. I think the least intrusive way is via the builtin tooltip to rc-slider:

gas-slider-tip


common/components/TXMetaDataPanel/components/SimpleGas.tsx
```diff
@@ -9,6 +9,7 @@ import { getGasLimitEstimationTimedOut } from 'selectors/transaction';
 import { connect } from 'react-redux';
 import { GasLimitField } from 'components/GasLimitField';
 import { getIsWeb3Node } from 'selectors/config';
+const SliderWithTooltip = Slider.createSliderWithTooltip(Slider);
```
dternyak: 🔥

PR #1017, opened 2018-02-07
Empty Public Path for Production Builds

skubakdj:
Change makes the production build compatible with GitHub Pages.


dternyak:
We will indeed use AWS for v4 as we do with v4, however we use it in a way
that requires GitHub pages as well.

skubakdj:
Thanks for the feedback @wbobeirne, the change isn't necessary.

Issue #835, opened 2018-01-15
Show loading indicator for nonce retrieval


dternyak:
Similar to gas estimation blocking transaction generation from being possible, nonce retrieval should have a spinner / loading indicator to inform users about what they are waiting for before they can proceed with transaction generation.

Loading indicator UI can be similar to gas estimate loading indicator.

Issue #632, opened 2017-12-19
Refresh Nonce Button


dternyak:
From https://github.com/MyEtherWallet/MyEtherWallet/issues/30

Allows users to get up-to-date nonce if they believe the most recently loaded nonce while building a transaction may be out of date.

tayvano:
If anyone has any insights on how to keep user connected to same node behind the load balancer, I'm all ears. This may also be resolved when we make some improvements to the backend / caching.

This is happening a lot lately due to (1) full tx pools (2) hitting different nodes which may not have received the latest tx yet (3) being 100% on parity ATM which handles grabbing transaction count (the nonce) a bit differently than geth.

@skubakdj can probably provide more insights, but ideally the user would never have to refresh the nonce so there wouldn't have to be a button. 😊

dternyak:
I don't know of a way to ensure that a request is routed to a specific node behind the load-balancer, but that's certainly an interesting concept.

Even in the case where the node returns the "correct" nonce after the user decrypts their wallet, they may have sent a tx from another wallet using the same private key. Unless we have a websocket that will push the updated nonce, or we poll, the user wouldn't be able to have their nonce automatically updated in this case (without taking action to refresh themselves).

PR #1021, opened 2018-02-07
Add nonce loading indicator & refresh button

james-prado:
Closes https://github.com/MyCryptoHQ/MyCrypto/issues/632 & closes https://github.com/MyCryptoHQ/MyCrypto/issues/835

- Add loading indicator for nonce field
- Add refresh button for nonce field
- Fix a tiny css issue for footer

- Add `what-input` to remove outlines on buttons when using mouse


package.json
```diff
@@ -50,6 +50,7 @@
     "scryptsy": "2.0.0",
     "uuid": "3.2.1",
     "wallet-address-validator": "0.1.1",
+    "what-input": "^5.0.5",
```
dternyak: version.

common/components/NonceField.tsx
```diff
@@ -1,38 +1,72 @@
...
 }
 
-const nonceHelp = (
-  <Help
-    size={'x1'}
-    link={'https://myetherwallet.github.io/knowledge-base/transactions/what-is-nonce.html'}
-  />
-);
+interface StateProps {
+  nonePending: any;
```
dternyak: type

common/components/NonceField.tsx
```diff
@@ -1,38 +1,72 @@
...
+}
+
+interface DispatchProps {
+  requestNonce: any;
```
dternyak: type

common/components/GasLimitField.tsx
```diff
@@ -1,66 +1,34 @@
...
-            />
-          )}
-        </React.Fragment>
-      )}
-    />
-  </React.Fragment>
+export const GasLimitField: React.SFC<Props> = ({ customLabel, disabled }) => (
+  <GasLimitFieldFactory
+    withProps={({ gasLimit: { raw }, onChange, readOnly, gasEstimationPending }) => (
+      <React.Fragment>
```
dternyak: useless fragement

common/Root.tsx
```diff
@@ -17,6 +17,7 @@ import { Store } from 'redux';
 import { pollOfflineStatus } from 'actions/config';
 import { AppState } from 'reducers';
 import { RouteNotFound } from 'components/RouteNotFound';
+import 'what-input';
```
dternyak: What is this package being used for? 

https://github.com/ten1seven/what-input gives some information, but I don't see why we need the functionality.

wbobeirne: Looks like it's used in sass/styles.scss to suppress the focus ring. I think the idea is to not show the focus ring if a user is clicking around, but to show it when they're navigating via keyboard. Haven't run the site yet, but I wonder if this may have some unintended consequences. But I like the concept.

james-prado: `what-input` lets you remove the big blue outline around buttons & links only if a user is using the mouse to navigate (so not to remove it for users who need that accessibility feature)

common/components/GasLimitField.tsx
james-prado: no it's required here. `withProps` has to render 1 dom element, so you need to wrap the `<div />` & `<input />` in either a `<div />` or `<React.Fragment />`

dternyak: Huh, learned something new. Thanks for the explanation.

dternyak:
During testing, I've found that the request can resolve so quickly that the user is given no feedback that a request was made.

I suggest adding a `race` or `delay` to ensure that the loading indicator is shown for at least 1s before disappearing.

dternyak:
I addressed my feedback in https://github.com/MyCryptoHQ/MyCrypto/pull/1021/commits/0e369b12bd20fa128fab8bb57333d5f45c6bac5f -- hope you don't mind! good to merge now 👍

PR #1035, opened 2018-02-09
Add regex for BIP44 Strings

james-prado:
- Add a regex to test dPath strings
	- The regex supports several different lengths, both with and without whitespace
	- checks for valid integers
	- checks for valid boolean (int of 0 or 1)
	- supports optional whitespace
- Add tests for the regex

Not sure if this is intentional, but the DPath value for `ETC_LEDGER` has a `'` after the `change` portion of the string, marking hardened derivation where it [should be public](https://github.com/bitcoin/bips/blob/master/bip-0044.mediawiki#change)

`m / purpose' / coin_type' / account' / change / address_index`

```
export const ETC_LEDGER: DPath = {
  label: 'Ledger (ETC)',
  value: "m/44'/60'/160720'/0'"
  // ----------------------- ^ right here
};
```

if this is intentional, I can add an exception, or maybe break out the regex into different sections to pick and choose from, ie `m / purpose' / coin_type' / account'` or `m / purpose' / coin_type' / account' / change / address_index`, and etc...


spec/config/dPathRegex/dPathRegex.spec.ts
```diff
@@ -0,0 +1,30 @@
...
+      expect(whitespaceDPathRegex.test(path)).toBeTruthy();
+    });
+  });
+
+  it("shouldn't match invalid strings", () => {
+    whitespaceInvalid.forEach(path => {
+      expect(whitespaceDPathRegex.test(path)).toBeFalsy();
+    });
+  });
+});
```
wbobeirne: I think more valuable than testing the regex's, we should be testing our validator function. We don't use the regexes in code directly anywhere, they're all hidden behind the one validator function.

With that in mind, this should probably expand on the existing tests in /spec/libs/validators.spec.ts

common/config/dpaths.ts
```diff
@@ -48,5 +48,19 @@ export const ETH_SINGULAR: DPath = {
...
+
+// m / purpose' / coin_type' / account' / change / address_index
+//   |          |            |          |        |
+//   | constant |   index    |  index   | 0 or 1 |
+//   |__________|____________|__________|________|
+
+// whitespace strings are evaluated the same way as nospace strings, except they allow optional spaces between each portion of the string
+// ie. "m / 44' / 0' / 0'" is valid, "m / 4 4' / 0' / 0'" is invalid
+export const dPathRegex = /m\/44'\/[0-9]+\'\/[0-9]+(\'+$|\'+(\/[0-1]+$))/;
+export const whitespaceDPathRegex = /m\s*\/\s*44'\s*\/\s*[0-9]+\'\s*\/\s*[0-9]+(\'+$|\'+\s*(\/\s*[0-1]+$))/;
```
wbobeirne: Does this get used in code anywhere, other than the unit tests? No use in testing code that doesn't actually run in the product.

spec/config/dPathRegex/dPathRegex.spec.ts
```diff
@@ -0,0 +1,30 @@
+import { dPathRegex, whitespaceDPathRegex } from 'config';
```
wbobeirne: Spec tests are traditionally supposed to match the file structure in common.

spec/config/dPathRegex/dPathRegex.spec.ts
wbobeirne: Yeah now that I'm looking at this more, there's actually a false positive here in having a test for the whitespace regex thinking that that would work for user input, but the isValidPath function only uses the other regex. So those would be invalid in the app.

common/config/dpaths.ts
james-prado: No, it's not actually used anywhere right now. It doesn't need to be tested, since currently it isn't used. It'd probably be better just to comment it out, if we ever plan to take user input and want to verify BIP44 strings, it'd be good to accept strings with whitespace.

wbobeirne: Ah, looks like we disabled custom paths in DeterministicWalletsModal.tsx. That's where we were allowing user input for these. But the code is still all there for it.

HenryNguyen5:
The coin type regex should be `60` according to https://github.com/satoshilabs/slips/blob/master/slip-0044.md

james-prado:
@HenryNguyen5 the regex allows any integer > 1 as a coin type

HenryNguyen5:
Oops sorry, I read the wrong index for it. Everything looks solid

Issue #913, opened 2018-01-24
Add Symbols to Equivalent Values


dternyak:
<img width="397" alt="screen shot 2018-01-24 at 4 35 13 pm" src="https://user-images.githubusercontent.com/7861465/35360795-913b7338-0124-11e8-9fd5-56645a2d375e.png">

- [ ] Add $, £ etc next to symbol code for each equivalent value.
     - Note: for crypto, many of the symbol assets are already included in source for shapeshift dropdown.

PR #1038, opened 2018-02-09
Add Symbols to Equivalent Values

james-prado:
Closes https://github.com/MyCryptoHQ/MyCrypto/issues/913

screen shot 2018-02-12 at 7 16 37 am

- Add some small css tweaks
- Add symbols for fiat currencies
	- CHF [doesn't have a proper symbol](https://en.wikipedia.org/wiki/Swiss_franc)
- Add coin & token symbols


common/components/ui/SelfHidingImg.tsx
```diff
@@ -0,0 +1,21 @@
...
+interface Props {
+  src: string;
+}
+
+export default class SelfHideImg extends React.Component<Props> {
+  public state = {
+    errored: false
+  };
+
+  public handleError = () => {
```
dternyak: When does this get called?

Issue #1016, opened 2018-02-07
Display 'ETH' instead of 'ether' 


james-prado:
We should remain consistent and use 'ETH' instead of 'ether'. I think the error occurs because the amount field dropdown has the 'ETH' option set to a value of 'ether', which is causing this inconsistency in the tx confirmation modal.

<img width="549" alt="screen shot 2018-02-07 at 11 35 28 am" src="https://user-images.githubusercontent.com/13836920/35928520-5e4a1efc-0bfb-11e8-9b9d-96cfb970c4ce.png">

wbobeirne:
In most places in the code, we do
```
unit === 'ether' ? network.unit : unit
```
so we should either move this into a common use formatter, make it a part of UnitDisplay, or make it a less user-displayable string like `default` or something like that. The reason we do this is because it's still `'ether'` on non-ETH networks, like ETC or UBIQ.

PR #1039, opened 2018-02-09
Use Network Symbol in Confirmation Modal

james-prado:
Closes https://github.com/MyCryptoHQ/MyCrypto/issues/1016
Closes https://github.com/MyCryptoHQ/MyCrypto/issues/1079

- `state.transaction.meta.unit` now defaults to an empty string
	- Set default network unit when `<Root />` component mounts
	- reset `transaction.unit` to network unit on reset via `sagas/transaction/reset.ts` saga
- `isEtherUnit()` is now `isNetworkUnit()`
	- this selector needs improvement but would require some significant refactoring & I think is beyond the scope of this issue
- fix incorect unit conversion for non-ethereum networks https://github.com/MyCryptoHQ/MyCrypto/issues/1079

todo: 
- mock `isNetworkUnit()` for tests
- fix tests


common/libs/units.ts
```diff
@@ -103,7 +103,7 @@ const fromTokenBase = (value: TokenValue, decimal: number) =>
 const toTokenBase = (value: string, decimal: number) =>
   TokenValue(convertedToBaseUnit(value, decimal));
 
-const isEtherUnit = (unit: string) => unit === 'ether' || unit === 'ETH';
+const isEtherUnit = (unit: string) => unit === 'ETH';
```
HenryNguyen5: this should be used everywhere instead of manually checking for `unit === 'ETH'` so we only need to change one function if we ever need to deal with different base units (like EXP / UBIQ). I'm not too sure if we still need to check for `ether`, as API calls or node calls might return that instead of `ETH`

james-prado: good catch, yea `isEtherUnit()` should just be used everywhere and replace the one off conditions.

common/reducers/transaction/meta/meta.ts
```diff
@@ -14,8 +14,8 @@ import {
 import { Reducer } from 'redux';
 
 const INITIAL_STATE: State = {
-  unit: 'ether',
-  previousUnit: 'ether',
+  unit: 'ETH',
+  previousUnit: 'ETH',
```
wbobeirne: I'm a little concerned about making this appear correctly, because I wouldn't want us to use `transaction.meta.unit` nakedly anywhere without turning it into network.unit. How would you feel about making this "default" or "network unit" or something like that, to make it clear to anyone that this shouldn't be used for display? That'd also mean we should rename the check function to isNetworkUnit or something like that, for non-Ethereum networks.

james-prado: Yea that's a good idea, I'll make those changes

wbobeirne:
@james-prado can you fix up the tests / conflicts on this? I'll review when CI is cleared up.

common/actions/transaction/actionCreators/meta.ts
```diff
@@ -5,23 +5,22 @@ import {
...
-
-const setTokenTo = (payload: SetTokenToMetaAction['payload']): SetTokenToMetaAction => ({
+export type TSetTokenTo = typeof setTokenTo;
+export const setTokenTo = (payload: SetTokenToMetaAction['payload']): SetTokenToMetaAction => ({
   type: TypeKeys.TOKEN_TO_META_SET,
   payload
 });
 
-const setTokenValue = (payload: SetTokenValueMetaAction['payload']): SetTokenValueMetaAction => ({
+export type TSetTokenBalance = typeof setTokenValue;
```
HenryNguyen5: Isnt part of this PR, but it should be 
```ts
export type TSetTokenValue typeof setTokenValue;
```

common/components/ui/Dropdown.tsx
```diff
@@ -27,11 +26,11 @@ export default class DropdownComponent<T> extends PureComponent<Props<T>, State>
   private dropdownShell: DropdownShell | null;
 
   public render() {
-    const { ariaLabel, color, size } = this.props;
-
+    const { ariaLabel, color, size, value } = this.props;
+    const labelStr = !!this.props.label ? this.props.label : '';
```
HenryNguyen5: Dont need the double negation here

common/libs/units.ts
```diff
@@ -103,7 +103,9 @@ const fromTokenBase = (value: TokenValue, decimal: number) =>
 const toTokenBase = (value: string, decimal: number) =>
   TokenValue(convertedToBaseUnit(value, decimal));
 
-const isEtherUnit = (unit: string) => unit === 'ether' || unit === 'ETH';
+const isNetworkUnit = (unit: string) =>
+  // TODO: find a better strategy for determining if unit is network unit
+  unit === 'ETH' || unit === 'ETC' || unit === 'UBQ' || unit === 'EXP';
```
HenryNguyen5: You should be able to derive the network unit by reducing the network config state

common/selectors/transaction/meta.ts
```diff
@@ -11,8 +11,8 @@ const getTokenValue = (state: AppState) => getMetaState(state).tokenValue;
 const getUnit = (state: AppState) => getMetaState(state).unit;
 const getPreviousUnit = (state: AppState) => getMetaState(state).previousUnit;
 const getDecimalFromUnit = (state: AppState, unit: string) => {
-  if (isEtherUnit(unit)) {
-    return getDecimalFromEtherUnit('ether');
+  if (isNetworkUnit(state, unit)) {
+    return getDecimal(state);
```
HenryNguyen5: This should be getting the decimal from the current unit passed instead, we dont have defined decimals for now (see https://github.com/MyCryptoHQ/MyCrypto/issues/1185) but, lets stick to `getDecimalFromEtherUnit('ether');` in the meanwhile

common/libs/units.ts
wbobeirne: Yep, unfortunately this wouldn't cover custom networks, so I think this'll need to take in the network object.

common/components/ui/Dropdown.tsx
wbobeirne: Even simpler, `this.props.label || ''`

common/libs/units.ts
```diff
@@ -103,7 +104,17 @@ const fromTokenBase = (value: TokenValue, decimal: number) =>
...
+  unit: string,
+  networks: AppState['config']['networks'],
+  nodes: AppState['config']['nodes']
+) => {
+  const selectedNode =
+    nodes.staticNodes[nodes.selectedNode.nodeId] || nodes.customNodes[nodes.selectedNode.nodeId];
+  const selectedNetwork =
+    networks.staticNetworks[selectedNode.network] || networks.customNetworks[selectedNode.network];
+  return unit === selectedNetwork.unit;
+};
```
wbobeirne: This seems like overkill, passing all of networks and nodes state. Shouldn't we just pass the current network? That's much less data for each component to have as a prop, they only need to pass the result of `getSelectedNetwork`.

wbobeirne: Oh I also see you made a `getNetworkUnit` selector, maybe even just that.

james-prado: yea actually that'd be a lot easier, forgot I even made a selector for that

PR #1040, opened 2018-02-09
Update Banners / Versions to Beta

dternyak:


dternyak:
I left the header $brand-warning - the blue doesn't work well.
screen shot 2018-02-09 at 2 01 11 pm

common/config/data.tsx
```diff
@@ -13,11 +13,11 @@ export const N_FACTOR = 8192;
 export const ANNOUNCEMENT_TYPE = 'warning';
 export const ANNOUNCEMENT_MESSAGE = (
   <div>
-    This is an Beta build of MyCrypto v4. Please only use for testing, or use production at{' '}
+    This is an Beta build of MyCrypto Beta. Please only use for testing, or use production at{' '}
```
wbobeirne: *a Beta build. And maybe this is just me, but a beta build of the beta reads a little redundantly to me.

wbobeirne: Maybe "pre-release" build or "early" build or "preview" build.

dternyak: definitely, good catch

PR #1050, opened 2018-02-10
Gas Price Estimates API

### LINKED ISSUE #1046: Integrate GasExpress
https://dev.blockscale.net/api/gasexpress.json

- [ ] Show marks on slider based on speed data retrieved from `gasexpress`. Default to `fast`
- [ ] Gracefully fail on gasexpress service being down / unavailable, and use hardcoded default gas price

wbobeirne:
Closes #1046. This hooks up to the new gas estimator API to fill in some better guesses for what you should set your gas price to when using the simple view. Here's what's implemented:

### Changes

* Adds a gas reducer, saga, and api for managing gas estimates.
    * This seemed worth breaking out into its own reducer, as it's pretty well separated from all other transaction and config logic, and may get more complicated if we implement time estimates.
    * Tests have also been written for reducer / saga behavior.
* Kicks off a fetch for gas price estimates whenever the slider mounts.
    * The response is cached for a minute, so subsequent loads won't keep re-requesting if it hasn't been too long.
    * If the request fails, falls back to hard-coded defaults in config/data.
* Sets the gas price to `response.fast`, and the upper and lower bounds to `response.fastest` and `response.safeLow` respectively.
* Shows `X Gwei (Recommended)` in the tooltip when you're on the estimator-set gas price.
* While fetching, shows a loader on the gas slider / transaction fee components.

### Screencap

gas-estimate


common/sagas/gas.ts
```diff
@@ -0,0 +1,53 @@
...
+  // Cache estimates for a bit
+  const oldEstimates: AppState['gas']['estimates'] = yield select(getEstimates);
+  if (oldEstimates && oldEstimates.time + gasEstimateCacheTime > Date.now()) {
+    yield put(setGasEstimates(oldEstimates));
+    return;
+  }
+
+  // Try to fetch new estimates
+  try {
+    const estimates: GasEstimates = yield call(fetchGasEstimates);
```
dternyak: I would like to include validation on the `estimates` value to ensure that we aren't setting unexpectedly high default gas prices for users in the event of an error in the API response.

We could create a utility function `checkValidEstimate` that ensures the default value (`fast`) is not above e.g. 250 gwei, and do some additional loose checking on the min/max values.

wbobeirne:
@dternyak great feedback on doing sanity checks, I added something to make sure that the fields are numeric, and not above a crazy amount, otherwise the api throws:

screen shot 2018-02-14 at 10 22 00 am

screen shot 2018-02-14 at 10 22 37 am

I'm not sure what you had in mind about the `low <= default >= fast`. If you want to clarify on Slack, that'd be probably be better.

PR #1054, opened 2018-02-11
Contributor Guidelines

### LINKED ISSUE #1028: Contributor Guidelines
https://github.com/davezuko/react-redux-starter-kit/blob/master/CONTRIBUTING.md

wbobeirne:
Closes #1028. The main contributor guidelines [live on the wiki](https://github.com/MyCryptoHQ/MyCrypto/wiki/Contributor-Guidelines) so they can be updated without needing to PR. I definitely expect edits and suggestions to this, so don't be shy.

One thing that was lacking was test-writing documentation. I think we should probably go into that more somewhere.

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -1,0 +1,1 @@
+Please see the [Contributor Guidelines on the wiki](https://github.com/MyCryptoHQ/MyCrypto/wiki/Contributor-Guidelines).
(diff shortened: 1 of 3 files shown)


dternyak:
Agreed about test-writing documentation. Some of that has begun at https://github.com/MyCryptoHQ/MyCrypto/wiki/Writing-reducer-saga-selector-tests, but we can go into significantly more depth.

PR #1082, opened 2018-02-13
Change in disclaimer modal

clrxbl:


common/components/Footer/DisclaimerModal.tsx
```diff
@@ -41,7 +41,9 @@ const DisclaimerModal: React.SFC<Props> = ({ isOpen, handleClose }) => {
         English and, because of this, the English version of our website is the official text.
       </p>
       <p>
-        <b>MIT License</b> Copyright © 2015-2017 MyCrypto LLC
+        <b>MIT License</b>
+        Copyright © 2015-2017 MyEtherWallet LLC
+        Copyright © 2018 MyCrypto, Inc.
```
dternyak: Prettier issues. 

This would also be formatted all on one line, which would look strange, so we want to fix that up.

dternyak:
Copyright changes have been made independently.

Issue #775, opened 2018-01-10
Check TX Status / TX Replace


dternyak:
- [ ] Port existing functionality from Check TX view on V3

dternyak:
@jclancy93 One more issue you might be interested in. This is more of a straight re-implementation, and could serve to get you familiar with the codebase.

jclancy93:
@dternyak Looks like it might be a good warmup for #774!

The windows build issue looks tricky and my windows dev experience is very limited so I don't think I'm your guy for that one.

gitcoinbot:
__This issue now has a funding of 0.23 ETH (272.2 USD) attached to it.__

 * If you would like to work on this issue you can claim it [here](https://gitcoin.co/funding/details?url=https://github.com/MyEtherWallet/MyEtherWallet/issues/775).
 * If you've completed this issue and want to claim the bounty you can do so [here](https://gitcoin.co/funding/details?url=https://github.com/MyEtherWallet/MyEtherWallet/issues/775)
 * Questions? Get help on the <a href='https://gitcoin.co/slack'>Gitcoin Slack</a>
 * $16166.51 more Funded OSS Work Available at: https://gitcoin.co/explorer

gitcoinbot:
__The funding of 0.23 ETH (270.11 USD) attached has been claimed by @jclancy93.__ 

 @jclancy93, please leave a comment to let the funder (@owocki) and the other parties involved your implementation plan.  If you don't leave a comment, the funder may expire your claim at their discretion. 

 * Learn more [on the gitcoin issue page](https://gitcoin.co/funding/details?url=https://github.com/MyEtherWallet/MyEtherWallet/issues/775)
 * Questions? Get help on the <a href='https://gitcoin.co/slack'>Gitcoin Slack</a>
 * $17821.51 more Funded OSS Work Available at: https://gitcoin.co/explorer

owocki:
@jclancy93 👋

vs77bb:
@jclancy93 You still planning on tackling this one? cc @owocki

jclancy93:
@vs77bb @dternyak sorry about the delay. I have a really big week of work so was hoping to work on this over the weekend. I can release the claim for the mean time if you guys would prefer.

dternyak:
No need, this weekend is fine to start on it

jclancy93:
Sweet. Thanks @dternyak

jclancy93:
[...]

PR #1099, opened 2018-02-15
Check Transaction page (Pt. 1 - The Basics)

wbobeirne:
Addresses the first part of #775, the check transaction page and component.

## Description

This implements the ability to search for a transaction by hash, and get some information about it. To keep changes small, this is only the transaction status part. No cancel functionality is a part of this (yet.) In addition, a few improvements have been made to the amount of information given by also fetching `eth_getTransactionReceipt` on mined transactions. Anything Etherscan shows that this doesn't show is because the information isn't available purely from JSON RPC api calls.

## Changes

* Adds a "Check TX" navigation item and page
* Adds `transactions` actions / reducer / saga for dealing with existing transactions
    * I was a little worried about this getting confused with the `transaction` ones, but couldn't come up with a different name. Maybe that should be renamed to `send`?
* Adds a stand-alone component called `TransactionStatus` that takes in a tx hash, kicks off a request for data, and renders that when it comes in.
    * Displays the following fields:
    * Status, which is SUCCESFUL, PENDING, FAILED, or UNKNOWN (For pre-byzantium transactions that don't have the `status` field.)
    * TX Hash (Linked to block explorer)
    * Block # when mined (Linked to block explorer)
    * From address (Linked to block explorer)
    * To address (Linked to block explorer)
    * Amount
    * Gas price
    * Gas limit
    * Gas used when mined
    * Transaction fee in network unit (gas used * gas price)
    * Deployed contract address (if applicable)
    * Nonce
    * Input Data
* Some of the copy from the old version was left out due to not having features / it being kinda verbose

## Future Features

1. Auto-reload for pending transactions
2. Identify contract sends (At least ERC-20) and show token values instead
3. Transaction canceling like prod
4. Re-send transaction where it auto-fills the transaction form with the past data
5. Maybe rendering the `TransactionStatus` component on Send, after you send, with auto-reloading.

## Steps to Test

1. Start the app on the `check-tx` branch
2. Go to `/tx-status`
3. Enter transaction hashes from https://etherscan.io/txs and confirm their information matches
4. Enter transaction hashes from https://etherscan.io/txsPending and confirm their information matches (and they show up as pending.)
5. Try to enter an invalid transaction and be unable to submit

### The Form

screen shot 2018-02-15 at 2 08 30 pm

### If You Enter an Address Instead

screen shot 2018-02-15 at 2 08 37 pm

### Succesful TX

screen shot 2018-02-15 at 2 06 55 pm

### Failed TX

screen shot 2018-02-15 at 2 07 21 pm

### Pending TX

screen shot 2018-02-15 at 2 07 45 pm

### Creating a New Contract

screen shot 2018-02-15 at 2 37 28 pm


wbobeirne:
Fixed conflicts. @HenryNguyen5 to reconcile the change between custom networks and hard-coded networks, I've made `blockExplorer` a required field of hard-coded networks, so that it's as easy as:
```
!network.isCustom && network.blockExplorer...
```
rather than
```
!network.isCustom && network.blockExplorer && network.blockExplorer...
```
I also think this is good, to require prospective chains to have a functioning block explorer before they can be added to MyCrypto's source.

PR #1144, opened 2018-02-19
Don't refresh on network change

### LINKED ISSUE #1047: Refactor `DeterministicWalletsModalClass` to allow network change without refresh
`DeterministicWalletsModalClass` keeps state in component as a opposed to redux, which means that we can't change network state without a window refresh.

- [ ] Move component state to redux

wbobeirne:
Closes #1047

### Description

Handles a lot of cases for network changes that would require refreshes of state or redux data.

### Changes

* Removes the reload behavior for network changes
* Re-fetches rates in TxMetaDataPanel and EquivalentValues components if network unit changed
* Re-fetches wallet balances if network changes
* Resets dPath state for Trezor, Ledger, and Mnemonic wallets if network dpath changes
    * The code here was a little copy-pasta-ish, would definitely accept some feedback about how to generecize.
* Re-fetches nonce if network changes
    * I changed nonce fetches to fail silently if there's no wallet inst, rather than displaying a warning, like if you're offline. Still warns if nonce fetch fails otherwise.

### Steps to Test

1. Run app with `network-change-no-refresh` branch
2. Unlock a wallet
3. Change to a different network (e.g. ETH -> ETC)
4. Confirm balance updated
5. Confirm nonce updated
6. Confirm equivalent values updated
7. Confirm tx fee USD updated
8. Change back to previous network, and confirm 4-7 again


dternyak:
Although I can't necessarily defend this belief, I think the behavior where the wallet instance is cleared and users are sent back to the account view is preferable, however without the window refresh that we were doing before.

wbobeirne:
I think that's super reasonable, and less error prone. I'll go ahead and do that.

wbobeirne:
Updated the behavior by adding a new saga function in `sagas/config/node.ts` called `handleNewNetwork` that we can use to do any other redux-based network changeovers. It resets the wallet now.

I left behind all of my other network checks though, so that those components will be resistant in any case.

PR #1146, opened 2018-02-19
Copy to clipboard

### LINKED ISSUE #1000: Copy Address to Clipboard
https://www.npmjs.com/package/react-copy-to-clipboard

- [ ] Document / copy icon next to account address that copies it to the clipboard

KanteLabs:
Closes #1000 

### Description

Allows users to copy their address, by clicking the copy icon. 

### Changes

Document/copy icon next to account address that copies it to the clipboard

### Steps to Test

1. Unlock wallet
2. Under account, address click copy icon
3. Paste in any text field to see if address was copied


wbobeirne:
Hey Kante, thanks for taking a crack at this. Right now we don't use yarn, and I want to make sure someone on our end does the checking in of a yarn.lock at some point. Would you mind removing it from your PR temporarily? I'll probably add it to `.gitignore` until we do check it in.

common/components/BalanceSidebar/AccountInfo.tsx
```diff
@@ -90,6 +104,24 @@ class AccountInfo extends React.Component<Props, State> {
...
+                  data-icon="copy"
+                  role="img"
+                  viewBox="0 0 448 512"
+                  className="svg-inline--fa fa-copy fa-w-14 fa-2x"
+                >
+                  <path
+                    fill={this.state.currentColor}
+                    d="M433.941 65.941l-51.882-51.882A48 48 0 0 0 348.118 0H176c-26.51 0-48 21.49-48 48v48H48c-26.51 0-48 21.49-48 48v320c0 26.51 21.49 48 48 48h224c26.51 0 48-21.49 48-48v-48h80c26.51 0 48-21.49 48-48V99.882a48 48 0 0 0-14.059-33.941zM266 464H54a6 6 0 0 1-6-6V150a6 6 0 0 1 6-6h74v224c0 26.51 21.49 48 48 48h96v42a6 6 0 0 1-6 6zm128-96H182a6 6 0 0 1-6-6V54a6 6 0 0 1 6-6h106v88c0 13.255 10.745 24 24 24h88v202a6 6 0 0 1-6 6zm6-256h-64V48h9.632c1.591 0 3.117.632 4.243 1.757l48.368 48.368a6 6 0 0 1 1.757 4.243V112z"
+                  />
+                </svg>
```
wbobeirne: You can use `<i className="fa fa-copy"/>` instead, we already use font awesome elsewhere this way.

common/components/BalanceSidebar/AccountInfo.tsx
```diff
@@ -90,6 +104,24 @@ class AccountInfo extends React.Component<Props, State> {
           </div>
           <div className="AccountInfo-address-wrapper">
             <div className="AccountInfo-address-addr">{address}</div>
+            <CopyToClipboard onCopy={this.onCopy} text={address}>
+              <div className="copy-clipboard" title="copy to clipboard">
```
wbobeirne: We use SuitCSS class naming for styling, you can read about that [more here](https://github.com/MyCryptoHQ/MyCrypto/wiki/Style-Guides#classnames). Would you mind changing this to adhere to that?

package.json
```diff
@@ -34,7 +34,8 @@
     "qrcode.react": "0.7.2",
     "query-string": "5.1.0",
     "rc-slider": "8.6.0",
-    "react": "16.2.0",
+    "react": "^16.2.0",
+    "react-copy-to-clipboard": "^5.0.1",
```
wbobeirne: Version numbers should be locked in, not `^`.

common/components/BalanceSidebar/AccountInfo.tsx
```diff
@@ -69,6 +74,15 @@ class AccountInfo extends React.Component<Props, State> {
     });
   };
 
+  public onCopy = () => {
+    this.setState(state => {
+      return {
+        copied: !state.copied,
+        currentColor: 'blue'
```
wbobeirne: It would be preferable if style related stuff was handled in CSS, using our sass variables instead of introducing new colors. Can you have a different class added based on `copied`'s state, and style it in `AccountInfo.scss`?

common/components/BalanceSidebar/AccountInfo.tsx
KanteLabs: okay so the better styling would be "AccountInfo-copy-icon" ?

KanteLabs:
okay i updated the branch with the latest changes

common/components/BalanceSidebar/AccountInfo.scss
```diff
@@ -2,6 +2,17 @@
 @import 'common/sass/mixins';
 
 .AccountInfo {
+  &-copy-icon{
+    width: 15px;
+    color: black;
+    &-copied{
+      color: red;
```
wbobeirne: Can you change `color: black;` and `color: red;` with some of our existing colors in `common/sass/variables/colors.scss`? The file's already imported in here. I'd suggest `$gray-light` by default, `$text-color` on hover, and `$brand-success` or `$brand-primary` for the copied state. We usually save red for errors.

common/components/BalanceSidebar/AccountInfo.tsx
```diff
@@ -90,6 +101,15 @@ class AccountInfo extends React.Component<Props, State> {
           </div>
           <div className="AccountInfo-address-wrapper">
             <div className="AccountInfo-address-addr">{address}</div>
+            <CopyToClipboard onCopy={this.onCopy} text={address}>
+              <div
+                className={`AccountInfo-copy-icon${this.state.copied ? '-copied' : ''}`}
+                title="Copy To Clipboard"
+              >
+                <i className="fa fa-copy" />
+                <p>{this.state.copied ? 'copied!' : null}</p>
```
wbobeirne: Right now this text appears on a second line, I think it'd look a little nicer if it was besides the button. Can you make this a span, and add a margin between them?

common/components/BalanceSidebar/AccountInfo.tsx
```diff
@@ -69,6 +72,14 @@ class AccountInfo extends React.Component<Props, State> {
     });
   };
 
+  public onCopy = () => {
+    this.setState(state => {
+      return {
+        copied: !state.copied
+      };
+    });
+  };
```
wbobeirne: Right now the only way to get rid of the copied text is to click it again. That also doesn't give exactly the right impression, since they just copied it again. I'd suggest leaving this text up just for a little bit, something like:

```
this.setState({ copied: true });

setTimeout(() => {
  this.setState({ copied: false });
}, 2000);
```

common/components/BalanceSidebar/AccountInfo.tsx
wbobeirne: Also, because some users may not understand the icon, it'd be nice if when it wasn't copied, there was some text that said `copy address`.

common/components/BalanceSidebar/AccountInfo.tsx
KanteLabs: I also realize that you can't really undo a copy to clipboard. I think I'll just remove that aspect, and keep the copied text displayed

KanteLabs:
Okay, everything should be up to date with the style guidelines.

dternyak:
Thanks a lot @KanteLabs! 👍

Issue #1083, opened 2018-02-13
Transaction History Component


dternyak:
### Description of the Feature

As part of the on-going work in https://github.com/MyCryptoHQ/MyCrypto/issues/774, @wbobeirne has mocked up a TX history component that will show progress as confirmations are broadcast during `sweeper` sweeps.

This TX history component should be generic enough to also apply to our Send view, so that we can display the transaction history for the unlocked account.

### Example(s) of Feature

Currently, we link out to [etherscan](https://etherscan.io/address/0xcfceab0b7380befc8952a674df2fcff0398bd2a7) so that users can view their transaction history.

wbobeirne:
So the mockups in #774 don't actually show what a history would look like, so design wise we'd have to start fresh.

Full history for an address is actually a tricky subject. Nodes don't offer a way of requesting all of the transactions for an address, only the information of individual transactions. This means that given a user's address, we don't have an easy, node-agnostic way of collecting all of this information. This leaves us with a few choices:

1. Leverage etherscan or some other explorer for getting this information. All we would need to grab is all the tx hashes for a given address, and we could let the node do the rest of the work.
2. Set up our own infrastructure for tracking all of these transactions. I think this could be pretty lightweight as a redis set, but it would definitely be a non-trivial amount of dev work.
3. Only track transactions sent through MEW, and store tx hashes in local storage. This is something I think I want to do as a part of #775 as-is, have a pre-filled dropdown of recent transactions for you to check, and we could extend that. However this may cause major user confusion, not seeing the same transactions that they see on Etherscan, if they've done some in a different browser / different client. Not very in favor of this one.

Let me know what you think. I think for our current use-cases, we can do the short-term same session transaction tracking for sweeper and check tx. But if we want a proper full history, we'll have to figure this out.

dternyak:
Session specific transaction history would be 👍 with me as an MVP.

I'm not sure that Etherscan could / would want to handle the additional load for retrieving the tx hashes for a given address, so we may have to hold off until geth / parity have that functionality natively.

I'd definitely like to stay away from (3), as I agree with you that users are likely to experience confusion.

PR #1147, opened 2018-02-19
Show Recent Txs on Check Tx Page

wbobeirne:
Pre-requisite to #1083, enhancement for #775.

### Description

When you send a transaction, it'll now be logged to local storage. Check Tx page uses this by displaying a select of your recent transactions.

### Changes

* Added `sagas/transaction.ts` function that listens to tx broadcasts, logs them to local storage
* Added select list to /tx-status, only visible when you have recent transactions
* Made /tx-status better at handling network changes
* Checksummed addresses shown on `components/TransactionStatus.tsx`

### Known Issues

* Right now the recent txs list is unlimited. We might want to put some kind of cap on it, like 100, and prune from it if transactions get old.
* The notification that pops up doesn't have a link to tx-status, I'll add that later.

### Steps to Test

1. Start app on the `save-transactions` branch
2. Send a transaction (Use testnet, works there too.)
3. Go to /tx-status
4. Confirm there is a dropdown
5. Select your TX from that dropdown
6. Confirm it was the correct transaction
7. Switch to a different network
8. Confirm that you have no selectable recent transactions

### Screenshots

tx-history


common/actions/transactions/constants.ts
```diff
@@ -1,5 +1,7 @@
 export enum TypeKeys {
   TRANSACTIONS_FETCH_TRANSACTION_DATA = 'TRANSACTIONS_FETCH_TRANSACTION_DATA',
   TRANSACTIONS_SET_TRANSACTION_DATA = 'TRANSACTIONS_SET_TRANSACTION_DATA',
-  TRANSACTIONS_SET_TRANSACTION_ERROR = 'TRANSACTIONS_SET_TRANSACTION_ERROR'
+  TRANSACTIONS_SET_TRANSACTION_ERROR = 'TRANSACTIONS_SET_TRANSACTION_ERROR',
```
HenryNguyen5: The above two action constants arent in this PR but, shouldn't this be the other way around like `TRANSACTION_TRANSACTION_DATA_SET`?

common/sagas/transactions.ts
```diff
@@ -34,6 +49,38 @@ export function* fetchTxData(action: FetchTransactionDataAction): SagaIterator {
   yield put(setTransactionData({ txhash, data, receipt, error }));
 }
 
+export function* saveBroadcastedTx(action: BroadcastTransactionQueuedAction) {
+  const txBuffer = action.payload.serializedTransaction;
```
HenryNguyen5: You could do variable reassigning in the object deconstruction here instead 
```ts
const { serializedTransaction: txBuffer, txIdx: indexingHash } = action.payload
```

common/sagas/transactions.ts
```diff
@@ -34,6 +49,38 @@ export function* fetchTxData(action: FetchTransactionDataAction): SagaIterator {
...
+    TxTypeKeys.BROADCAST_TRASACTION_FAILED
+  ]);
+
+  // If our TX succeeded, save it and update the store.
+  if (
+    res.type === TxTypeKeys.BROADCAST_TRANSACTION_SUCCEEDED &&
+    res.payload.indexingHash === txIdx
+  ) {
+    const tx = new EthTx(txBuffer);
+    saveRecentTransaction(res.payload.broadcastedHash, tx);
```
HenryNguyen5: should be done via `yield call( saveRecentTransaction, res.payload.broadcastedHash, tx); `

common/sagas/transactions.ts
```diff
@@ -34,6 +49,38 @@ export function* fetchTxData(action: FetchTransactionDataAction): SagaIterator {
...
+    yield call(loadRecentTxs);
+  }
+}
+
+export function* loadRecentTxs() {
+  const recentTxs = loadRecentTransactions();
```
HenryNguyen5: should be done via `yield call`

common/utils/localStorage.ts
```diff
@@ -62,3 +66,40 @@ async function getWalletConfigKey(wallet: IWallet): Promise<string> {
...
+    const txObj: SavedTransaction = {
+      hash,
+      to: toChecksumAddress(fields.to),
+      from: toChecksumAddress(from),
+      value: fields.value,
+      chainId: fields.chainId,
+      time: Date.now()
+    };
+    const recentTxs = [txObj, ...loadRecentTransactions()];
+    localStorage.setItem('recent-transactions', JSON.stringify(recentTxs));
```
HenryNguyen5: Can this be made into an enum instead?

common/sagas/transactions.ts
```diff
@@ -34,6 +49,38 @@ export function* fetchTxData(action: FetchTransactionDataAction): SagaIterator {
...
+    res.type === TxTypeKeys.BROADCAST_TRANSACTION_SUCCEEDED &&
+    res.payload.indexingHash === txIdx
+  ) {
+    const tx = new EthTx(txBuffer);
+    saveRecentTransaction(res.payload.broadcastedHash, tx);
+    yield call(loadRecentTxs);
+  }
+}
+
+export function* loadRecentTxs() {
```
HenryNguyen5: confusing naming scheme here, maybe rename `loadRecentTransactions` to `loadRecentTransactionsFromLS`

common/sagas/transactions.ts
```diff
@@ -34,6 +49,38 @@ export function* fetchTxData(action: FetchTransactionDataAction): SagaIterator {
...
+  ]);
+
+  // If our TX succeeded, save it and update the store.
+  if (
+    res.type === TxTypeKeys.BROADCAST_TRANSACTION_SUCCEEDED &&
+    res.payload.indexingHash === txIdx
+  ) {
+    const tx = new EthTx(txBuffer);
+    saveRecentTransaction(res.payload.broadcastedHash, tx);
+    yield call(loadRecentTxs);
```
HenryNguyen5: Instead of re-loading the all tx's again to rehydrate the redux store, this should just dispatch an action to the reducer `TRANSACTIONS_SET_RECENT_TRANSACTIONS`

common/utils/localStorage.ts
HenryNguyen5: On second thought, this should probably just be stored within the redux state itself under `recent`, avoiding the use of separate keys and binding the data to the redux state

common/actions/transactions/constants.ts
wbobeirne: Do we have a history of doing NAMESPACE_NOUN_VERB? Most places I see have the noun last. Either way, we should probably document nomenclature if we have a preference.

common/sagas/transactions.ts
wbobeirne: Oh cool, didn't know you could rename with destructures.

common/sagas/transactions.ts
wbobeirne: I use `tx` as shorthand in all of these saga function names, so I'll leave that for consistency, but I'll make it clear it's from LS.

wbobeirne:
@HenryNguyen5 refactored recent transactions to just have one action that pushes onto the list. I also refactored all of the transactions types into `shared/types/transactions.ts`. Finally, I limited the recent transactions list to 50.

wbobeirne:
It looks like CI is failing due to circular class dependencies, I'll have to see where that was introduced.

I've got to say, circular dependencies have burned me 3 times in the past week. We might want to consider getting more strict on them.

wbobeirne:
@dternyak thanks for merging in the tx subtab. Were there any comments from that PR leftover that you wanted to see implemented before we do this one?

dternyak:
@wbobeirne Everything looks good! We can discuss more about moving transactions into the main tab at a later time.

PR #1149, opened 2018-02-20
[WIP] Desktop App Download Page

wbobeirne:
Prerequisite to #557

### Description

Adds a download page at /download that describes the app, and links to downloads for it. Most of the text here is just placeholder to give a sense for layout.

### Changes

* Adds a new route at /download that renders the DownloadPage component.
* Adds a prop to TabSection `hideHeader` that makes it not render the header.

### TODOs

* Figure out if we're still going to offer the HTML version
* Provide screenshots for all platforms
* Finalize copy / what we say about it
* Fetch version number / direct download link from Github release

### Screenshots

download-page


wbobeirne:
* The gradient transition was intentional, but by no means has to stay. Just trying things out.
* I also considered a separate domain, since I felt that this page really didn't "fit" in our template. It would be easy to port to a separate project.
* Agreed, I didn't put too much thought into these info blocks, they were mostly FPO.
* That text is dynamic. I'll need to get screenshots for all OS's, but it's detected from the browser info: https://github.com/MyCryptoHQ/MyCrypto/pull/1149/files#diff-b0b9d12848a680b1032597632b7c90b4R50

wbobeirne:
Closed in favor of https://github.com/MyCryptoHQ/download.mycrypto.com

PR #1150, opened 2018-02-20
Align footer to bottom

james-prado:
No issue currently open for this.

### Description

Align the footer to the bottom of the page 

### Changes

- Add flexbox styles to `.page-layout` to align the footer to the bottom of the page
	- added `display: flex; flex-direction: column; flex-wrap: nowrap;`
	- added `<div className="flex-spacer" />` to fill empty space and push the footer (and prefooter) to the bottom of the page
	- reduce the margin on `<Prefooter />` from `5rem` to `2rem`

Before: 
screen shot 2018-02-20 at 2 36 02 pm

After: 
screen shot 2018-02-20 at 2 42 22 pm

### DIFF
--- a/common/sass/styles/scaffolding.scss
+++ b/common/sass/styles/scaffolding.scss
@@ -34,6 +34,20 @@ body {
   height: 100%;
 }
 
+#app {
+  height: 100%;
+}
+
+.page-layout {
+  height: 100%;
+  display: flex;
+  flex-direction: column;
+  flex-wrap: nowrap;
+  > .flex-spacer {
+    flex-grow: 2;
+  }
+}
+
 input,
 button,
 select,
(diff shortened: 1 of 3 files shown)


common/sass/styles/scaffolding.scss
```diff
@@ -34,6 +34,20 @@ body {
   height: 100%;
 }
 
+#app {
+  height: 100%;
+}
+
+.page-layout {
+  height: 100%;
```
wbobeirne: Shouldn't this and #app be `min-height`? Will anything weird happening with overflow if it exceeds the page height?

james-prado: Nope, nothing weird would happen. Typically, all components get added to `<body />` and when it overflows, a scrollbar appears on `<body />`. Since it's a react app, we also wrap all our components in `<div id="app" />`. Then everything is wrapped once more in `<div className="page-layout" />`. Adding the height 100% to these 3 DOM elements ensures the content will be the height of the viewport, and any overflow will show a scrollbar. 

Really, the `height: 100%;` on `.page-layout` and `#app` is the same as inheriting the height from `<body />` with `height: inherit;`

Issue #1153, opened 2018-02-21
Adjust Wallet Interface 


dternyak:
```
interface IBaseWallet {
  isReadOnly?: boolean;
  getAddressString(): Promise<string> | string;
  getPrivateKeyString?(): string;
}
```

`getAddressString` was originally implemented to return a promise, but in practice it always returns a synchronous value since address is kept in state after the original async unlock.

- [ ] convert interface and interface implementations to type signature of string return type
- [ ] convert callers to stop `await`ing

Issue #1137, opened 2018-02-19
Display Checksummed Addresses


dternyak:
- [ ] Create component that ensures the address that is passed by prop is checksummed and then renders it.
- [ ] Convert all instances of address being shown in ^ component for display

wbobeirne:
We should probably checksum everywhere we display an address, for consistency.

dternyak:
@wbobeirne Definitely, we should make a component that accepts the address and enforces that it's checksummed.

PR #1154, opened 2018-02-21
Adjust Wallet Interface

dternyak:
Closes https://github.com/MyCryptoHQ/MyCrypto/issues/1153

To test:

1. test generate keystore works (from https://github.com/MyCryptoHQ/MyCrypto/pull/1154/files#diff-da809197b595406e304acd50516913eaR62)
2. test web3 unlock works (https://github.com/MyCryptoHQ/MyCrypto/pull/1154/files#diff-6e02304097be0810552c805f4537b71bR18)
3. Test signing works for all formats (https://github.com/MyCryptoHQ/MyCrypto/pull/1154/files#diff-14d64f27973fa652c812d4bf57981601R28)
4. Test payment requests work (https://github.com/MyCryptoHQ/MyCrypto/pull/1154/files#diff-3b47f7a518452473a0a206bddecfd7b9R144)
5. Test creating a signed message on all wallet formats (https://github.com/MyCryptoHQ/MyCrypto/pull/1154/files#diff-4922ccfe05f56e9512da9a9c6571d78dR52

### DIFF
--- a/common/containers/Tabs/SignAndVerifyMessage/components/SignMessage/SignButton.tsx
+++ b/common/containers/Tabs/SignAndVerifyMessage/components/SignMessage/SignButton.tsx
@@ -25,7 +25,7 @@ export default class SignMessageButton extends React.Component<Props, {}> {
 
     try {
       const signedMessage: ISignedMessage = {
-        address: await wallet.getAddressString(),
+        address: wallet.getAddressString(),
         msg: message,
         sig: await wallet.signMessage(message),
         version: '2'
--- a/common/utils/localStorage.ts
+++ b/common/utils/localStorage.ts
@@ -40,16 +40,16 @@ export async function saveWalletConfig(
   wallet: IWallet,
   state: Partial<WalletConfig>
 ): Promise<WalletConfig> {
-  const oldState = await loadWalletConfig(wallet);
+  const oldState = loadWalletConfig(wallet);
   const newState = { ...oldState, ...state };
-  const key = await getWalletConfigKey(wallet);
+  const key = getWalletConfigKey(wallet);
   localStorage.setItem(key, JSON.stringify(newState));
   return newState;
 }
 
-export async function loadWalletConfig(wallet: IWallet): Promise<WalletConfig> {
+export function loadWalletConfig(wallet: IWallet): WalletConfig {
   try {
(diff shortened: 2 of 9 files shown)


common/containers/Tabs/SignAndVerifyMessage/components/SignMessage/SignButton.tsx
```diff
@@ -25,7 +25,7 @@ export default class SignMessageButton extends React.Component<Props, {}> {
 
     try {
       const signedMessage: ISignedMessage = {
-        address: await wallet.getAddressString(),
+        address: wallet.getAddressString(),
```
wbobeirne: This function no longer needs to be async.

common/utils/localStorage.ts
```diff
@@ -40,16 +40,16 @@ export async function saveWalletConfig(
   wallet: IWallet,
   state: Partial<WalletConfig>
 ): Promise<WalletConfig> {
-  const oldState = await loadWalletConfig(wallet);
+  const oldState = loadWalletConfig(wallet);
   const newState = { ...oldState, ...state };
-  const key = await getWalletConfigKey(wallet);
+  const key = getWalletConfigKey(wallet);
```
wbobeirne: this function no longer needs to be async

common/containers/Tabs/SignAndVerifyMessage/components/SignMessage/SignButton.tsx
dternyak: It does due to `sig: await wallet.signMessage(message),` unless I am mistaken

common/utils/localStorage.ts
dternyak: thanks, fixed.

dternyak:
@wbobeirne Perhaps https://github.com/MyCryptoHQ/MyCrypto/issues/1137 could be a connected component that retrieves address, checksums, and then renders? Either way I suggest we handle that in subsequent PR.

common/containers/Tabs/SignAndVerifyMessage/components/SignMessage/SignButton.tsx
wbobeirne: Well butter my biscuits, you're right. My bad.

PR #1162, opened 2018-02-22
Minor swap rates refactor

### LINKED ISSUE #1104: Gracefully Fail With Shapeshift
(User bug report)

Reproduction steps are hard to provide since the issue is dependent on specific network settings for the user, however the issue is described as:

*ShapeShift doesn't load, and just sits in a constant loading state. Every 30 seconds I get an alert saying that swap is reverting to Bity, but nothing changes.*

- [ ] Gracefully fail when shapeshift is unable to load
- [ ] Remove bity fallback and associated messaging.

wbobeirne:
Closes #1104

### Description

Reworks swap rates to be more self-sufficient, and a little smarter about notifying

### Changes

* Reworks the rates sagas to only notify on failure once, but to keep continuing to try to fetch rates after failure
* Makes `CurrentRates.tsx` a connected component that handles all rates redux actions, which means we stop fetching rates after step 1 is finished.
* Refactors all swap saga files to export one kickoff function like other sagas, rather than a bunch of individual saga functions.
* Refactors some typings and action names for consistency and stricter typing
* Reworks error notifications to log error data, but display simple failure copy

### Steps to Test

1. Run app on `swap-refactor` branch
2. Open up Swap, confirm rates fetch
3. Do something to cause a rates fetch to fail (Go offline, throttle network, block requests to shapeshift etc.)
4. Confirm error notification shows
5. Wait 30s, confirm that no other notification shows
6. Restart swap allowing rates to work correctly now
7. Go to step 2 of swap, confirm rates stop fetching
8. Finish swap as normal, confirming swap still works

### Other Notes

I was unable to test a swap end to end because Shapeshift is blocked in New York. Would love for someone to run through and make sure everything's OK.


dternyak:
Tested, confirmed working! 👍

Issue #969, opened 2018-01-30
Move config data to API, request as needed


wbobeirne:
This is something I've been kicking around for a while. One annoying thing about the current MyEtherWallet site is that in order to do something as simple as add a gas limit for an ICO, change gas price ranges, add a new token, or change a node's URL, we need to push a whole new release. If it's something pretty critical, such as updating a token's contract address, old versions that have been downloaded are left high and dry.

I propose that we move out the following items to an API that loads the data from https://github.com/MyEtherWallet/ethereum-lists for the following things:
* Announcement banner, requested immediately
* Node list, requested immediately
* Gas defaults, requested immediately (Assuming we don't migrate to ethgasstation or something)
* Network tokens, requested on account unlock
* Contract ABIs, requested on contracts page

However, I think we should continue to ship with the data in the bundle for offline cases, or if the API is down. We can pop up a message that lets the user know that some data may be out of date, and just revert back to the old data. That way whatever version we ship will just have whatever data is in ethereum-lists at that time.

Let me know what y'all think.

HenryNguyen5:
Definitely in favour of this. But maybe with a more decentralised solution such as IPFS + Smart contract

wbobeirne:
While I'm down for rock-solid solutions for the critical stuff, I'd like to avoid a painful deploy process for the less critical things:

* Announcements - Definitely don't want to start pointing people towards some scam ICO, so yeah, this would be a good choice for that.
* Node list - I don't feel like a malicious node does a terrible amount of damage, and could be caught pretty fast. Perhaps we hard-code the default node, but allow the list to be loaded dynamically to mitigate any damage?
* Gas defaults - Worst it could do is make unobservant people overpay / fail to transact. Not a big deal.
* Network tokens - Since all you're sending is data, not ETH, I don't think this can be used maliciously? Someone correct me if I'm wrong.
* Contract ABIs - Definitely a danger zone here, since someone could change the contract address, and would be pretty subtle.

dternyak:
@wbobeirne Changing the contract address for default network tokens would be a rather large attack vector.

HenryNguyen5:
We can instead pull from parity's token registry, which also lets us display icons for the tokens to the user (horray ux!)
https://paritytech.github.io/wiki/Token-Registry

dternyak:
For the initial MVP, we'll be exploring a multi-sig smart contract IPFS hash pointer to `ethereum-lists`'s token lists.

[...]

PR #1247, opened 2018-03-04
Create script to update tokens list from ethereum-lists/tokens

wbobeirne:
Triage temporary solution for #969.

### Description

Right now we've just been keeping up to date with MyCryptoHQ/mycrypto.com for token lists changes, but the future plan is to use [ethereum-lists/tokens](https://github.com/ethereum-lists/tokens) as our source of tokens. The problem is its format is kind of spread out and not easy to import directly to our repo. This PR adds a script that fetches its output from IPFS, imports and formats the tokens to match our format, and stages the tokens for commit. This won't stop us from manually updating, but should make the process much easier.

One concern of this change is that our token list has ballooned from ~90 to 513 with this. This'll be a much larger load on our node, and probably includes a lot of junk tokens. Not to mention, some of them don't use proper symbol acronyms, and instead have rather long names. We may want to either audit `ethereum-lists/tokens` and clean out any useless ones, or implement a blacklist of tokens to exclude.

~Also of note, I'm not sure decimals or amounts are being handled properly right now with tokens. See screenshot below.~ Fixed, see comment below.

### Changes

* Add a script to a new `scripts/` folder that fetches tokens.
* Updated tokens list with what was fetched from the aforementioned script.

### Steps to Test

1. Checkout `eth-lists-script`
2. Run `npm run update:tokens`
3. Confirm the script runs without fail
4. Confirm no changes are made, since the token lists should be the same

### Screenshots

screen shot 2018-03-04 at 5 01 05 pm

screen shot 2018-03-04 at 5 07 15 pm

^ Are these amounts correct? Is something wrong with decimals?


wbobeirne:
Looks like CI is failing with the new tokens, which probably means we'll need to clean up `ethereum-lists/tokens` before we can fully rely on it.

wbobeirne:
I fixed the incorrect decimals issue, the json from IPFS had decimals as strings, not ints. Fixed that up during import:

screen shot 2018-03-04 at 5 24 24 pm

PR #1257, opened 2018-03-05
Derive available tx metadata

### LINKED ISSUE #1200: Broadcast transaction confirmation doesn't show "from"
### Description of the Issue

When broadcasting a signed transaction, the "From" portion of it doesn't show up

### Steps to Reproduce

1. Sign a transaction
2. Go to /pushTx
3. Broadcast transaction
4. See missing info

screen shot 2018-02-28 at 2 28 15 pm

HenryNguyen5:
Previously we were getting some data such as `to` and `unit` from redux state even when the transaction in context was available. This PR changes the selectors so that they return derived data from the current loaded transaction if available. This isolates the confirmation modal to only derived data when possible.

Closes #1200

### DIFF
--- a/common/selectors/transaction/meta.ts
+++ b/common/selectors/transaction/meta.ts
@@ -22,7 +22,9 @@ const getFrom = (state: AppState) => {
       if (from) {
         return toChecksumAddress(from.toString('hex'));
       }
-    } catch {}
+    } catch (e) {
+      console.warn(e);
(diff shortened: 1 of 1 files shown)


dternyak:
screen shot 2018-03-05 at 10 54 36 pm

During testing, noticed that `from` is checksummed but `to` is not (on broadcast)

PR #1263, opened 2018-03-06
Enable no-implicit-any

### LINKED ISSUE #373: Enable strict mode in tsc
- [ ] fix all implicit any types
- [x] enable strict mode in tsconfig

HenryNguyen5:
This is the new branch for `no-implicit-anys`, the previous one got stale.

Big thanks to @aitrean for helping me finish this one.

Closes #373


common/api/gas.ts
```diff
@@ -1,4 +1,5 @@
 import { checkHttpStatus, parseJSON } from './utils';
+import { Omit } from 'react-redux';
```
wbobeirne: Does this need to be imported? I see it's not here https://github.com/MyCryptoHQ/MyCrypto/pull/1263/files#diff-0fb63f2ecec1a1d8998af37ee70ac8e7R20

HenryNguyen5: It is imported in that file too

electron-app/main/window.ts
```diff
@@ -6,7 +6,7 @@ import { APP_TITLE } from '../constants';
 const isDevelopment = process.env.NODE_ENV !== 'production';
 
 // Cached reference, preventing recreations
-let window;
+let window: any;
```
wbobeirne: Could by typed as `BrowserWindow`.

common/store/store.ts
```diff
@@ -65,10 +65,10 @@ const configureStore = () => {
     ...rehydrateConfigAndCustomTokenState()
   };
 
-  store = createStore(RootReducer, persistedInitialState, middleware);
+  store = createStore<AppState>(RootReducer, persistedInitialState as any, middleware);
```
wbobeirne: I think if persistedInitialState is typed as a partial AppState, this should work.

common/sagas/transaction/current/currentValue.ts
```diff
@@ -19,7 +19,7 @@ export function* setCurrentValue(action: SetCurrentValueAction): SagaIterator {
   return yield call(valueHandler, action, setter);
 }
 
-export function* valueHandler({ payload }: SetCurrentValueAction, setter) {
+export function* valueHandler({ payload }: SetCurrentValueAction, setter: any) {
```
wbobeirne: `setter` could be `setValueField | setTokenValue`

wbobeirne: Or just a function shape, `setter(field): void`

common/sagas/config/web3.ts
```diff
@@ -24,8 +25,13 @@ export function* initWeb3Node(): SagaIterator {
...
 
 // unset web3 as the selected node if a non-web3 wallet has been selected
-export function* unsetWeb3NodeOnWalletEvent(action): SagaIterator {
+export function* unsetWeb3NodeOnWalletEvent(
+  action: ResetWalletAction | SetWalletAction
+): SagaIterator {
   const node = yield select(getNodeId);
+  if (action.type === WalletTypeKeys.WALLET_RESET) {
+    return;
+  }
```
wbobeirne: Why even `takeEvery(WalletTypeKeys.WALLET_RESET, unsetWeb3NodeOnWalletEvent)` if we just return?

common/containers/Tabs/Contracts/components/Interact/components/InteractForm/index.tsx
```diff
@@ -140,7 +140,9 @@ class InteractForm extends Component<Props, State> {
     );
   }
 
-  private handleInput = name => (ev: React.FormEvent<HTMLInputElement | HTMLTextAreaElement>) => {
+  private handleInput = (name: any) => (
```
wbobeirne: `name: string`

common/store/store.ts
HenryNguyen5: this doesn't work because `createStore` expects a full AppState, when it should be a DeepPartial of AppState

common/sagas/config/web3.ts
HenryNguyen5: was not part of this PR but good catch, ill remove it anyway

common/containers/Tabs/Contracts/components/Interact/components/InteractForm/index.tsx
HenryNguyen5: this throws an error because we dont narrow `string` down to `keyof State`

electron-app/main/window.ts
HenryNguyen5: addressed

common/store/store.ts
wbobeirne: Ah, I knew this was familiar. They updated `index.d.ts` in `redux` to allow for deep partials (https://github.com/reactjs/redux/commit/cbdca6215e1891e97120ac05c971618182455e54), but they haven't updated the npm module yet, so it's got old typings. We'll just have to remember to do this the next time redux is updated.

dternyak:
🔥

Issue #493, opened 2017-11-29
Accessability


dternyak:
Ensure that at a minimum, aria-label attributes are at parity with V3.

Notes:
![-2017-11-29 at 2 19 40 pm](https://user-images.githubusercontent.com/7861465/33402154-4b5bd2b0-d521-11e7-8b23-a10929ce25ec.png)

tayvano:
Reference: http://simplyaccessible.com/article/react-a11y/

wbobeirne:
I've got some experience with accessibility (contracts with required WCAG compliance, grueling stuff) and know how to use a screen reader, I'll definitely give this a once-over when we get closer to public beta. But of course, any additional eyes and ears are helpful!

tayvano:
FYI:

> Oh not a problem at all, it is my pleasure. Happy to help with disability/readability issues anytime for free, you are welcome to @ me, dm me, or email me > honeypot (at) http://promether.com 

https://twitter.com/MadameHoneyPot/status/936002085870587904

wbobeirne:
Did a pretty quick audit and ran into plenty of issues. Bold ones are high priority, italicized ones are not the most important. All in all though, the site seemed entirely accessible to me, with not too many pain points.

### Global Components
* **Modals and screen covers (Like the alpha agreement) need to use the [dialog attribute](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_dialog_role)**
* **Modal close button needs descriptive text**
* Footer social icons need descriptive text
* **More alerts should be using `aria-live`, `"assertive"` if it's very important and `"polite"` if it can wait.** Notifications do this, and it works very well.

### Generate Wallet
* Private key in disabled input has aria-label of [Object object]
* Print paper wallet page has multiple `h1` elements
* "Print Paper Wallet" button has aria-label of [Object object]

### Send Transaction
* Online / Offline image needs alt text
* Promo images need alt text
* Promo buttons need descriptive text
* Token row balance title attribute is in wei, very hard to understand, and says to double click, which doesn't do anything on this element
* Close wallet selector X button needs descriptive text
* _Generate transaction button should change to indicate it's loading, there's no feedback to the screen reader_

### Swap
* Bity logo needs alt text
* Multiple h1 elements

dternyak:
[...]

PR #1267, opened 2018-03-06
improve accessibility (a11y)

james-prado:
Closes https://github.com/MyCryptoHQ/MyCrypto/issues/493

### Description

This pr makes various a11y improvements throughout the site. Included are improvements to focus management, new aria-labels, new alt texts, and improving the use of semantic html across the site

#### Global Components

- [x] Add dialog attribute to modals
- [x] Add descriptive text to modal close button 
- [x] Add descriptive text to footer social media icons
- [x] Add `aria-live` to alerts & notifications
- [x] Provide feedback to screen readers when something is loading (`aria-busy`)

#### Generate Wallet

- [x] Update private key disabled input aria-label
- [x] Remove multiple `<h1 />`'s on Print paper wallet view
- [x] Update "Print Paper Wallet" button aria-label

#### Send Transaction

- [x] Add alt text to promo images

#### Swap

- [x] Add alt text to Shapeshift logo

#### Other
- Fix Broken Shapeshift link (was still pointing to bity)

### How To Test

Install a [screen reader](https://chrome.google.com/webstore/detail/chromevox/kgejglhpjiefppelpmljglcjbhoiplfn) and go around the site looking for missing or misleading information.

The site can be tested for ease of use with keyboard navigation by using (shift) tab. Helps if you have experience testing this sort of interaction.


common/components/WalletDecrypt/components/WalletButton.tsx
```diff
@@ -86,24 +86,47 @@ export class WalletButton extends React.PureComponent<Props> {
       >
         <div className="WalletButton-inner">
           <div className="WalletButton-title">
-            {icon && <img className="WalletButton-title-icon" src={icon} />}
+            {icon && <img className="WalletButton-title-icon" src={icon} alt="Logo" />}
```
wbobeirne: This alt should probably be dynamic, e.g. `alt={'${name} Logo'}`

common/components/WalletDecrypt/components/WalletButton.tsx
```diff
@@ -86,24 +86,47 @@ export class WalletButton extends React.PureComponent<Props> {
...
+              aria-hidden={!isSecure}
+            >
+              {description}
+            </div>
+          )}
+          {example && (
+            <div
+              className="WalletButton-example"
+              aria-label="example"
+              aria-hidden={!isSecure ? true : false}
```
wbobeirne: `!!isSecure`

common/components/WalletDecrypt/components/WalletButton.tsx
```diff
@@ -86,24 +86,47 @@ export class WalletButton extends React.PureComponent<Props> {
...
             <span>{name}</span>
           </div>
 
-          {description && <div className="WalletButton-description">{description}</div>}
-          {example && <div className="WalletButton-example">{example}</div>}
+          {description && (
+            <div
+              className="WalletButton-description"
+              aria-label="description"
+              aria-hidden={!isSecure}
```
wbobeirne: Why would this be aria hidden if it's insecure?

common/components/WalletDecrypt/components/WalletButton.tsx
wbobeirne: Also, why would this be hidden if it's secure?

common/components/WalletDecrypt/components/WalletButton.tsx
```diff
@@ -86,24 +86,47 @@ export class WalletButton extends React.PureComponent<Props> {
...
+              {example}
+            </div>
+          )}
 
           <div className="WalletButton-icons">
             {icons.map(i => (
               <span className="WalletButton-icons-icon" key={i.icon} onClick={this.stopPropogation}>
                 {i.href ? (
-                  <NewTabLink href={i.href} onClick={this.stopPropogation}>
+                  <NewTabLink href={i.href} onClick={this.stopPropogation} aria-label="More Info">
```
wbobeirne: This label should probably be a prop on the icons list.

common/components/WalletDecrypt/components/WalletButton.tsx
```diff
@@ -86,24 +86,47 @@ export class WalletButton extends React.PureComponent<Props> {
...
                   </NewTabLink>
                 ) : (
-                  <i className={`fa fa-${i.icon}`} />
+                  <i
+                    className={`fa fa-${i.icon}`}
+                    aria-label={isSecure ? 'Secure wallet type' : 'Insecure wallet type'}
+                  />
+                )}
+                {!isDisabled && (
+                  <Tooltip size="sm" aria-hidden={isDisabled}>
```
wbobeirne: This doesn't render if isDisabled, so no need for aria hidden.

common/components/ui/Identicon.tsx
```diff
@@ -24,6 +24,7 @@ export default function Identicon(props: Props) {
         <React.Fragment>
           <img
             src={identiconDataUrl}
+            alt="Unique Address Image"
```
wbobeirne: How much of a pain in the ass would it be for users to hear "Address image for 0x124ab21f...", because I feel like this alt text doesn't tell you what the image is of.

common/components/ui/Modal/ModalBody.tsx
```diff
@@ -0,0 +1,122 @@
...
+
+    // Convert NodeList to Array
+    this.firstTabStop = this.focusableElements[0];
+    this.lastTabStop = this.focusableElements[this.focusableElements.length - 1];
+
+    // Focus first child
+    this.firstTabStop.focus();
+
+    this.modal.addEventListener('keydown', this.keyDownListener);
+  }
```
wbobeirne: `focusableElementsString`, `focusableElements` are only used in this function, should be `const`.

common/components/ui/Modal/ModalBody.tsx
```diff
@@ -0,0 +1,122 @@
...
+        }
+
+        // TAB
+      } else {
+        if (document.activeElement === this.lastTabStop) {
+          e.preventDefault();
+          this.firstTabStop.focus();
+        }
+      }
+    }
```
wbobeirne: This is super clever, I'll have to test this out but if it works as I'd expect I'm a huge fan.

common/components/ui/Modal/ModalBody.tsx
```diff
@@ -0,0 +1,122 @@
...
+  };
+
+  public render() {
+    const { title, children, modalStyle, hasButtons, handleClose } = this.props;
+    return (
+      <div
+        className="Modal"
+        style={modalStyle}
+        role="dialog"
+        aria-labelledby="Modal-header-title"
```
wbobeirne: `aria-labelledby` must be an ID, not a class. You may want to generate a unique ID for this in case there are multiple modals.

common/containers/OnboardModal/index.tsx
```diff
@@ -180,7 +184,7 @@ class OnboardModal extends React.Component<Props, State> {
     localStorage.setItem(ONBOARD_LOCAL_STORAGE_KEY, String(nextSlideNum));
     this.props.incrementSlide();
     if (this.modal) {
-      this.modal.scrollContentToTop();
+      this.modal.modalBody.scrollContentToTop();
```
wbobeirne: Can you just bubble this up to `modal` by giving it a public function of the same name that calls `this.modalBody.scrollContentToTop`? I'd like to encapsulate as much of the modal changes as possible.

common/components/ui/Modal/index.scss
```diff
@@ -45,6 +34,17 @@ $m-anim-speed: 400ms;
...
+  &-overlay {
+    position: fixed;
+    top: 0;
+    left: 0;
+    right: 0;
+    bottom: 0;
+    background-color: rgba(#000, 0.54);
+    z-index: $zindex-modal-background;
+    display: block;
+  }
```
wbobeirne: This element isn't a child of `.Modal`, which is why it had its own classname.

common/components/ui/OfflineSymbol.tsx
```diff
@@ -26,7 +26,7 @@ const OfflineSymbol = ({ offline, size }: OfflineSymbolProps) => {
       break;
   }
 
-  return <img src={offline ? wifiOff : wifiOn} width={width} height={height} />;
+  return <img src={offline ? wifiOff : wifiOn} alt="wifi status" width={width} height={height} />;
```
wbobeirne: We actually don't use this component anywhere anymore, it can be removed entirely.

common/components/ui/QRCode.tsx
```diff
@@ -35,6 +35,7 @@ export default class QRCode extends React.PureComponent<Props, State> {
     return (
       <img
         src={qr}
+        alt="QR Code"
```
wbobeirne: Should probably require a prop for this, QR code of what?

common/components/ui/Spinner.tsx
```diff
@@ -11,7 +11,7 @@ interface SpinnerProps {
 const Spinner = ({ size = 'x1', light = false }: SpinnerProps) => {
   const color = light ? 'Spinner-light' : 'Spinner-dark';
   return (
-    <svg className={`Spinner Spinner-${size} ${color}`} viewBox="0 0 50 50">
+    <svg className={`Spinner Spinner-${size} ${color}`} viewBox="0 0 50 50" aria-busy="true">
```
wbobeirne: I like the idea of this, but this is actually supposed to mark the container as busy, not the loader. Perhaps this should mark the parent element as busy, and unmark it when this component unmounts?

common/components/ui/SwapDropdown.tsx
```diff
@@ -18,7 +18,7 @@ interface Props<T> {
 const ValueComp: React.SFC = (props: any) => {
   return (
     <div className={`${props.className} swap-option-wrapper`}>
-      <img src={props.value.img} className="swap-option-img" />
+      <img src={props.value.img} className="swap-option-img" alt="Logo" />
```
wbobeirne: Logo for what?

common/components/ui/SwapDropdown.tsx
```diff
@@ -46,7 +46,7 @@ const OptionComp: React.SFC = (props: any) => {
       onMouseEnter={handleMouseEnter}
       onMouseMove={handleMouseMove}
     >
-      <img src={props.option.img} className="swap-option-img" />
+      <img src={props.option.img} className="swap-option-img" alt="Logo" />
```
wbobeirne: Logo for what?

common/containers/Tabs/GenerateWallet/components/CryptoWarning.tsx
```diff
@@ -53,7 +53,7 @@ const CryptoWarning: React.SFC<{}> = () => (
             className="CryptoWarning-browsers-browser"
           >
             <div>
-              <img className="CryptoWarning-browsers-browser-icon" src={browser.icon} />
+              <img className="CryptoWarning-browsers-browser-icon" src={browser.icon} alt="Logo" />
```
wbobeirne: Logo for what?

common/containers/Tabs/GenerateWallet/components/Keystore/EnterPassword.scss
```diff
@@ -3,6 +3,10 @@
 $pw-max-width: 40rem;
 
 .EnterPw {
+  input {
+    border-top-right-radius: 0;
+    border-bottom-right-radius: 0;
+  }
```
wbobeirne: Can we get a fix for this that works for all input groups, rather than fixing this one instance? e.g.
```
.input-group {
  > * {
    border-radius: 0;

    &:first-child {
      // bottom left / top left have border radius
    }

    &:last-child {
      // top right / bottom right have border radius
    }
  }
}
```

common/containers/Tabs/GenerateWallet/components/Keystore/PaperWallet.scss
```diff
@@ -1,13 +1,21 @@
...
+    max-width: 680px;
     margin: 0 auto $space * 3;
+    margin-top: 12px;
+    > h1.input-group-header {
+      margin-bottom: 1rem;
+      > span {
+        font-size: 2rem;
+        margin: auto;
+      }
+    }
```
wbobeirne: Can these get proper classnames?

common/containers/Tabs/Swap/components/ShapeshiftBanner.tsx
```diff
@@ -0,0 +1,17 @@
...
+      <p>
+        <b className="ShapeshiftBanner-banner-new">New Feature:</b>
+        Exchange coins & tokens with
+      </p>
+      <img src={shapeshiftSvg} alt="Shapeshift Logo" />
+    </div>
+  </div>
+);
+
+export default ShapeshiftBanner;
```
wbobeirne: This component's been removed, if you pull develop.

common/containers/Tabs/Swap/components/SwapInfoHeaderTitle.tsx
```diff
@@ -29,11 +29,11 @@ export default class SwapInfoHeaderTitle extends PureComponent<SwapInfoHeaderTit
         <div className="col-xs-3">
           <a
             className="SwapInfo-top-logo"
-            href={bityReferralURL}
+            href={shapeshiftReferralURL}
```
wbobeirne: This should either be a variable based on provider, or we should remove the provider code altogether.

common/containers/Tabs/Swap/components/SwapInfoHeaderTitle.tsx
```diff
@@ -29,11 +29,11 @@ export default class SwapInfoHeaderTitle extends PureComponent<SwapInfoHeaderTit
...
             target="_blank"
             rel="noopener noreferrer"
           >
-            <img className="SwapInfo-top-logo-img" src={logoToRender} />
+            <img className="SwapInfo-top-logo-img" src={logoToRender} alt="Shapeshift Logo" />
```
wbobeirne: Same as above, alt should either be var or remove providers.

wbobeirne:
One other thought I had with a lot of the alt tags: If they're only for visuals, and don't provide extra context, I think it's OK to use aria-hidden. That's not just meant for visually hidden elements, but visually non-useful elements as well. I'll go back and tag my review with things I think could be hidden.

common/components/WalletDecrypt/components/WalletButton.tsx
james-prado: I tried doing `alt={'${name} Logo'}` but it felt a bit weird because I'd get something like 'bitcoin logo bitcoin' which just felt off. Personally I think just `alt="logo"` is the best option

common/components/WalletDecrypt/components/WalletButton.tsx
james-prado: I found the descriptions on the insecure wallet types to a bit annoying when using a screen reader. For example, the private key description is just a private key. Someone using a screen-reader would just head a series of numbers and letters, and have to piece together it's an example of a private key. 

I think think what you're getting at is that this could be handled better, and I agree. It would make more sense to add an aria-label describing the example private key instead of hiding it or providing it to the user with no context.

common/components/WalletDecrypt/components/WalletButton.tsx
james-prado: ^ same as previous comment

common/components/ui/Identicon.tsx
james-prado: I agree, it would improve the context for the user by providing the address. I'm not sure how hard that'd be but I'll look into it.

common/components/ui/Modal/ModalBody.tsx
james-prado: I saw that you merged a pr a few days ago to solve the issue of `esc` always toggling the modal. This should also provide a cleaner solution to that, as all the keydown events can be handled by the modal itself, and all in the same function.

common/components/ui/Modal/index.scss
james-prado: Yup, totally missed that, ty for pointing it out.

common/components/ui/QRCode.tsx
james-prado: Not sure what you mean by 'require a prop for this'. Do you mean it should just provide more context like what the QR code is for?

common/components/ui/Spinner.tsx
james-prado: That might be a little bit difficult to properly find the correct parent element every time the spinner is mounted and unmounted. I think it would make more sense to just add the `aria-busy` attribute manually to all the parent elements that use the spinner.

Just like spinners are conditionally rendered with `{isLoading && <Spinner />}`, `aria-busy={isLoading}` can be added to the parent element.

common/containers/Tabs/GenerateWallet/components/Keystore/EnterPassword.scss
james-prado: This fix was actually specific just to the password inputs when generating a wallet because they're the only inputs with an inline button.

I'll add a proper class to cover this case for any future inputs with inline buttons. Also the border radius for all inputs is 2px.

Before:
screen shot 2018-03-07 at 9 38 16 am

After:
screen shot 2018-03-07 at 9 39 03 am

common/containers/Tabs/GenerateWallet/components/Keystore/PaperWallet.scss
james-prado: yes

common/components/WalletDecrypt/components/WalletButton.tsx
wbobeirne: Example is only used by insecure wallets, and description only by secure. So I think you can make this `true` and just never aria-hidden the description. I think it's fair to assume that examples will always be not-useful, and descriptions will always be useful.

common/components/WalletDecrypt/components/WalletButton.tsx
wbobeirne: See above, just remove `aria-hidden` attribute.

common/components/ui/Modal/ModalBody.tsx
wbobeirne: All my change did was add a `&& this.props.isOpen` check before calling the close function for keycode 27, you should do the same.

common/components/ui/QRCode.tsx
wbobeirne: I mean an `alt` prop for the QRCode component. So something like:
```
<QRCode
  data={wallet.getAddressString()}
  alt="Your Address"
/>
```
and then do
```
alt={`QR Code for ${this.props.alt}`}
```
or something.

common/components/ui/Spinner.tsx
wbobeirne: Yeah I think that'd have to be how it's done. I'm not sure we can easily enforce that, but I'm also not sure how good the benefits are for a screen reader.

common/containers/Tabs/GenerateWallet/components/Keystore/EnterPassword.scss
wbobeirne: This style of addon next to input is actually used in a bunch of places, just grep for `input-group-addon`. Mnemonic generation, ENS, and anything that uses the TogglablePassword component. So I think a holistic fix would be ideal.

I was under the impression Bootstrap did this for us (See the `input-group` example here http://getbootstrap.com/docs/3.3/css/#forms-inline), so we may be overriding that somewhere.

james-prado: Nope, since updating the input styles, we've moved on from using the bootstrap styles

wbobeirne: Oh right on. Well maybe we can carry over what they're doing in that example for border radii.

PR #1268, opened 2018-03-06
Fix recommended gas price breaking on subtab switch

### LINKED ISSUE #1258: Gas Price resets on sub-tab change
Reproduction Steps:

1. unlock wallet
2. check that gas price is "recommended"
3. Change to other subtab
4. Change back to account view
5. Notice that gas price is reset to 20

HenryNguyen5:
Closes #1258 
@wbobeirne could you make sure the logic is correct here?
Instead of checking if the current gas price is higher or lower than the fastest/slow settings on the api call, I just force set it to recommended, since that'll always be in range.


wbobeirne:
fixGasPrice is also meant for going from advanced to simple, in case you set the price outside the range of the slider. This also won't account for when the component re-mounts, componentWillReceiveProps isn't called on initial props, so it wouldn't reset it to fast.

Issue #1213, opened 2018-03-01
View Address Unlock shows account view


dternyak:
When unlocking a view address wallet, the account view should be locked and users should be taken the view info view.

Currently, the account view subtab is locked, but users are still taken to the account view immediately after unlock and are able to interact with the transaction fields / build up a tx.

- [ ] Show view wallet info after unlocked view address wallet instead of account view

Issue #1230, opened 2018-03-02
Center / Adjust Token Scanner Buttons


dternyak:
<img width="432" alt="screen shot 2018-03-02 at 1 58 50 am" src="https://user-images.githubusercontent.com/7861465/36888934-4ec90f32-1dbd-11e8-8a7a-1dff03c2985c.png">

Adexbafo:
I think this needs to be reviewed as the button meant for scanning tokens which I believe has to be automatic, didn't perform that functions at all. So please kindly check this and work on it, as adding manually is strenuous and time consuming.

james-prado:
Closed by https://github.com/MyCryptoHQ/MyCrypto/pull/1276

PR #1276, opened 2018-03-06
Update account subtab routing & minor css tweaks

james-prado:
Closes https://github.com/MyCryptoHQ/MyCrypto/issues/1213, https://github.com/MyCryptoHQ/MyCrypto/issues/1230

### Description

Fix routing issue where user's are able directed to a disabled view. Also made some minor CSS tweaks.

### Changes

- Add an optional `redirect` property for wallets in `<WalletDecrypt />`
	-  when a redirect string is provided, and a user unlocks their wallet they'll be redirected. This is useful for sending users to specific subtabs depending on what type of wallet they log in with.
- Update button styles in token balances

### Steps to Test

1. Go to https://localhost:PORT/account/send 
2. View Address

On the beta right now, doing this will send you to a disabled tab. The new `<WalletDecrypt />` redirect changes fixes this.


common/components/WalletDecrypt/WalletDecrypt.tsx
```diff
@@ -259,21 +262,30 @@ export class WalletDecrypt extends Component<Props, State> {
...
+                    ? this.props.isWalletPending
+                    : undefined
+                }
+                isPasswordPending={
+                  this.state.selectedWalletKey === InsecureWalletName.KEYSTORE_FILE
+                    ? this.props.isPasswordPending
+                    : undefined
+                }
+              />
+            )}
```
wbobeirne: You can't redirect any unlock to /account/send, because the wallet may be unlocked on a different page. Also from a separation of concerns standpoint, I don't like the idea of the standalone `<WalletDecrypt/>` component being responsible for /account's behavior.

This should most likely be handled somewhere in the `containers/SendTransaction` domain. Perhaps the Send form should do the redirect, if the wallet it receives is read-only.

common/components/BalanceSidebar/TokenBalances/index.scss
```diff
@@ -34,6 +34,12 @@
   }
 
   &-buttons {
+    display: flex;
+    flex-wrap: wrap;
+    justify-content: center;
+    > button {
```
wbobeirne: target `&-btn`, not the HTML element

Issue #1284, opened 2018-03-07
Coinbase widget address hardcoded


MrLuit:
### Description of the Issue

The ethereum address that gets sent to coinbase when you click the 'Buy ETH with USD' widget is hardcoded in `common/components/BalanceSidebar/PromoComponents/Coinbase.tsx`:
```
<a
    className="Promos-promo Promos-Coinbase"
    target="_blank"
    rel="noopener noreferrer"
    href="https://buy.coinbase.com?code=60c05061-3a76-57be-b1cd-a7afa97bcb8c&address=0xA7DeFf12461661212734dB35AdE9aE7d987D648c&crypto_currency=ETH&currency=USD"
  >
```

For comparison, this is how it's coded in v3:

```
<a href="https://buy.coinbase.com?code=60c05061-3a76-57be-b1cd-a7afa97bcb8c&address={{wallet.getChecksumAddressString()}}&crypto_currency=ETH&currency=USD"
     ng-show="ajaxReq.type=='ETH' && slide==2"
     class="block swap--usd"
     target="_blank" rel="noopener noreferrer">
```

### Steps to Reproduce

Access a random wallet, and inspect the coinbase widget element using something like Dev Tools. Compare beta and production, beta will always use `0xA7DeFf12461661212734dB35AdE9aE7d987D648c` as the address.

### Description of Your Machine

n/a

### Console Logs / Screenshots

![](https://i.imgur.com/unGdLg0.png)
(on beta)

![](https://i.imgur.com/IraBFfj.png)
(on prod, the address here reflects my wallet's address)

PR #1286, opened 2018-03-07
Fix Coinbase Widget

dternyak:
Closes https://github.com/MyCryptoHQ/MyCrypto/issues/1284

The Coinbase widget does not currently pass the correct instantiated wallet address, and instead passes a hardcoded address (presumably added during testing).

This PR connects `Promos` to redux state to grab the wallet instance, and then passes the correct address to `Coinbase` when the wallet instance exists.

### DIFF
--- a/common/components/BalanceSidebar/Promos.tsx
+++ b/common/components/BalanceSidebar/Promos.tsx
@@ -2,8 +2,9 @@ import React from 'react';
 import { TransitionGroup, CSSTransition } from 'react-transition-group';
 import { HardwareWallets, Coinbase, Shapeshift } from './PromoComponents';
 import './Promos.scss';
-
-const promos = [HardwareWallets, Coinbase, Shapeshift];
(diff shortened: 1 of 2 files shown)


common/components/BalanceSidebar/Promos.tsx
```diff
@@ -2,21 +2,30 @@ import React from 'react';
...
+// Don't change Coinbase index
+const promos = [HardwareWallets, Coinbase, Shapeshift];
+
 interface State {
   activePromo: number;
 }
 
-export default class Promos extends React.PureComponent<{}, State> {
+interface StateProps {
+  wallet: IWallet | undefined | null;
```
HenryNguyen5: Use reference to app state instead `AppState['wallet']['inst']`

dternyak: thanks, good catch

PR #1307, opened 2018-03-09
Allow 0 ETH Transactions

### LINKED ISSUE #1275: Cannot Make Transaction with 0 ETH
### Description of the Issue

When attempting to make a transaction with 0 ETH being sent, the interface prevents me from signing the transaction, regardless of the presence of data in the transaction.

### Steps to Reproduce

Unlock a wallet with a Ledger, then attempt to make a self transaction with 0 ETH in the amount.  

Google Chrome Version 64.0.3282.186 (Official Build) (64-bit)
Windows 10

[Screenshot of the issue](https://imgur.com/a/aiKOB)

wbobeirne:
Closes #1275

### Description

Allows for zero ETH transactions on the send tab. Does not allow for zero token transactions, or zero value transaction requests.

### Changes

* Adjusted `validNumber` validator to allow for `>= 0`
* Added `validPositiveNumber` for cases that should be `> 0`
* Updated saga logic to selectively use those two based on whether or not it's an ETH transaction
* Added tests to validate zero eth transaction is valid, zero token transactions are not

### Steps to Test

1. Run app with the `allow-zero` branch
2. Unlock a wallet with tokens on `/send`
3. Enter `0` as the value, confirm form allows for this (generate the tx)
4. Change unit to a token, confirm field becomes invalid
5. Change value to a positive number, confirm field becomes valid again


wbobeirne:
AWS CodeBuild is likely failing due to #1308.

dternyak:
screen shot 2018-03-05 at 3 46 07 pm

I don't think we want to allow users to generate transactions if amount is undefined in addition to explicitly 0.

wbobeirne:
Great catch, the number casting we were doing `+value` turns `''` into `0`. I swapped those out for parseInts, so that it'll only give us a value if they typed an actual number.

Issue #1285, opened 2018-03-07
Update Translations


james-prado:
With the release of the beta, we've added and update a lot of copy throughout the site. To provide translations, all the new 'raw' text needs to be added to the translation files.

wbobeirne:
Likewise, this would be a good opportunity to clear out unused translations. I'd like to see the following things done as a part of this cleanup, please comment if you think any of these are out of scope:

1. Wrap all user-facing copy in `translate` and `translateRaw` calls
2. Identify as many common use string as possible, and re-key those as `x_NameOfString`. Some of our common use strings are in really weird keys, like `"OFFLINE_Step2_Label_3": "Gas Price"`. This should just be `x_GasPrice`.
3. Create a way to pass variables to translations. Right now the way we do untranslated dynamic copy is usually something like `You are about to send ${amount}  to ${address}`, and for translated stuff, we usually just tack it on to the end. What would work 100x better for localization is if we allow tokens that get replaced in translations. For instance, something like `You’re sending $amount to $address" so that other languages that have different grammar can re-arrange those tokens as needed. It could be called like
  ```
  translate('SEND_summary', {
    amount: tx.amount,
    address: tx.to
  });
  ```
4. Allow for some kind of configurable way to highlight translated copy. This should help with visual auditing to make sure everything is wrapped. This probably doesn't need to be fancy, it could just be a dev-only query param.

james-prado:
Those are all good suggestions. All the translations will be cleaned up, renamed (for those that need it), old strings removed. Adding tokens to translation strings is a great idea, I'll update the translate functions and any existing strings to use string literals for variables like amounts or addresses

I'm not sure if there's anyway I can properly highlight every translated string. If we were using just `translate` everywhere, a class could be appended onto the parent text's parent `<span />`, but then we'd have a bunch of extra `<span />`'s everywhere

wbobeirne:
We could choose to only wrap with spans when translation is enabled. For translateRaw, we could probably add some kind of unique character to the translated text to show it's been translated, like everyone's favorite unicode snowman `☃`

That could totally be a future PR though.

james-prado:
naw, I think it would be helpful to do it in this pr. I'll probably just use emoji's though, more fun. Something like 🎉

PR #1323, opened 2018-03-14
Translation Updates

james-prado:
Closes https://github.com/MyCryptoHQ/MyCrypto/issues/1285

#### Description

Move raw text to translation json files. Update translation functions. Remove old translation strings.

#### Changes

- Update `translate()` and `translateRaw()` functions
	- nows accept variables in translation strings
	- variables can be placed into any translation string, and update before being rendered as markdown or raw text
- Move all* strings of text to `translation/*.json` files
	- Some names and technical terms have purposely not been translated. MyCrypto, Ledger, Infura, etc... are specific names and don't need to be translated. ABI, bytes32, uint256, etc... are technical terms that will be **very** difficult to translate properly (maybe even impossible) without professional translation
- Replace all raw text with new translation functions
- Get selected language from localstorage instead of redux 
	- some strings are translated before the redux store has actually initialized. Trying to access the redux store before this will throw a vague error msg and stop everything from rendering. To avoid this, I get the selected language from localstorage, the same value that gets put into the rehydrated redux state. 

#### Testing 

Go to https://mycryptobuilds.com/39ebd4f226963999582759e3166751c7528b5065

I've added emoji's to the end of translated and non-translated strings to visually spot raw text strings. 

🎉 = translated 
☠ = not translated

This method does produces false negatives when putting a variable or link in a string to be translated, but it's the best method I've got for finding untranslated strings. 

Strings with ☠ at the end, or no emoji at all definitely aren't translated. Just let me know if you find one and where the string is, I'll add it into the translation json files.


common/components/Footer/index.tsx
```diff
@@ -127,6 +129,23 @@ export default class Footer extends React.PureComponent<Props, State> {
     isDisclaimerOpen: false
   };
 
+  public componentDidMount() {
```
dternyak: Was this a debug tool?

james-prado: yes

common/components/Footer/index.tsx
```diff
@@ -165,30 +184,27 @@ export default class Footer extends React.PureComponent<Props, State> {
...
-                <a onClick={this.toggleModal}>Disclaimer</a>
+                <a onClick={this.toggleModal}>{translateRaw('DISCLAIMER')}</a>
               </div>
               <div className="Footer-about-legal-text">v{VERSION}</div>
             </div>
           </div>
 
           <div className="Footer-support Footer-section">
-            <h5 className="Footer-support-title">Support Us & Our Friends</h5>
+            <h5 className="Footer-support-title">{translateRaw('FOOTER_AFFILIAT_TITLE')}</h5>
```
wbobeirne: *AFFILIATE

common/components/Footer/index.tsx
```diff
@@ -198,12 +214,16 @@ export default class Footer extends React.PureComponent<Props, State> {
             </div>
 
             <div className="Footer-support-donate">
-              <div className="Footer-support-donate-currency">Donate ETH</div>
+              <div className="Footer-support-donate-currency">
+                {translateRaw('DONATE_CURRENCY') + ' ETH'}
```
wbobeirne: This is a case for variables, some languages may put the subject (ETH) before the verb (Donate)

common/components/GenerateKeystoreModal/index.tsx
```diff
@@ -91,7 +91,7 @@ export default class GenerateKeystoreModal extends React.Component<Props, State>
               className="GenKeystore-button btn btn-primary btn-block"
               disabled={!isPrivateKeyValid || !isPasswordValid}
             >
-              {translate('Generate Keystore File')}
+              {translateRaw('Generate Keystore File')}
```
wbobeirne: These should be translation keys, not raw text.

common/components/GenerateKeystoreModal/index.tsx
```diff
@@ -108,7 +108,7 @@ export default class GenerateKeystoreModal extends React.Component<Props, State>
               aria-describedby={translateRaw('x_KeystoreDesc')}
               download={keystoreFile.filename}
             >
-              {translate('Download Keystore File')}
+              {translateRaw('Download Keystore File')}
```
wbobeirne: Should be translation keys, not raw text.

common/components/TXMetaDataPanel/TXMetaDataPanel.tsx
```diff
@@ -114,7 +114,7 @@ class TXMetaDataPanel extends React.Component<Props, State> {
             <div className="help-block">
               <a className="Gas-toggle" onClick={this.toggleAdvanced}>
                 {showAdvanced
-                  ? `- ${translateRaw('Back to simple')}`
+                  ? `- ${translateRaw('ACTION_4')}`
                   : `+ ${translateRaw('Advanced Settings')}`}
```
wbobeirne: Should be a translation key and not raw text.

common/components/TXMetaDataPanel/components/AdvancedGas.tsx
```diff
@@ -74,7 +74,7 @@ class AdvancedGas extends React.Component<Props, State> {
             <div className="input-group-wrapper AdvancedGas-gas-price">
               <label className="input-group">
                 <div className="input-group-header">
-                  {translate('OFFLINE_Step2_Label_3')} (gwei)
+                  {translateRaw('OFFLINE_Step2_Label_3')} (gwei)
```
wbobeirne: I'd like to see a better key for this one, the `OFFLINE_stepX` keys are hard to parse at a glance.

common/components/TXMetaDataPanel/components/SimpleGas.tsx
```diff
@@ -108,8 +108,8 @@ class SimpleGas extends React.Component<Props> {
               disabled={isGasEstimating}
             />
             <div className="SimpleGas-slider-labels">
-              <span>{translate('Cheap')}</span>
-              <span>{translate('Fast')}</span>
+              <span>{translateRaw('Cheap')}</span>
+              <span>{translateRaw('Fast')}</span>
```
wbobeirne: Should be keys and not raw strings.

common/components/WalletDecrypt/WalletDecrypt.tsx
```diff
@@ -133,311 +135,319 @@ const SECURE_WALLETS = Object.values(SecureWalletName);
...
-      helpLink: 'https://doc.satoshilabs.com/trezor-apps/mew.html'
-    },
-    [InsecureWalletName.KEYSTORE_FILE]: {
-      lid: 'x_Keystore2',
-      example: 'UTC--2017-12-15T17-35-22.547Z--6be6e49e82425a5aa56396db03512f2cc10e95e8',
-      component: KeystoreDecrypt,
-      initialParams: {
-        file: '',
-        password: ''
+const WalletDecrypt = withRouter<Props>(
```
wbobeirne: You can spare yourself the indentation by moving the withRouter call to the export below, ala connect.

common/components/WalletDecrypt/components/InsecureWalletWarning.tsx
```diff
@@ -61,38 +62,31 @@ export class InsecureWalletWarning extends React.Component<Props, State> {
...
+              var_run_local_article: knowledgeBaseURL + HELP_ARTICLE.RUNNING_LOCALLY
+            })}
           </li>
           <li>
-            Reading{' '}
-            <HelpLink article={HELP_ARTICLE.SECURING_YOUR_ETH}>
-              How to Protect Yourself and Your Funds
-            </HelpLink>
+            {translateMarkdown('INSECURE_WALLET_RECOMMEND_3', {
+              var_secure_your_eth_article: knowledgeBaseURL + HELP_ARTICLE.SECURING_YOUR_ETH
```
wbobeirne: I've noticed most of these translation variables are prefixed with `var_`, is there any reason for that?

common/components/WalletDecrypt/components/Mnemonic.tsx
```diff
@@ -78,7 +78,7 @@ class MnemonicDecryptClass extends PureComponent<Props, State> {
               className="btn btn-primary btn-lg"
               disabled={!isValidMnemonic}
             >
-              {translate('Choose Address')}
+              {translateRaw('Choose Address')}
```
wbobeirne: Should be a key, not a raw string.

common/components/WalletDecrypt/components/Web3.tsx
```diff
@@ -11,14 +11,14 @@ export const Web3Decrypt: React.SFC<Props> = ({ onUnlock }) => (
...
-        {translate('ADD_MetaMask')}
+        {translateRaw('ADD_MetaMask')}
       </button>
     </div>
 
     <div>
       <NewTabLink
         className="Web3Decrypt-install btn btn-sm btn-default btn-block"
-        content={translate('Download MetaMask')}
+        content={translateRaw('Download MetaMask')}
```
wbobeirne: Should be a key, not a raw string. Might be a good idea to make this a variable'd translation so we could make it "Download $thing"

common/components/ui/UnlockHeader.tsx
```diff
@@ -44,7 +44,7 @@ export class UnlockHeader extends React.PureComponent<Props, State> {
             >
               <span>
                 <span className="hidden-xs UnlockHeader-open-text">
-                  {translate('Change Wallet')}
+                  {translateRaw('Change Wallet')}
```
wbobeirne: Should be a key, not raw text

common/config/data.tsx
```diff
@@ -48,7 +48,7 @@ export const gasPriceDefaults: GasPriceSetting = {
 
 export const MINIMUM_PASSWORD_LENGTH = 12;
 
-export const knowledgeBaseURL = 'https://support.mycrypto.com';
+export const knowledgeBaseURL = 'https://support.mycrypto.com/';
```
wbobeirne: Double check this one, you're going to end up with a _lot_ of double slashes in a lot of places, which will be an invalid URL. I'd advise leaving it as-is.

common/containers/Tabs/Contracts/components/Deploy.tsx
```diff
@@ -26,12 +26,12 @@ class DeployClass extends Component<DispatchProps> {
       <main className="Deploy Tab-content-pane" role="main">
         <button className="Deploy-field-reset btn btn-default btn-sm" onClick={this.changeWallet}>
           <i className="fa fa-refresh" />
-          {translate('Change Wallet')}
+          {translateRaw('Change Wallet')}
```
wbobeirne: Should be a key, not raw text

common/containers/Tabs/Contracts/index.tsx
```diff
@@ -18,11 +18,11 @@ interface Props {
 const tabs = [
   {
     path: 'interact',
-    name: translate('Interact')
+    name: translateRaw('Interact')
   },
   {
     path: 'deploy',
-    name: translate('Deploy')
+    name: translateRaw('Deploy')
```
wbobeirne: Should be keys, not raw text

common/containers/Tabs/GenerateWallet/components/Keystore/EnterPassword.tsx
```diff
@@ -108,11 +110,11 @@ export default class EnterPassword extends Component<Props, State> {
       const { password, passwordValidation } = this.state;
 
       if (password.length < MINIMUM_PASSWORD_LENGTH) {
-        feedback = `Password must be ${MINIMUM_PASSWORD_LENGTH}+ characters`;
+        feedback = translateRaw('INPUT_PASSWORD_PLACEHOLDER');
       } else if (passwordValidation && passwordValidation.feedback) {
-        feedback = `This password is not strong enough. ${passwordValidation.feedback.warning}.`;
+        feedback = translateRaw('WEAK_PASSWORD') + ' ' + passwordValidation.feedback.warning;
```
wbobeirne: We should potentially figure out a way to localize these password feedback warnings.

common/containers/Tabs/SendTransaction/components/RequestPayment.tsx
```diff
@@ -123,7 +123,7 @@ class RequestPayment extends React.Component<Props, {}> {
 
           {!!eip681String.length && (
             <div className="row form-group">
-              <label className="RequestPayment-title">{translate('Payment QR & Code')}</label>
+              <label className="RequestPayment-title">{translateRaw('Payment QR & Code')}</label>
```
wbobeirne: Should be a key, not raw text

common/containers/Tabs/SendTransaction/index.tsx
```diff
@@ -39,24 +39,24 @@ class SendTransaction extends React.Component<Props> {
...
       {
         path: 'send',
-        name: translate('NAV_SendEther'),
+        name: translateRaw('NAV_SendEther'),
         disabled: !!wallet && !!wallet.isReadOnly
       },
       {
         path: 'request',
-        name: translate('Request Payment'),
+        name: translateRaw('Request Payment'),
```
wbobeirne: Should be a key and not raw text

common/containers/Tabs/SendTransaction/index.tsx
```diff
@@ -39,24 +39,24 @@ class SendTransaction extends React.Component<Props> {
...
-        name: translate('NAV_ViewWallet')
+        name: translateRaw('NAV_ViewWallet')
       }
     ];
 
     return (
       <TabSection>
         <section className="Tab-content">
-          <UnlockHeader title={translate('Account')} showGenerateLink={true} />
+          <UnlockHeader title={translateRaw('Account')} showGenerateLink={true} />
```
wbobeirne: Should be a key and not raw text

common/containers/Tabs/SignAndVerifyMessage/components/SignMessage/index.tsx
```diff
@@ -52,20 +51,20 @@ export class SignMessage extends Component<Props, State> {
               onClick={this.changeWallet}
             >
               <i className="fa fa-refresh" />
-              {translate('Change Wallet')}
+              {translateRaw('Change Wallet')}
```
wbobeirne: Should be a key and not raw text

common/containers/Tabs/Swap/components/LiteSend/Fields.tsx
```diff
@@ -31,7 +31,7 @@ class FieldsClass extends Component<Props> {
               onClick={this.changeWallet}
             >
               <i className="fa fa-refresh" />
-              {translate('Change Wallet')}
+              {translateRaw('Change Wallet')}
```
wbobeirne: Should be a key and not raw text. Might also consider making the change wallet button a component if that makes any sense, I see it a whole lot.

common/utils/findandreplace.js
```diff
@@ -0,0 +1,648 @@
...
+          lastNode.parentNode.replaceChild(matchEndNode, lastNode);
+        });
+
+        return lastNode;
+      }
+    }
+  };
+
+  return exposed;
+});
```
wbobeirne: This looks a lot like a module. Should we be pulling in a dependency instead?

common/translations/lang/en.json
```diff
@@ -193,10 +222,12 @@
     "BULK_Label_1": "Number of Wallets To Generate ",
     "BULK_Label_2": "Generate Wallets ",
     "BULK_SuccessMsg": "Success! Your wallets have been generated. ",
-    "SEND_addr": "To Address ",
+    "SEND_addr": "To ",
```
wbobeirne: We should add a `SEND_addr_short` instead, I think we may want to keep this key around.

common/components/WalletDecrypt/WalletDecrypt.tsx
james-prado: I tried that but I wasn't able to get the types to work properly, not sure why :/ I'll have another look and see if I can clean it up

common/components/WalletDecrypt/components/InsecureWalletWarning.tsx
james-prado: just to denote that that represents a variable. Using `var_` as a prefix is just a preference.

common/utils/findandreplace.js
james-prado: It's just used for debugging. It helps adds the emoji's to strings. It'll be removed with the rest of the debugging stuff.

james-prado:
@wbobeirne fixed up most of the suggested changes.

I'd like to standardize the naming of translation keys but right now I can't think of any effective way to do it at scale.

I was able to make all the keys uppercase with some regex find and replace, and a little manual work, but we'll need to come up with a better method to update the rest of the strings

james-prado:
Addressing @wbobeirne comment on having a conditional return type for `translate()`, ie. 
```
function translate(markdown: true): React.Element<any>;
function translate(markdown: false): string;
```

I agree, ideally we shouldn’t have to know the content of the translation string to translate it properly and provide the correct return type, but I can’t find any better way to solve this issue. Returning `React.Element<any>` all the time doesn’t work because other components / module require strings as arguments, and returning strings all the time doesn’t work because then you can’t return proper markdown. There really isn’t any good solution to this other than changing the return type if you know you’re using a markdown string.

wbobeirne:
I think the solution for that issue was what we had before. `translate` is always a React element, `translateRaw` is always a string. If there's a component that was only accepting strings before, it should accept React elements as well if and only if it's rendering that prop in a React element friendly context. Otherwise, it's good that it only accepts a string because we should be using `translateRaw` in that instance.

```jsx
// If it goes in a <p>, it can _always_ be a react element.
<p>{translate('x_Description')}</p>

// Placeholders should _always_ be strings. If this were a prop,
// it should be typed to only accept strings.
<input placeholder={translateRaw('x_Placeholder')}/>
```

React elements for all translated text is the pattern that most libraries use, like [react-intl](https://github.com/yahoo/react-intl) and [react-i18n-next](https://github.com/i18next/react-i18next). They also typically provide an escape hatch string returning function for things like placeholder text.

I feel like if we're having difficulty coming to consensus, it might be time to use a third party library to solve our translation rather than trying to reinvent the wheel.

dternyak:
I've taken some time to review both approaches, and I have a few thoughts. 

It took me some time to try to determine the objective benefits and drawbacks at each approach, but I believe I can summarize them as follows. I may have missed a few items, so please feel free to add as needed:

### Single `translate` function
Benefits:
1. No need for multiple imports of `translate` and `translateRaw`
2. [I'm sure I'm missing something here, but I couldn't determine additional benefits]

Drawbacks:
1. Call signatures have to include `, {}, true` to support markdown
2. (subjective) Slightly less clear overall API. `translate` always returning a component and `translateRaw` always returning a string gives you a clear reason for choosing each, whereas a parameter signature determination of return value is less clear.

### Existing approach.
Benefits:
1. Clearer API (call signature is simpler, return value is simpler as described above)
2. Supporting markdown if the origin string changes requires no changes unless that string is being used as placeholders / string required properties. I think the existing API could be modified to include `forceString` as an optional argument defaulting to false if this becomes a significant hurdle.

Drawbacks:
1. Requires multiple imports of `translate` and `translateRaw`
1. [Please insert what I'm missing]

Overall, with what information I've been able to gather, it seems like the changes are not overall beneficial and in some cases are regressions. I would prefer to stick with the current approach as well, but I think there are some points I am probably missing, so it would be worth having @wbobeirne and @james-prado help updates the above breakdown to ensure we have all available information.

dternyak:
screen shot 2018-03-21 at 1 09 33 pm

Issue on the account view.

Issue #1228, opened 2018-03-02
Update Code Styles


dternyak:
Certain text boxes for code styles don't have the `inset` style after https://github.com/MyCryptoHQ/MyCrypto/pull/1169.

<img width="1199" alt="screen shot 2018-03-02 at 12 46 41 am" src="https://user-images.githubusercontent.com/7861465/36886688-35f82bb4-1db3-11e8-81c1-9cdfeeca129e.png">
<img width="386" alt="screen shot 2018-03-02 at 12 46 34 am" src="https://user-images.githubusercontent.com/7861465/36886689-3607e586-1db3-11e8-902a-2d654f417332.png">

- [x] Create styles for code textareas

PR #1333, opened 2018-03-16
Update code blocks & generate / send tx buttons

james-prado:
Closes https://github.com/MyCryptoHQ/MyCrypto/issues/1228, Closes https://github.com/MyCryptoHQ/MyCrypto/issues/1340

### Description

Update inline code and code block styles. Update generate & send tx buttons. Fix a few other minor issues.

### Changes

- Update inline code, and code block styles
	- All disabled `<textarea />`s are now proper `<Code />` blocks
- Small css tweaks
	- Adjust the height of the dropdown components
	- Add optional className prop to `<SendButton />` to fix misaligned send buttons
- Fix `<WalletDecrypt />` types
- Fix Modal animation
	- focusing the first element before the animation finishes breaks the start of the animation
- Make contract interact dropdowns clearable & searchable
- Generate tx button is now for offline only
- Send tx will generate a tx, and open the confirm tx modal when transaction is signed
- Confirmation modal now shows raw transaction (hex) and serialized transaction
- Fix confirmation modal address styles
	- now properly aligns addresses

### Screenshots

Inline Code
screen shot 2018-03-16 at 9 37 20 pm

Code Block
screen shot 2018-03-21 at 6 22 04 am

### Testing generate & send tx buttons

https://mycryptobuilds.com/627acc99bc90f0ba5e4b74ce8afff1884290de9f

1. Test sending a transaction (online and offline)
2. Test broadcasting a signed transaction
3. Test 'contracts/deploying' & 'contracts/intereacting'
	- heres some example bytecode to test with `0x6001600054016000556000600060006000600073bbbf5374fce5edbc8e2a8697c15331677e6ebf0b610401600054046001036127105a0302f16001556103e860005402600101600255`
4. Test lite send


dternyak:
Manual testing is 👍. The only issue I ran into is that it is non-obvious why an offline transaction is invalid. The nonce isn't set by default (since it can't be requested), but isn't highlighted in red / otherwise indicated to the user that it is invalid.

An issue will be created for:
- [ ] Amount, To, and Nonce field should have error validation for offline transaction generation

common/components/ui/Code.scss
```diff
@@ -1,14 +1,42 @@
-pre {
+@import 'common/sass/variables';
+@import 'common/sass/mixins';
+
+code {
```
wbobeirne: Since this is componentized, we should target classnames here. I wouldn't want someone relying on using the raw `<code/>` and `<pre/>` tags without going through the component.

james-prado:
Seperate issue for form validation [here](https://github.com/MyCryptoHQ/MyCrypto/issues/1361)

common/components/ui/Code.scss
```diff
@@ -1,14 +1,42 @@
...
-  border-radius: 5px;
+  border-radius: 2px;
+  &.wrap {
+    code {
+      white-space: normal;
+    }
+  }
   code {
+    display: block;
+    overflow: scroll;
```
wbobeirne: Prefer `overflow: auto;` to `overflow: scroll;`. Operating systems that don't have invisible scrollbars (e.g. Windows) will show scrollbars for both vertical and horizontal scrolling all of the time if you force scroll.

I also think it might look nicer to have the scrolling on the `pre` rather than `code`. Right now the scrollbar kind of floats in the middle of the pre.

common/containers/Tabs/SignAndVerifyMessage/components/SignMessage/index.tsx
```diff
@@ -11,6 +11,7 @@ import SignButton from './SignButton';
 import { isWalletFullyUnlocked } from 'selectors/wallet';
 import './index.scss';
 import { TextArea } from 'components/ui';
+import Code from 'components/ui/Code';
```
wbobeirne: I've seen this a bunch so just dropping a note, but you can just add `Code` to the bulk import of `components/ui` above instead of adding a new import each time. Just add the export to `components/ui/index.tsx`

james-prado: forgot to do that, I've just been having vscode auto import and hadn't noticed

PR #1337, opened 2018-03-18
Update blockie package in webpack config

skubakdj:


dternyak:
Thanks for the catch! Indeed a l77t PR.

PR #1349, opened 2018-03-20
Parity Signer

### LINKED ISSUE #1056: Integrate Parity Signer
### Disclaimer

This is kind of an extension of: https://github.com/kvhnuke/etherwallet/issues/551

Also I'm actually happy to sit down and implement it myself, but I'd like to hear if it's something you'd be willing to include in the first place.

Parity has developed a Parity Signer mobile app a while ago, it's available both on [iOS](https://itunes.apple.com/de/app/parity-signer/id1218174838?l=en&mt=8) and [Android](https://play.google.com/store/apps/details?id=com.nativesigner). The app can be a low entry barrier alternative to hardware wallets (especially if you have an old phone that you can turn completely offline and run the app on it).

The app in combination with the Parity client UI allows you to store the private key on your phone, recover wallet from Parity phrases (which solves the MEW issue linked above for us), and sign transactions done in the browser via QR codes - the raw transaction is presented as a QR code, scanned by the app, signed in the app, and then a QR code with the signed transaction can be scanned by the browser via webcam.

I reckon it should be reasonably easy to port the QR scanning code to MyCrypto, which would offer the users a more secure way of handling their private keys than using keystore files / phrases in the browser directly. It can also help with phishing since the app displays the transaction details from the QR code to the user (how much ether and to whom is going to be transferred, regardless of what the web UI displays).

Yay/nay?

maciejhirsz:
Closes #1056

### Description

Introduces Parity Signer as another secure wallet option on the main screen. Account is unlocked by scanning a QR code of the address from the mobile app. Signing transactions is done in a modal overlaying the UI that first displays the QR code of the RLP of the transaction and then scans for the signature QR code from the mobile app.

### Changes

* New secure wallet has been added.
* New modal has been added to allow signing of the transaction while the `Promise` is being resolved.
* New reducers and actions have been added in order to pipe through the transaction from the UI to the `Promise` required by the `IFullWallet` interface.
* Added `@parity/qr-signer` as a new dependency.

### Steps to Test

1. Download the Parity Signer mobile app ([iOS](https://itunes.apple.com/us/app/parity-signer/id1218174838) / [Android](https://play.google.com/store/apps/details?id=com.nativesigner)). Create an account on it, send some ETH to the account (works with testnets).
1. Click on the "Parity Signer" option on the main screen.
1. Scan the account QR code from Parity Signer to unlock the wallet.
1. Create any kind of a transaction.
1. Sign the transaction using the app.
1. Make sure that the transaction goes through and is mined on chain.

Due to interactions with the webcam, integration tests might be tricky unless a lot of the stuff gets mocked away.

### TODO / Known issues

* [x] Needs a proper icon on the home screen.
* [x] `ParitySigner.tsx` needs to be moved to the containers.
* [x] UI needs some love (things are off-center, some paddings would be nice, some copy explaining what it is with links to the app would be nice).
* [x] Parity Signer needs a help website (linking to README.md for now...).


dternyak:
@maciejhirsz Are there any plans to support testnets on the parity mobile signer? 

Additionally, we'll want to disable Parity when non-ethereum networks are selected (e.g. disable parity option when network is ETC)

maciejhirsz:
Testnets are already supported, or did you run into issues? The UI doesn't separate accounts by network, but the chainId is correctly handled when signing the transaction.

Signing messages should be possible with the Parity Signer, I haven't investigated that yet, the UI (on the mobile app) might be a bit dumb about it, so will disable the panel there for now. Also will look at disabling it for non-Ethereum networks.

dternyak:
Got it. I didn't attempt a testnet tx as I didn't see a way to select testnet on the app, but it's good to see that's supported.

👍 to message signing as a later functionality.

maciejhirsz:
@dternyak I've been looking into it, and I don't really see any issues why the signer shouldn't be able to handle ETC and other networks as well. Basically anything that can be handled by keystore file / privatekey currently, should work with the app, unless there are some differences in how those networks sign txs/messages that I'm unaware of?

maciejhirsz:
Updated the PR:
* All UI stuff is fixed, latest `develop` merged back.
* Disabled Parity Signer for signing transactions (subject to future PR).
* Removed the **[WIP]** :tada:.

common/components/Footer/index.tsx
```diff
@@ -219,6 +220,7 @@ export default class Footer extends React.PureComponent<Props, State> {
           </div>
         </footer>
 
+        <QrSignerModal />
```
wbobeirne: Would prefer this component only be rendered where it contextually makes sense, somewhere in the send form.

common/components/WalletDecrypt/disables.ts
```diff
@@ -22,9 +22,10 @@ export const DISABLE_WALLETS: { [key in WalletMode]: DisabledWallets } = {
     }
   },
   [WalletMode.UNABLE_TO_SIGN]: {
-    wallets: [SecureWalletName.TREZOR, MiscWalletName.VIEW_ONLY],
+    wallets: [SecureWalletName.TREZOR, SecureWalletName.PARITY_SIGNER, MiscWalletName.VIEW_ONLY],
     reasons: {
       [SecureWalletName.TREZOR]: 'This wallet can’t sign messages',
+      [SecureWalletName.PARITY_SIGNER]: 'This wallet can’t sign messages',
```
wbobeirne: Not your responsibility, but just a reminder that we should be translating this.

common/containers/ParitySignerDecrypt/index.scss
```diff
@@ -0,0 +1,14 @@
+.ParitySignerUnlock {
+    a img {
+        display: inline-block;
+        height: 3em;
+        margin: 0 0.4em;
+    }
+
+    .qr-bounds {
```
wbobeirne: We use [SuitCSS naming convention](https://github.com/MyCryptoHQ/MyCrypto/wiki/Style-Guides#classnames) to avoid class name conflicts. Likewise, the `a img` above would be better suited by class names.

common/containers/ParitySignerDecrypt/index.tsx
```diff
@@ -0,0 +1,67 @@
...
+          <QrSigner
+            size={300}
+            scan={true}
+            onScan={this.unlockAddress}
+            styles={{ display: 'inline-block' }}
+          />
+        </div>
+        <p>{translate('ADD_PARITY_2')}</p>
+        <p>
+          <a href="https://itunes.apple.com/us/app/parity-signer/id1218174838" rel="nofollow">
```
wbobeirne: Use the `NewTabLink` component here.

common/containers/ParitySignerDecrypt/index.tsx
```diff
@@ -0,0 +1,67 @@
...
+      <div className="ParitySignerUnlock">
+        <div className="qr-bounds">
+          <QrSigner
+            size={300}
+            scan={true}
+            onScan={this.unlockAddress}
+            styles={{ display: 'inline-block' }}
+          />
+        </div>
+        <p>{translate('ADD_PARITY_2')}</p>
```
wbobeirne: At some point we should put together a help article that explains this process, or have some additional descriptive copy here.

common/containers/ParitySignerDecrypt/index.tsx
```diff
@@ -0,0 +1,67 @@
...
+        </p>
+      </div>
+    );
+  }
+
+  private unlockAddress = (address: string) => {
+    if (!isValidETHAddress(address)) {
+      console.error('Invalid address!');
+      return;
+    }
```
wbobeirne: What's the user experience for when this happens? Does the signer keep watching for a potentially valid address?

common/containers/QrSignerModal/index.scss
```diff
@@ -0,0 +1,8 @@
+.QrSignerModal {
+  .qr-bounds {
```
wbobeirne: Prefer SuitCSS naming convention.

common/containers/QrSignerModal/index.tsx
```diff
@@ -0,0 +1,120 @@
...
+
+    const { scan } = this.state;
+    const { tx, from } = this.props;
+
+    // Poor man's serialize without signature.
+    // All those values are later overriden by actual signature
+    // values in: wallets/non-deterministic/parity.ts
+    tx.v = Buffer.from([tx._chainId]);
+    tx.r = Buffer.from([0]);
+    tx.s = Buffer.from([0]);
```
wbobeirne: Looks like you're mutating the `tx` props object here. If you want to alter something passed in as props, you'll need to copy it and save it in state. Something like...

```.tsx
public componentDidMount() {
  if (this.props.tx) {
    this.makeStateTx(this.props);
  }
}

public componentWillReceiveProps(nextProps: Props) {
  if (nextProps.tx && this.props.tx !== nextProps.tx) {
    this.makeStateTx(nextProps);
  }
}

private makeStateTx(props: Props) {
  const tx = { ...props.tx };
  tx.v = Buffer.from([tx._chainId]);
  tx.r = Buffer.from([0]);
  tx.s = Buffer.from([0]);
  this.setState({ tx });
}
```

common/libs/wallet/non-deterministic/parity.ts
```diff
@@ -0,0 +1,59 @@
...
+        }
+
+        tx.r = sigBuf.slice(0, 32);
+        tx.s = sigBuf.slice(32, 64);
+        tx.v = Buffer.from([v]);
+
+        resolve(tx.serialize());
+      };
+
+      this.setWalletQrTransaction(tx, from, onSignature, onCancel);
```
wbobeirne: Is there a reason this is a callback argument in constructor, instead of just resolving the promise with these variables? Would be nice if the wallet was more agnostic to where it gets used, so that you could use the same wallet instance to sign transactions with two different implementations.

common/reducers/wallet.ts
```diff
@@ -30,9 +35,21 @@ export interface State {
...
+  tx: QrTransaction;
+  from: QrAddress;
+  onSignature(signature: QrSignature): void;
+  onCancel(): void;
+}
+
 export const INITIAL_STATE: State = {
   inst: null,
   config: null,
+  signViaQr: null,
```
wbobeirne: IMO (Would definitely like additional opinions) this should probably be its own reducer with its own set of actions / sagas. I think this is pretty separate from the other wallet processes, and is specific to only the parity signer.

dternyak: I definitely see maintainability improvements with such a separation of concerns 👍

common/containers/QrSignerModal/index.tsx
maciejhirsz: The values aren't read anywhere down the line, so it shouldn't really matter. Ideally there would be a way to serialize a transaction to RLP in ethereumjs without including the signature (`tx.serialize(false)` to match `tx.hash(false)`), but that's not the option. I can make a helper function for it here, and make a PR to ethereumjs.

wbobeirne: The problem is it makes for non-deterministic redux state, makes it so there's no way to determine if the tx object has changed (without checking deep equality) and [in general is just advised against in React](https://reactjs.org/docs/components-and-props.html#props-are-read-only).

If you want to avoid the state boilerplate, it could at least do
```
const tx = { ...this.props.tx };
```
in `render`.

common/libs/wallet/non-deterministic/parity.ts
maciejhirsz: Yeah, this is the part I'm least happy about, and it goes along with how `QrSignerModal` is done. The callbacks go up to the container and are in turn placed in redux state for the modal to use.

The component that instantiates the wallet (`ParitySignerDecrypt` container in this case) has no knowledge of the modal, and is unmounted after unlocking so it cannot host the `QrSignerModal` itself, hence I'm passing these things over redux, which works but also feels over-engineered for my tastes.

Ideally I'd be able to pass the callbacks from the wallet to the modal as props without going through all these gymnastics, but I haven't been able to figure out how to make it do it any cleaner so far (my understanding of the existing code base is still pretty shallow).

common/containers/ParitySignerDecrypt/index.tsx
maciejhirsz: Yep, it will keep on scanning until a QR code with a valid address is presented.

common/libs/wallet/non-deterministic/parity.ts
wbobeirne: We came up with a migration strategy to move these things to Redux via slack DMs. Those changes probably won't land for a little while though.

common/containers/ParitySignerDecrypt/index.tsx
maciejhirsz: We are working on a proper help page that describes how to use the signer, I'll add a link to it here once it's done (ideally before publicity but can be done after the PR is merged).

common/components/WalletDecrypt/WalletDecrypt.tsx
```diff
@@ -177,7 +177,7 @@ const WalletDecrypt = withRouter<Props>(
         component: ParitySignerDecrypt,
         initialParams: {},
         unlock: this.props.setWallet,
-        helpLink: 'https://github.com/paritytech/parity-signer/blob/master/README.md'
+        helpLink: 'https://wiki.parity.io/Parity-Signer-Mobile-App-MyCrypto-tutorial.md'
```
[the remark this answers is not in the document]
maciejhirsz: [Yup!](https://github.com/paritytech/wiki/pull/54)

PR #1350, opened 2018-03-20
Fixed offline broadcast link

### LINKED ISSUE #1336: Offline broadcast tx link is incorrect
### Description of the Issue

When you generate a transaction offline, it presents you with a link to the broadcast page. The link goes to `/account/pushTx` though, not `/pushTx`

### Steps to Reproduce

1. Go offline
2. Generate transaction
3. Click "Broadcast Transaction" at the bottom
4. End up on 404 page

hahnmichaelf:
Closes #1336

### Changes

Changed link to broadcast page that pops up when user is offline to be `/pushTx` instead of `/account/pushTx`

### Steps to Test

1. Access your address through the 'account view and send' tab. 
2. Go offline
3. Submit information for new transaction
4. URL below 'raw transaction' and 'sign transaction' boxes in the following string: "To broadcast this transaction, paste the above into the 'Broadcast Transaction tab'".


common/components/SendButtonFactory/OfflineBroadcast.tsx
```diff
@@ -20,7 +20,7 @@ export const OfflineBroadcast = connect((state: AppState) => ({ offline: getOffl
 const BroadCast: React.SFC<{}> = () => (
   <p>
     To broadcast this transaction, paste the above into the{' '}
-    <Link to="pushTx">Broadcast Transaction tab</Link> or{' '}
+    <Link to="../pushTx">Broadcast Transaction tab</Link> or{' '}
```
wbobeirne: Don't do `..`, we just want it to be `/pushTx` so that it goes to the root. `..` wouldn't work if it were nested 2 levels deep, e.g. `/account/send/confirm` or something like that.

hahnmichaelf: It didn't occur to me that it would need to be referenced from anywhere else. Thanks 💃

PR #1351, opened 2018-03-20
Custom DPaths, Improvements, and Fix SingularDTV

### LINKED ISSUE #1311: Deterministic wallet dPath improvements
A couple of issues that have come in from user feedback:

* [ ] Show what the actual dPath value is somewhere. I think it's unclear to users what this select is, and which each path is.
* [ ] Re-add the ability to set custom dPaths. Ideally this value could be saved to local storage for future use. I think we don't have to add full on saving like tokens and nodes, just save the value between page loads.

Custom dPaths are on the live site right now, so I'm considering this one a blocker since we shouldn't ship with fewer features.

wbobeirne:
Closes #1311.

### Description / Changes

* Re-enables the ability to set a custom dpath.
* Shows the dpath of the hard-coded options in lighter text next the the label.
* Loosens up the dPath regex to allow for SingularDTV wallets. This is not BIP-44 compliant, but maintains functionality from the current prod site.

### Steps to Test

1. Run the app on `dpath-improvements` branch
2. Enter a mnemonic phrase, trezor, or ledger
3. Confirm that dpath dropdown shows dpaths as options
4. Select the "Custom" option from the dpath dropdown
5. Enter a custom dpath, confirm the addresses look correct (This is easily done by just punching in one of the hardcoded ones.)
6. Switch to SingularDTV, confirm that works.

### Screenshots

screen shot 2018-03-20 at 12 56 25 pm

screen shot 2018-03-20 at 1 27 06 pm

screen shot 2018-03-20 at 1 21 03 pm

Just as a note, the reason there's a button and it doesn't just auto-refresh when a valid path is entered is because hardware wallets require you to confirm on-device. This would be annoying for typing it out manually, as your device would prompt you with every keystroke.


james-prado:
Looks good 👍

PR #1355, opened 2018-03-20
Update keystore visual indicator

### LINKED ISSUE #1260: Update Keystore Password UI Behaviour
Right now on beta, the keystore field looks unresponsive because we're not giving any visual queues when the user picks a keystore. In an old build, the password box would appear after we attempted to unlock a keystore with a blank password (which that a password was necessary). I'm not sure why we removed that, but I think it would be ideal, or at least some other visual queue.

hahnmichaelf:
Closes #1260 

### Description

Updates the keystore access method to indicate that a keystore file has been input, by disabling the password field until a valid keystore has been selected.

### Changes

1. Makes the keystore access method have a default-disabled password field.
2. Enables the password field based on whether or not a keystore file is selected.

### Steps to Test

1. Go to `/account`
2. Select Keystore access method.
3. Fill out disclaimers/acknowledgments and click 'Continue'.
4. See that password field is disabled
start

5. Select option to select a wallet file and select the file. Password field is now enabled.
secondstep

6. Type in password and the unlock button becomes enabled.
stepthree


dternyak:
@hahnmichaelf Thanks for the PR. I don't think this fully addresses the issue, as we want an indication when the keystore is uploaded. Currently, there is no real difference in the UI between un-uploaded keystore and uploaded keystore but no password yet entered.

hahnmichaelf:
@dternyak I tried to add further visual indication by truncating the filename below "Select Wallet File" button.

how is this

dternyak:
@hahnmichaelf Solid approach! I think it might make sense to show the keystore file reversed and truncated, since the first bit of the keystore filename will be the same across all keystores (with the ending portion being unique).

PR #1386, opened 2018-03-26
Upgrade to Webpack 4

### LINKED ISSUE #1191: Upgrade to Webpack 4
Webpack 4 was [just released](https://medium.com/webpack/webpack-4-released-today-6cdb994702d4), and will require some config migration. This issue should contain most discussion of any problems that come up with doing so.

I think it would be ideal to work out any of the kinks here _before_ launch, as build bugs can be pretty hard to pin down, and this'll be a pain to do when we have considerably more users and devs.

wbobeirne:
Closes #1191 

### Description

Upgrade from Webpack 3 to Webpack 4, and update a whole lotta modules along the way. Build times aren't that different (See https://github.com/MyCryptoHQ/MyCrypto/issues/1191#issuecomment-376197713 for more.) However, the size savings are pretty sweet due to being able to use all uglifier optimizations, and some other optimizations that came with it.

| File   | Old    | New    |
|--------|--------|--------|
| Client File(s) | 6.79mb | 4.59 MiB | 
| Vendor File(s) | 2.28mb | 1.48 MiB |

Note that I said file(s), because webpack has split out client and vendor JS into multiple files using its voodoo magic. I'm not 100% sure how all of that works, but it doesn't seem to have an effect on DOMContentLoaded happening any faster.

Also of note, one of Webpack 4's optimizations had to be disabled due to our dependency graph being a little messed up (See again link above), but I think we got the bulk of its savings.

### Changes

* Updated more or less all modules with `webpack`, `loader`, or `plugin` in the name
* Updated TypeScript to `v2.7.2`
    * This resulted in the need for a few code changes, mostly just typings
* Removed a bunch of unused modules from package.json
* Use webpack's default minifier instead of `babel-minify-webpack-plugin`.
    * Looks like the latest version of Webpack has no issues with mangling and all that

### Steps to Test

1. Run app in dev, make sure it works
2. Build prod version, make sure it works
3. Build downloadable version, make sure it works
4. Build electron version, make sure it works
5. Build freezer, make sure it works


dternyak:
There were a few (maybe?) issues I noticed while testing.

- Electron build did not contain icons (this is likely stemming from the actual electron integration, not the changes here)
- I'm no longer getting a hash when running the `downloadable` version, although mycryptobuilds.com did host a `HashRouter` based build, so it worked for it.

Issue #1414, opened 2018-03-29
gastracker address url is error


VieYang:
### Description of the Issue

ETC Transaction History Address URL is error.
MyCrypto show me the url https://gastracker.io/address/0x4bbeEB066eD09B7AEd07bF39EEe0460DFa261520,  but https://gastracker.io/addr/0x4bbeEB066eD09B7AEd07bF39EEe0460DFa261520 is correct

dternyak:
Thanks for the report! We'll be sure to get this patched up in the next release.

PR #1429, opened 2018-04-02
Fix gastracker url path

dternyak:
Closes https://github.com/MyCryptoHQ/MyCrypto/issues/1414


common/reducers/config/networks/staticNetworks.ts
```diff
@@ -23,12 +23,16 @@ export type State = { [key in StaticNetworkIds]: StaticNetworkConfig };
 // Must be a website that follows the ethplorer convention of /tx/[hash] and
 // address/[address] to generate the correct functions.
 // TODO: put this in utils / libs
-export function makeExplorer(name: string, origin: string): BlockExplorerConfig {
+export function makeExplorer(
+  name: string,
+  origin: string,
+  addressPath: string = 'address'
```
wbobeirne: I'd like to see this become a config, rather than a bunch of misc. args, but I'd say that can come the next time we need to make an explorer more configurable.

PR #1445, opened 2018-04-04
Deterministic Webpack 4

dternyak:
This PR re-introduces the changes made in https://github.com/MyCryptoHQ/MyCrypto/pull/1386, but also contains changes that ensure that builds continue to build deterministically.


webpack_config/makeConfig.js
```diff
@@ -28,6 +28,14 @@ const DEFAULT_OPTIONS = {
...
   const options = Object.assign({}, DEFAULT_OPTIONS, opts);
   const isDownloadable = options.isHTMLBuild || options.isElectronBuild;
+  const clientHash = hashFiles.sync({ files: __dirname + '/../common/**/*' });
+  console.log('clientHash', clientHash);
+  const vendorHash = hashFiles.sync({
+    files: __dirname + '/../node_modules/**/index.js'
+  });
+  console.log('vendorHash', vendorHash);
+  const cssHash = hashFiles.sync({ files: __dirname + '/../common/*.(css|scss)' });
+  console.log('cssHash', cssHash);
```
wbobeirne: Console logs

webpack_config/makeConfig.js
```diff
@@ -278,7 +282,7 @@ module.exports = function(opts = {}) {
   // ====================
   const output = {
     path: path.resolve(config.path.output, options.outputDir),
-    filename: options.isProduction ? '[name].[chunkhash:8].js' : '[name].js',
+    filename: options.isProduction ? `[name].${clientHash}.js` : '[name].js',
```
wbobeirne: This will apply clientHash to both client and vendor JS, so it will be the same name even if vendor changes (and client does not.)

wbobeirne:
Alright, did some digging on this one and it turns out it's `FaviconsWebpackPlugin` that's causing the uncertainty in the build. I think the sort cache plugin we had made before was fixing this, but maybe the hashing has changed that it hashes _before_ we do the sorting. Either way, I'll try to have a PR up to sort this out.

PR #1466, opened 2018-04-06
No Private Keys Online

### LINKED ISSUE #557: Discourage Private Keys / Keystore

wbobeirne:
### WHY?

Over the past few months, you've probably noticed warnings and notices when using wallet formats marked as unsafe. 

Between hardware wallets like Ledger and TREZOR, applications like MetaMask and Parity Signer, and now the MyCrypto Desktop application, we feel that there are sufficient and low-friction alternatives to direct private key access.

Although our upcoming production release of MyCrypto Desktop will support Keystore, Mnemonic, and Private Key formats, we continue to suggest users switch to Hardware Wallets, MetaMask, and Parity Signer. 

Closes #557

### Description

It's finally here, no more private keys through the site. This both disables generating and unlocking private keys when accessing through the website (and, currently, in dev. I may add an escape hatch for that.)

When unlocking, users are presented the same view, but prompted to download when they select an insecure wallet.

When creating a new wallet, users are presented alternatives with hardware wallets, metamask, and the parity signer (#1349). They are also given the option to download the app. If they re-visit this screen on the app, the metamask option will be replaced with the option to create new wallets. They'll then see the old view with keystore file and mnemonic options.

Note that when testing, mycryptobuilds.com runs the downloadable HTML version of the site and therefore will allow you to use private keys, thinking it's running on your machine.

### Changes

* Rework InsecureWalletWarning to be a total blocker, not a step before unlocking.
* Rework WalletTypes to present alternatives, and have an option for showing generateable wallets.
* Temporarily add in some parity signer assets and translations from #1349. Will resolve conflicts when that PR is merged.

#### No Unlocking

1. Run the production or developer versions of the app (Not mycrypto build)
2. Attempt to unlock a private key, mnemonic phrase, and keystore file
3. Confirm you are prompted to download for each
4. Run the electron app
5. Confirm you are able to use those 3 without warning

#### No Generating

1. Run the production or developer versions of the app (Not mycrypto build)
2. Go to "Create New Wallet"
3. Confirm your options do not include generated wallets
4. Run the electron app
5. Go to "Create New Wallet"
6. Confirm you are presented the generatable wallets
7. Confirm they still correctly generate wallets

### Blocked Until the Following

* [x] https://github.com/MyCryptoHQ/download.mycrypto.com has to come online
* [x] Parity signer needs to be merged (#1349)
* [x] (Opinion) We need to reinstate the downloadable HTML release

^ To the above, I think that we should start posting the HTML downloadable versions of the site again. I think they're a great feature for people who don't want to grant us access to their computer via Electron, or want a portable build of the app to put on a USB drive. While I'm confident in the security of our Electron app, I understand people don't want to necessarily have something running natively on their machine. If we do add the HTML version back in, I'll probably want to update the download page to reflect that (Though still preferring Electron.)

### Create New Wallet - Web

new-generate-web

### Create New Wallet - Running Locally

new-generate-electron

### Insecure Wallet - Web

screen shot 2018-04-06 at 1 56 13 pm

I'm not wild about the design here, I think some kind of illustration or some helpful links could go a long way to making this more friendly and less annoying to users.


dternyak:
While reviewing, I noticed that it could be difficult to differentiate the "Create Wallet" options between desktop and web.

In order to make this more immediately clear to users, as well as provide users more guidance about recommended wallet options, I'd like to recommend that we add a color code to each wallet creation option.

wbobeirne:
Added a splash of color, hope this differentiates enough without overpowering the view. Also added some color to the warning on generating a wallet:

screen shot 2018-04-12 at 6 32 47 pm

screen shot 2018-04-12 at 6 32 27 pm

jspence425:
Is the text under MetaMask applicable to MetaMask only or all Web3? 

What text will appear if MetaMask is not installed, or if there are other web providers in the future re: #1700

wbobeirne:
This is a suggestion for creating a new wallet, not for unlocking one, so it wouldn't be dynamic based on the provider. Also, if they're already using a different provider, they probably don't want a new wallet.

jspence425:
For now it makes sense to only say MetaMask, but is there a way to make it dynamic? I think it would be beneficial to develop as such. The only real upcoming use case is with Blue's wallet but I imagine more and more will be implemented in the future and it would be beneficial (in the future) to not recommend one over another.

wbobeirne:
Updated with latest develop. Removed the deprecation warnings (Since only users with downloadable builds will see those now, they don't need the warning.) Also added an override button while in dev mode, in case you need to test insecure wallets and don't want to build / run `dev:electron`.

PR #1498, opened 2018-04-12
Update ETH Token List

dternyak:
This PR uses the script created by @wbobeirne in https://github.com/MyCryptoHQ/MyCrypto/pull/1247 to generate an up-to-date list of default tokens for the ETH network. The [ethereum-lists](https://github.com/ethereum-lists) repo includes duplicate symbols in it's list, so I have manually edited conflicting tokens and appended (x) to their symbol name as needed.

It also adjusts the "Tokens JSON" spec to gather all errors instead of immediately existing by throwing on the first error.


dternyak:
Example of the new test output: 

screen shot 2018-04-11 at 10 41 30 pm

wbobeirne:
@dternyak check out #1508, maybe we should go that route instead?

dternyak:
@wbobeirne Yup! I'll update this PR with the output of that script, and then merge.

PR #1501, opened 2018-04-12
Mnemonic Redesign

### LINKED ISSUE #1279: Mnemonic UI Redesign
### Why?
Our current practice for verifying that the user has copied the correct mnemonic includes re-typing the mnemonic in its entirety. This leaves users open to keyloggers that may copy the mnemonic, and is more time intensive than alternative verification methods.

A common practice for verifying mnemonic (e.g. Decred, MyMonero) includes a drag and drop (gif below). A redesign of our verification and switching to a drag and drop approach will benefit users by making verification simpler and more secure.

- [ ] Drag and drop word phrases in order instead of re-typing.

ConnorBryan:
Closes #1279

### Description

Daniel T. noted the inherent insecurity of typing the the mnemonic words after revelation, and suggested a move to a shuffled, click-based system to avoid keylogging.

### Changes

*Before*: User generates a new wallet via mnemonic phrase. In the confirmation stage, each word is entered via keyboard in order prior to proceeding.

*Now*: User generates a new wallet via mnemonic phrase. In the confirmation stage, the words are shuffled and turned into buttons, and the user must click them in the correct order.

### Steps to Test

1. From the home page, click "Create New Wallet".
2. Select the option for "Generate a Mnemonic Phrase".
3. Select the option for 'Confirm Phrase'.
4. Select the correct first word. It should flash green, become disabled, and show an index representing its place.
5. Select an incorrect next word. It should flash red, indicating that it was not the correct choice.
6. Press the 'Reveal Next Word' button. The word that is next in sequence should flash green, indicating it is the next valid choice.
7. Select all words and ensure the created wallet is accessible.

### Screenshots

_Fresh screen_
fresh-confirming

_Flashing an error on incorrect click_
flashing-error

_Flashing when being revealed_
being-revealed

_Successful selection_
screen shot 2018-04-12 at 2 26 36 pm


common/containers/Tabs/GenerateWallet/components/Mnemonic/Mnemonic.tsx
```diff
@@ -129,18 +125,89 @@ export default class GenerateMnemonic extends React.Component<{}, State> {
...
+    }
+  };
+
+  private getWordConfirmed = (word: string) => this.state.confirmValues.includes(word);
 
   private skip = () => {
     this.setState({ isConfirmed: true });
   };
+
+  private getShuffledWords = (words: string[]) => {
```
dternyak: would prefer to move this out to `utils`

ConnorBryan: @dternyak 
I can do that, do you anticipate it being used elsewhere? Seems fairly specific to this component. If I move it to `/utils`, should it gets its own file, or go in the `helpers` file (which seems to be mainly for BLL`

dternyak: I don't have a strong opinion on this, and I do agree it's likely this won't be used elsewhere. That said, moving it out to `utils` and then adding a test for it seems valuable to me.

dternyak:
Some of the styling could use a bit of improvement.

I would like to see each word option have an equal width, and keep the size identical before and after selection. Furthermore, the numbering isn't centered with respect to the word option.

common/containers/Tabs/GenerateWallet/components/Mnemonic/Mnemonic.tsx
```diff
@@ -113,7 +104,13 @@ export default class GenerateMnemonic extends React.Component<{}, State> {
     if (this.state.isConfirming) {
       this.setState({ isConfirmed: true });
     } else {
-      this.setState({ isConfirming: true });
+      const shuffledWords = shuffle([...this.state.words]);
```
wbobeirne: Shuffle already creates a new array, no need to `...`

common/utils/shuffle.ts
```diff
@@ -0,0 +1,17 @@
...
+  while (finalArray.toString() === array.toString()) {
+    // Fisher-Yates shuffle
+    for (let i = finalArray.length - 1; i > 0; i--) {
+      const j = Math.floor(Math.random() * (i + 1));
+      [finalArray[i], finalArray[j]] = [finalArray[j], finalArray[i]];
+    }
+  }
+
+  return finalArray;
+}
```
wbobeirne: We already use lodash in a few places, so might just be easier to `import shuffle from 'lodash/shuffle';`

https://lodash.com/docs#shuffle

common/containers/Tabs/GenerateWallet/components/Mnemonic/Mnemonic.tsx
ConnorBryan: Right, forgot I moved that to the function.

common/utils/shuffle.ts
ConnorBryan: @wbobeirne 
Excellent, I removed shuffle and replaced with `lodash`.

dternyak:
Excellent job @ConnorBryan! I appreciate the final tweaks 👍

PR #1505, opened 2018-04-12
Electron Redesign

wbobeirne:
Not tracked as an issue.

### Description

This PR aims to make our electron app feel more like just that, an app. It moves the navigation off to the side, and gets rid of the header and footer. There was also a bit of refactoring that went into this to make certain components more agnostic about how they're used.

The PR looks a lot bigger than it really is, everything here is just styling and component changes, no changes to redux / libs / utils.

### Changes

* Electron app now has a dope sidebar that handles navigation, node switching, and language switching.
* OSX app no longer has a frame
    * This was a little too tricky to implement and test on Windows and Linux in the short time I had, but I plan to revisit those platforms at a later date.
* Added a "Support Us" page only on the electron app that has all of the information in the footer.
* Modals now use [React portals](https://reactjs.org/docs/portals.html), so that they always render correctly
    * This allows elements that have opacity or overflow hidden to render modals with no problem. This also makes it so modals are unaffected by the styling of the component they're contained in.
* `TabSection` now uses either `WebTemplate` or `ElectronTemplate` depending on environment
* Navigation links and footer links have been moved into `config/navigation.ts` and `config/links.ts` for reusability
* Exposes `$is-electron` variable to sass.
    * This is primarily to make breakpoints happen at larger sizes, due to the sidebar
* Added an external link icon to "Help" in the navigation

### Steps to Test

1. Run the electron app in dev, provide feedback
2. Run the web app, make sure nothing changed
3. Try to build the electron app for your platform, make sure it's not busted

Some areas that I'm either concerned about, or would like feedback on:

* Not wild about the language / network buttons on the frame, but couldn't come up with anything better
* Modals currently cover everything, including the nav. Does that look weird?
* The navigation text is quite small to keep footprint low, does this render alright on all operating systems?

### OSX, Homepage

screen shot 2018-04-12 at 5 46 00 pm

### OSX, Swap, Changing Networks

screen shot 2018-04-12 at 5 48 37 pm

More platform screenshots forthcoming


dternyak:
🎉 This is excellent!

FYI: This is the generate wallet view with `no-private-keys` merged.

screen shot 2018-04-12 at 4 53 22 pm

wbobeirne:
Yuck, that wrapping-by-default is pretty nasty. I wonder if that can be improved.

dternyak:
@wbobeirne It's actually not too bad imo. The layout is better than the current generate wallet view, so I wanted to show that it's an improvement 😁

wbobeirne:
Reminder: Replace step 1 with opening the app

screen shot 2018-04-12 at 6 27 52 pm

wbobeirne:
Added a beta banner
screen shot 2018-04-13 at 2 12 06 pm

common/Root.tsx
```diff
@@ -95,18 +98,41 @@ class RootClass extends Component<Props, State> {
...
+        </Provider>
+        <div id="ModalContainer" />
+      </React.Fragment>
     );
   }
+
+  private addBodyClasses() {
+    const classes = [];
+
+    console.log(process);
```
HenryNguyen5: got a log here

common/components/ElectronNav/NetworkStatus.tsx
```diff
@@ -0,0 +1,43 @@
...
+}
+
+const NetworkStatus: React.SFC<StateProps> = ({ isOffline, isChangingNode, network }) => {
+  let statusClass: string;
+  let statusText: string;
+  const $network = network.isCustom ? network.unit : network.name;
+
+  if (isChangingNode) {
+    statusClass = 'is-connecting';
+    statusText = 'NETWORK_STATUS_CONNECTING';
```
HenryNguyen5: Would prefer enums here

common/config/navigation.ts
```diff
@@ -0,0 +1,53 @@
...
+    to: '/pushTx'
+  },
+  {
+    name: 'NAV_SUPPORT_US',
+    to: '/support-us',
+    disabled: !process.env.BUILD_ELECTRON
+  },
+  {
+    name: 'NAV_HELP',
+    to: `${knowledgeBaseURL}`,
```
HenryNguyen5: dont need back ticks here (not sure if part of PR)

PR #1518, opened 2018-04-14
Add Ethereum Commonwealth ETC node

wbobeirne:
It's on MyEtherWallet and ClassicEtherWallet, and should improve stability for ETC users.


dternyak:
I can pretty reliably get https://mycryptobuilds.com/e22536a0d555f64c15761441bdf5167ee2777155/#/account/info to flash offline and back online after refreshing with ETC Commonwealth node selected.

@HenryNguyen5 Do you mind looking into this?

Issue #1117, opened 2018-02-16
Integrate Existing Helpers View into Beta


hahnmichaelf:
- [ ] Create React Implementation of [helpers](https://www.mycrypto.com/helpers.html); integrate into beta

https://beta.mycrypto.com/helpers.html leads to a blank page.

In the footer there is a link to the helpers page, but it leads to the MyCrypto production version's helpers page and when you navigate off that, you end up on the prod. version of MyCrypto.

dternyak:
We'll likely just link out to the helpers page once we move current production to classic.mycrypto.com

gitcoinbot:
__This issue now has a funding of 0.25 ETH (135.28 USD @ $541.12/ETH) attached to it.__

 * If you would like to work on this issue you can claim it [here](https://gitcoin.co/issue/MyCryptoHQ/MyCrypto/1117).
 * If you've completed this issue and want to claim the bounty you can do so [here](https://gitcoin.co/issue/MyCryptoHQ/MyCrypto/1117)
 * Questions? Get help on the <a href='https://gitcoin.co/slack'>Gitcoin Slack</a>
 * $3379.55 more Funded OSS Work Available at: https://gitcoin.co/explorer

tommycox:
I can get going on this right away! Will let you know if I have any questions once I get setup.

gitcoinbot:
__Work has been started on the 0.25 ETH (135.28 USD @ $541.12/ETH) funding by__: 
 1. [@tommycox](https://gitcoin.co/profile/tommycox) 

 __Please work together__ and coordinate delivery of the issue scope. Gitcoin doesn't know enough about everyones skillsets / free time to say who should work on what, but we trust that the community is smart and well-intentioned enough to work together.  As a general rule; if you start work first, youll be at the top of the above list ^^, and should have 'dibs' as long as you follow through. 

 On the above list? Please leave a comment to let the funder (@owocki) and the other parties involved what you're working, with respect to this issue and your plans to resolve it.  If you don't leave a comment, the funder may expire your submission at their discretion. 

 * Learn more [on the gitcoin issue page](https://gitcoin.co/issue/MyCryptoHQ/MyCrypto/1117)
 * Questions? Get help on the <a href='https://gitcoin.co/slack'>Gitcoin Slack</a>
 * $3379.55 more Funded OSS Work Available at: https://gitcoin.co/explorer

tommycox:
[...]

PR #1521, opened 2018-04-14
Helpers Page

oojr:
The input spacing and size is different in the new design causing the user to see fewer conversions without scrolling, there is also convertUnit/unitToUnit calculations for the different units in the angular version, I began to port that over. This is a work in progress but I wanted to update you guys on direction

screen shot 2018-04-14 at 12 27 00 pm


dternyak:
Hey @oojr - great start! 

How are things progressing on your end?

oojr:
@dternyak I am trying to figure out how to do ENS, I am not sure if I need to create new services I saw that the contract abi is already included in the project? Are there reducer actions I can leverage I saw some Interfaces that are close to what I need, how would you approach this?

dternyak:
Hey @oojr -- just left a comment for you on the initial bounty thread https://github.com/MyCryptoHQ/MyCrypto/issues/1117.

PR #1532, opened 2018-04-16
Add MUSIC Network

### LINKED ISSUE #1474: [Coin Req] Add MUSIC to MyCrypto
Musicoin recently had a hardfork which enables EIP 155 and that should ideally make it work with MyCrypto (ref. https://github.com/kvhnuke/etherwallet/issues/912 ) I can PR back with the code if the devs are okay with adding Musicoin back, so do let me know reg this. Thanks!

wbobeirne:
Closes #1474

### Description

Adds in the MUSIC network and a node for it. Would love a confirmation from @varunram and @frankurcrazy that this looks like it's configured alright. One thing of note was I found that the address I got from entering my mnemonic on the Music Wallet app wasn't the same as the one I got with the dpath provided.

### Steps to Test

1. Select MUSIC from network dropdown
2. Make sure it works?

### DIFF
--- a/common/config/dpaths.ts
+++ b/common/config/dpaths.ts
@@ -55,7 +55,7 @@ export const ELLA_DEFAULT: DPath = {
 
 export const MUSIC_DEFAULT: DPath = {
   label: 'Default (MUSIC)',
-  value: "m/44'/184'/0'/0"
+  value: "m/44'/60'/0'/0"
 };
 
 export const ETH_SINGULAR: DPath = {
--- a/common/config/dpaths.ts
+++ b/common/config/dpaths.ts
@@ -53,6 +53,11 @@ export const ELLA_DEFAULT: DPath = {
   value: "m/44'/163'/0'/0"
 };
 
+export const MUSIC_DEFAULT: DPath = {
+  label: 'Default (MUSIC)',
+  value: "m/44'/184'/0'/0"
+};
+
 export const ETH_SINGULAR: DPath = {
   label: 'SingularDTV',
   value: "m/0'/0'/0'"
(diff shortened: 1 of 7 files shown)


common/config/dpaths.ts
```diff
@@ -53,6 +53,11 @@ export const ELLA_DEFAULT: DPath = {
   value: "m/44'/163'/0'/0"
 };
 
+export const MUSIC_DEFAULT: DPath = {
+  label: 'Default (MUSIC)',
+  value: "m/44'/184'/0'/0"
```
Varunram: Apologies for the incorrect info, it is `m/44'/60'/0'/0`

Varunram:
Thanks for taking the effort out to make the PR. I can confirm that

1. Unlocking the wallet works
2. Tx sending works:https://explorer.musicoin.org/tx/0x116a260698d0121ea31a4cc61377cdad56ad6347d227db1aeb5b380777c9d4c7
3. Viewing addresses' history works
4. Signing messages works
5. Tx status works

I assume mnemonic generation would work fine after the update, will confirm regarding that one the update is made.

wbobeirne:
I've updated the dpath, you'll find the new build at http://mycryptobuilds.com/1a5b290fc27e03e71908f281c6071166f914e0c5 shortly (Takes a bit to build.) I also went ahead and enabled TREZOR unlocking, sine I was able to generate a transaction no problem on my TREZOR. Unfortunately, Ledger doesn't work as it won't sign with the correct chain ID.

Varunram:
@wbobeirne Awesome, addresses match on my side but I don't have a trezor, so can't test regarding that.

Varunram:
@wbobeirne Just checking back on the status of this PR, thanks!

dternyak:
Hey @Varunram. Thanks for your assistance on this PR.

We're planning to merge this, along with ETSC, after we get the Beta out to production and include it with the subsequent release which will also continue our network dropdown redesign.

Timeline will be end of May.

Just wanted to keep you in the loop!

PR #1603, opened 2018-04-18
Node Refactor

### LINKED ISSUE #1353: Shepherd Network Selection Dropdown
With shepherd integration, we'll want to redesign our node dropdown.

- [ ] Investigate new UX of selecting networks / nodes

### LINKED ISSUE #1455: Add "Experimental" section to node dropdown
Networks which are newer, or networks that our team doesn't have sufficient context on, should be separated from the mainstream networks we support through an "experimental" sub-section.

wbobeirne:
Closes #1353. Closes #1455.

### Description

~This is the first of potentially a few node refactor PRs that will make it easier to work with nodes & shepherd. This one cleans up some of the leftovers from the shepherd conversion, and reduces the number of places where we have to define nodes. It also reduces some of the reliance on implicit config naming.~

Additional PRs have been merged, see #1660, #1654, #1652 & #1640 for those. This PR contains them all, in addition to the below changes.

### Changes

* Consolidate static node definitions from being split across `libs/nodes/index.ts`, `reducers/config/nodes/staticNodes.ts`, and `types/node.d.ts` to all being located in `libs/nodes/configs.ts`
* Create `makeNodeName` and `makeAutoNodeName` functions to reduce chance of making inconsistent node names
    * This required renaming a few nodes, only for the alternative networks & testnets
* Shepherd auto configs are now automatically created for each network, instead of hard coded
    * Shepherd node configs also have an `isAuto` property, for future use
* Remove unused `lib` property from `StaticNodeConfig` and `CustomNodeConfig`

### Steps to Test

1. Connect to various nodes, ensure they work
2. Connect using web3, ensure it still works


common/components/CustomNodeModal/CustomNodeModal.tsx
```diff
@@ -329,7 +328,7 @@ class CustomNodeModal extends React.Component<Props, State> {
         : {})
     };
 
-    return { ...node, lib: shepherdProvider };
+    return { ...node };
```
HenryNguyen5: Can just return `node` here

common/reducers/config/nodes/staticNodes.ts
```diff
@@ -1,191 +1,36 @@
...
+import { StaticNetworkIds } from 'types/network';
+
+function makeStateFromNodeConfigs(prev: Partial<StaticNodesState>, network: StaticNetworkIds) {
+  // Auto network
+  prev[makeAutoNodeName(network)] = {
+    network,
+    isAuto: true,
+    isCustom: false,
+    service: 'AUTO',
+    estimateGas: NODE_CONFIGS[network].reduce((bool, node) => bool || node.estimateGas, false)
```
HenryNguyen5: Do we actually use this property anywhere? Seems like its more suited towards network configs

wbobeirne: Huh, I had assumed that had something to do with estimating gas limit and that some nodes didn't support it. But you're right, absolutely nothing is using it.

PR #1640, opened 2018-04-19
Node Refactor (Pt. 2 - Web Node Selector Rework)

wbobeirne:
Closes the web portion of #1353 & #1455.

### Description

This rebuilds the web network selector to be a totally custom component that emphasizes a network-first view of nodes. When opening the dropdown now, you'll be presented with each network as an option, instead of all of the nodes. Selecting a network will choose the auto config for that network. You can expand each network to be a list of nodes, and select individuals from there.

Includes some minor refactors of types, as well as some change in behavior for custom nodes.

### Changes

* Change the node selector from a ColorDropdown to a custom component called NetworkSelector (Gif below)
    * Presents networks as options, individual nodes hidden in collapsibles
    * Changes auto network displays to show `NETWORK (X Nodes)` instead of `NETWORK (AUTO)`
* Removed the ColorDropdown component
* Added changeNetworkIntent action that selects the auto node for a given network, or the first node if it has no auto (e.g. custom network)
    * My vision for the future is that this dropdown is _only_ network selection, and that you specify on an options page which nodes a given network should use. So `changeNode*` functions will disappear, and we'll only change networks, which will be configured independently.
    * In a future refactor PR, I'd like to make `node.name` be display-only, and instead us `node.id` for logic stuff. Right now we use `node.name` to identify static nodes, and `node.chainId` to identify custom nodes. This is a huge pain in the ass.
* Added an `id` attribute to `StaticNodeConfig`s that matches their key on the staticNodes reducer
* Fix `removeCustomNode` from resetting you to ETH network all the time

### Steps to Test

1. Open up the new network selector
2. Click on a network, confirm you are on the auto node of that network
3. Expand the dropdown of a network and select an individual node, confirm you are only connected to that node
4. Add a custom node, confirm it shows up in the correct network
5. Add a custom node with a custom network, confirm the network shows up with that node
6. Remove the custom node, confirm it no longer shows up in the menu

## Screenshots

network-selector


common/components/ui/DropdownShell.tsx
```diff
@@ -14,7 +14,7 @@ interface State {
   expanded: boolean;
 }
 
-export default class DropdownComponent extends PureComponent<Props, State> {
+export default class DropdownComponent extends Component<Props, State> {
```
wbobeirne: Since DropdownComponent takes in a function to render its content, it's not pure. The reference to the function may not change, but its return value could, which it will fail to re-render with since it won't rerun the render functions.

dternyak:
Haven't gotten a chance to review the implementation, but I did get this exception (and red error screen) when unlocking MetaMask with ETH - 4 nodes.

```
TypeError: Cannot read property 'name' of undefined
    at renderLabel (client.c6ae6a371f15440f11b064b1ca87fe30cc0ca563.js:1)
    at i.render (client.c6ae6a371f15440f11b064b1ca87fe30cc0ca563.js:1)
    at p (vendors~client~vendor.c6ae6a371f15440f11b064b1ca87fe30cc0ca563.js:95)
    at beginWork (vendors~client~vendor.c6ae6a371f15440f11b064b1ca87fe30cc0ca563.js:95)
    at i (vendors~client~vendor.c6ae6a371f15440f11b064b1ca87fe30cc0ca563.js:95)
    at a (vendors~client~vendor.c6ae6a371f15440f11b064b1ca87fe30cc0ca563.js:95)
    at M (vendors~client~vendor.c6ae6a371f15440f11b064b1ca87fe30cc0ca563.js:95)
    at v (vendors~client~vendor.c6ae6a371f15440f11b064b1ca87fe30cc0ca563.js:95)
    at b (vendors~client~vendor.c6ae6a371f15440f11b064b1ca87fe30cc0ca563.js:95)
    at _ (vendors~client~vendor.c6ae6a371f15440f11b064b1ca87fe30cc0ca563.js:95)
```

wbobeirne:
Thanks for the error message, turns our web3 nodes prefix network with `WEB3_` and that magic gets handled in the selector. I wasn't using the selector to get the current network, so network was null. I'm using the selector now.

common/actions/config/actionCreators.ts
```diff
@@ -69,6 +69,16 @@ export function changeNodeForce(payload: string): interfaces.ChangeNodeForceActi
   };
 }
 
+export type TChangeNetworkIntent = typeof changeNetworkIntent;
```
HenryNguyen5: I'd really rather stick to the `REQUESTED` `SUCCEEDED` `FAILED` convention here. When debugging node changing in the past, it does get confusing when you see `NODE_CHANGE` but that was actually just for the reducers, and `NODE_CHANGE_INTENT` is what the sagas acted on. Just seems more obvious with the latter 3 postfixes.

common/components/Header/components/NetworkSelector/NetworkOption.tsx
```diff
@@ -0,0 +1,73 @@
...
+              />
+            ))}
+          </div>
+        )}
+      </div>
+    );
+  }
+
+  private handleSelect = () => {
+    this.props.selectNetwork(this.props.network);
```
HenryNguyen5: You could instead make selectNetwork a HOF in the parent component to avoid making extra methods here (and actually you could convert this entire component to a HOF by doing so)

common/components/Header/components/NetworkSelector/NetworkOption.tsx
```diff
@@ -0,0 +1,73 @@
...
+  };
+
+  private handleToggleExpand = () => {
```
HenryNguyen5: Ditto

common/components/Header/components/NetworkSelector/NetworkSelector.tsx
```diff
@@ -0,0 +1,227 @@
...
+        {isShowingAltNetworks && options.alt}
+        <button className="NetworkSelector-add" onClick={this.openCustomNodeModal}>
+          <i className="fa fa-plus" />
+          {translate('NODE_ADD')}
+        </button>
+      </div>
+    );
+  };
+
+  private selectNetwork = (net: NetworkConfig) => {
```
HenryNguyen5: e.g. the function signature would be `(net: NetworkConfig) => () = > void)`. And then in the rendering function call `selectNetwork(allNetworks[netKey])`. That way you can just call it in `NetworkOption` without creating a wrapper function inside it

common/components/Header/components/NetworkSelector/NetworkSelector.tsx
```diff
@@ -0,0 +1,227 @@
...
+
+interface OwnProps {
+  openCustomNodeModal(): void;
+}
+
+interface StateProps {
+  nodeSelection: string;
+  node: NodeConfig;
+  network: NetworkConfig;
+  allNodes: { [key: string]: NodeConfig };
```
HenryNguyen5: We really should just make a generic for this, like https://github.com/MyCryptoHQ/shepherd/blob/master/src/types/index.d.ts#L41

common/components/Header/components/NetworkSelector/NetworkOption.tsx
```diff
@@ -0,0 +1,73 @@
...
+import NodeOption from './NodeOption';
+import './NetworkOption.scss';
+
+interface Props {
+  nodes: NodeConfig[];
+  network: NetworkConfig;
+  nodeSelection: string;
+  isNetworkSelected: boolean;
+  isExpanded: boolean;
+  selectNode(node: NodeConfig): void;
```
HenryNguyen5: Another cool thing, but not related: you can get the method types of a class by using accessor syntax so you dont have to re-write these. E.g ```ts
import {NetworkSelector /* Assuming you've exported the non-connected version of the class */}
interface Props {
 selectNode: NetworkSelector["selectNode"]
}
```

common/actions/config/actionCreators.ts
wbobeirne: Agreed, I'll bring this and `nodeChangeIntent` inline with that.

common/components/Header/components/NetworkSelector/NetworkOption.tsx
wbobeirne: Definitely makes refactoring easier, but it does introduce more cyclic dependencies, which eventually I'd like to get rid of.

HenryNguyen5: Yeah, only way to avoid that is by having the class implement an interface, but at that point its a little verbose

common/actions/config/actionCreators.ts
wbobeirne: Addressed in https://github.com/MyCryptoHQ/MyCrypto/pull/1652

PR #1654, opened 2018-04-20
Node Refactor (Pt. 3 - Node & Network Cleanup)

wbobeirne:
### Description

The next in a series of node refactors (Don't worry, I think this is the second to last one.) This one does some cleanup to make nodes and networks more consistent, and makes them a lot easier to manage between static and custom ones. It also adjusts some display, and prevents some errorous situations.

### Changes

* Networks now have an `id` field that works much like `node.id`. It's what should be used for anything logic-based, and is always the key to the in the reducer state object.
* Networks are now unique by chainId. You can't add a custom network whose chainId collides with another network, static or custom.
* Custom network and node actions have been simplified to pass object for add, id for remove
* Network names have been changed to their full name, e.g. `"ETH"` -> `"Ethereum"`
* Network selector now shows `(Testnet)` next to testnet networks.

### Steps to Test

1. Change networks and make sure it all works fine
2. Unlock a web3 wallet and make sure it switches to the web3 node
3. Add a custom node, confirm that works
4. Connect to the custom node, confirm that works
5. Remove a custom node, confirm it's gone
6. Add a custom node with a custom network, confirm that works
8. Connect to the custom node, confirm that works
7. Remove a custom node with a custom network, confirm that both the node and the network are gone

### No conflicting chainid

screen shot 2018-04-20 at 1 34 06 pm

### Testnet badges

screen shot 2018-04-20 at 4 35 43 pm


common/actions/config/actionTypes.ts
```diff
@@ -61,25 +61,25 @@ export interface ChangeNetworkRequestedAction {
...
   type: TypeKeys.CONFIG_ADD_CUSTOM_NODE;
-  payload: { id: string; config: CustomNodeConfig };
+  payload: CustomNodeConfig;
 }
 
 /*** Remove Custom Node ***/
 export interface RemoveCustomNodeAction {
   type: TypeKeys.CONFIG_REMOVE_CUSTOM_NODE;
-  payload: { id: string };
+  payload: string;
```
HenryNguyen5: Not sure how i feel about changing this to not be object based, it loses context of what you're supposed to pass in the action creator when you use it.

wbobeirne: Do you think typing this as `CustomNodeConfig['id']` alleviate that? All of the object creation / destructuring was getting on my nerves between this and `removeCustomNetwork`

HenryNguyen5: its not that, its when you call an action creator such as `removeCustomNode(myVar)` then its not easy at a quick glance to see what's passed in and why. `removeCusomNode({id: myVar})` makes it more legible and keeps context imo

wbobeirne: I think the current invocations are clear from a reading standpoint, since they're always `*.id`, here are all of their uses

> yield put(removeCustomNetwork(customNetwork.id));
> removeCustomNetwork(firstCustomNetwork.id)

> this.props.removeCustomNode(this.props.node.id);
> removeCustomNode(firstCustomNode.id)

But yeah, from someone looking to call it for the first time, they might not be sure what to pass. Maybe the best alternative is to pass in the entire config, rather than a string id.

PR #1663, opened 2018-04-24
Enable Parity Signer Message Signing

maciejhirsz:
### Description

Allows to use the Parity Signer to sign messages :).

### Changes

* The `Sign Message` button now triggers a `SIGN_MESSAGE_REQUESTED` action.
* `IFullWallet.signMessage` is now being used by `sagas/message/signing` for non-Signer wallets.
* Parity Signer wallet handles the flow within the saga and ignores the `signMessage` implementation, analog to `signRawTransaction`.

### Steps to Test

1. Download the Parity Signer from the App Store / Google Play. Create an account on it.
2. Attempt to sign a message with the account on the signer.
3. Verify that the signature is valid.
4. Make sure that signing a message by other means works.
5. (Optional sanity check) check that sending a transaction still checks.

### TODO

* [x] Verify that the signatures are valid (in case the scanned QR code is not a signature or not a valid signature).


common/actions/paritySigner/actionTypes.ts
```diff
@@ -1,9 +1,19 @@
...
+    isMessage: false;
+    data: string;
+    from: string;
+  };
+}
+
+export interface RequestMessageSignatureAction {
+  type: TypeKeys.PARITY_SIGNER_REQUEST_MSG_SIGNATURE;
+  payload: {
+    isMessage: true;
```
wbobeirne: This is fine for now, but if there's ever another signature type, we should make this an enum, e.g. `payload.type === SIGNATURE_DATA.MESSAGE` or something.

common/components/WalletDecrypt/disables.ts
```diff
@@ -22,10 +22,9 @@ export const DISABLE_WALLETS: { [key in WalletMode]: DisabledWallets } = {
     }
   },
   [WalletMode.UNABLE_TO_SIGN]: {
-    wallets: [SecureWalletName.TREZOR, SecureWalletName.PARITY_SIGNER, MiscWalletName.VIEW_ONLY],
+    wallets: [SecureWalletName.TREZOR, MiscWalletName.VIEW_ONLY],
     reasons: {
       [SecureWalletName.TREZOR]: 'This wallet can’t sign messages',
-      [SecureWalletName.PARITY_SIGNER]: 'This wallet can’t sign messages',
       [MiscWalletName.VIEW_ONLY]: 'This wallet can’t sign messages'
```
wbobeirne: Not for this PR, but general reminder that we need to localize these.

common/reducers/message.ts
```diff
@@ -0,0 +1,26 @@
+import { MessageAction, SignLocalMessageSucceededAction, TypeKeys } from 'actions/message';
+import { ISignedMessage } from 'libs/signing';
+
+export interface State {
+  signed?: ISignedMessage | null;
```
wbobeirne: Would love to see an `isSigning` boolean here, and stop users from trying to sign again / providing a loader while waiting for signature.

common/sagas/message/signing.ts
```diff
@@ -0,0 +1,79 @@
...
+  yield put(
+    showNotification('success', translate('SIGN_MSG_SUCCESS', { $address: action.payload.address }))
+  );
+}
+
+export const signing = [
+  takeEvery(TypeKeys.SIGN_MESSAGE_REQUESTED, handleMessageRequest),
+  takeEvery(TypeKeys.SIGN_LOCAL_MESSAGE_SUCCEEDED, notifySuccess)
+  // TODO:
+  // takeEvery(TypeKeys.SIGN_LOCAL_TRANSACTION_SUCCEEDED, verifySignature)
```
wbobeirne: In order to verify, we'll need to store the requested data to sign in the reducer, right?

maciejhirsz: It stores the entire JSON (`address`, `sig` and `msg`) already, so should be easy.

common/reducers/message.ts
maciejhirsz: :+1: not strictly necessary for the QR codes since the modal takes over, but I can see how it can be a thing with hardware wallets.

maciejhirsz:
Merging `develop` in somehow broke unlocking the account. I don't think it's me since it's borked for unlocking with private key as well as the signer.

Also the spinner on signing isn't there yet, should I do it or do you want to tackle it @wbobeirne?

dternyak:
@maciejhirsz Do you mind elaborating on what is broken? I just tested all of the unlock options on develop and this branch, and I didn't run into any issues.

dternyak:
I was also able to sign successfully on the latest commit.

maciejhirsz:
Signing messages works fine.

Signing *transactions* (or rather just unlocking the account to sign transaction) breaks:

```
Uncaught TypeError: this.props.resetTransactionRequested is not a function
    at ProxyComponent.WalletDecryptClass.onUnlock (WalletDecrypt.tsx?d638:451)
    at Object.onUnlock (WalletDecrypt.tsx?d638:281)
```

PR #1665, opened 2018-04-24
Electron Alpha Prep

wbobeirne:
### Description

Reworks language, warnings, and some new behavior to make Electron ready for an alpha release.

**Copy is _not_ localized because of its temporary nature.**

### Changes

* Adds a warning each time the user opens the alpha.
* Adds a hard coded expiration date to alpha builds (in `config/data.tsx`) Once this date has passed, the alpha will brick itself.
* Adds an update check done in the webview. Pings github releases, and does a semver check against the current version. If it's lower, pops open a simple modal requesting an update.
* Changes 'BETA RELEASE' banner to say 'ALPHA RELEASE'
* Disables native updater from running at all (for now)

### Steps to Test

1. Run the electron app
2. Confirm you get alpha warning
3. Edit `APP_ALPHA_EXPIRATION` to an earlier date, confirm you get bricked view
4. Edit `VERSION_RC` to a lower version (e.g. 0.9.0), confirm you are prompted to update

### Alpha Notice

screen shot 2018-04-24 at 6 11 15 pm

### Bricked App

screen shot 2018-04-24 at 5 32 27 pm

### Update Modal

screen shot 2018-04-24 at 4 40 26 pm

### DIFF
--- a/common/components/AppAlphaNotice/AppExpired.scss
+++ b/common/components/AppAlphaNotice/AppExpired.scss
@@ -1,0 +1,44 @@
+@import 'common/sass/variables';
+@import 'common/sass/mixins';
+
+.AppExpired {
+  @include cover-message;
+  background: $brand-danger;
+  display: flex;
+  align-items: center;
+
+  &-content {
+    padding-bottom: 60px;
+
+    h2 {
+      text-align: center;
+    }
+
+    p {
+      text-align: justify;
+    }
+
+    &-btn {
+      display: block;
+      width: 100%;
+      max-width: 280px;
+      margin: 40px auto 0;
+      text-align: center;
+      border: none;
+      padding: 0;
+      transition: $transition;
+      height: 60px;
+      line-height: 60px;
+      font-size: 22px;
+      background: rgba(#fff, 0.96);
+      color: $gray-dark;
+      border-radius: 4px;
+      text-shadow: none;
+
+      &:hover {
+        background: #fff;
+        color: $gray-darker;
+      }
+    }
+  }
+}
--- a/common/Root.tsx
+++ b/common/Root.tsx
@@ -15,6 +15,7 @@ import ErrorScreen from 'components/ErrorScreen';
 import PageNotFound from 'components/PageNotFound';
 import LogOutPrompt from 'components/LogOutPrompt';
 import QrSignerModal from 'containers/QrSignerModal';
+import NewAppReleaseModal from 'components/NewAppReleaseModal';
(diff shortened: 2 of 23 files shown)


common/components/AppAlphaNotice/AppExpired.scss
```diff
@@ -0,0 +1,44 @@
...
+      border-radius: 4px;
+      text-shadow: none;
+
+      &:hover {
+        background: #fff;
+        color: $gray-darker;
+      }
+    }
+  }
+}
```
wbobeirne: Styles for these two are total copy / pastes of BetaAgreement styles. I didn't want to refactor into reusable component because this is so temporary. Sorry for my laziness 🙏

dternyak:
Can confirm the above as well! Thanks a lot @wbobeirne! 

I did notice the download link takes you to https://github.com/MyCryptoHQ/MyCrypto/releases/tag/1.0.0-RC, but I think we should have it link out to download.mycrypto.com instead. Let me know if you had a reason for linking to the GitHub release instead @wbobeirne, otherwise I think that should be swapped out before merge.

wbobeirne:
That site just links out to GitHub anyway. And if you already have the app, you don't need the sales pitch IMO, you just want the download link.

dternyak:
@wbobeirne The changes that @james-prado is working on will change that.

tayvano:
> I did notice the download link takes you to https://github.com/MyCryptoHQ/MyCrypto/releases/tag/1.0.0-RC

Just to confirm, the actual link being used is https://github.com/MyCryptoHQ/MyCrypto/releases/latest, not https://github.com/MyCryptoHQ/MyCrypto/releases/tag/1.0.0-RC, right?

dternyak:
> Just to confirm, the actual link being used is https://github.com/MyCryptoHQ/MyCrypto/releases/latest, not https://github.com/MyCryptoHQ/MyCrypto/releases/tag/1.0.0-RC, right?

@tayvano Yup, correct. It redirects to https://github.com/MyCryptoHQ/MyCrypto/releases/tag/1.0.0-RC after hitting `/latest`. Either way, I think we're changing this to just point to https://download.mycrypto.com though, unless we have a good reason not to.

jspence425:
Agreed on adding HackerOne text on navy blue screen.
Agreed on adding small text below "Get the New Version" - added text should say "It only takes a few clicks!"
Leave "maybe later," but change capitalization to "Maybe Later" and "Get the New Version"

Issue #1570, opened 2018-04-18
No loading indicator for tokens in DeterministicWalletModal


HenryNguyen5:
### Description of the Issue

Lack of loading indicator on tokens leads user to thinking that nothing is happening

### Steps to Reproduce

1. Go to `Send & Verify` tab
2. Choose `Ledger` wallet
3. Sign in `Ledger` wallet
4. Observe that ETH has loading indicators
5. Choose a token
6. Observe that token amounts will stay `???` until loaded, no loading indicator

### Console Logs / Screenshots

![2018-04-17-182706_820x573_scrot](https://user-images.githubusercontent.com/6404866/38905989-09b696c8-4282-11e8-86fa-6f875698709b.png)

wbobeirne:
Fixed via #1682

PR #1682, opened 2018-04-26
Fix Duplicate DPaths & Deterministic Modal Improvements

### LINKED ISSUE #1563: Duplicate entries in deriviation path
### Description of the Issue

When selecting the deriviation path dropdown in the DeterministicWalletModal, duplicate entries can appear

### Steps to Reproduce

1. Get to DeterministicWalletModal component via View & Send -> Ledger
2. Click on deriviation path dropdown
3. Multiple Ledger(ETH) entries appear

(Console logs or screenshots go here, if applicable)

2018-04-17-184254_647x218_scrot
2018-04-17-184248_647x205_scrot

wbobeirne:
Closes #1563

### Description

Removes duplicate dpaths, and treats each option as a unique option rather than always selecting the first instance of the dpath. Also made some quick improvements to the modal while I was here.

### Changes

* Fixes de-duplication in the dpaths selector function
* Pass around full `DPath` objects to wallet decrypt components & deterministic modal, rather than just the dpath string.
    * This makes it so that if you select `Default (POA)`, that it shows up as the selected option, whereas before it would should `Default (ETH)` since they have the same dPath, and ETH is the first in the list.
* Improve deterministic wallet modal UI
    * When no token is selected, show small, light `N/A` text instead of `???`
    * Token balances now show spinner while loading
    * Gets rid of constantly persistent scrollbar on table, even though it doesn't scroll

### Steps to Test

1. Enter a mnemonic, ensure there are no duplicate dPaths and it functions correctly
2. Unlock a TREZOR, ensure there are no duplicate dPaths and it functions correctly
3. Unlock a Ledger, ensure there are no duplicate dPaths and it functions correctly

### Screenshots

screen shot 2018-04-26 at 12 07 49 pm

### DIFF
--- a/common/components/WalletDecrypt/components/DeterministicWalletsModal.tsx
+++ b/common/components/WalletDecrypt/components/DeterministicWalletsModal.tsx
@@ -305,16 +305,16 @@ class DeterministicWalletsModalClass extends React.PureComponent<Props, State> {
           />
         </td>
         <td>
-          {token ? (
+          {desiredToken ? (
             <UnitDisplay
-              decimal={token.decimal}
-              value={token.value}
+              decimal={token ? token.decimal : 0}
+              value={token ? token.value : null}
               symbol={desiredToken}
               displayShortBalance={true}
               checkOffline={true}
             />
           ) : (
-            '???'
+            <span className="DWModal-addresses-table-na">N/A</span>
           )}
         </td>
         <td>
--- a/common/components/WalletDecrypt/components/DeterministicWalletsModal.tsx
+++ b/common/components/WalletDecrypt/components/DeterministicWalletsModal.tsx
@@ -65,7 +65,7 @@ class DeterministicWalletsModalClass extends React.PureComponent<Props, State> {
     selectedAddrIndex: 0,
     isCustomPath: false,
     customPath: '',
-    currentDPath: this.props.dPaths[0],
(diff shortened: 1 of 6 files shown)


HenryNguyen5:
This also closes https://github.com/MyCryptoHQ/MyCrypto/issues/1570

common/components/WalletDecrypt/components/DeterministicWalletsModal.tsx
```diff
@@ -42,11 +42,11 @@ interface Props {
 
   onCancel(): void;
   onConfirmAddress(address: string, addressIndex: number): void;
-  onPathChange(path: string): void;
+  onPathChange(dPath: DPath): void;
 }
 
 interface State {
-  currentLabel: string;
+  currentDPath: DPath;
```
HenryNguyen5: This seems like an invalid type since if the user uses a custom dpath, it's not in our enumeration of dpaths.

dternyak:
👍 Can confirm no duplicate dpath options.

common/components/WalletDecrypt/components/DeterministicWalletsModal.tsx
wbobeirne: `DPath` isn't an enum, it's just an interface. So it's OK if it's not one of our hardcoded ones.

PR #1683, opened 2018-04-26
Allow Network Change Offline

### LINKED ISSUE #1681: Can't offline sign non-mainnet tx
### Description of the Issue

Version tested: 1.0.0rc1

When using MyCrypto offline it's not possible to sign txs for anything but the mainnet. 

The reason is that it's impossible to switch to another Eth node due to the connection check preventing it. From that it follows that it's not possible to use anything else than the mainnet since the network id is coupled to the node selection.

### Steps to Reproduce

- Use MyCrypto on a system without internet connection
- Try to switch to anything that is not mainnet (ropsten, kovan, ...)
- See that it's not possible

wbobeirne:
Closes #1681

### Description

I think this is a fair compromise between the UX of not letting users switch to nodes that won't connect vs allowing users to switch to other networks when offline. Allows a user to switch without fail if they're currently offline. Still attempts to fetch a block and get them back online, even if offline though.

However, maybe the UX tradeoff isn't worth it, and we should just let users switch to offline nodes no matter what. I'm definitely open to that discussion!

### Steps to Test

1. Load the page and make sure you're connected to a network
2. Attempt to switch to another node, make sure that works
3. Set yourself offline, and quickly switch to another node (Simulating connecting to an "offline node"), confirm it doesn't allow you to connect
4. Get yourself into the offline state (triggered by causing a node request, e.g. balance)
5. Attempt to switch to another node again, confirm it allows you to connect

### DIFF
--- a/common/sagas/config/node.ts
+++ b/common/sagas/config/node.ts
@@ -160,6 +160,7 @@ export function* handleNodeChangeIntent({
     );
   }
 
+  const isOffline = yield select(getOffline);
   if (isAutoNode(nodeIdToSwitchTo)) {
     shepherd.auto();
     if (getShepherdNetwork() !== nextNodeConfig.network) {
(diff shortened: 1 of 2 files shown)


common/sagas/config/node.ts
```diff
@@ -160,26 +160,29 @@ export function* handleNodeChangeIntent({
...
-      yield apply(shepherd, shepherd.manual, [nodeIdToSwitchTo, false]);
+      yield apply(shepherd, shepherd.manual, [nodeIdToSwitchTo, isOffline]);
     } catch (err) {
       console.error(err);
       return yield* bailOut(translateRaw('ERROR_32'));
     }
   }
 
-  let currentBlock;
+  let currentBlock = '???';
```
HenryNguyen5: Hm, maybe better to keep the current block as a number, since once we introduce stricter typing to sagas this'll throw. Instead could set it to `-1` and display on the UI side `???` if the block number is negative

wbobeirne: Block number is a string in the reducer, the actions all take in a string.

wbobeirne: https://github.com/MyCryptoHQ/MyCrypto/blob/develop/common/reducers/config/meta/meta.ts#L8

https://github.com/MyCryptoHQ/MyCrypto/blob/develop/common/actions/config/actionCreators.ts#L113

HenryNguyen5: whoops, my bad

wbobeirne: No worries, might be worth refactoring at some point if we ever need to treat the block number as a number.

dternyak:
👍 I think the UX is optimal in that online users will still get notifications about offline nodes, but offline users can switch without delay.

PR #1690, opened 2018-04-27
Ledger types

HenryNguyen5:
Adds types for packages under https://github.com/LedgerHQ/ledgerjs
- `@ledgerhq/hw-app-eth` https://github.com/LedgerHQ/ledgerjs/tree/master/packages/hw-app-eth
- `@ledgerhq/hw-transport-node-hid` https://github.com/LedgerHQ/ledgerjs/tree/master/packages/hw-transport-node-hid
- `@ledgerhq/hw-transport` https://github.com/LedgerHQ/ledgerjs/tree/master/packages/hw-transport
-  `@ledgerhq/hw-transport-u2f` https://github.com/LedgerHQ/ledgerjs/tree/master/packages/hw-transport-u2f


package.json
```diff
@@ -129,6 +132,7 @@
     "tslint-react": "3.5.1",
     "types-rlp": "0.0.1",
     "typescript": "2.8.1",
+    "u2f-api": "1.0.6",
```
wbobeirne: Do we need to add this dependency just for types? Or is there a way around that?

HenryNguyen5: Yeah, the package is written in TS so you have to have it as a dep to get the types.

HenryNguyen5: Ideally we can merge these types upstream so it won't have the extra package

dternyak:
In the long-term, it will be nice to have these upstreamed. In the interim, I'll merge.

PR #1710, opened 2018-05-02
Use proxy.mycryptoapi.com for rates

dternyak:
No [public] issue associated with this PR.

### What?
Replaces cryptocompare with proxy.mycryptoapi.com for rate requests.

### Why?
In an effort to increase user privacy, an effort is made to reduce requests to third parties directly.

### How to test
1. Open dev tools
2. Unlock any wallet format
3. Check to make sure equivalent values still load
4. Verify that requests is are being made proxy.mycryptoapi.com instead of min-api.cryptocompare.com


dternyak:
Approved by @HenryNguyen5

PR #1717, opened 2018-05-04
Adding network unit to account balance

### LINKED ISSUE #1715: Show network unit in Account Balance

It would be more meaningful if in Account Balance is shown not only the network name but the network unit.

### Example(s) of Feature

Now: 0.1234 RSK_TESTNET
Proposal: 0.1234 SBTC (RSK_TESTNET)

I understand that keeping network name is important to avoid mistakes between different networks with the same ETH unit.

alepc253:
Closes #1715 

### Description

Adds network unit to the account balance.

### Changes

String concatenation of network unit and network name.

### Steps to Test

1. Unlock any account and check visually the account balance.


wbobeirne:
Hey @alepc253, thanks for the suggestion. Given that I think the goal of this is to avoid testnet mistakes, it might be good to only show this in the case of testnets. We have a `.isTestnet` property on networks, so maybe it could be used in that case.

alepc253:
So, you are saying to show unit + network name if it is a testnet and just unit if it is mainnet, is it right?

dternyak:
Thanks for the PR @alepc253!

We're currently in the midst of moving our beta into production, so I wanted to give you a heads up that this will likely not be merged in time of the full release.

r/e your last message to @wbobeirne, I think showing network name next to the unit is indeed mostly valuable on testnets, and not mainnet.

alepc253:
Agreed @dternyak, that's how it is done in the PR. Thanks!

dternyak:
Thanks @alepc253 ! 👍 This is solid addition :)

PR #1718, opened 2018-05-05
Don't require leading 0 in amount field

### LINKED ISSUE #1629: Don't require leading 0 in amount field
(User request)

The amount field requires a leading 0 when sending fractions of ether (e.g. 0.01 is valid, .01 is not).

screen shot 2018-04-19 at 12 48 00 pm

- [ ] Allow decimals without leading 0

srowhani:
Closes #1629 
### Description

Amount field no longer requires leading zero.
image

### Changes

`parseInt(".01", 10) -> NaN`
`parseInt("0.01", 10) -> 0`
`Number(".01") -> 0.01`

- Use number constructor vs parseInt


dternyak:
Thanks for the PR @srowhani! Nice quality of life improvement indeed. Welcome to the project!

PR #1722, opened 2018-05-07
Disable tabs by network configuration

### LINKED ISSUE #1721: Hide unsupported tabs by network

It should be possible to disable or hide some tabs that doesn't make sense in a selected network.

alepc253:
Closes #1721

### Description

Gives the possibility of disable navigation tabs that don't make sense in a certain network.

### Changes

* Addition of new parameter `unsupportedTabs` in network configuration. 
* Addition of new prop `isNotEnabled` in NavigationLink (this prop determines the rendering of the component).

### Steps to Test

1. Choose a network config and add, for example, `unsupportedTabs: ['NAV_ENS', 'NAV_SWAP']` in file `staticNetworks.ts`
2. Select that network
3. Check visually that this tabs are not displayed


common/components/NavigationLink.tsx
```diff
@@ -8,13 +8,18 @@ interface Props extends RouteComponentProps<{}> {
   link: NavigationLink;
   isHomepage: boolean;
   className: string;
+  isNotEnabled: boolean;
```
dternyak: Can we make this an optional boolean? If its not set, it should be falsy anyway.

common/components/NavigationLink.tsx
```diff
@@ -8,13 +8,18 @@ interface Props extends RouteComponentProps<{}> {
...
 }
 
 class NavigationLinkClass extends React.PureComponent<Props, {}> {
   public render() {
-    const { link, location, isHomepage, className } = this.props;
+    const { link, location, isHomepage, className, isNotEnabled } = this.props;
     let isActive = false;
 
+    if (isNotEnabled) {
+      return '';
```
dternyak: `return null` is more conventional imho.

common/components/Header/components/Navigation.tsx
```diff
@@ -57,6 +58,7 @@ export default class Navigation extends PureComponent<Props, State> {
                 link={link}
                 isHomepage={link === navigationLinks[0]}
                 className="NavigationLink"
+                isNotEnabled={unsupportedTabs != null && unsupportedTabs.includes(link.name)}
```
dternyak: nitpick: `unsupportedTabs && unsupportedTabs.includes(link.name)` is more concise

common/components/Header/components/Navigation.tsx
```diff
@@ -5,6 +5,7 @@ import './Navigation.scss';
 
 interface Props {
   color?: string | false;
+  unsupportedTabs?: string[];
```
dternyak: Would prefer to create a union of all tabs and so we can get some type safety around what tabs will be unsupported:
```
type TABS = 'CREATE' | 'SWAP' | 'CONTRACT' ...

const unsupportedTabs: TABS[] = ['SWAP'];
```

HenryNguyen5: Would prefer enums here (can be string enums if needed) such as 
```ts
enum TAB {
 CREATE,
 SWAP, 
}
const unsupportedTabs: TAB[] = [TAB.CREATE]
```

alepc253:
@dternyak @HenryNguyen5 thanks for your feedback!
I don't understand what happened but I've lost this branch so I'm creating a new pull request, setting a reference to this one.
Very sorry for the inconvenience.

PR #1730, opened 2018-05-08
Show version in Electron

### LINKED ISSUE #1728: Add "version" text somewhere in Windows/Linux version of Desktop App
User pointed out that they have no idea how to check what version they have - turns out only OSX has the "About" window so there is in fact no way to check your version on Windows or Linux.

Might be best to add it to the "Support Us" page or somewhere else, dunno.

@wbobeirne

wbobeirne:
Closes #1728

### Description

Adds version to help menu and on support us page that more or less has all of the same footer information.

### Screenshots

screen shot 2018-05-08 at 4 12 03 pm

screen shot 2018-05-08 at 4 00 47 pm


dternyak:
Can confirm version shows up in footer!

PR #1777, opened 2018-05-13
Improve electron versioning

### LINKED ISSUE #1773: Fix Electron Update Bug
[TBD]

HenryNguyen5:
Closes #1773

### Description
- Changes the package.json so that both web and electron version are separated.
- Exports a single variable that is either the web or electron version depending on appropriate context
- Updates exported electron version to RC2.
- Adds version display to update window so the user knows what they're updating to/from

### Electron versioning schema
- Adheres to SemVer
- Comes in string format of `<PLATFORM>_<VERSION>_MyCrypto.<EXT>`
    - Ex: `mac_1.0.0-alpha.4_MyCrypto.dmg`
    - Platforms can be any of the following (case sensitive): `'mac'|'windows'|'linux-i386'|'linux-x86-64'`

### Electron version algorithm
1. Grab all releases from github
2. For each release `r`
3. For each asset `a` in `r`
4. Parse each asset name
5. Determine if `<VERSION>` in asset name is greater than `electron-version` in `package.json`
6. If so, prompt user to update
7. Otherwise, do not prompt

### Uploading binaries to release page
- Make sure the binary name adheres to the Electron Versioning schema
- Make sure you upload it in the binary section of the release page
- Make sure to use the `SemVer` package to determine that the release is newer than the current one in `package.json[electron-version]` so that users will be prompted to update

### Steps to Test

1. Change version in package json
2. Validate that web build version in footer changes accordingly
3. Change electron-version in package json to something other than RC2
4. Validate that the update window pops up with the correct version comparision
5. Validate that the `help` menu shows the right electron version
6. Change electron-version in package json to RC2
7. Rebuild electron
8. Validate that the update window no longer shows, and the `help` menu displays the correct version

### Screenshots
2018-05-13-122505_676x394_scrot


dternyak:
We'll be going with a new approach that checks filenames in uploaded bins associated with a release to determine if the local application is outdated. This helps us in decoupling the github release version (web) and electron release version, which we can grab from the filename.

PR #1829, opened 2018-05-21
Custom node defaults

### LINKED ISSUE #1804: Auto search common urls for custom nodes

Currently its somewhat of a chore to add a custom node when you have one running locally on default settings. We can instead scan a common list of endpoints and display them to the user as a list of available custom nodes.

HenryNguyen5:
Closes #1804

### Description
Currently this only adds very basic support (just searches for `http://localhost:8545` since thats the most common node setup) but it lays down the groundwork for future things like scanning across ranges of ports/urls + ipc/ws/https support + showing node status reactively to the custom node modal input.


dternyak:
I'll have to republish `eth-exists` before merge, but otherwise looks good!

PR #1830, opened 2018-05-22
Add Electron Build to Travis

skubakdj:
Implements a build matrix, causing Travis to run our tests in both Linux and OSX environments in parallel. The `.travis.yml` file has also been modified to run our traditional test suite when in Linux, and build our Electron app for all platforms when in OSX.

A downside to this approach is that (I think) Travis provides fewer OSX machines to execute CI, so the Electron build tasks may take a little longer to actually spin up (and thus increase the time it takes to complete the Travis pass/fail status check).  Seems like an acceptable tradeoff, however.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,35 +1,69 @@
-dist: trusty
-sudo: required
+
 language: node_js
 
+matrix:
+  include:
+    - os: osx
+      osx_image: xcode9.3
+      env:
+        - ELECTRON_CACHE=$HOME/.cache/electron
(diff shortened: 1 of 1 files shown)


skubakdj:
Looks like `yarn` is failing on `travis-ci/pr`. Might need to clear the cached files since this PR switches Travis from `npm` to `yarn` for module install.

dternyak:
@skubakdj Just cleared all caches.

skubakdj:
@dternyak looks like that did the trick

PR #1832, opened 2018-05-23
Fetch tokens from IPFS hash on load / node change

HenryNguyen5:
If you're contributing translations, please follow the contributor guidelines
at https://github.com/MyCryptoHQ/MyCrypto/wiki/Contributor-Guidelines#providing-translations

---

If you're contributing code, please link to the issue it resolves and follow the
contributor guidelines at https://github.com/MyCryptoHQ/MyCrypto/wiki/Contributor-Guidelines#contributing-code and fill out the following template:

Closes #(issue number goes here)

### Description

(Description goes here)

### Changes

* High level
* Changes that
* You Made

### Steps to Test

1. Steps
2. To
3. Test

### Screenshots

(Only if applicable)


common/config/data.tsx
```diff
@@ -5,6 +5,10 @@ import { GasPriceSetting } from 'types/network';
 import { makeExplorer } from 'utils/helpers';
 import NewTabLink from 'components/ui/NewTabLink';
 
+// tslint:disable-next-line:no-http-string
+export const IPFS_GATEWAY = 'http://127.0.0.1:8080';
```
dternyak: NOTE: Update before merge.

common/config/data.tsx
```diff
@@ -5,6 +5,10 @@ import { GasPriceSetting } from 'types/network';
...
+export const IPFS_MULTISIG_CONTRACT_ADDRESS = '0x230ead3dbbd50248456a57afdd6570cdc8b4e439';
```
dternyak: NOTE: Update before merge.

dternyak:
> The key-value pair for the resolver is as follows :.

I think you missed a section here

HenryNguyen5:
@dternyak ah, the markdown was escaping what i typed there :+1:

common/config/data.tsx
skubakdj: @HenryNguyen5 we should consider having support for multiple endpoints in case one of them goes down.

skubakdj: I think we might be defaulting to `https://ipfs.infura.io` to kick things off.

HenryNguyen5: Cool, should i change this now then? @skubakdj  @dternyak

HenryNguyen5:
@skubakdj Concerning the two comments above:
1) Agreed, I just kept it simple for now though for MVP. Maybe we can have an even more generic shepherd in the future for IPFS :)
2) I dug into this but it surprisingly is not as simple as it seems, the lowest friction way would be to just use an IPFS node, which would also solve 1). We would have to test how much it bloats bundle size though.

common/config/data.tsx
skubakdj: Yes, IMO. I believe we confirmed with @tayvano on using Infura in the MVP.

tayvano: Yup, I'm cool using Infura for MVP. Ideally we would set up our own stuff and use that but this is fine for the time being, especially until we have Mia and DS back 100%. 👍

dternyak:
We'll have to do some training to make sure parties are able to successfully update tokens on an on-going basis, which is holding up the merge here. I'm going to start driving towards building that up, so we can hopefully merge this soon!

Issue #1621, opened 2018-04-19
Improve Copy / UX on Account View for Desktop Applications


dternyak:
A large portion of the space on the default view of the MyCrypto Desktop applications goes to showing disabled wallet formats.

- [ ] Determine a better UX that displays valid formats while making better use of space.

jspence425:
Highlighting "Web3" on the Desktop app displays "This wallet is not supported in the MyCrypto app"
Highlighting "Ledger" on the Desktop app displays " This wallet is only supported at MyCrypto.com"

I understand there's a difference between the two but the messaging either needs to be the same for both (a large majority of our users do not care for the specific reason why they can't access in these manners, they'll just care that it can or cannot be used) or different for both but explained in detail for why you cannot access using these routes.

I would recommend saying "This wallet type is not supported on the MyCrypto app" as a catch-all.

wbobeirne:
I think the goal by the time we launch this is to have these two sorted out. #921 should allow for Ledger on the app, and we're also eyeing https://github.com/MetaMask/mascara which could replace the Web3 option.

dternyak:
Mascara is really early, so unfortunately it's not a good option for the
foreseeable future.

We can look into #921, but I don't see that viable in the next month, but
perhaps a separate ledger implementation is.

wbobeirne:
I think I'd sooner whip up a temporary Ledger implementation that's not #921, than ship without supporting it. We should probably get that on the roadmap.

dternyak:
@wbobeirne Yeah, absolutely. I think the node-hid approach might be the way to go for desktop apps, so we'll just have a separate implementation for desktop.

PR #1836, opened 2018-05-23
Electron Ledger Support

### LINKED ISSUE #1616: Current Ledger dependency unmaintained
### Description of the Issue

Related to #921 

We currently depend on `ledgerco` which was a monolithic package that is now unmaintained. We should switch over to ledger's new packages hosted in their monorepo https://github.com/LedgerHQ/ledgerjs/tree/master/packages

This would also allow us to keep up to date with their packages, and handle transport layers much more easily (u2f vs nodehid), letting us implement hardware wallet signing for electron.

wbobeirne:
Closes half of #1621, closes #1616.

### Description

This PR is the first of a few to add hardware wallet support to Electron with a new library dubbed Enclave (props to @HenryNguyen5 for the name.) You can read more about its specific approach here: https://github.com/MyCryptoHQ/MyCrypto/blob/enclave-willo-protocol/shared/enclave/README.md. While it's being developed, I'd like to keep it in the MyCrypto repo. But once we're in a steady place with it, I think it could easily be its own module.

Likewise, I think it's still a little rough around the edges. There's maybe too much boilerplate both on the server and client side and the types could probably be better, I'm definitely expecting some feedback and changes on the pr. There's also _huge_ amount of repeated code between `common/libs/wallets` and `enclave/server/wallets`. We could potentially expand Enclave to handle web wallets as well, and just have a single API for interacting with wallets, regardless of platform.

I also adjusted the Ledger unlock screen with persistent information (screenshot below.)

### Changes

* Made Enclave lib with Ledger support, Trezor & KeepKey stubs
* Changed Ledger libs over to their `@ledgerhq/*` packages
* Changed our Ledger and Trezor libraries to defer to EnclaveAPI in Electron
* Change Ledger tip to display persistently and have messaging about browser support and which Ledger app to open

### Steps to Test

1. Unlock ledger on web
2. Display address on device with ledger on the web
3. Send a transaction with ledger on web
4. Sign a message with ledger on the web
5. Do all of the above in Electron

### Screenshots

screen shot 2018-05-23 at 5 12 40 pm


common/index.tsx
```diff
@@ -24,3 +25,7 @@ if (module.hot) {
 if (process.env.NODE_ENV === 'production') {
   consoleAdvertisement();
 }
+
+if (process.env.BUILD_ELECTRON) {
+  (window as any).EnclaveAPI = EnclaveAPI;
+}
```
wbobeirne: Reminder to remove, this was just for testing in console purposes.

common/libs/wallet/deterministic/trezor.ts
```diff
@@ -2,24 +2,51 @@ import BN from 'bn.js';
...
+import { HardwareWallet, ChainCodeResponse } from './hardware';
 import { getTransactionFields } from 'libs/transaction';
 import mapValues from 'lodash/mapValues';
-
 import { IFullWallet } from '../IWallet';
 import { translateRaw } from 'translations';
+import EnclaveAPI, { WalletTypes } from 'shared/enclave/client';
 
 export const TREZOR_MINIMUM_FIRMWARE = '1.5.2';
+const TrezorConnect = TC as any;
```
wbobeirne: I should probably just d.ts this hot boi.

common/libs/wallet/deterministic/hardware.ts
```diff
@@ -0,0 +1,17 @@
...
+
+export interface ChainCodeResponse {
+  chainCode: string;
+  publicKey: string;
+}
+
+export class HardwareWallet extends DeterministicWallet {
+  // @ts-ignore
+  public static getChainCode(dpath: string): Promise<ChainCodeResponse> {
+    throw new Error(`getChainCode is not implemented in ${this.constructor.name}`);
```
HenryNguyen5: Do we want to have these i18n'd?

common/libs/wallet/deterministic/ledger.ts
```diff
@@ -1,80 +1,142 @@
...
+        dpath
+      });
+    }
+
+    return makeApp()
+      .then(app => app.getAddress(dpath, false, true))
+      .then(res => {
+        return {
+          publicKey: res.publicKey,
+          chainCode: res.chainCode as string
```
HenryNguyen5: What's the reason for the typecast here?

common/libs/wallet/deterministic/ledger.ts
```diff
@@ -1,80 +1,142 @@
...
+        };
+      })
+      .catch((err: any) => {
```
HenryNguyen5: Rather have `err as string`

common/libs/wallet/deterministic/ledger.ts
```diff
@@ -1,80 +1,142 @@
...
     return translateRaw('X_LEDGER');
   }
 }
+
+async function makeApp() {
+  const transport = await TransportU2F.create();
+  return new LedgerEth(transport);
+}
+
+function ledgerErrToMessage(err: any) {
```
HenryNguyen5: Would be nice to have this properly typed

common/libs/wallet/deterministic/trezor.ts
HenryNguyen5: Agreed

common/libs/wallet/deterministic/trezor.ts
```diff
@@ -53,28 +80,21 @@ export class TrezorWallet extends DeterministicWallet implements IFullWallet {
...
-      index = this.index;
-    }
-
-    return new Promise((resolve, reject) => {
-      (TrezorConnect as any).ethereumGetAddress(
-        dPath + '/' + index,
+  public displayAddress(): Promise<any> {
+    return new Promise(resolve => {
+      TrezorConnect.ethereumGetAddress(
+        this.dPath + '/' + this.index,
```
HenryNguyen5: Mostly stylistic, but string template here would be better

common/libs/wallet/deterministic/trezor.ts
```diff
@@ -53,28 +80,21 @@ export class TrezorWallet extends DeterministicWallet implements IFullWallet {
...
         (res: any) => {
           if (res.error) {
-            reject(res.error);
+            resolve(false);
```
HenryNguyen5: you can eliminate the branching logic here by just returning `resolve(!res.error)`

electron-app/main/index.ts
```diff
@@ -20,3 +22,6 @@ app.on('activate', () => {
 app.on('ready', () => {
   getWindow();
 });
+
+// Register enclave protocol
+registerServer(app);
```
HenryNguyen5: Maybe a better name would be `initEnclave`

shared/enclave/README.md
```diff
@@ -0,0 +1,34 @@
+# ETH Enclave
+
+Enclave is the communication layer between hardware wallets and the Electron
+web view. This layer is necessary if you've disabled node integration, and
+enabled context isolation on your webview ([Which is something you should do.](https://github.com/electron/electron/blob/master/docs/tutorial/security.md))
+
+Enclave uses Electron's Protocol API to open up an HTTP-like communication layer
```
HenryNguyen5: I think request-response would be better in lieu of HTTP here to keep it more generalized

shared/enclave/client/index.ts
```diff
@@ -0,0 +1,33 @@
...
+  SignTransactionParams,
+  SignTransactionResponse,
+  SignMessageParams,
+  SignMessageResponse,
+  DisplayAddressParams,
+  DisplayAddressResponse
+} from 'shared/enclave/types';
+
+const api = {
+  getChainCode(params: GetChainCodeParams) {
```
HenryNguyen5: We've got a lot of duplication here, could probably cut it down quite a bit with utility functions

shared/enclave/client/requests.ts
```diff
@@ -0,0 +1,19 @@
...
+  return fetch(`${PROTOCOL_NAME}://${type}`, {
+    method: 'POST',
+    body: JSON.stringify(params)
+  })
+    .then(res => res.json())
+    .then((res: EnclaveResponse<T>) => {
+      if (res.error) {
+        throw new Error(res.error.message);
+      } else if (res.data) {
+        return res.data;
```
HenryNguyen5: Can this be cut down to 
```ts
const {error, data} = res
const UNKNOWN_ERR  = 'Unknown response from server'
if (data) {
return data
}
throw Error(error.message ||  UNKNOWN_ERR)

File: shared/enclave/server/handlers/displayAddress.ts [hunk 13]
```diff
@@ -0,0 +1,7 @@
+import { getWalletLib } from 'shared/enclave/server/wallets';
+import { DisplayAddressParams, DisplayAddressResponse } from 'shared/enclave/types';
+
+export default function(params: DisplayAddressParams): Promise<DisplayAddressResponse> {
```
HenryNguyen5: Not really a fan of default exports especially for functions that'll be used often, as it makes refactor-by-reference (powered by ts) not possible as the default import name is user defined

shared/enclave/server/handlers/index.ts
```diff
@@ -0,0 +1,18 @@
+import getChainCode from './getChainCode';
+import signTransaction from './signTransaction';
+import signMessage from './signMessage';
+import displayAddress from './displayAddress';
+import { EnclaveMethods, EnclaveMethodParams, EnclaveMethodResponse } from 'shared/enclave/types';
+
+const handlers: {
+  [key in EnclaveMethods]: (
```
HenryNguyen5: Prefer to not have the interface inline

shared/enclave/server/handlers/displayAddress.ts
HenryNguyen5: This goes for the functions below too

shared/enclave/server/index.ts
```diff
@@ -0,0 +1,68 @@
...
+    protocol.registerStringProtocol(PROTOCOL_NAME, async (req, cb) => {
+      let res: EnclaveResponse;
+
+      try {
+        const method = getMethod(req);
+        const params = getParams(method, req);
+        const data = await handlers[method](params);
+        res = { data };
+      } catch (err) {
+        console.error(`Request to '${req.url}' failed with error:`, err);
```
HenryNguyen5: Should stringify the error here incase it contains an object

shared/enclave/server/index.ts
```diff
@@ -0,0 +1,68 @@
...
+  const data = req.uploadData.find(d => !!d.bytes);
+
+  if (!data) {
+    throw new Error(`No data provided for '${method}'`);
+  }
+
+  try {
+    // TODO: Validate params based on provided method
+    const params = JSON.parse(data.bytes.toString());
+    return params as EnclaveMethodParams;
```
HenryNguyen5: Should turn this into a validation function that performs the job of a user defined type guard instead of casting it

shared/enclave/server/index.ts
```diff
@@ -0,0 +1,68 @@
...
+  } catch (err) {
+    throw new Error(`Invalid JSON blob provided for '${method}'`);
```
HenryNguyen5: Should we include the actual error in here too?

shared/enclave/server/wallets/keepkey.ts
```diff
@@ -0,0 +1,21 @@
+import { WalletLib } from 'shared/enclave/types';
+
+const KeepKey: WalletLib = {
```
HenryNguyen5: Instead of the repetition of implementing a wallet lib here, we should make a base class that has all of the methods throw the unimplemented error

shared/enclave/server/wallets/ledger.ts
```diff
@@ -0,0 +1,85 @@
+import EthTx from 'ethereumjs-tx';
+import { addHexPrefix, toBuffer } from 'ethereumjs-util';
+import { WalletLib } from 'shared/enclave/types';
+import TransportNodeHid from '@ledgerhq/hw-transport-node-hid';
+import LedgerEth from '@ledgerhq/hw-app-eth';
+let transport: any;
```
HenryNguyen5: Prefer to have this typed

shared/enclave/server/wallets/ledger.ts
```diff
@@ -0,0 +1,85 @@
...
+  try {
+    if (!transport) {
+      transport = await TransportNodeHid.create();
+      transport.on('disconnect', () => (transport = null));
+    }
+    return new LedgerEth(transport);
+  } catch (err) {
+    if (err && err.message && err.message.includes('cannot open device with path')) {
+      throw new Error(
+        'Your Ledger is currently in use with another application. Please wait, or close other wallet applications before trying again.'
```
HenryNguyen5: Should have this i18n'd i think since its a user error

common/components/WalletDecrypt/components/LedgerNano.tsx
```diff
@@ -58,14 +49,14 @@ class LedgerNanoSDecryptClass extends PureComponent<Props, State> {
...
-    const { dPath, publicKey, chainCode, error, isLoading, showTip } = this.state;
+    const { dPath, publicKey, chainCode, error, isLoading } = this.state;
     const showErr = error ? 'is-showing' : '';
 
     if (!dPath) {
       return <UnsupportedNetwork walletType={translateRaw('x_Ledger')} />;
     }
 
-    if (window.location.protocol !== 'https:') {
+    if (!process.env.BUILD_ELECTRON && window.location.protocol !== 'https:') {
```
HenryNguyen5: We use this env variable everywhere in the codebase, perhaps use another level of indirection by hiding it behind a function like `isElectron` so we can change it if need be in the future? We could do this for any other environment variables too.

common/libs/wallet/deterministic/hardware.ts
wbobeirne: Nah, this is meant for developers. I wanted to make this abstract, but TypeScript doesn't like `static abstract` methods, so I wasn't sure what to do.

HenryNguyen5:
I think that the easiest way to remedy the code duplication is to have each duplicated method its own function, then use that function where it used to be duplicated.

common/libs/wallet/deterministic/ledger.ts
wbobeirne: `app.getAddress` thinks this can be undefined, but that's only if you pass `false` as third arg or don't pass a third arg. This could be fixed in our typings by looking at that arg in the typing, I think we discussed this over DM a bit. I can't remember where we left that.

common/libs/wallet/deterministic/ledger.ts
wbobeirne: Not a string tho. It's either an `Error`, or it's an object with keys that `ledgerErrToMessage` handles.

shared/enclave/README.md
wbobeirne: Do you mean just calling it that in docs, or restructuring this? It uses things like HTTP methods and HTTP status codes, so it's pretty grounded in HTTP.

shared/enclave/server/index.ts
wbobeirne: It's the second arg in console.error which does the stringifying for you, not a string interpolated thing.

HenryNguyen5: Wouldn't it print [Object object] in some cases?

shared/enclave/README.md
HenryNguyen5: Just calling it in the docs. Since this could technically use a number of transport layers like when we were using ipcRenderer and postMessage

shared/enclave/server/wallets/ledger.ts
wbobeirne: Absolutely, though I have no way of getting at redux state in here. Probably what I should do is replace all of these human readable errors with ERROR_CODES to be `translate`'d on the frontend.

common/libs/wallet/deterministic/trezor.ts
wbobeirne: 👍 will clean these up in #1946.

common/libs/wallet/deterministic/trezor.ts
wbobeirne: 👍 will clean these up in #1946.

shared/enclave/server/index.ts
wbobeirne: No, it's the same as doing
```
console.log({ my: 'object' });
```
or
```
console.log(new Error('test'));
```
it'll always print a "nice" version of it.

shared/enclave/server/index.ts
wbobeirne: I'm not sure what that would look like, can you describe it a bit more?

shared/enclave/server/wallets/keepkey.ts
wbobeirne: I initially decided against classes in favor of singletons since there aren't multiple instances of any of these, they're moreso handlers than instances. However I could still make them classes and just export an instance of them, so I guess that makes more sense.

The util function makes sense though, I'll try to work that out.

shared/enclave/server/wallets/ledger.ts
```diff
@@ -0,0 +1,85 @@
...
+  async getChainCode(dpath) {
+    const app = await getEthApp();
+    try {
+      const res = await app.getAddress(dpath, false, true);
+      return {
+        publicKey: res.publicKey,
+        chainCode: res.chainCode as string
+      };
+    } catch (err) {
+      throw new Error('Failed to connect to Ledger');
```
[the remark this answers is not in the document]
wbobeirne: No way to do that & do i18n at the same time, so I'm going to settle for a console.error. In general we'll probably need to come up with a way to get console logs from users if they experience issues electron-side.

PR #1896, opened 2018-06-05
Night Mode Theme

### LINKED ISSUE #1549: Dark Theme
[Approach TBD by @wbobeirne or @james-prado]

wbobeirne:
Closes #1549

### Description

Adds the much sought after, long awaited night theme. It comes with a slew of changes to the way we do colors in our styles. Previously, while we mostly adhered to the palette, there were a few random colors floating around, and there wasn't any consistency about when we used opacity vs color to lighten or darken things. Now everything is more or less consistent:

* Colors are assigned using the `color(name-of-color)` function. For most cases, you should be running things through `color()`
* Colors output CSS variables instead of hex values
* Things that should just be darker or lighter will use `shade-dark(0.5)` and `shade-light(0.5)` functions.
    * In night mode, shade-dark makes things lighter and shade-light makes them darker.

Some rules of thumb I came across while converting things:

* Variable overuse is rough, if you have a distinct component, don't make it `color(gray-light)`, make a new variable like `color(my-component-bg)`
* Some things are just meant to be black or white, and don't need to be flipped between themes. Shadows are always block, text on color should always be white.
* Sometimes alterations are needed beyond color, or the color should be completely different across themes, so we have a mixin `@include theme(dark) { /* ... */ }` that only applies rules to a particular theme.

If you need to make other alterations

### Changes

* Converted all color variables to use the `color` function or `shade-dark / shade-light` functions
* Added a theme toggle to the header on web
* Added a theme toggle to the logo in Electron
* Persisting `state.config.meta.theme` in local storage
* Added a palette page at `/dev/palette` in dev to view the colors all in one spot

### TODOs

* [x] Check and fix shapeshift (I can't in NYC)
* [x] Prevent component refresh with theme change
* [ ] Come up with a better theme toggle on Electron (Clicking on logo is temporoary)
* [ ] (Future task?) Make a full-fledged settings screen

### Steps to Test

Browse the whole site in both themes, let me know if anything looks wrong: https://mycryptobuilds.com/34c45afbdcc3d5f1a91363fbd459f0aeaf697567/
Try to run the app as well.

### Web

screen shot 2018-06-05 at 3 31 45 pm

### Electron

screen shot 2018-06-05 at 3 41 13 pm


dternyak:
I identified some text color issued on the third step of ShapeShift:

screen shot 2018-06-12 at 10 57 57 am
screen shot 2018-06-12 at 10 57 50 am

wbobeirne:
Good catches @dternyak. Just adjusted the styles a bit to work for both. I ended up removing the circle's individual colors to be easier on the dark theme, and I think it's got more visual clarity this way. Let me know if you want anything reverted:

### Light

screen shot 2018-06-13 at 11 48 05 am


### Dark

screen shot 2018-06-13 at 11 47 47 am

wbobeirne:
Alright, I've made the following changes:

* Merged develop back in and resolved all conflicts
* Moved the web toggle to be in the footer
* Fixed up dark theme styles for Swap
* Reduced brightness of disabled buttons

I think this should be good to look at. The Electron toggle issue is still outstanding, but I think that will be better addressed by a proper settings screen at some point in the future.

avanimiriyala:
Only comment that I have is to decrease size of the switch for night mode slightly! I think if we decrease to be 75-80% of the size it is now, we'll be good to release. 

@dternyak

PR #1917, opened 2018-06-07
Revamp Offline Status Checker

### LINKED ISSUE #1915: Offline Transaction Signing Regression
1. Signing button is disabled when fields are filled out properly but balances / network requests haven't resolved.
2. The user gets several error messages that should be silenced when the user is offline.
3. The online status indicator takes a long time to resolve to the correct state (10-20 seconds, should be 1-2 seconds).

### Steps to Reproduce

1. Load mycrypto.com or the electron application.
2. Go offline.
3. Login with your mnemonic / private key / keystore on the account view.
4. Notice that spinners are still loading
5. Notice that you are unable to sign a transaction due to signing being disabled

Chrome, macOS

HenryNguyen5:
Closes #1915

### Changes

* Change offline detection method to be event emitter based, so it instantly reacts to offline states 

### Steps to Test

1. Load app while online
2. Online indicator should be green
3. Go to dev tools
4. Check `offline` box
5. App should go offline, but it'll do a check to see if it can reach any nodes it's connected to
6. If it can reach a node, itll go back online
7. Uncheck `offline` box, app should still remain online if steps 6 completed successfully
8. Completely disable your network connection 
9. App should go offline, node checks should fail, so app should stay offline while network connection is disabled
10. Re-enable your network connection
11. After a short period, if a node is online, app should regain connection to network and display being connected back to network
12. Login with wallet
13. Notice that nonce field is highlighted with a red border and is visible immediately.


james-prado:
Fixed the issue of invalid borders not showing up properly.

kapture 2018-06-08 at 8 51 19

common/components/ui/Input.scss
```diff
@@ -81,7 +81,7 @@
       color: $input-color-placeholder;
     }
     &:not([disabled]):not([readonly]) {
-      &.invalid.has-blurred.has-value {
+      &.invalid.has-blurred {
```
dternyak: We actually want the red border even when the field hasn't yet blurred, as it's a common pattern for users to be stuck and not realize they have to manually fill out nonce while offline.

@wbobeirne Any ideas on the best pattern here that won't affect all other inputs? Nonce is really a special case

wbobeirne: Seems to me the only fix here is a separate class and prop for this, since we need different functionality in different cases. Something like `showErrorWhileEmpty`, so that the selector is:
```
&.invalid.has-blurred.has-value,
&.invalid.has-blurred.show-error-empty {
  // ...
}
```

Alternatively, we could just make it so that the invalid boolean is always shown when passed, and change cases where we don't want it to while it doesn't have value to not pass it. I think both approaches are valid.

PR #1926, opened 2018-06-11
Fix paper wallet address & icon

### LINKED ISSUE #1824: Creating Keystore Wallet - Identicon does not display
### Description of the Issue

When you create a new wallet using Keystore file method, the Identicon is not loaded on the paper wallet, even when you download it. It does not appear. 

### Steps to Reproduce
Create a New wallet via Keystore file

image

wbobeirne:
Closes #1824

### Description

Fixes the paper wallet by hex prefixing and checksumming the address

### Steps to Test

1. Generate keystore file wallet
2. Proceed to print wallet phase
3. Confirm address is hex prefixed, checksummed, and has an identicon

### DIFF
--- a/common/components/ui/Identicon.tsx
+++ b/common/components/ui/Identicon.tsx
@@ -11,7 +11,6 @@ interface Props {
 export default function Identicon(props: Props) {
   const size = props.size || '4rem';
   const { address, className = '' } = props;
-  // FIXME breaks on failed checksums
   const identiconDataUrl = isValidETHAddress(address) ? makeBlockie(address) : '';
   return (
     // Use inline styles for printable wallets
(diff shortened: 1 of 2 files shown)


common/components/ui/Identicon.tsx
```diff
@@ -11,7 +11,6 @@ interface Props {
 export default function Identicon(props: Props) {
   const size = props.size || '4rem';
   const { address, className = '' } = props;
-  // FIXME breaks on failed checksums
```
wbobeirne: This comment isn't true, it was only on addresses with no hex prefix (which happens incredibly infrequently) so I felt it wasn't very actionable.

PR #1927, opened 2018-06-11
Bug Fix for #1813

### LINKED ISSUE #1813: [Regression] Pre-filled transaction values are no longer pre-filled
https://twitter.com/Mfourgeaud/status/996685708948574208?s=19

@wbobeirne -
"Yep, looks like a regression. Data and gas limit are filling though so it's not all the params that are busted. Looks like it's just to and value."

skubakdj:
Moves `resetTransactionRequested()` call from `componentDidMount` to `componentWillMount` in `TXMetaDataPanel.tsx`. Closes #1813.

Query params are parsed and set inside components, so resetting transaction state _after_ render was wiping it out.


Azarielle:
@skubakdj you are given this after you log into the wallet. I dont think you need the "Unlock your wallet to get started." since you've already unlocked your wallet 
image

wbobeirne:
^As an aside, I think it'd be good if that alert was either in the white box of content, or if it had no margin below it. Kind of weird that it detaches the tabs. Just while we're in the neighborhood...

PR #1933, opened 2018-06-11
Update ReadMe instructions with yarn requirement

dternyak:


dternyak:
Good catch @HenryNguyen5

PR #1892, opened 2018-06-04
Add RSK network with checksum

alepc253:
Closes #626

Description
Integrating RSK testnet network and a different checksum for it. Now, depending on the chainId it applies ethereumjs-utils checksum for every ETH addresses or rskjs-util checksum for RSK addresses.

Changes
New configuration. Instead of calling 'toChecksumAddress" now it calls 'toChecksumAddressByChainId' and instead of calling 'isValidETHAddress' now it calls 'isValidAddress' (with chainId as parameter).

Steps to Test
1) Connect to RSK network.
2) Verify that ETH addresses are rightly checksumed.
3) Verify that RSK addresses are rightly checksumed.
4) Verify that a ETH address doesn't work in RSK and vice versa.

PR #1939, opened 2018-06-12
Add RSK network w/ network agnostic refactors

### LINKED ISSUE #1937: Address Book allows for multiple addresses with different casing
### Description of the Issue

Looks like the comparison doesn't happen consistently for duplicate address, so multiple addresses are allowed.

### Steps to Reproduce

1. Save an address with the checksummed version of the address
2. Attempt to save an address with the same address, lowercased
3. Receive no error of duplicate address

screen shot 2018-06-12 at 11 18 00 am

wbobeirne:
Replaces #1892 and closes #1937.

### Description

This pull request is a refactor of #1892 to be done in a more redux-y way. Rather than passing a chainId around everywhere and putting it into util functions, this implements selector functions for grabbing validation and checksum functions. Components that need to use these functions just get them as props from selectors. Would like some feedback about this approach, as I think it will come in handy as more networks differ. But maybe it's unintuitive or cumbersome, so let me know.

To simplify the address book, this also refactors it to store all addresses as lower-case'd, and only checksum them at render. In general we should always checksum at render time, and do comparisons as lower case.

### Changes

* Create selectors `getChecksumAddressFn` and `getIsValidAddressFn` and use those where necessary.
* Rework addressBook to lowercase addresses whenever storing or comparing them (cc @ConnorBryan)
* Adds all of the RSK info from #1892 (cc @alepc253)

### Steps to Test

1. Connect to RSK network.
2. Verify that ETH addresses are rightly checksummed.
3. Verify that RSK addresses are rightly checksummed.
4. Verify that an ETH address doesn't work in RSK and vice versa.


alepc253:
Awesome news @wbobeirne! Happy to test this version!

dternyak:
@alepc253 Much appreciated! Let us know how your test goes. If everything looks good on your end, I'll do a final pass and then we can merge!

dternyak:
Thanks for helping to test. What issues did you encounter there?

alepc253:
@dternyak this is a screenshot of what I see. Hope it helps.

print_wallet_error

wbobeirne:
Thanks for the catch @alepc253. The paper wallet's been our most brittle component for a long time, and seeing as it's currently broken in the desktop app right now, I think it's time to better decouple it from the main app so it doesn't break all the time. This will probably happen in a separate PR on top of this one, which might hold this up for just a bit longer.

alepc253:
Just to clarify the PR: this RSK network configuration is for **testnet**. After this integration, the next step is going for mainnet (we are setting up a public mainnet node for MyCrypto).

common/components/PrintableWallet/index.tsx
```diff
@@ -1,53 +1,53 @@
...
+  public state: State = {
+    paperWalletImage: ''
+  };
+
+  private paperWallet: PaperWallet | null;
+
+  public componentDidMount() {
+    setTimeout(() => {
+      if (!this.paperWallet) {
+        return this.componentDidMount();
```
HenryNguyen5: What's this supposed to do?

common/components/ui/Address.tsx
```diff
@@ -1,40 +1,46 @@
...
-
-const Address: React.SFC<Props> = props => {
-  let addr = '';
-  if (isAddressProps(props)) {
-    addr = props.address;
-  } else {
-    addr = props.wallet.getAddressString();
+type Props = BaseProps & StateProps;
+
+// const isAddressProps = (props: Props): props is Props =>
```
HenryNguyen5: Can this be removed now?

common/components/ui/Address.tsx
```diff
@@ -1,40 +1,46 @@
...
+//   typeof (props as AddressProps).address === 'string';
+
+export class Address extends React.PureComponent<Props> {
+  public render() {
+    const { wallet, address, explorer, toChecksumAddress } = this.props;
+    let addr = '';
+    if (address != null) {
```
HenryNguyen5: This should be using strict inequality

common/components/ui/Address.tsx
```diff
@@ -1,40 +1,46 @@
...
+      addr = address;
+    } else {
+      addr = wallet != null ? wallet.getAddressString() : '';
```
HenryNguyen5: Ditto

common/components/ui/Address.tsx
```diff
@@ -1,40 +1,46 @@
...
-    addr = props.wallet.getAddressString();
+type Props = BaseProps & StateProps;
+
+// const isAddressProps = (props: Props): props is Props =>
+//   typeof (props as AddressProps).address === 'string';
+
+export class Address extends React.PureComponent<Props> {
+  public render() {
+    const { wallet, address, explorer, toChecksumAddress } = this.props;
+    let addr = '';
```
HenryNguyen5: Better variable naming here other than addr/address would be preferred

common/components/ui/Address.tsx
```diff
@@ -1,40 +1,46 @@
...
-  if (props.explorer) {
-    return <NewTabLink href={props.explorer.addressUrl(addr)}>{addr}</NewTabLink>;
-  } else {
-    return <React.Fragment>{addr}</React.Fragment>;
-  }
-};
+}
 
-export default Address;
+//export default Address;
```
HenryNguyen5: Can this be removed?

common/components/ui/Identicon.tsx
```diff
@@ -1,49 +1,43 @@
...
+  isValidAddress: ReturnType<typeof getIsValidAddressFn>;
+}
+
+type Props = OwnProps & StateProps;
+
+const Identicon: React.SFC<Props> = props => {
   const size = props.size || '4rem';
-  const { address, className = '' } = props;
-  const identiconDataUrl = isValidETHAddress(address) ? makeBlockie(address) : '';
+  const { address, isValidAddress, className = '' } = props;
```
HenryNguyen5: Prefer using the supplied defaultProps property rather than supplying a default destructure value here

HenryNguyen5: Noticed this isnt part of the PR, my bad

common/containers/Tabs/Swap/components/LiteSend/Fields.tsx
```diff
@@ -103,6 +103,9 @@ class FieldsClass extends Component<Props> {
 }
 
 export const Fields = connect(
-  (state: AppState) => ({ unit: getUnit(state), currentBalance: getCurrentBalance(state) }),
+  (state: AppState) => ({
```
HenryNguyen5: Prettier conflict?

common/libs/contracts/ABIFunction.ts
```diff
@@ -1,6 +1,7 @@
 import abi from 'ethereumjs-abi';
-import { toChecksumAddress, addHexPrefix, stripHexPrefix } from 'ethereumjs-util';
+import { addHexPrefix, stripHexPrefix } from 'ethereumjs-util';
 import BN from 'bn.js';
+import { toChecksumAddress } from 'utils/formatters';
```
HenryNguyen5: This could potentially get very confusing considering that the same function name exists in multiple places

common/components/PrintableWallet/index.tsx
wbobeirne: This _should_ never happen, but typescript complained that paperwallet could be null, so this is just in case that were to happen, it would try again in a sec. The initial setTimeout is to make sure everything  has a chance to load properly before PNGifying it (Things like  tags can take a blip.)

common/components/ui/Address.tsx
wbobeirne: Surprised the linter didn't bark at me for this.

common/containers/Tabs/Swap/components/LiteSend/Fields.tsx
wbobeirne: Not sure, I might have just indented for some reason while I was tinkering in here but ultimately made no changes.

PR #1942, opened 2018-06-13
[Fix] Estimate Gas on Value Field Change

### LINKED ISSUE #1873: Incorrect gas limit estimation when sending to contract
### Description of the Issue

The gas limit estimated when sending to a contract is not always correct 

### Steps to Reproduce

1. Go to send a transaction via MyCrypto

2. Enter the address `0x37f25bd26848cf02614c29338dd71a4d03b62218`

3. A gas limit of `21065` is estimated

4. If you send with this gas limit, the transaction will fail

For reference, this is the gnosis multisig wallet.

MyEtherWallet estimates `22511` 

MetaMask estimates `33766`.

This essentially makes MyCrypto unusable for sending to certain contracts and makes MEW/MetaMask a better choice. Sending with a gas limit of `22511` results in a successful transaction.

skubakdj:
Forces gas estimation on value field change. Closes #1873.

Main change here is having the `shouldEstimateGas` saga take on `VALUE_FIELD_SET`. However, making only this change created an infinite loop (the `revalidateCurrentValue` saga fires on gas estimate completion and sets the value field, which would kick off gas estimation again, etc). 

Fix was to have `revalidateCurrentValue` set a new value _only_ if it differs from the previous one. Logic for that is contained in the `isValueDifferent` function.


spec/reducers/config/__snapshots__/config.spec.ts.snap
```diff
@@ -5,7 +5,7 @@ Object {
   "@@redux-saga/IO": true,
   "SELECT": Object {
     "args": Array [
-      "ELLA",
+      "CLO",
```
skubakdj: Not sure why this snapshot had to be updated. I didn't touch anything related to network config. . .

PR #1945, opened 2018-06-13
[Parity Signer] Add future support for ERC-681

maciejhirsz:
### Description

Prepares the Parity Signer address scanning for the future versions that will use ERC-681 formatted strings ([more context](https://github.com/paritytech/parity-signer/issues/103)).

This change is backwards compatible, old (current) versions of Parity Signer should work without changes.

### Changes

* Updated `@parity/qr-signer` to `0.3.1`
* `onScan` will now callback with an object containing `address` and `chainId`, so made appropriate adjustments in `ParitySigner.tsx`.

### Steps to Test

1. Send a tx using Parity Signer.
2. Try to unlock an account by showing a constructed QR code containing ERC-681 URL, such as `ethereum:0x5555555555555555555555555555555555555555@42` (chainId is being parsed, but we throw it away right now)


maciejhirsz:
Note: I got tscheck errors when trying to push, but the same errors were thrown on `develop` branch without changes

HenryNguyen5:
If you're using npm its known to throw ts errors due to version mismatch, yarn should work fine

common/containers/QrSignerModal/index.tsx
```diff
@@ -94,7 +94,7 @@ class QrSignerModal extends React.Component<Props, State> {
       return;
     }
 
-    this.props.finalizeSignature(signature);
+    this.props.finalizeSignature('0x' + signature);
```
dternyak: TODO (for myself): circle back and provide utility function for this.

wbobeirne: `addHexPrefix` does this from `ethereumjs-util`

dternyak: @maciejhirsz You cool with making this small tweak? No reason not to merge after imo.

maciejhirsz: Sure!

PR #1946, opened 2018-06-13
Electron TREZOR Support

wbobeirne:
Builds off of #1836, finishes the other half of #1621.

### Description

Implements the TREZOR side of hardware wallet integration in Electron using the enclave library introduced in #1621. Similar to that PR, there's an unfortunate amount of code duplication between the two wallet implementations. Any feedback on how to better unite those would be appreciated.

See "Steps to Test" for note about pending PR. I have personally tested that this all works.

### Changes

* Implemented TREZOR on Enclave side
* Re-enabled TREZOR wallet in Electron environment
* Typed up all TREZOR libraries (including the vendor'd `TrezorConnect`.)

### Steps to Test

**NOTE:** This is still pending on https://github.com/trezor/trezor-link/pull/9. If you wish to test this PR locally, you will have to manually make the changes proposed their to `node_modules/trezor-link/lib/http.js`.

1. Start electron app
2. Unlock a TREZOR wallet address
3. Display the address on your device
4. Sign a transaction, confirm it looks good
5. Repeat those steps in the web app


common/libs/wallet/deterministic/trezor.ts
```diff
@@ -42,8 +41,24 @@ export class TrezorWallet extends HardwareWallet implements IFullWallet {
 
   public signRawTransaction(tx: EthTx): Promise<Buffer> {
     return new Promise((resolve, reject) => {
-      const { chainId, ...strTx } = getTransactionFields(tx);
+      const txFields = getTransactionFields(tx);
+
+      if (process.env.BUILD_ELECTRON) {
```
wbobeirne: It would be great to make a class that extends HardwareWallet that does this for you. @HenryNguyen5 how would you suggest this best be done?

package.json
```diff
@@ -89,6 +91,7 @@
     "@types/webpack-env": "1.13.4",
     "@types/zxcvbn": "4.4.0",
     "autodll-webpack-plugin": "0.3.9",
+    "bchaddrjs": "0.2.1",
```
wbobeirne: Note to self, this was fixed with https://github.com/trezor/trezor.js/pull/58 so I can remove this.

common/libs/wallet/deterministic/trezor.ts
HenryNguyen5: I'd probably just make a utility HOF instead that does the branching for you

PR #1971, opened 2018-06-20
Allow PIN Entry for TREZOR One in App

### LINKED ISSUE #1960: TREZOR One stuck entering PIN on desktop app
### Description of the Issue

TREZOR One (the original two button model) devices get stuck at the "enter PIN" phase of doing anything with the wallet. Because the electron app doesn't use `trezor-connect`, there's no UI for entering a PIN, so this will have to be done on our end.

### Steps to Reproduce

1. Attempt to unlock a TREZOR one on the desktop app
2. Be told to enter PIN with no prompt

wbobeirne:
Closes #1960.

### Description

All testing for TREZOR integration was initially done on a TREZOR T, which has on-device PIN entry. After finishing it, we realized that the TREZOR One needs a way to enter it in the app. This allows for that.

I'll start off by saying that the code for this is _very_ ratchet. Because Enclave is very separate from the main codebase as it's more of an API than anything, presenting the user with UI-layer things is rather difficult. The way that's done here is by going directly to Electron to pop up a separate window. Since we don't have access to Webpack to output files using the nice things we're used to, like SCSS and JSX, I instead opted to make the window as bare-bones as possible with raw CSS, HTML, and Javascript. As far as I can tell, the security of this window shouldn't be an issue since it loads only local content, and has a CSP that ensures no additional scrips run.

Another possible option here would have been to reject the request if a PIN is required, and have them resubmit the request with a PIN entered attached (After parsing the error.) I didn't go this route for a few reasons: One is that this is very TREZOR specific handling which would have bled into other wallet implementations. Another is that this is app-specific, but would have to be mixed with the web unlock flow. However, if the feeling is that this approach is much preferred, I'd be happy to rework it.

### Changes

* Added a PIN handler to the Enclave TREZOR wallet
* Switched from `mycrypto-trezor.js` back to `trezor.js`

### Steps to Test

1. Plug in your TREZOR One device
2. Open the app on the `trezor-native-pin` branch
3. Attempt to open the TREZOR wallet
4. Be prompted for the PIN
5. Close the window, ensure you get a proper rejection error.
6. Try again, this time entering the wrong PIN, and ensure you get another error.
7. Try again, this time entering the right PIN, and ensure it succeeds.

### Screenshots

screen shot 2018-06-20 at 12 07 41 pm

### DIFF
--- a/shared/enclave/server/views/pin.ts
+++ b/shared/enclave/server/views/pin.ts
@@ -1,0 +1,246 @@
+import { BrowserWindow, ipcMain, IpcMessageEvent } from 'electron';
+
+const EVENT = 'enclave:pin';
+const scriptNonce = Math.floor(Math.random() * 1000000000000);
+const html = `
+<html>
+  <head>
+    <title>TREZOR - Enter PIN</title>
+    <meta http-equiv="Content-Security-Policy" content="script-src 'self' 'nonce-${scriptNonce}'">
+    <style>
+      body {
+        color: #FFF;
+        padding: 20px;
+        margin: 0;
+        font-family: 'Helvetica Neue', Helvetica, Arial, sans-serif;
+        box-sizing: border-box;
+      }
+      body * {
+        box-sizing: inherit;
+      }
+
+      h1,
+      h2 {
+        text-align: center;
+      }
+      h1 {
+        font-size: 28px;
+        font-weight: 100;
+        margin-bottom: 3px;
+        letter-spacing: 1.2px;
+      }
+      h2 {
+        opacity: 0.5;
+        font-size: 13px;
+        font-weight: 400;
+      }
+
+      .pin-buttons {
+        display: flex;
+        flex-wrap: wrap;
+        justify-content: space-between;
+        align-items: center;
+        margin: 0 auto;
+        width: 180px;
+        height: 180px;
+      }
+      .pin-button {
+        position: relative;
+        width: 54px;
+        height: 54px;
+        border-radius: 4px;
+        border: 1px solid #FFF;
+        background: rgba(255, 255, 255, 0.1);
+        transition: all 120ms ease;
+        opacity: 0.4;
+      }
+      .pin-button:hover,
+      .pin-button:active {
+        opacity: 0.8;
+      }
+      .pin-button:after {
+        content: '';
+        position: absolute;
+        width: 8px;
+        height: 8px;
+        top: 50%;
+        left: 50%;
+        transform: translate(-50%, -50%);
+        border-radius: 100%;
+        background: #FFF;
+      }
+
+      .pin-controls {
+        position: relative;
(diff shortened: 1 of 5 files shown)


shared/enclave/server/views/pin.ts
```diff
@@ -0,0 +1,246 @@
+import { BrowserWindow, ipcMain, IpcMessageEvent } from 'electron';
+
+const EVENT = 'enclave:pin';
+const scriptNonce = Math.floor(Math.random() * 1000000000000);
+const html = `
+<html>
```
wbobeirne: If anyone can think of a reasonable way to put this into its own HTML file and require it without having some webpack HTML handler, I'm all ears.

HenryNguyen5: Is node `fs` disabled here? Could just read it in as a text file

shared/enclave/server/views/pin.ts
```diff
@@ -0,0 +1,246 @@
...
+
+      close.addEventListener('click', function() {
+        closeWindow();
+      });
+    </script>
+  </body>
+</html>
+`;
+
+export async function showPinPrompt(): Promise<string> {
```
HenryNguyen5: No need for the `async` keyword here

shared/enclave/server/views/pin.ts
```diff
@@ -0,0 +1,246 @@
...
+  return new Promise<string>((resolve, reject) => {
```
HenryNguyen5: Shouldnt need to specify the generic here either, can keep it in the function return statement (or even leave that empty too and let TS infer it)

shared/enclave/server/wallets/trezor.ts
```diff
@@ -25,6 +25,15 @@ async function getSession() {
       currentSession = null;
     }
   });
+  device.on('pin', (_, cb: (err: Error | undefined, pin?: string) => void) => {
```
HenryNguyen5: Might as well keep it consistent and have `err` parameter use the optional type syntax as used in `pin`

wbobeirne: My thought was that I didn't want you ever calling the callback with neither error nor pin, so I figured making err explicitly undefined would be good for that. But it's definitely more succinct to go optional, so I'll do that.

shared/enclave/server/views/pin.ts
wbobeirne: I thought so too, but it was barking at me that it was implicitly any so I just typed it to stop it. Not sure why it wasn't inferring it here.

wbobeirne: Nvm removing it gave me no errors. Thanks!

wbobeirne:
Thanks for the feedback @HenryNguyen5, everything's been addressed.

PR #1982, opened 2018-06-22
Add additional validation to Amount input field

### LINKED ISSUE #1789: Request Payment - Amount can be 3.4.56.6
---

If you've experienced a bug, please provide the following information:

Amount field for Request payment

image

---

ConnorBryan:
Closes #1789

### Description
Users were able to enter a value with multiple decimal separators without being shown an error.

### Changes

* The `validDecimal` validator now checks for only a single decimal.
* `AmountField`s will now show an error after being touched if there is no value.

### Steps to Test

1. Navigate to the "Request Payment" tab in "Account".
2. Enter a value with multiple decimals (such as "1.2.3").
3. The field should become red after blurring, indicating an error.

### Screenshots
screen shot 2018-06-22 at 5 16 19 pm


ConnorBryan:
@dternyak 

I reverted most changes to `ui/Input` and instead opted to allow `<AmountField />` to pass down `showInvalidWithoutValue` if desired. That way, the fields will error when empty once they've been touched.

This issue stems from the fact that entering `1.2.3` into an `<input type="number" />` is the same thing as entering an empty string. When the value goes from `""` to `"1.2.3"`, the reducer doesn't make any changes, and the validity logic for input was checking only for `!hasValue && !showInvalidWithoutValue`, which is usually going to be the case.

dternyak:
@ConnorBryan The steps are as follows:

1. Load the initial view.
2. Go offline on your network.
3. Using an offline supported format (e.g. mnemonic), unlock your wallet.
4. Notice that the nonce field is highlighted red.

dternyak:
screen shot 2018-06-27 at 12 16 23 am

ConnorBryan:
@dternyak 
screen shot 2018-06-29 at 5 32 21 pm

It looks like the latest commit is making it work as expected.

PR #1988, opened 2018-06-26
Handle MetaMask network change

### LINKED ISSUE #1966: Handle MetaMask deprecating page reload on network change
https://medium.com/metamask/breaking-change-no-longer-reloading-pages-on-network-change-4a3e1fd2f5e7

ConnorBryan:
Closes #1966

### Description

Because MetaMask has chosen to stop reloading pages on network change, we need to implement our own solution when a change is detected.

See: https://medium.com/metamask/breaking-change-no-longer-reloading-pages-on-network-change-4a3e1fd2f5e7

### Changes

* Once per second, MetaMask's network is compared to the local version.
* If the versions don't match, the user will be logged out and a notification alerting them to the change will appear.

### Steps to Test

1. Access a wallet using MetaMask.
2. Switch the network in the MetaMask browser extension.
3. The wallet should be reset and you should be logged out.

**NOTE: This is going to be hard to test while the extension still refreshes the page, as it seems to take precedence over everything else. I tested this "in code" by changing window.web3 to return an arbitrary chain ID, but I'm not sure how we can test this before the change to the extension.**


common/features/handleMetaMaskPolling.spec.ts
```diff
@@ -0,0 +1,108 @@
...
+      getState: noop,
+      dispatch: noop
+    };
+    const result = await handleMetaMaskPolling(store as any);
+
+    expect(result).toBe(false);
+    done();
+  });
+
+  it('should display a notification and reset the wallet if the network has changed', async done => {
```
dternyak: This test should be adjusted with the recent page refresh afaict.

common/features/handleMetaMaskPolling.ts
```diff
@@ -0,0 +1,52 @@
...
+    );
+  });
+
+/**
+ * @desc
+ * MetaMask no longer refreshes the page automatically on network change,
+ * so we must poll to ensure the network is the same as the locally stored version.
+ * @see https://medium.com/metamask/breaking-change-no-longer-reloading-pages-on-network-change-4a3e1fd2f5e7
+ */
+export default async function handleMetaMaskPolling(store: Store<AppState>): Promise<boolean> {
```
dternyak: is a return value useful here?

ConnorBryan: It's certainly not harmful, and prevents me from doing ridiculous mocking to make `window.location.reload` testable (due to inherent limitations with the headless chrome used by jest)

common/features/handleMetaMaskPolling.spec.ts
ConnorBryan: Addressing

PR #1997, opened 2018-06-30
Allow contract value to be cleared

### LINKED ISSUE #1903: Selecting Existing Contract - x (clear value does not work)
The x to clear selected data in the field does not work. 

image

mwbailey:
Closes #1903

### Description

Fixes an issue where clicking the 'x' to clear the selected contract has no effect. This also allows a user to enter a custom contract after an existing contract had been selected. Previously, the ABI/JSON Interface field would never switch back to a text field once it had become a code block (when an existing contract is chosen).

### Changes

* When passing props to the Select component, move `{...this.props}` above `value` and `onChange` to avoid overriding those props with previous values.
* Add parameter when calling `onChange()` so that InteractForm can update the contract address and ABI/JSON when the contract is changed.
* Handle condition in InteractForm where there is no selected contract. The contract address and ABI/JSON values are cleared, and ABI/JSON is reset to a text field.
* Initialize Dropdown's `selectedOption.value` to `undefined` so the initial display is the placeholder ("Select a contract...") instead of the first option (Custom).

### Steps to Test

1. Go to Contracts tab
2. Select any existing contract
3. Click on cross button and confirm that all fields are cleared and the ABI/JSON Interface is now a text field


dternyak:
Thank you!

PR #2011, opened 2018-07-04
Hide unsupported tabs by RSK

### LINKED ISSUE #2010: RSK unsupported tabs

To hide tabs that don't make sense (up to now) in RSK network by using configuration.

alepc253:
Closes #2010

### Description

This change hides tabs not supported (up to now) by both RSK testnet and mainnet. It avoids errors and misundertstandings.

### Changes

* Adds `TAB.ENS` and `TAB.SWAP` to `unsuppportedTabs` field in `StaticNetworksState`

### Steps to Test

1. Select RSK network and check that ENS and Swap tabs are not visible.
2. Change to other network and check that ENS and Swap tabs do are visible.

### DIFF
--- a/common/features/config/networks/static/reducer.ts
+++ b/common/features/config/networks/static/reducer.ts
@@ -29,6 +29,7 @@ import {
 } from 'config/dpaths';
 import { makeExplorer } from 'utils/helpers';
 import { StaticNetworksState } from './types';
+import { TAB } from 'components/Header/components/constants';
 
 const testnetDefaultGasPrice = {
   min: 0.1,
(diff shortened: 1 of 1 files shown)


common/features/config/networks/static/reducer.ts
```diff
@@ -393,7 +394,8 @@ export const STATIC_NETWORKS_INITIAL_STATE: StaticNetworksState = {
       min: 0.183,
       max: 1.5,
       initial: 0.183
-    }
+    },
+    unsupportedTabs: [TAB.SWAP, TAB.ENS]
```
dternyak: swap should be supported still - no reason to disable it if people want to swap between various other coins/tokens :)

alepc253: @dternyak we are trying to avoid misunderstandings about the network, but it's true that this feature considers other networks

PR #2020, opened 2018-07-06
Make SimpleGas default to recommended gas price on initial load

### LINKED ISSUE #1999: Gas slider defaults to far left, fixes when checking/unchecking Advance
OS: Windows
Browser: Chrome Version 67.0.3396.99 (Official Build) (64-bit)
via: MetaMask

Steps to replicate:
Go to MyCrypto.com
Click MetaMask
"Connect to MetaMask"

### **Upon getting to the view/send page, I see the slider is set to the far left price (currently 50 gwei):**
ss 2018-07-01 at 08 13 34

### **Upon clicking the "Advanced" button, the gwei does not change from 50 gwei:**
ss 2018-07-01 at 08 13 42

### **Upon clicking the "Simple" button to close the advanced section, the gas price seems to recalculate (or actually calculate for the first time):**
ss 2018-07-01 at 08 13 47

ConnorBryan:
Closes #1999

### Description

On wallet selection, the gas slider was defaulting to the lowest price in the gas estimation, instead of the recommended price.

### Changes

* `SimpleGas` now passes a value that is always within the calculated boundaries.
* `rc-slider` automatically triggers the `onChange` callback when the passed value is out-of-bounds. This was inadvertently triggers a gas price change based on an old value.

### Steps to Test

1. Unlock a wallet.
2. The gas price slider should jump to the recommended price after loading.

### Other
Notes from Slack:

So it turns out:

1. The wallet loads.
2. Because the SimpleGas slider loads, the gas estimates are fetched.
3. During the initial render, the boundaries for the slider are set. Because there are no values for reference, the bounds are *1* - *60*. The value is set to the initialValue in the TransactionFields reducer, which is 20.
4. Once the gas estimates fetch is retrieved, the gasEstimates are set, causing SimpleGas to re-render. When it calculates the new boundaries, it uses a new min and max, based on the gas estimates. The min is now 41, and the max becomes 85. *BUT WAIT.* Because the current value (20) is less than the lower boundary (41), *THE SLIDER AUTOMATICALLY MOVES ITSELF*. Moving the slider triggers a `setGasPriceIntent`, which attempts to set the new gas price to `41`, even though `41` was never selected in the first place.
5. The original calculation from the gas estimation retrieval completes, setting the TransactionFields gasPrice property to 55, which is the `fast` property in the estimates.
6. The newer `handleGasFieldChange` from the `setGasPriceIntent` has finished firing with that minimum boundary of 41, so TransactionFields gasPrice goes back to 41.
7. Any other action with the slider occurs as expected, including loading Advanced then back to Simple.


common/components/TXMetaDataPanel/components/SimpleGas.tsx
```diff
@@ -76,6 +76,16 @@ class SimpleGas extends React.Component<Props> {
...
 
+    /**
+     * @desc On retrieval of gas estimates,
+     *  the current gas price may be lower than the lowest recommended price.
+     *  `rc-slider` will force the onChange if the value is too low, so we
+     *  ensure it at least passes the lower boundary.
+     *  When this occurs, the logic in `UNSAFE_componentWillReceiveProps` fires,
+     *  and it cannot happen again from that point forward.
+     */
+    const actualGasPrice = Math.max(this.getGasPriceGwei(gasPrice.value), bounds.min);
```
ConnorBryan: I believe this will work fine, because the time between the value still being the initial default and it being changed as a result of `componentWillReceiveProps` in `SimpleGas` is almost instant.

PR #2051, opened 2018-07-11
Adds support for checking addresses against the MetaCert Protocol registry

kams451:
### Description

Adds support for checking addresses against the MetaCert Protocol registry that provides a visual indicator of the safety of an address before a user sends their cryptocurrency.

### Changes

* Query the MetaCert Protocol registry via API request to return the state of an address inputted into a field, which includes 4 states:
  * Verified
  * NOT Verified
  * Phishing
  * Pending Investigation
* Add visual indicator overlay to Ethereum identicons when an address is inputted into a field

### Steps to Test

1. Log in to an account.
2. Input an address into the "To Addresss" field
3. Alternatively, input an address into the Address Book (e.g., 0x4bbeeb066ed09b7aed07bf39eee0460dfa261520 MyCryptoDonate)
4. Visual indicator is displayed overlayed on the Ethereum identicon

**Note:** Visual indicator will also be displayed on MyCrypto user addresses in the Request Payment and Wallet Info tabs.

### Screenshots

mycrypto_verified

mycrypto_phishing

mycrypto_notverified

mycrypto_address_book


tayvano:
We will need to have devs do an actual code review before merging, but a few things up front:

1. Make icons a bit smaller

2. Remove icon if doesn't have a definitive status (red or green)

Rationale: we are actively working on improving how we educate users about the blockies/identicons and covering them with another icon is going to lead to confusion. I'm 100% for having the red / green icons and educating them about verified and not verified but the gray doesn't add anything to the average users' experience.

Something closer to this size:

2018-07-17 at 6 03 00 pm

3. http://middleware.seeksmarterdev.com/check is not SSL—it will need to be. Getting an error on https://mycryptobuilds.com/e7323b179a29bcfba9ae47360381ec1f1e2613d6/

4. We will actually cache / proxy this endpoint behind our infra as we do with all APIs we rely on. (Danny S or Mia will need to set this up)

5. If the request does fail, it shouldn't cause the identicon to not show at all, it should only not show the Metacert icon but the addressicon / blockie should remain.

6. Ensure that even if endpoint is compromised, nothing could be affected on our website / our users. Should ensure strict tests / checks for any and all calls out are adhered to.

MrLuit:
So basically while a feature like this would probably be very useful to the end user, some problems I see with this are:

- The addresses are sent directly to the API which means (theoretically) a database could be built linking IP addresses to Ethereum addresses, which is obviously dangerous. This could possibly be solved by hashing every address (and taking the first x chars?) and maybe proxying/caching all requests. Even in that case hashes could very easily be bruteforced by grabbing recent transaction data and the proxy would still have the users IP address

- As Taylor said, the API doesn't use https, which makes it vulnerable for man-in-the-middle attacks that could read or modify data sent from user to API and from API to user. This could be a major problem on public wifi networks

- The data is not open source and doesn't return any reason as to why an address is marked as malicious. If all the data were to be hosted on GitHub, more people could help out on identifying and adding addresses to the database and you would also create another level of transparency

I honestly would like to see this implemented but maybe it would be better off to be implemented in Cryptonite so users aren't forced to use it on MyCrypto

kams451:
Thanks @tayvano and @MrLuit  for feedback. We will have an updated pull request in next few days based on the feedbacks.

Azarielle:
[Added to Asana](https://app.asana.com/0/763966770926912/779983315837069)

CryptoQueen628:
----

PR #2056, opened 2018-07-13
Discard signed message after leaving tab

### LINKED ISSUE #2035: Sign message code box state persists even when switching tabs
Signing a message and then switching tabs and doing other stuff, when you return, your signed message still appears. 
image

mwbailey:
Closes #2035

### Description

Fixes an issue where the previously signed message was still visible after leaving the "Sign Message" tab and then returning.

### Changes

* Added new Redux action to clear `state.message.signed`
* Clear the signed message whenever the SignMessage component is unmounted or the "Change Wallet" button is clicked, so the previous value won't get passed in as a prop to SignMessage
* Clear the message input field when "Change Wallet" is clicked, so the placeholder text is displayed instead of the previous input

### Steps to Test

1. Go to Sign Message tab
2. Sign a message
3. Click on any other tab
4. Go back to Sign Message and confirm that the previous signed message isn't there


dternyak:
Thanks for the excellent PR @mwbailey!

mwbailey:
Thanks, @dternyak! Glad to help.

PR #2071, opened 2018-07-14
Remove assumption about property names in Dropdown's options

### LINKED ISSUE #2042: Selecting contract function does not appear in textbox
View video for demonstration. This happens on all contracts and even on different nodes. 

https://azarielle.tinytake.com/sf/Mjc1MDA2NV84MjQ4MzEw

mwbailey:
Closes #2042 

### Description

The Dropdown component is given an array of options as a prop. If these option objects don't have a `value` property, the currently selected option isn't being passed as a prop to Select, which results in the selected option not being displayed (though functionally, it still works).

### Changes

This PR removes the assumption about property names by passing in the entire `selectedOption` as Select's initial value. The react-select documentation isn't very explicit about this, but it handles the case where the initial value is an object instead of a string or number.

### Steps to Test

1. Go to Contracts tab
2. Select any contract and choose any function
3. The function name should be displayed in the Read/Write Contract dropdown menu


dternyak:
@mwbailey 👍 Thanks a lot! If you're interested, email me at dt at mycrypto and I'll get you into our slack! We really appreciate you taking the time to close these issues.

PR #2132, opened 2018-08-24
Purge unused components

Sonicrida:
### Description

Purged unused components from the codebase

### Changes

* Removed 11 components and related references.
* Removed two files that had "not yet implemented" functions

### Steps to Test

1. Checking that builds work
2. Smoke Test


Sonicrida:
I had some trouble building the downloadable locally. Might be a windows thing 😞 but the dev server ran fine and was able to go through flows there.

The `donate.tsx` component that was removed also had references to some translations. Should those be removed as well?

HenryNguyen5:
Hey @Sonicrida, I think it would make sense to remove the translations if they're not used anywhere else.

Sonicrida:
Thanks Henry! I merged in those new changes and also removed the translation text.

PR #2146, opened 2018-09-01
Release/1.2.3.3

ConnorBryan:
Onboarding modal could theoretically load in dark theme if a previous user switched to dark theme and then upgraded their app. Also show the onboarding modal on desktop.


Azarielle:
1bd88a54987f43b9b6d582695b8e022d49b53608f1eedf5dc52a25df2ada35a7

PR #2160, opened 2018-09-12
Remove (X) button when changing wallets

Sonicrida:
### Description

Removes unnecessary (X) button on the wallet selection screen after a user clicks "Change Wallet". Users should not be able to log back in without first going through the proper method.

### Changes

* Removed the (X) button from UnlockHeader.tsx
* Removed styles for the (X) button since they aren't being used anywhere else

### Steps to Test

1. Login with the wallet of your choice
2. Click change wallet
3. The (X) button should be gone

### Screenshots

*Before*
image

*After*
image

### DIFF
--- a/common/components/ui/UnlockHeader.scss
+++ b/common/components/ui/UnlockHeader.scss
@@ -14,26 +14,4 @@
       padding-right: 8px;
     }
   }
-
-  &-close {
-    @include reset-button;
-    position: absolute;
-    top: 6px;
-    right: 6px;
-    margin: 0.5rem 1rem;
-    z-index: 1;
-    line-height: 24px;
-    font-size: 24px;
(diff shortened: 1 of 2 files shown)


mkrufky:
I have found the (x) button to be very helpful when I've accidentally clicked on "Change Wallet" and just want to get back to where I was beforehand.

I am new here, but personally, I'd vote to keep the button intact.

Azarielle:
@mkrufky I understand the ease of coming back to your wallet if you accidentally click "Change wallet", however, the main concern is security. If you were on your bank's website, would you really want it to be easy to log back into your account? I know I'd prefer to go through the security motions again.

mkrufky:
I understand - thank you for the explanation.  :-)

PR #2188, opened 2018-09-28
Fix ENS subdomain resolution

### LINKED ISSUE #1977: Correct the way features/ENS handles ENS subdomains
The current methodology for resolving a subdomain is not correct. As it stands, the entire entry (including period delimiters) is tested as a single unit, instead of looking at the subdomain registry for a given entry.

M-Picco:
Closes #1977 

### Description

As of now, the ENS resolution process doesn't work for subdomains (e.g. "wallet.mydomain.eth"). This PR aims to solve this issue.

Additionally, the changes allow new tlds to be added (e.g. [.xyz](https://medium.com/the-ethereum-name-service/announcing-support-for-xyz-on-ens-7f5bc7fe1b24)), since resolution is no longer forced under the `.eth` tld.

### Changes

Don't split the input address into `{name}.{tld}` in `/transaction/sagas.ts` and leave it to `/ens/sagas.ts`. This change also removes the hardcoded `.eth` used for the resolution process.

Add an auxiliary function to extract the domain from the address to perform the auction status check with.

### Steps to Test

1. Go to `account/send`
2. Type `fantastique.eth` (or any other known domain) in the address input. It should resolve to an address.
3. Type `wallet.etherbase.eth` (or any other known subdomain) in the address input. It should resolve to an address.

### Screenshots

domain_resolution

subdomain_resolution

### DIFF
--- a/common/features/ens/helpers.ts
+++ b/common/features/ens/helpers.ts
@@ -5,7 +5,7 @@ import ethUtil from 'ethereumjs-util';
 import networkConfigs from 'libs/ens/networkConfigs';
 import { INode } from 'libs/nodes/INode';
 import ENS from 'libs/ens/contracts';
-import { IDomainData, NameState, getNameHash, IBaseDomainRequest } from 'libs/ens';
+import { IDomainData, NameState, getNameHash, IBaseDomainRequest, extractDomain } from 'libs/ens';
 import * as configNodesSelectors from 'features/config/nodes/selectors';
 
 //#region Make & Decode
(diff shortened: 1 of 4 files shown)


common/features/ens/helpers.ts
```diff
@@ -89,9 +89,10 @@ const modeMap: IModeMap = {
...
 
-export function* resolveDomainRequest(name: string): SagaIterator {
-  const hash = ethUtil.sha3(name);
-  const nameHash = getNameHash(`${name}.eth`);
+export function* resolveDomainRequest(fullDomain: string): SagaIterator {
+  const domain = extractDomain(fullDomain);
+  const hash = ethUtil.sha3(domain);
+  const nameHash = getNameHash(fullDomain);
 
   const domainData: typeof ENS.auction.entries.outputType = yield call(makeEthCallAndDecode, {
```
Arachnid: Perhaps this should only be done if the name is a direct subdomain of 'eth'?

M-Picco: I thought it wouldn't be necessary to force the 'eth' tld since there are already checks [here](https://github.com/MyCryptoHQ/MyCrypto/blob/develop/common/components/AddressFieldFactory/AddressInputFactory.tsx#L89) and [here](https://github.com/MyCryptoHQ/MyCrypto/blob/develop/common/features/transaction/sagas.ts#L47) that make sure the name to resolve is under 'eth'.

Besides, current implementation forces an auction check regardless the tld is managed by an auction registrar or no. Future works (e.g. using a `tld => resolve function` map) could use the actual tld.

PR #2194, opened 2018-10-05
Require Node 8 (bans Node 9) and remove npm from engines

nickmccurdy:
### Description

Our engines field of package.json currently allows Node 9 to be installed, which is no longer supported. Unfortunately because upath's engines do not support Node 10 (we may be able to work around this with a simple `yarn upgrade` though), Node 9 is the latest version of Node that a user may try, even though it's not a good idea in terms of security and stability. As a precaution I limited allowed Node versions to 8 only, reducing confusion, increasing consistency, and preventing issues with Node 9's lack of maintenance.

### Changes

* Node 9 is not allowed (Node 10 already isn't allowed because of engines in one of our dependencies, upath)
* Remove npm from engines for consistency, since we're using Yarn


nickmccurdy:
What's the current status of this, should I work on it more? Node 8 and 9 do work with UI. It would be convenient if we could update to latest, though I'm not sure how much work that would involve and if it's worth prioritizing over features.

nickmccurdy:
I want to update Node anyway, so this probably isn't necessary anymore.

PR #2204, opened 2018-10-11
Bug/Don't attempt to modify state with timeout when <WalletDecrypt /> is unmounted

Sonicrida:
Closes Core-43

### Description

Stops a timeout in `<WalletDecrypt />` from attempting to `setState` after the component has been unmounted

### Changes

* Added a type to the state interface to store the timeout ID
* Added logic to clear the timeout attempting to modify state when `<WalletDecrypt />` is unmounted using the stored timeout ID

### Steps to Test

1. Add debug code to display when a timeout is running and cleared
2. Login to a wallet 
3. Log out of the wallet to see if the timeout was cleared when the `<WalletDecrypt />` component was unmounted


common/components/WalletDecrypt/WalletDecrypt.tsx
```diff
@@ -213,7 +214,8 @@ const WalletDecrypt = withRouter<Props>(
     public state: State = {
       selectedWalletKey: null,
       isInsecureOverridden: false,
-      value: null
+      value: null,
+      walletTimeoutID: undefined
```
ConnorBryan: @Sonicrida 
The component doesn't utilize walletTimeoutId as part of it's visual output, so it can be an instance variable, `this.walletTimeoutID`. Also, default it to `null`; `undefined` is semantically better for values that you haven't defined -- defining an `undefined`.

common/components/WalletDecrypt/WalletDecrypt.tsx
```diff
@@ -74,6 +74,7 @@ interface State {
   selectedWalletKey: WalletName | null;
   isInsecureOverridden: boolean;
   value: UnlockParams | null;
+  walletTimeoutID: number | undefined;
```
ConnorBryan: @Sonicrida 

This should be `number | null` as an instance variable, because it doesn't affect the UI output. Also, undefined is semantically better for values that you haven't defined -- defining an `undefined`.

If you need to denote that a variable may not be coming in, you can use `exampleVariable?: number`. The `?` lets it know it may be `undefined`.

common/components/WalletDecrypt/WalletDecrypt.tsx
```diff
@@ -226,6 +228,10 @@ const WalletDecrypt = withRouter<Props>(
       }
     }
 
+    public componentWillUnmount() {
+      window.clearTimeout(this.walletTimeoutID);
```
Sonicrida: I am getting this error here and wasn't quite sure about how to handle this 

image

PR #2220, opened 2018-10-18
[Parity Signer] Do not ask to scan an address for a known account twice

fckt:
### Description

This PR adds an ability to skip the initial `scan QR` to copy a sender address from the Parity Signer app if its was already scanned before. Currently it uses the `address book` as it is on the `View & Send -> View Address` screen. 

Also this PR fixes the wrong messaging: instead of saying `Unlock your Parity Signer` (which is not truth, we do not "unlock" anything on that stage), I've changed it to `Select an account from your Parity Signer`. 
Here is a reasoning behind that change: the process is different from a case with hardware wallets: in a case of Singer, the actual `account unlocking` happens right (and only) at the moment of signing on the device, never before and never after. Suggestions on further improving of the messages are welcome!

### Steps to Test

1. Go to `View & Send`
2. Select `Parity Signer`
3. Scan a QR of account, stored in your Parity Signer
4. Click on `Change Wallet`
5. Go to `View & Send`
6. Select `Parity Signer`
7. Tap on `Use a recent account`
8. There should be an account you're scanned recently - you'll be able to select an account with address you've scanned recently.

### Screenshots
screen shot 2018-10-18 at 3 57 17 pm

### Further improvements

- Use a distinct `account suggestions list`, which is specific to the Signer. E. g. show addressed which was populated only using the QR scanned on the `Parity Signer` screen.
- Use ERC-681 to automatically change network for the specific account. It applies to the `account suggestions list` as well as to `QR scanner`.

### Related issues:
https://github.com/MyCryptoHQ/MyCrypto/pull/1349
https://github.com/MyCryptoHQ/MyCrypto/pull/1945
https://github.com/paritytech/parity-signer/issues/103

### DIFF
--- a/common/components/ParityQrSigner.scss
+++ b/common/components/ParityQrSigner.scss
@@ -2,6 +2,7 @@
 
 .ParityQrSigner {
   display: flex;
+  z-index: 0;
   align-items: center;
   justify-content: center;
   width: 300px;
--- a/common/components/WalletDecrypt/WalletDecrypt.tsx
+++ b/common/components/WalletDecrypt/WalletDecrypt.tsx
@@ -268,7 +268,7 @@ const WalletDecrypt = withRouter<Props>(
             <i className="fa fa-arrow-left" /> {translate('CHANGE_WALLET')}
           </button>
           <h2 className="WalletDecrypt-decrypt-title">
-            {!selectedWallet.isReadOnly &&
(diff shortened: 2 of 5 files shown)


fckt:
BTW, why `AWS CodeBuild us-east-2 (mycryptobuilds-develop)` is failing (introduced by [this](https://github.com/MyCryptoHQ/MyCrypto/pull/2220/commits/8f1f40fbc63f67f29d2f8ce39095f7adfb651f3d))? Build restart is needed?

PR #2233, opened 2018-10-25
Allow BIP49 dpath

### LINKED ISSUE #2232: Only supports BIP44 dpaths (not BIP49)
### Description of the Issue

It's necessary for RSK platform to access accounts that are derivated from BTC segwit addresses. Those dpath start like `m/49'/` and MyCrypto only supports `m/44'/`.

### Steps to Reproduce

Select `Custom` in `Select an Address`. Then enter a dpath that starts with `m/49'/` and it won't be consider as valid.

alepc253:
Closes #2232

### Description

It's necessary for RSK platform to access accounts that are derivated from BTC segwit addresses. Those dpath start like `m/49'/` and MyCrypto only supports `m/44'/`.

### Changes

* A Regex expression that validate custom dpaths

### Steps to Test

1. Go to "Select an Address"
2. Select "Custom" from the combo
3. Write a dpath that starts with `m/49'/` and check it gets validated
4. Write a dpath that starts with `m/44'/` and check it still gets validated


alepc253:
Can you help me with that failing build? I don't have any problem building locally

tayvano:
Thanks for the PR. We're going to try to get this into our next release. I did have one question though: are there any unforeseen consequences to allowing / supporting 44 and 49? Is it simply adding an additional path, not a different derivation method?

alepc253:
Thanks for the consideration! As far as we understand, there is no difference but that "purpose" number in the derivation path (check this [link](https://github.com/bitcoin/bips/blob/master/bip-0049.mediawiki)).

PR #2258, opened 2018-11-13
[BUGSQUASH] Disable contract interaction fields when they are pending

ConnorBryan:
Solution to https://mycrypto.atlassian.net/browse/CORE-99.


ConnorBryan:
Let's try this again later.

PR #2275, opened 2018-11-20
Make selecting navigation dropdown items easier

ConnorBryan:
A complaint was made that the navigation dropdown items were hard to select. This change expands their hitbox and changes the background color when hovered.

screen shot 2018-11-20 at 12 15 22 pm


Azarielle:
[Jira task](https://mycrypto.atlassian.net/browse/CORE-152)

PR #2276, opened 2018-11-20
Display a warning notification when trying to use a wallet feature while not signed in

ConnorBryan:
When the user attempts to click a dropdown item from the header for a wallet option, but is not logged in, a customized message will appear that will notify them they will need to access their wallet. If the logo is clicked, or if a wallet is successfully unlocked, the message will disappear.

screen shot 2018-11-20 at 1 25 10 pm
screen shot 2018-11-20 at 1 25 19 pm


Azarielle:
[Jira task awaiting merge](https://mycrypto.atlassian.net/secure/RapidBoard.jspa?rapidView=19&projectKey=CORE&modal=detail&selectedIssue=CORE-153)

PR #2280, opened 2018-11-21
[DO NOT MERGE] 0x Integration

ConnorBryan:
This is the first attempt at utilizing `@0x/instant` to provide a decentralized exchange for MyCrypto to use.

Notes:
* This is using the CDN version of `@0x/instant`, which is subject to quick iteration.
* A hacky method is used to provide our own styles for the wrapping of the widget.

## After clicking `Swap ETH & Tokens` under `Buy & Exchange`
screen shot 2018-11-21 at 12 57 27 pm

## After clicking the card named `0x`
screen shot 2018-11-21 at 12 57 44 pm


ConnorBryan:
Part of Swap Overhaul.

PR #2281, opened 2018-11-21
Enable ShapeShift swapping on Desktop

ConnorBryan:
This PR adds logic to `ipcMain` and `ipcRenderer` to handle the 0Auth process for ShapeShift when in an `Electron` context.


ConnorBryan:
Part of Swap Overhaul.

PR #2364, opened 2019-01-24
Update Node

nickmccurdy:
Closes UI-155

### Description

This repo currently works with Node 8 (maintenance LTS) and 9 (unsupported). This means we're missing out on some new features, performance improvements, and security features in Node. Additionally, it is more difficult to work on dependency packages (including [@mycrypto/ui](https://github.com/MyCryptoHQ/ui)) that have different Node versions, and it requires a version manger like nvm (which isn't supported on my shell, fish). After updating dependencies, we can use any Node version between 8 and latest (currently 11), though 10 LTS is recommended for production deployments and builds.

### Steps to Test

I was able to use `yarn` and `yarn dev` on Node 8, 10, and 11 without issues. Please test run the usual testing and Electron building processes, I do not know how to run them and my Electron install is not working. Please also make sure that Electron supports all these Node versions.


nickmccurdy:
Travis CI isn't starting builds, should wait to merge until I figure this out

PR #2376, opened 2019-02-06
Update node 2

nickmccurdy:
Continued from #2364 to fix a CI issue, see previous PR for description


3nprob:
@tayvano @nickmccurdy Could you give some insight into why this was closed? I understand the intention is to bring v2 to Electron but in the meantime, the latest release of the desktop app is still on nodejs v8, which has been EoL for 2 years next week. I'd consider that a potential security issue by now and it would be helpful with some insight into why this effort has been abandoned and if a new PR addressing that would be considered.

PR #2387, opened 2019-02-19
Remove unused 'node' variable in domain resolution

sudoryan:
### Description

resolveDomainRequest only takes in one argument but was being called with two arguments.

### Changes

Removed the second argument "node" that wasn't being used in the function call.


Azarielle:
Merged in https://github.com/MyCryptoHQ/MyCrypto/pull/2410

PR #2397, opened 2019-02-26
[EPIC] Global Account Unlock v1

nickmccurdy:


nickmccurdy:
This is not a user story. Closing. Let's keep using the branch as we have been though.

SharonManrique:
This is the Epic keeping track of all progress. I think we should keep this open since in the end this gets merged in Develop when GAU is ready for release.

nickmccurdy:
We can still use the same branch even if this is closed. My intention was to only have PRs for Clubhouse issues, but we can keep this open if you want us to be able to comment and communicate about the entire branch overall on GitHub.

PR #2398, opened 2019-02-26
Account Functionality (WIP)

blurpesec:


tsconfig.json
```diff
@@ -21,7 +21,7 @@
     "noUnusedParameters": true,
     "pretty": true,
     "noErrorTruncation": true,
-    "noImplicitAny": true,
+    "noImplicitAny": false,
```
nickmccurdy:
```suggestion
    "noImplicitAny": true,
```
This setting would lower the type safety of the entire repo by allowing types that aren't inferred or explicitly typed to be `any` (which disables typechecking). We should re-enable this, but let me know if you need help with any TypeScript errors as a result.

Issue #2394, opened 2019-02-26
CSP Error with Trezor on new TrezorConnect


blurpesec:
### Description of the Issue

Trezor won't connect on web due to CSP error.

### Steps to Reproduce

1. Try to access Trezor device on web.
2. Should fail to connect. Popup will open and spin for a while, then time out and close.

### Console Logs / Screenshots
![image](https://user-images.githubusercontent.com/29407814/53418519-fdaef700-39a5-11e9-8ad5-e417d5a4ba97.png)

PR #2403, opened 2019-02-28
update trezor-connect@7

szymonlesisz:
Hi,
After you solve this issue: https://github.com/MyCryptoHQ/MyCrypto/issues/2394 I would like to ask you to update trezor-connect version to 7.0.1
There were some major changes in ethereum protobuf messages in new Trezor firmware (1.8.0/2.1.0) and connect@6 will probably fail on "signTransaction" and "getAddress" methods.
There is also a new, required method "manifest" which needs to be added. We would like you to provide a contact to reach you in case of any required maintenance.
```javascript
Trezor.manifest({
      email: 'info@mycrypto.com',
      appUrl: 'https://www.mycrypto.com'
});
```

you can read more here:
https://github.com/trezor/connect/blob/develop/docs/index.md#trezor-connect-manifest

I've replaced 'vendor/trezor-connect' by 'trezor-connect' from npm


blurpesec:
Hey @szymonlesisz, thanks for the PR. Does the static v7 file in the README here: https://github.com/trezor/connect fix this issue? Asking the team, but we'd probably prefer to continue loading it manually as opposed to introducing the additional dependency.

Also, just as a heads up, the reason the CI failed is because we require dependencies to be pinned. If the answer to the above question is "no" then we'll probably ask you to pin the dependency you added so we can pass CI. Thanks for your help!

szymonlesisz:
Yes, the file from [https://connect.trezor.io/7/trezor-connect.js](https://connect.trezor.io/7/trezor-connect.js) could be used as well and it should be working.
I've made this change because i was thinking it will be easier for you to maintain 'trezor-connect' dependency from yarn instead of copy-pasting this file.

I've added another commit with pinned version.
However final decision how to maintain this dependency is up to you.

PR #2404, opened 2019-02-28
Add analytics and events

amacar:


Sonicrida:
This looks good to me! I like how easy it is to add tracking throughout. Being able to add additional categories easily will be useful too.

PR #2416, opened 2019-03-04
Made name lookup generic with Namicorn

sudoryan:
### Description
Restructured ENS to Name Services so that future crypto TLDs can be added more easily. 

### Changes
Added Namicorn dependency to handle name resolution for owned ENS domain. Placed ENS into Name Service folder. Renamed store and actions to NameService.


blurpesec:
Heya @sudoryan. Thanks for the PR. I have a couple of questions:

- Why does this require so many dependencies? Is there any way to decrease this?
- It seems like you've wrapped the ens resolution task in a domain resolution task to make it more interoperable with other networks ? What other changes does this make?
- What do these changes address?
- What do these changes not address?

Thanks again, and I may have some followup :)

sudoryan:
**Why does this require so many dependencies? Is there any way to decrease this?**

Good catch! There were a few dev dependencies in there we removed.  There are now only 2 dependencies, Ethers.js and Punycode
- Ethers - We use ethers for all our ethereum interactions.
- Punycode - We use this to enable punycode in Namicorn

**It seems like you've wrapped the ens resolution task in a domain resolution task to make it more interoperable with other networks ? What other changes does this make?**

Yes that’s correct, see below.

**What do these changes address?**

- **Makes name resolution more extensible**.  Some naming services may have complex interaction. For example Grin or Iota may be quite involved, and enriched lookups will be difficult to support on a one off basis. It makes sense to abstract these lookups out to the library.
- **Support several naming services like .crypto, .luxe, and .zil**. This supports .eth and has placeholders for the others. Will add .luxe next, followed by .zil, etc. This PR allows us to resolve these TLDs in just a few lines of code. 
- **Support enriched naming resolution features**.  Return a photo or twitter account associated with a name.  See example [here](https://docs.google.com/document/d/1x6krnlp3nz4ZMOvbn-y27CR-mIOpNG-Hw0HUsYB-DMY/edit?usp=sharing). This way if MyCrypto wants to show more information associated with a name in the future it can.

**What do these changes not address?**

.RSK name resolution. We're working to get this done asap and we'll let you know once it's ready to roll out.

Let me know if you have any more questions. Thanks for reading through the PR!

theporpoise:
.RSK name resolution added to Namicorn library.  @sudoryan take a look and let us know if you need anything else (like deed owner, etc.).  Thanks!

common/features/nameService/helpers.ts
```diff
@@ -24,39 +24,12 @@ export function* makeEthCallAndDecode({ to, data, decoder }: Params): SagaIterat
...
-      data: ENS.resolver.addr.encodeInput({ node: nameHash }),
-      decoder: ENS.resolver.addr.decodeOutput
-    });
-
-    resolvedAddress = result.ret;
-  }
-
+function* nameStateOwned({}, domain: string) {
+  const namicorn = new Namicorn({ debug: true, disableMatcher: true });
+  namicorn.use(namicorn.middleware.ens({ url: 'https://mainnet.infura.io/mycrypto' }));
```
blurpesec: So we generally don't hardcode jsonrpc urls. We use a client-side node-balancer for interacting with the chain. Normally with these types of interactions we would specify it as a new "Namicorn" API [here](https://github.com/MyCryptoHQ/MyCrypto/tree/develop/common/api) and interact with it in that way.

@nickmccurdy @Sonicrida Do you guys have opinions on maintaining existing structure for things like this?

nickmccurdy: I'm not sure, it sounds like a good idea, but does everything need a load balancer? Maybe this could be extracted in such a way that we could add one later.

perfect-cents: @blurpesec We can sync with the current node url inside the redux state if that's better. Namicorn is just a wrapper around calls to the blockchain. However, Namicorn just calls one method to resolve all names. It seems excessive to put it in the apis.

blurpesec: @nickmccurdy Ideally, we wouldn't want all of ENS functionality to drop because infura has a bad day >.>

@perfect-cents That should be fine :-)

Azarielle:
I am bumping this to release 1.6.2

common/features/nameService/helpers.ts
perfect-cents: We are going to make some major updates on Namicorn soon that will enable us to integrate into the MyCrypto name system. After that we will be ready, for another review.

SharonManrique:
@sudoryan Hey! We appreciate your contribution and wish to get this PR merged in for next release. Is there any update as to when this will be ready for a review by our devs?

sudoryan:
Hi @SharonManrique, this code is a bit stale since it's been a few months so it's not yet ready for release. Our team is going to come back to this PR next week and make the necessary updates. Thanks!

vilsbole:
closing in favour of #2986

PR #2441, opened 2019-03-19
update ledger packages

tayvano:
due to https://github.com/LedgerHQ/ledgerjs/blob/master/docs/migrate_eth_app.md


nickmccurdy:
I fixed the build, but this seems to have some environment issues so we should still test it manually.

PR #2443, opened 2019-03-20
Fix for Trezor Firmware 1.8.0 bug.

### LINKED ISSUE #2439: Updated Trezor firmware breaks compatibility with MyCryptoWallet
### Updated Trezor firmware breaks compatibility with MyCryptoWallet

Hello, I have tried sending a transaction recently with the latest firmware update on trezor, but for whatever reason I'm getting a safety check failed red notification error when running in the yarn electron:dev mode. I also tested this issue with a fully built version of MyCrypto application, and to no avail. I also tested this against an older trezor with older firmware, and it seems to be fine there. For whatever reason, it seems that the ethereum signing check on the trezor fails.

https://github.com/trezor/trezor-mcu/blob/master/firmware/ethereum.c

If you want to check out the link above and go through it with a simple ctrl-f of "safety check failed" will show you where in the code the error is coming from. 

### Steps to Reproduce

1. Update trezor one to firmware version 1.8.0 
2. Try to send a transaction
3. Fail sending transaction due to safety check notification error.

Running a 13 inch 2017 Macbook Pro with MacOS Mojave Version 10.14.3 (18D109)

image

I modified the trezor.js node_modules to print out the request and responses going to the trezor, and here are the console logs

image

blurpesec:
Closes #2439

### Description

Fixes Trezor Firmware bug that leads to `Safety Check Failed`

### Changes

* Change method for initializing Trezor session with config.

### Steps to Test

1. Access account with trezor device on electron developer build `yarn run dev:electron`
2. Try to send transaction with Trezor account. Sign and submit transaction.

### DIFF
--- a/shared/enclave/server/wallets/msg-config.ts
+++ b/shared/enclave/server/wallets/msg-config.ts
@@ -1,0 +1,9986 @@
+export default JSON.stringify({
+  package: null,
+  messages: [
+    {
+      name: 'MultisigRedeemScriptType',
+      fields: [
+        {
+          rule: 'repeated',
+          options: {},
+          type: 'HDNodePathType',
+          name: 'pubkeys',
+          id: 1
+        },
+        {
+          rule: 'repeated',
+          options: {},
+          type: 'bytes',
+          name: 'signatures',
+          id: 2
+        },
+        {
+          rule: 'optional',
+          options: {},
+          type: 'uint32',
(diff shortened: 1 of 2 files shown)


nickmccurdy:
Thanks for the contribution @andreweximchain! You'll be credit for the commits on GitHub (even though we made the pull request).

PR #2454, opened 2019-03-25
Feature/homepage

amacar:
This PR includes a homepage with analytics. ~~Download desktop app feature is also merged due to shared code.~~


common/v2/features/Home/routes.ts
```diff
@@ -6,5 +6,11 @@ export default [
     path: '/',
     exact: true,
     component: Home
+  },
+  {
+    name: 'Home',
+    path: '/home',
+    exact: true,
+    component: Home
```
nickmccurdy: Why do we need to add this when `/` already routes to `Home`?

nickmccurdy:
Nice work overall, thanks. I noticed some of the buttons aren't working, is this something we should fix after merging?

common/v2/features/Home/routes.ts
amacar: This is a requirement from Jira; we can also remove that.

amacar:
> Nice work overall, thanks. The page looks great though we may want to do a design review to be sure. I noticed some of the buttons aren't working, is this something we should fix after merging?

Buttons which redirect to nonexisting features are of course not working yet. We should add missing redirects after features are implemented.

common/v2/features/Home/routes.ts
nickmccurdy: We haven't decided how to handle routing yet, though it's easy enough to change later, so this is fine.

nickmccurdy:
We have some feedback from our designer:
- the top banner doesn’t look right? the icons / arrows are not the right colors 😞let me know if you need guidance on this
- there should be a line break after the word “mycrypto” in the 2nd line on top section: “It’s Time To Use MyCrypto For Your Crypto.” [if we can figure out a way to translate the line break positions between languages]
- image on top section seems slightly too big. can you scale down by 10-15%?
- try using hover state for panels
- “view your dashboard” txt needs to be right aligned
- testimonials section is missing yellow graphics
- Can you unbold “Have Questions? We’re Here to Help!” at bottom?
- we are missing the graphics on the “Ready to start
managing your funds safely and confidently? ” section on mobile. Can we put these on Desktop too?

I also noticed that the "Fully Compatible" and testimonial sections have low contrast text.

amacar:
> Thanks, these improvements are looking really good. At this point I think we just need to make the icon colors in the navigation match the designs (blue on desktop and white on mobile).

I already did this but forgot that styled-components overwrite css, so I rewrite top section using styled-components.

PR #2455, opened 2019-03-25
Fix error when committing with lint-staged hook

nickmccurdy:


nickmccurdy:
I need to merge this, I can't commit on other branches with my editor

PR #2458, opened 2019-03-26
Provide ERC-20 token information to Ledger wallets

Mrtenz:
### Description

Ledger removed the hardcoded ERC-20 token info from the Ethereum app to reduce the size. This update implements the `provideERC20TokenInformation` function to replace this. More info on this change [here](https://github.com/LedgerHQ/ledgerjs/blob/master/docs/migrate_eth_app.md).

### Changes

* Updated `@ledgerhq/hw-app-eth` to `4.48.0`
* Added TypeScript types for the new functions
* Implemented the `provideERC20TokenInformation` function

### Steps to Test

1. Update the Ethereum app to at least v1.2.2. This update is currently available in Ledger Live only when launching with `FORCE_PROVIDER=4`, but should be available without pretty soon.
2. Generate a token transaction in MyCrypto.
3. Check if the token symbol, amount and destination address are shown on the Ledger device.

### DIFF
--- a/common/libs/wallet/deterministic/ledger.ts
+++ b/common/libs/wallet/deterministic/ledger.ts
@@ -6,6 +6,7 @@ import LedgerEth from '@ledgerhq/hw-app-eth';
 import { translateRaw } from 'translations';
 import { getTransactionFields } from 'libs/transaction';
 import { HardwareWallet, ChainCodeResponse } from './hardware';
+import { byContractAddress } from '@ledgerhq/hw-app-eth/erc20';
 
 // Ledger throws a few types of errors
 interface U2FError {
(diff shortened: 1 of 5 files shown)


nickmccurdy:
The client bundle is about 410 KB larger, is this acceptable or do we need to trim it down?

nickmccurdy:
Using a Ledger Nano S with firmware 1.5.5 and Ethereum app 1.2.2, I only see the following fields to confirm, missing the token symbol and amount:
1. Selector
2. Field 1 (address)
3. Field 2 (hex string with colons)
4. Amount (says 0 ETH so it's incorrect)
5. Maximum fees (accurately based on gas price)

blurpesec:
> I only see the following fields to confirm, missing the correct token symbol and amount:


Desktop app or web? @nickmccurdy

nickmccurdy:
> > I only see the following fields to confirm, missing the correct token symbol and amount:
> 
> Desktop app or web? @nickmccurdy

Desktop

nickmccurdy:
This works as far as I can tell but I want to make sure @blurpesec is OK with it first

SharonManrique:
Merged into https://github.com/MyCryptoHQ/MyCrypto/pull/2447

PR #2461, opened 2019-03-29
Update Master to reflect current state of develop

SharonManrique:
Updating Master to be even with Develop. 

Master will now be our stable/default branch, while Develop is reserved for dev-ing


nickmccurdy:
Can we set the default branch to master and still base our pull requests out of develop by default? Without making this change, GitHub users may be confused by seeing newer develop branch code that hasn't been released yet.

PR #2474, opened 2019-03-30
fix drawbridge

tayvano:
to better emulate what we actually do when we build, update that shit.

also, no more using npm commands. npm is funky.


nickmccurdy:
I'd like to try this without `rm -rf node_modules` initially. Yarn is supposed to install the same dependencies with or without `node_modules` but keeping it will make deploys faster.

SharonManrique:
Merged into https://github.com/MyCryptoHQ/MyCrypto/pull/2476

PR #2475, opened 2019-04-01
Fix Electron builds for Windows

Mrtenz:
Based on #2471

### Description

Fixes Electron builds for Windows, by removing the `node-usb` dependency (from `@ledgerhq/hw-transport-node-hid`).

### Changes

* Replaced `@ledgerhq/hw-transport-node-hid` with `@ledgerhq/hw-transport-node-hid-noevents` to remove the `node-usb` dependency.
* Removed `electron-rebuild`, as `electron-builder` suggests using `electron-builder install-app-deps` instead.

### Steps to Test

1. Build the Electron app for Windows on another platform (macOS or Linux)
2. Run the app on Windows
3. Verify that you're able to send a transaction with a Ledger


SharonManrique:
Merged into https://github.com/MyCryptoHQ/MyCrypto/pull/2476

PR #2477, opened 2019-04-04
Update Address Field Placeholder to include ENS name example

tayvano:
Makes it more obvious that you can input an ENS name or an Ethereum address in the placeholder field. 

In future, probably only do this on networks that support ENS.

### DIFF
--- a/common/components/AddressField.tsx
+++ b/common/components/AddressField.tsx
@@ -35,7 +35,7 @@ const AddressField: React.SFC<Props> = ({
   showLabelMatch,
   toChecksumAddress,
   showIdenticon,
-  placeholder = donationAddressMap.ETH,
+  placeholder = "donate.mycryptoid.eth or 0x4bbeEB066eD09B7AEd07bF39EEe0460DFa261520",
   showInputLabel = true,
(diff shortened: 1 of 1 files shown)


common/components/AddressField.tsx
```diff
@@ -35,7 +35,7 @@ const AddressField: React.SFC<Props> = ({
   showLabelMatch,
   toChecksumAddress,
   showIdenticon,
-  placeholder = donationAddressMap.ETH,
+  placeholder = 'donate.mycryptoid.eth or 0x4bbeEB066eD09B7AEd07bF39EEe0460DFa261520',
```
nickmccurdy: We can continue using `donationAddressMap` to make this easier to update.
```suggestion
  placeholder = `donate.mycryptoid.eth or ${donationAddressMap.ETH}`,
```
It may also be helpful to move `'donate.mycryptoid.eth'` to the config if it's going to be used in other places.

SharonManrique:
Merged into Release 1.6.4 https://github.com/MyCryptoHQ/MyCrypto/pull/2479

PR #2496, opened 2019-04-11
Feature/screen lock

PeterZlodej:


common/v2/providers/LockScreenProvider/LockScreenProvider.tsx
```diff
@@ -0,0 +1,189 @@
...
+  };
+
+  public resetInactivityTimer = () => {
+    clearTimeout(inactivityTimer);
+    inactivityTimer = setTimeout(this.startLockCountdown, inactivityTime);
+  };
+
+  public startLockCountdown = () => {
+    //Start the lock screen countdown only if user is on one of the dashboard pages
+    if (!window.location.href.includes('/dashboard')) {
```
nickmccurdy: Let's use `this.props.location` from React Router

PeterZlodej:
Currently, you can manually navigate to /screen-lock/new, create a password, lock the screen and then test the unlock. To test the countdown screen, the only way is to be inactive for 5 minutes on the dashboard. To test it quickly locally, the inactivityTime value can be changed in the LockScreenProvider.tsx.

common/translations/lang/en.json
```diff
@@ -786,6 +786,35 @@
...
-    "DOWNLOAD_APP_FOOTER_INFO_LINK": "Learn more about our desktop app."
+    "DOWNLOAD_APP_FOOTER_INFO_LINK": "Learn more about our desktop app.",
+    "SCREEN_LOCK_NEW_HEADING":"Create Password to Lock Your Wallet",
+    "SCREEN_LOCK_NEW_DESCRIPTION":"We want to help you keep your funds safe! Please create a password so you can securly view your wallet when you come back to MyCrypto.",
+    "SCREEN_LOCK_NEW_PASSWORD_LABEL":"Password (min 8 characters)",
+    "SCREEN_LOCK_NEW_CONFIRM_PASSWORD_LABEL":"Confirm password",
+    "SCREEN_LOCK_NEW_CREATE_PASSWORD_BUTTON": "Create Password",
+    "SCREEN_LOCK_NEW_PASSWORD_TOO_SHORT":"Password must be at least 8 characters long",
+    "SCREEN_LOCK_NEW_PASSWORDS_DONT_MATCH":"Passwords don't match",
+    "SCREEN_LOCK_FORGOT_PASSWORD_HEADING":"Forgot Screen Lock Password?",
```
nickmccurdy: I think this should just be "Forgot Password?"

amacar: There is "Forgot Password" in designs but "Forgot Screen Lock Password" in Jira card. We can change it to whatever you think is better.

tayvano: I would prefer "Forgot Screen Lock Password" so users don't get confused with their keystore file password (or Facebook password 😉😅😂).

PR #2500, opened 2019-04-12
Attempt routing for mnemonic unlocking

nickmccurdy:
This is just to show my progress, it does not work yet


nickmccurdy:
We have decided to implement routing after GUA MVP, and the code will probably change significantly by then.

PR #2503, opened 2019-04-14
Implement basic analytics

wtzb:
### Description

This PR implements basic analytics (based on Povio's GAU analytics implementation) on the non-GAU (current) codebase.
- Count usage between browser and desktop application
- Which pages are visited
- Browser software
- Browser lang

### Steps to Test

Open developer tools, go to network tab, see requests being made.

### Issues

- Since the desktop application is also a Chrome installation, this will count toward the browser software and browser lang statistics in Matomo.

### Todo

- [x] Differentiate between browser/desktop app


common/v2/services/Analytics/constants.ts
```diff
@@ -0,0 +1,9 @@
+// The root URL for Analytics API requests.
+export const ANALYTICS_API_URL = 'https://d2xxi6ebw5.execute-api.us-west-2.amazonaws.com/latest';
```
409H: This doesn't resolve for me - don't we have a subdomain under mycryptoapi?

common/v2/services/Analytics/Analytics.ts
```diff
@@ -0,0 +1,55 @@
...
+  public trackPageVisit(pageUrl: string): Promise<any> {
+    const customParams: Params = {
+      local: isDevelopment().toString()
+    };
+
+    const cvar: object = this.mapParamsToCvars(customParams);
+
+    const analyticsId = isDesktop() ? ANALYTICS_ID_DESKTOP : ANALYTICS_ID_SITE;
+
+    const params: object = {
```
409H: Should we add a `timeout` here so we don't run into a similar issue as MetaMask? https://medium.com/metamask/incident-postmortem-slow-confirmation-popups-23108bc954bc

wtzb: Added in newest commit 👍

nickmccurdy:
> Since the desktop application is also a Chrome installation, this will count toward the browser software and browser lang statistics in Matomo.
I think we can use Matomo's custom variables to filter what's coming from Electron. Alternatively if we want to organize Electron as a separate browser rather than being Chrome-like, we should be able to change or observe  its user agent to be different from Chrome's.

common/Root.tsx
```diff
@@ -152,6 +153,17 @@ class RootClass extends Component<Props, State> {
...
 }
 
+let previousURL = '';
+const PageVisitsAnalytics = withRouter(({ history, children }) => {
+  history.listen(() => {
+    if (previousURL !== window.location.href) {
+      AnalyticsService.instance.trackPageVisit(window.location.href);
+      previousURL = window.location.href;
+    }
+  });
```
nickmccurdy: Side effects should not be called in function components (except hooks). Alternatively, I think this should be extracted into a function that's called in componentDidMount/constructor in other components.

common/v2/services/Analytics/types.ts
```diff
@@ -0,0 +1,7 @@
+export interface Params {
+  [key: string]: any;
+}
```
nickmccurdy: This should be strongly typed based on the types used in the `params` variable

common/v2/services/Analytics/types.ts
```diff
@@ -0,0 +1,7 @@
...
+
+export interface CvarEntry {
+  [key: string]: [string, string];
+}
```
nickmccurdy: This should be a mapped type derived from the `Params` type

common/v2/services/Analytics/Analytics.ts
```diff
@@ -0,0 +1,56 @@
...
```
nickmccurdy:
```suggestion
    const params: Params = {
```

nickmccurdy:
https://app.clubhouse.io/mycrypto/story/693/implement-basic-analytics

nickmccurdy:
I already reviewed this

nickmccurdy:
@wtzb I fixed the side effect in render and made some minor improvements to the types. If you know what specific keys are allowed in Params, I can help you make the types more accurate and useful (also for CvarEntry).

PR #2511, opened 2019-04-17
Add createWithID Method

blurpesec:
### Changes

* Added createWithID method.

TODO:
- [ ] Test using NodeOptions

### DIFF
--- a/common/v2/services/LocalCache/LocalCache.ts
+++ b/common/v2/services/LocalCache/LocalCache.ts
@@ -199,6 +199,20 @@ export const create = <K extends CollectionKey>(key: K) => (
   setCache(newCache);
 };
 
+export const createWithID = <K extends CollectionKey>(key: K) => (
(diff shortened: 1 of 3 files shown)


common/v2/services/LocalCache/LocalCache.ts
```diff
@@ -199,6 +199,20 @@ export const create = <K extends CollectionKey>(key: K) => (
...
+  value: LocalCache[K][keyof LocalCache[K]],
+  id: string
+) => {
+  const uuid = id;
+  if (getCache()[key][uuid] === undefined) {
+    const newCache = getCache();
+    newCache[key][uuid] = value;
+    setCache(newCache);
+  } else {
+    console.log('Error: key already exists in createWithID');
```
nickmccurdy: I'd prefer to throw an error or return a success boolean so that this can be error handled from services/providers.

PR #2514, opened 2019-04-20
Update Common/index.html BadBrowser check for Brave Browser users

jspence425:
### Description

Updates BadBrowser copy to inform Brave Browser users of the issue if/when they see this roadblock. This roadblock will appear to those users if they are blocking ALL cookies, though thankfully that is not a default setting for Brave.

### Changes

Added two lines of copy inside the BadBrowser desktop check.


common/index.html
```diff
@@ -51,6 +51,10 @@ <h2>Your Browser is Out of Date</h2>
           MyCrypto requires certain features that your browser doesn't offer. Your browser may also
           be missing security updates that could open you up to vulnerabilities. Please update your
           browser, or switch to one of the following browsers to continue using MyCrypto.
+
```
nickmccurdy: HTML ignores extra whitespace characters, so it would appear that "If you're using..." is part of the same paragraph even though there is a blank line in the code. Instead, you need to start a new `<p>` (paragraph) tag like this:

```suggestion
</p>
<p class="is-desktop">
```

jspence425:
Thank you Nick!

nickmccurdy:
Looks great now.
image

SharonManrique:
Merged into Release 1.6.5 https://github.com/MyCryptoHQ/MyCrypto/pull/2527

PR #2528, opened 2019-04-26
Add-account flow default changes

blurpesec:
### Description

Edits the defaults for account object in add-account flow.

TODO: 
- [x] Change accountData information depending on network - @blurpesec - #2528
- - [x] Asset (should be set to base asset of each network)
- - [x] Label (get feedback on this, currently set to "New Account") *Leaving this alone for now - make a task for this in PM tool*
- - [x] LocalSettings (figure out some default value for this and set it)
- - [x] TransactionHistory (leave as empty string until filled in by user..?) *Leaving this alone*

### DIFF
--- a/common/v2/features/AddAccount/AddAccount.tsx
+++ b/common/v2/features/AddAccount/AddAccount.tsx
@@ -275,30 +275,18 @@ const WalletDecrypt = withRouter<Props>(
       return this.WALLETS[selectedWalletKey];
     }
 
-    public handleCreateAccount = (createAccount: any) => {
+    public handleCreateAccount = (createAccount: ((newAccount: Account) => void)) => {
       const { accountData } = this.state;
       const network: NetworkOptions | undefined = getNetworkByName(accountData.network);
-      if (!network) {
-        const newAccount: Account = {
-          ...accountData,
-          assets: 'DefaultAsset',
-          value: 0,
-          label: 'New Account',
-          localSettings: 'default',
-          transactionHistory: ''
-        };
-        createAccount(newAccount);
-      } else {
-        const newAccount: Account = {
-          ...accountData,
-          assets: network.unit,
-          value: 0,
-          label: 'New Account',
-          localSettings: 'default',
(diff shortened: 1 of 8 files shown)


common/v2/libs/networks/networks.ts
```diff
@@ -1,19 +1,40 @@
...
+        if (network.name === name) {
+          resolve(network);
+        }
+      });
+      reject();
+    } catch (e) {
+      reject();
+    }
+  });
 };
```
nickmccurdy: The Promise constructor's resolve function can only be called once, but it's called in map, which would get called for each network in networks. What is your intention here? Do you want to resolve the promise with an array of all networks, or just get the first one? Either way, you shouldn't need to use the Promise constructor for syncronous code.

common/v2/features/AddAccount/AddAccount.tsx
```diff
@@ -275,23 +275,23 @@ const WalletDecrypt = withRouter<Props>(
       return this.WALLETS[selectedWalletKey];
     }
 
-    public handleCreateAccount = async (createAccount: any) => {
+    public handleCreateAccount = (createAccount: any) => {
```
nickmccurdy: Let's type this parameter so we don't accidentally misuse it: `(createAccount: Account => void)` (means a function that takes an `Account` and returns anything or nothing)

common/v2/features/AddAccount/AddAccount.tsx
```diff
@@ -275,23 +275,23 @@ const WalletDecrypt = withRouter<Props>(
...
+      if (!network) {
         const newAccount: Account = {
           ...accountData,
-          assets: network.unit,
+          assets: 'DefaultAsset',
           value: 0,
           label: 'New Account',
           localSettings: 'default',
           transactionHistory: ''
         };
```
nickmccurdy: We could simplify this if/else statement by replacing it with `const newAccount: Account = { assets: network ? network.unit : 'DefaultAsset, ... }'`

PR #2529, opened 2019-04-27
redirect user from dashboard if no account exists

SharonManrique:
Before fix: 
If user deleted all accounts while on dashboard they wouldn't be redirected. 

After fix:
Redirects user to no-accounts if they delete all their accounts while on dashboard.

### DIFF
--- a/common/v2/features/Dashboard/components/AccountList.tsx
+++ b/common/v2/features/Dashboard/components/AccountList.tsx
@@ -1,11 +1,18 @@
 import { Address, Button, CollapsibleTable, Icon, Network, Typography } from '@mycrypto/ui';
 import React from 'react';
 import { Redirect } from 'react-router-dom';
+
 import styled from 'styled-components';
 import { truncate } from 'v2/libs';
 import { ExtendedAccount } from 'v2/services';
 import './AccountList.scss';
 import DashboardPanel from './DashboardPanel';
+import { translateRaw } from 'translations';
+
+const DeleteButton = styled(Button)`
+  align-self: flex-start;
+  margin-left: 1em;
+`;
 
 type DeleteAccount = (uuid: string) => void;
 interface AccountListProps {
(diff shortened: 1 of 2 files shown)


common/v2/features/Dashboard/components/AccountList.tsx
```diff
@@ -18,45 +18,72 @@ const DeleteButton = styled(Button)`
...
+          <Network key={`account-network-${account.network}`} color="#a682ff">
+            {account.network}
+          </Network>,
+          <Typography key={`account-value-${account.value}`}>{account.value}</Typography>,
+          <DeleteButton
+            key={`account-id-${account.uuid}`}
+            onClick={this.handleAccountDelete(account.uuid)}
+            icon="exit"
+          />
+        ];
```
nickmccurdy: I don't think you need the key prop here because the Table component maps over children and adds them automatically (but our linter doesn't know this so it gets confused)

common/v2/features/Dashboard/components/AccountList.tsx
```diff
@@ -1,11 +1,11 @@
...
-import './AccountList.scss';
-import { ExtendedAccount } from 'v2/services';
+import { Address, Button, CollapsibleTable, Icon, Network, Typography } from '@mycrypto/ui';
+import React, { Component } from 'react';
+import { Redirect } from 'react-router-dom';
 import styled from 'styled-components';
 import { truncate } from 'v2/libs';
+import { ExtendedAccount } from 'v2/services';
+import './AccountList.scss';
+import DashboardPanel from './DashboardPanel';
```
nickmccurdy: Can we keep these two import groups separate? It looks like your editor auto formatted them.

common/v2/features/Dashboard/components/AccountList.tsx
```diff
@@ -18,45 +18,72 @@ const DeleteButton = styled(Button)`
...
+        sortableColumn: 'Address',
+        sortFunction: (a: any, b: any) => {
+          const aLabel = a.props.label;
+          const bLabel = b.props.label;
+          return aLabel === bLabel ? true : aLabel.localeCompare(bLabel);
+        },
+        hiddenHeadings: ['Favorite', 'Delete'],
+        iconColumns: ['Favorite', 'Delete']
+      }
+    };
```
nickmccurdy: Table content should be translatable.

common/v2/features/Dashboard/components/AccountList.tsx
```diff
@@ -18,45 +18,72 @@ const DeleteButton = styled(Button)`
...
+  }
+  public render() {
+    const { className } = this.props;
+
+    return (
+      <DashboardPanel
+        heading="Your Accounts"
+        action="Add Account"
```
nickmccurdy: Dashboard heading and action should be translatable.

PR #2540, opened 2019-05-01
Make each build fail as soon as a job fails

nickmccurdy:


nickmccurdy:
https://app.clubhouse.io/mycrypto/story/690/make-each-build-fail-as-soon-as-a-job-fails

.travis.yml
```diff
@@ -24,6 +24,7 @@ matrix:
       dist: trusty
       sudo: required
       services: docker
+  fast_finish: true
```
vilsbole: fantastically simple!
Is it normal that the indenting is different?

nickmccurdy: It's indented only one level (two spaces) below `matrix:` because it's a setting on matrix, not a specific OS in `include:`

nickmccurdy:
Ready to merge

PR #2541, opened 2019-05-02
Chore/rhl

vilsbole:
### Closes
[branch ch702]
[branch ch704]
[branch ch701]

### Description
Current RHL doesn't preserve State.
Preserving component state between page changes allows us to save time.

### Changes
- ~~set node version to `lts/carbon`~~
- use babel to transpile typscript during dev.
- use webpack-dev-server for development to enable RHL
- enable https for dev server to simplify setup.
- remove HMR noise from console
- update react root to use RHL. By default any change to a file which isn't `hot` triggers a page refresh so we can remove the condition `module.hot`.
- configure yarn to use the new dev-server
- remove deprecated npm scripts and README instructions.

### Steps to Test
1. `yarn install`
2. `yarn dev` 🔥


.nvmrc
```diff
@@ -1 +1 @@
-8.9.4
\ No newline at end of file
+v8.16.0
```
nickmccurdy: Is there a specific reason why we need to update node 8? We've done a lot of testing on Node 8.9.4 and we're going to have to do a major node update later this year anyway, so I'd rather not change this unless there's a significant tooling advantage.

README.md
```diff
@@ -74,18 +74,6 @@ yarn test
...
-Some parts of the site, such as interacting with a Ledger hardware wallet, require an HTTPS environment to work. To develop on HTTPS:
-
-1.  Create your own SSL Certificate (Heroku has a [nice guide here](https://devcenter.heroku.com/articles/ssl-certificate-self))
-2.  Move the `.key` and `.crt` files into `webpack_config/server.*`
-3.  Run the following command:
-
-```bash
-yarn dev:https
-```
-
```
nickmccurdy: I always like to see docs deleted when you don't have to worry about them any more 😉 nice job

package.json
```diff
@@ -73,10 +75,18 @@
     "trezor.js": "6.17.5",
     "uuid": "3.2.1",
     "wallet-address-validator": "0.2.4",
+    "webpack-dev-server": "3.2.1",
     "whatwg-fetch": "2.0.3",
     "zxcvbn": "4.4.2"
```
nickmccurdy: It seems like these new dependencies should only be used in development. If so, they also need to be conditionally disabled in production builds (we should have Webpack set up with `NODE_ENV`).

webpack_config/webpack.rhl.js
```diff
@@ -0,0 +1,68 @@
+const merge = require('webpack-merge');
```
nickmccurdy: Rather than using a separate merge package can we use our existing webpack config function?

.nvmrc
vilsbole: This is mainly a convenience.
Since we're stuck with a legacy version I found it best to move to a long term support version for less errors. I'm guessing this is used mainly by the electron app for wallet  handling?

README.md
vilsbole: 😎

package.json
vilsbole: Yes indeed! Correcting right away.

webpack_config/webpack.rhl.js
vilsbole: I would like to get rid of the imperative style of `makeConfig.js`. It's a collection of conditional statements which makes it hard to optimise a build.

I'm counting the need for at least 6 different builds.
- browser:dev
- browser:prod
- browser:test
- electron:dev
- electron:prod
- electron:test
And see that there are multiple others...
- freezer?
- derivation-checker?
- html?

`webpack-merge` allows us to use a composition pattern for configuration files. It enables us to remove the need for flags and condition checks by simply having a `webpack-common.js` file that we extend or override with a target specific configuration e.g. `webapack-electron-dev.js`.
The execution path becoming simple to read, we will be able to modify and optimise each one in a declarative manner.

vilsbole:
> Thanks, these improvements seem really useful and I'm excited to try them out. I have some questions about the tooling changes that aren't directly related to my source comments:
> 
> * Does the dev server (which is now using babel) still do typescript typechecking? I find that very useful, and I'd personally rather have typechecking integration than faster builds.
> * Are there any browser support issues with the dev server's HTTPS cert that we wouldn't have with a custom self-signed cert?
> * Did you mean to remove any scripts from package.json? I'm only seeing a change to the dev script.

1. Currently `typescript-loader` doesn't catch type errors in development. The errors appear when we run `tslint` in our commit-hooks and during testing. `babel-loader` works the same way. The most useful place for Ts handling is directly in the IDE where your editor displays the interface of the arguments! https://code.visualstudio.com/docs/languages/typescript
2. I'm not aware of browser support errors with https of webpack-dev-server, if need be we can also specify a self-signed cert instead of the one generated by the WDS.
3. No removals for the moment. This will happen once I move `dev:electron` to WDS.

.nvmrc
nickmccurdy: I thought Node 8.9.4 was still an LTS release, is it?

nickmccurdy:
> The most useful place for Ts handling is directly in the IDE where your editor displays the interface of the arguments! 
VSCode (which I think we all use) does not show all TypeScript errors in the project by default. Is there a way for us to change this for all devs?

.nvmrc
vilsbole: Interesting! So I've learned that Node 8 entered LTS with [8.9.0](https://nodejs.org/en/download/releases/). It appears that lts/carbon covers 8.9.0 till 8.16.0.
Node uses an [even numbered vs an odd numbered](https://nodesource.com/blog/understanding-how-node-js-release-lines-work/) cycle where 
> you can effectively equate even-numbered releases with LTS release lines, and odd-numbered releases with non-LTS release lines.

So in fact they are both LTS. I initially thought it was only 8.16.0 since it's the version that nvm proposes for lts/carbon.

Looking at the [changelog](https://github.com/nodejs/node/blob/master/doc/changelogs/CHANGELOG_V8.md) for Node 8, it seems that a lot of the minor version changes are security fixes. Combined with the ease of setting `nvm use lts/carbon` I still suggest that we bump the version.

If you feel otherwise, let me know and I'll revert the change.

nickmccurdy:
Note: We need to decide if it's safe to upgrade Node before merging

vilsbole:
> > The most useful place for Ts handling is directly in the IDE where your editor displays the interface of the arguments!
> 
> VSCode (which I think we all use) does not show all TypeScript errors in the project by default. Is there a way for us to change this for all devs?

I'm still on Atom and haven't tried VSC yet. The docs say they support error checking by default.
This SO mentions that there is flag in VCS for validation:
https://stackoverflow.com/questions/43908813/how-to-setup-vscode-to-show-typescript-error-inline

vilsbole:
> Before merging please check if we can fix this peer dependency: `warning " > sass-extract-loader@1.1.0" has incorrect peer dependency "webpack@^3 || ^2 || ^2.2.0-rc.0 || ^2.1.0-beta || ^1.12.6".`

The warning appears also on `overhaul/global-account-unlock-v1` and is not introduced by this merge-request. The project is already using `webpack@4.5.0` so the warning is expected and is easy to fix. I'm adding a card to clubhouse [ch1001] to address this issue in another MR.

nickmccurdy:
> > > The most useful place for Ts handling is directly in the IDE where your editor displays the interface of the arguments!
> > 
> > 
> > VSCode (which I think we all use) does not show all TypeScript errors in the project by default. Is there a way for us to change this for all devs?
> 
> I'm still on Atom and haven't tried VSC yet. The docs say they support error checking by default.
> This SO mentions that there is flag in VCS for validation:
> https://stackoverflow.com/questions/43908813/how-to-setup-vscode-to-show-typescript-error-inline

VSCode does check errors by default, but only on open files. I realized after making this comment that it's not really an issue though, since our commit hooks still typecheck the entire project and you can run `yarn tscheck` manually.

I would recommend trying out VSCode, it's not necessarily required, but it has better TypeScript support than Atom, and we use other extensions and tools that work well with it.

SharonManrique:

nickmccurdy:
QA results:
`yarn dev`: server doesn’t start
`yarn dev:electron`: RHL error in console

vilsbole:
@nickmccurdy @SharonManrique I've updated the scripts.
1. `yarn dev` will launch an https server at https://localhost:3000
2. `yarn dev:electron` runs with RHL and the renderer content is served at http://localhost:3000

nickmccurdy:
It seems like http://localhost:3000 is not redirecting to https://localhost:3000

vilsbole:
> http://localhost:3000 should redirect to https://localhost:3000

@nickmccurdy I attempted adding a redirect middleware to webpack-dev-server, but it wouldn't trigger on un-secure requests and I don't want to add another server to our dev environment. 
If you have a quick solution using webpack-dev-server please share, otherwise it seems like a low priority and I prefer not to spend any more this. Our team will have to adopt the good practice of typing `https`.

PR #2542, opened 2019-05-03
Fix incorrect URL for non-auctioned ENS domains

nickmccurdy:
Reported by @jspence425 

Please review as soon as possible, ENS manager updates on May 4th and it would be great to release this hotfix today before users will start relying on it more.

https://mycryptobuilds.com/hotfix/incorrect-url-for-non-auctioned-ens-domains/#/ens

## Expected behavior
Visiting a non-auctioned ENS domain goes to a valid ENS manager URL ending in `.eth`.

## Actual behavior
Visiting a non-auctioned ENS domain goes to an invalid URL not ending with `.eth` which has JS errors that lead to a blank page.


nickmccurdy:
https://app.clubhouse.io/mycrypto/story/758/fix-incorrect-url-for-non-auctioned-ens-domains

PR #2543, opened 2019-05-03
Add WalletType filtering for networks tab

blurpesec:
###### To play around on this: 
https://mycryptobuilds.com/network-filter-unsupported-wallets

### Description

Only shows supported networks when the user selects a wallet type

### Changes

* Added dPath shit to local cache initialization.
* Added filtering for networks where wallet type is supported. *This depends on the presence of a DPath option in the Network's Config List*.

### Steps to Test

1. Navigate to http://localhost/add-account. Access a ledger device.
2. Try to select the Webchain network. Note that it does not exist in the list.

### DIFF
--- a/common/v2/features/AddAccount/AddAccount.tsx
+++ b/common/v2/features/AddAccount/AddAccount.tsx
@@ -51,7 +51,7 @@ import * as WalletActions from 'v2/features/Wallets';
 import { NetworkOptionsContext, AccountContext } from 'v2/providers';
 import { Account } from 'v2/services/Account/types';
 import { Web3Decrypt } from 'components/WalletDecrypt/components/Web3';
-import { getNetworkByName } from 'v2/libs';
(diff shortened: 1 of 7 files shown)


SharonManrique:
@blurpesec The Network dropdown does not list networks available when I'm on ETH

blurpesec:
> @blurpesec The Network dropdown does not list networks available when I'm on ETH (used Ledger)
> Select address panel - Networks are indeed filtered. But would need to test this amongst all other Networks from the first panel

@SharonManrique I think i'm having trouble understanding. You're saying that network dropdown does not list available networks when accessing using a Ledger device?

nickmccurdy:
The filtering affects the dpath dropdown in select address, right?

blurpesec:
> The filtering affects the dpath dropdown in select address, right?

Nope. It's the network's displayed in the network dropdown - network select view.

PR #2546, opened 2019-05-04
Added handling for new-registrar ens name resolve

### LINKED ISSUE #2545: Names Migrated to new ENS Registrar don't resolve
### Description of Issue

After migrating an ens name to the new registrar, it no longer resolves on MyCrypto.

### What exactly were you doing when you encountered this issue?

View-Only address page in https://localhost:3443/account
Type in `mycryptoid.eth` and it resolves correctly.
Type in `harrydenley.eth` and it does not resolve correctly.

blurpesec:
### Closes #2545 
🎉 🎉 🎉

### Description

Re-enables ENS resolution for registrar 

### Changes

* Enables handling of ens names that have been migrated to the new ENS registrar.

### Steps to Test

1. Access http://localhost/account. Select "View Only Account"
2. In the bottom field, enter "mycryptoid.eth" this should resolve correctly. - This is on the old registrar
3. In the bottom field, enter "harrydenley.eth" this should resolve correctly. - This is on the new registrar
4. In the bottom field, enter "mycryptoidddddd.eth" this should *NOT* resolve correctly. - This is not registered

### DIFF
--- a/common/features/ens/helpers.ts
+++ b/common/features/ens/helpers.ts
@@ -122,6 +122,24 @@ export function* resolveDomainRequest(name: string): SagaIterator {
   const nameStateHandler = modeMap[domainData.mode];
   const result = yield call(nameStateHandler, domainData, nameHash);
 
+  if (!result.hasOwnProperty('resolvedAddress')) {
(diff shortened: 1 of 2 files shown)


SharonManrique:

PR #2549, opened 2019-05-07
Add assetOptions in LocalCache initialization

blurpesec:
### Description

Adds default tokens to cache on cache initialization.

### Changes

* Added a method to add default tokens to cache on initialization
* Copied over config folder to `v2`.
* Added a method to write tokens from `ethereum-lists/tokens` repo to `v2/config` folder with a different structure in addition to the old folder (I included `name` field to conform to `AssetOption` type for the new folder's token data structure).

### Steps to Test

1. Delete localcache in browser, Hard reset page. Open dev tools in browser, navigate to *Application* tab, click on *Local Storage* then localhost and see if the tokens were added to the local cache under the MyCryptoCache key


SharonManrique:
@nickmccurdy Its quite easy to tell in this pr what was copied over and what has been changed. I do agree that in the future, we should differentiate between setting up foler/file structure and code. But for now, I think it's safe to merge. Its functional and changes look good.

PR #2554, opened 2019-05-09
.nvmrc use lts/carbon

vilsbole:
nvm offers aliases for lts versions and lts/carbon ^8.9.0 points to 8.16.0.
Since most of the changes are security patches we should update the version.


409H:
I've no objections to this being merged - a minor update is backwards compatiable and brings in some security updates. :+1: from me

nickmccurdy:
I checked with a bunch of people, this should be safe to update.

SharonManrique:

PR #2558, opened 2019-05-10
Feature/dashboard notifications

PeterZlodej:


amacar:
> Nice work, I like the props API of NotificationWrapper.
> 
> Do you think we could set most of the Image and ResourceItem component styles in that component so individual instances of notifications would just pass data and handler props (rather than custom styling their contents)?

We changed left Image component, and it now accepts props. 
For resources, we don't think that creating a general component is the best idea because on the right side of notification could be buttons (some even have icons), images with text beneath and with actions, etc. In this case, a lot of things would need to be passed. In the future, there would probably exist more different notifications, and resource component would become even more complex with more props.

PR #2440, opened 2019-03-18
Upgrade packages to pass yarn audit

nickmccurdy:

PR #2565, opened 2019-05-15
Fix yarn audit for production/Electron package dependencies

nickmccurdy:
This replaces #2440, but it only updates dependencies of the following packages, which are either used in production or affect Electron builds.

- electron
- electron-builder
- electron-updater
- mycrypto-shepherd


nickmccurdy:
https://app.clubhouse.io/mycrypto/story/684/upgrade-packages-to-pass-yarn-audit

nickmccurdy:
I fixed the TS error, something else is going on with AWS

nickmccurdy:
It's pretty unlikely that this would break anything because all our top-level dependencies are locked to specific version numbers in `package.json`, and I didn't change them. If a nested dependency breaks something for end users of the dependent package, it's supposed to change its version, so we'd be safe.

Do you think we should still do some basic smoke testing to be extra careful, or merge as-is?

409H:
Yeah, I agree it's unlikely though I remember @SharonManrique mentioning something about a hardware wallet not functioning correctly with MyCrypto on a dependancy update but I forgot which, so thought I'd mention something.

As long as Sharon can say the hardware wallet functions good, I vote to merge.

SharonManrique:
@409H I will test this for the release tagging this week.

PR #2569, opened 2019-05-16
Add WebUSB support for Ledger devices

Mrtenz:
### Description

This PR adds [WebUSB](https://wicg.github.io/webusb/) support using Ledger's `@ledgerhq/hw-transport-webusb`. This solves issues like the [U2F timeout](https://support.ledger.com/hc/en-us/articles/360018810413-U2F-timeout-in-Chrome-browser) caused by a recent Chromium update. Currently WebUSB is only supported by Chrome and Opera. The "old" U2F method is used in browsers that don't support WebUSB. This only applies to the web version of MyCrypto, the desktop version is unchanged.

I also updated other Ledger libraries, to make sure we don't have any duplicates.

### Changes

* Use `@ledgerhq/hw-transport-webusb` as transport method if it's supported.

### Steps to Test

1. Access a Ledger device with MyCrypto as usual.
2. Select your device (see screenshot below).
3. Try sending a transaction, signing a message, etc.


common/typescript/ledger/hw-transport-webusb.d.ts
```diff
@@ -0,0 +1,74 @@
...
+
+    /**
+     *
+     * @description Close the exchange with the device.
+     * @returns {Promise<void>}
+     * @memberof TransportUSB
+     */
+    public close(): Promise<void>;
+  }
+}
```
nickmccurdy: If these types are stable and complete, we should consider contributing them to [DefinitelyTyped](https://github.com/DefinitelyTyped/DefinitelyTyped) so any package can install them from `@types/ledgerhq__hw-transport-webusb`.

nickmccurdy:
Awesome, this should be really helpful for Ledger users. Can U2F still be used on browsers that don't support WebUSB?

Mrtenz:
The code checks if WebUSB is supported and will fall back to U2F if it's not.

The current typings aren't fully accurate, I copied them from the `hw-transport-u2f` and modified it a bit so it works for us. I'll consider adding it to DefinitelyTyped.

PR #2573, opened 2019-05-21
Attempt to use yarn pnp for faster installs

nickmccurdy:
Currently failing at `yarn install`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -35,6 +35,7 @@ build/Release
 # Dependency directories
 node_modules
 jspm_packages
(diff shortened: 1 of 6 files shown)


nickmccurdy:
I tried this on a personal project with `create-react-app` and it still seems to be pretty unstable, I also got errors on `yarn install`. I may look into this more, but it may require our dependencies updating to support pnp.

nickmccurdy:
Looks like `electron-builder` doesn't support this.

PR #2575, opened 2019-05-22
set state default to enable development

vilsbole:


nickmccurdy:
What's this for, does it solve an issue with local storage/cache migration?

Also the formatter (Prettier) is failing, please run `yarn formatAll` and use a Prettier extension in your editor in the future.

vilsbole:
This was just a quick fix to share with the send-assets team to get the flow to work.
Deleting because it's no longer needed.

NB. I setup the extension yesterday but seem to still have issues. formatAll changes files I've never touched... investigating today

PR #2587, opened 2019-05-24
ev/update dependency

vilsbole:
After pulling the latest changes webpack would fail to launch the dev server with this error:
https://github.com/mzgoddard/hard-source-webpack-plugin/issues/316

Updating the package to 0.13.1 fixes the issue.


miagx:
Staging build: [MyCryptoBuilds](https://mycryptobuilds.com/pr/2587)

PR #2595, opened 2019-05-28
Sign transaction with SafeTmini styles

SharonManrique:
Web views only
https://app.clubhouse.io/mycrypto/story/553/step-2-sign-transaction-safe-t-mini-styles

image

### DIFF
--- a/common/v2/features/SendAssets/components/SignTransactionWallets/Ledger.scss
+++ b/common/v2/features/SendAssets/components/SignTransactionWallets/Ledger.scss
@@ -1,0 +1,58 @@
+.SignTransactionLedger-panel {
+  width: 500px;
+  height: 600px;
+  padding-left: 40px;
+  padding-right: 40px;
+}
+
+.SignTransactionLedger-title {
+  height: 78px;
+  font-family: Lato;
+  font-size: 32px;
+  font-weight: bold;
+  text-align: center;
+  margin-top: 40px;
+}
+
+.SignTransactionLedger-instructions {
+  padding-top: 2em;
+  font-size: 18px;
+  line-height: 1.5;
+  text-align: center;
+}
+
+.SignTransactionLedger-without-pass {
+  margin-top: 5em;
+}
+
+.SignTransactionLedger-img {
+  display: flex;
+  justify-content: center;
+  margin: 3em;
+}
+
+.SignTransactionLegder-description {
+  font-size: 18px;
+  line-height: 1.5;
+  text-align: center;
+  margin-bottom: 1em;
+}
+
+.SignTransactionLedger-footer {
+  width: 100%;
+}
+
+.SignTransactionLedger-button {
+  width: 100%;
+}
+
+.SignTransactionLedger-help {
+  margin-top: 1em;
+  text-align: center;
+  font-size: 16px;
+}
+
+.SignTransactionLedger-referal {
+  text-align: center;
+  font-size: 16px;
+}
--- a/common/v2/features/SendAssets/components/SignTransactionWallets/Metamask.tsx
+++ b/common/v2/features/SendAssets/components/SignTransactionWallets/Metamask.tsx
@@ -1,9 +1,47 @@
 import React, { Component } from 'react';
 import MetamaskSVG from 'common/assets/images/wallets/metamask-2.svg';
 import './MetaMask.scss';
+import * as ethers from 'ethers';
+import { AsyncSendable } from 'ethers/providers/web3-provider';
+import { ISendState } from '../../types';
 
-export default class SignTransactionMetaMask extends Component {
+type AsyncSendableMetamask = AsyncSendable & { enable(): Promise<boolean> };
+
+declare global {
+  interface Window {
+    ethereum?: AsyncSendableMetamask;
+    web3: any;
+  }
+}
+
+interface Props {
+  stateValues: ISendState;
+}
+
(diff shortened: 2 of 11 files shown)


common/v2/features/SendAssets/components/SignTransactionWallets/Metamask.tsx
```diff
@@ -27,4 +98,61 @@ export default class SignTransactionMetaMask extends Component {
...
+      this.checkAddressMatches(checksumAddress, this.props.stateValues);
+      console.log(this.state);
+    } else {
+      const defaultProvider = ethers.getDefaultProvider(DEFAULT_NETWORK_FOR_FALLBACK);
+      this.setState({ provider: defaultProvider });
+    }
+  }
+
+  private watchForAccountChanges(ethereum: NonNullable<Window['ethereum']>) {
+    ethereum.on('accountsChanged', this.getMetaMaskProviderAndSigner);
```
vilsbole: We don't need to reconnect to the network on each accountChange.
We can simplify by only getting the address, thus removing the recursive call.

common/v2/features/SendAssets/components/SignTransactionWallets/Ledger.scss
```diff
@@ -0,0 +1,58 @@
+.SignTransactionLedger-panel {
+  width: 500px;
+  height: 600px;
+  padding-left: 40px;
+  padding-right: 40px;
+}
+
+.SignTransactionLedger-title {
```
vilsbole: Thanks for using explicit classnames!

package.json
```diff
@@ -28,10 +28,12 @@
     "crypto-js": "3.1.9-1",
     "electron-updater": "2.21.10",
     "ethereum-blockies-base64": "1.0.2",
-    "ethereumjs-abi": "git://github.com/ethereumjs/ethereumjs-abi.git#09c3c48fd3bed143df7fa8f36f6f164205e23796",
+    "ethereumjs-abi":
```
vilsbole: Did you intentionally add a new line for each of these?

common/v2/features/SendAssets/components/SignTransactionWallets/Metamask.tsx
```diff
@@ -14,6 +76,15 @@ export default class SignTransactionMetaMask extends Component {
           <img src={MetamaskSVG} />
         </div>
         <div className="SignTransactionMetaMask-input">
+          {this.state.metaMaskAccountMatches === false && (
+            <div className="SignTransactionMetaMask-wrong-address">
+              {' '}
```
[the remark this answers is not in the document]
SharonManrique: This space was automatically inserted by prettier.

common/v2/features/SendAssets/components/SignTransactionWallets/Metamask.tsx
```diff
@@ -1,9 +1,71 @@
...
+  account: string;
+  network: string;
+  metaMaskSigner: any;
+  metaMaskProvider: Web3Provider | null;
+  metaMaskAccountMatches: boolean;
+  metaMaskNetworkMatches: boolean;
+  initAccountPolling: null;
+}
+
+const DEFAULT_NETWORK_FOR_FALLBACK = 'ropsten';
```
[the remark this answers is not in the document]
SharonManrique: Networks are going to be apart of the signing transaction component for other wallets, except MetaMask. Still figuring this one out.

common/v2/features/SendAssets/components/SignTransactionWallets/Metamask.tsx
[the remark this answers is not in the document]
nickmccurdy: Is the space between this child and the beginning of the container necessary? It's probably not, in which case it can be removed.

package.json
nickmccurdy: We seem to have some weird formatter clashes with package.json. Whenever I edit it, it often toggles hard wrapping these key/value pairs on/off. I have a feeling that Prettier turns it on and yarn turns it off.

common/v2/features/SendAssets/components/SignTransactionWallets/Metamask.tsx
```diff
@@ -27,4 +98,61 @@ export default class SignTransactionMetaMask extends Component {
...
+    setInterval(this.getMetaMaskProviderAndSigner, 2000);
+  }
+
+  private checkAddressMatches(metaMaskAddress: string, stateValues: ISendState) {
+    const desiredAddress = utils.getAddress(stateValues.transactionFields.senderAddress);
+
+    if (metaMaskAddress === desiredAddress) {
+      return this.setState({ metaMaskAccountMatches: true });
+    } else {
+      return this.setState({ metaMaskAccountMatches: false });
```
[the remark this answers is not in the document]
SharonManrique: Its false again if user for some reason switches accounts from correct one back to a different one.

package.json
vilsbole: It seems that we have 4 different configs for Prettier (tests, formatAll, travis, and VSC).
We need to bring it down to a single one.

nickmccurdy: All those scripts and tools should be using the same Prettier config: https://github.com/MyCryptoHQ/MyCrypto/blob/develop/.prettierrc

I think the issue is with Yarn, because it has its own JSON formatter that doesn't use Prettier. I agree that we should make it consistent, but I'm not sure what the best approach is.

PR #2596, opened 2019-05-28
Update pull request template

nickmccurdy:
## Changes

* Adds commonly used sections from core and UI
* Explains Clubhouse usage for employees
* Merges common information from employees and contributors
* Catches some common quality assurance needs (more to be added in the future as our branching model and test plan evolve)

## Quality Assurance

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -29,6 +29,7 @@
 <!-- Upload screenshots here. -->
 
 ## Quality Assurance
+
 - [ ] The branch name is in lowercase-kebab-case with no prefix (unless it was created from Clubhouse)
 - [ ] The base branch is develop or gau (no nested branches)
 - [ ] This is related to a maximum of one Clubhouse story or GitHub issue
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,29 +1,29 @@
-*To help us understand & test your amazing contribution, please include the below information with your PR. 🙏 Thanks!*
+<!-- Employees: Please use Clubhouse's "Open PR" button from the relevant story or include links to relevant Clubhouse stories in your branch name, commit messages, or pull request comments. Do not add links to your pull request description, they will be ignored. https://help.clubhouse.io/hc/en-us/articles/207540323-Using-The-Clubhouse-GitHub-Integration -->
 
-
-### Closes #ISSUE_NUMBER_GOES_HERE
+<!-- Employees: Delete this section. -->
+## Closes #ISSUE_NUMBER_GOES_HERE
 
 🎉 🎉 🎉
 
-### Description
+## Description
 
 ......
 
-
-### Changes
+## Changes
 
 * High level
 * changes that
 * you made
 
+### Reusable Code/Components
 
-### Steps to Test
+<!-- Employees: Delete this section if there's a clear spec or test plan in Clubhouse. -->
+## Steps to Test
 
 1. Steps
 2. to
 3. test
 
-
-### Screenshots
-
-...only if applicable...
+<!-- Contributors: Delete this section. -->
+## [Zeplin Design](https://app.zeplin.io/project/5b0334f5e91e8c481645ad56)
+<!-- Upload screenshots here. -->
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -17,7 +17,7 @@
 
 ### Reusable Code/Components
 
-<!-- Employees: Delete this section if there's a clear spec or test plan in Clubhouse. -->
+<!-- Preferably, include automated tests instead. -->
 ## Steps to Test
 
 1. Steps
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -27,3 +27,10 @@
 <!-- Contributors: Delete this section. -->
 ## [Zeplin Design](https://app.zeplin.io/project/5b0334f5e91e8c481645ad56)
 <!-- Upload screenshots here. -->
+
+## Quality Assurance
+- [ ] The branch name is in lowercase-kebab-case with no prefix (unless it was created from Clubhouse)
+- [ ] The base branch is develop or gau (no nested branches)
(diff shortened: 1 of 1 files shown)


nickmccurdy:
https://app.clubhouse.io/mycrypto/story/647/propose-pull-request-template-for-core

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -1,29 +1,37 @@
...
 
+<!-- Contributors: Delete this section. -->
+## [Zeplin Design](https://app.zeplin.io/project/5b0334f5e91e8c481645ad56)
+<!-- Upload screenshots here. -->
 
-### Screenshots
+## Quality Assurance
 
-...only if applicable...
+- [ ] The branch name is in lowercase-kebab-case with no prefix (unless it was created from Clubhouse)
```
vilsbole: I'm for prefixing a PR branch with our initials. eg. `ev/sender-address` it has the benefit of making it's author responsible for it's branch and encourage us to reduce it's life span.

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -1,29 +1,37 @@
...
+- [ ] The base branch is develop or gau (no nested branches)
+- [ ] This is related to a maximum of one Clubhouse story or GitHub issue
+- [ ] Types are safe (avoid TypeScript/TSLint features like any and disable, instead use more specific types)
```
vilsbole: This should be handled by our linter. I don't see why we need a reminder here.

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -1,29 +1,37 @@
...
+- [ ] If code is copied from existing directories, there is an explanation of why this is necesary in the description/changes, and all copying is done in separate commits to make them easy to filter out
```
vilsbole: I don't understand the filtering requirement? To simplify code review? Even a copy/paste from an existing codebase should be subject to our new standards.

.github/PULL_REQUEST_TEMPLATE.md
nickmccurdy: I think this could get confusing when stories are reassigned between developers. You could always see the latest author of a particular branch with `git show`, and GitHub's branch page also includes authors if you want to prune branches.

.github/PULL_REQUEST_TEMPLATE.md
nickmccurdy: Our linter allows unsafe types like `any` and `{}` if they're used explicitly. This is still dangerous, but changing the linter setting would require us to fix a lot of existing TypeScript types. In this case it's easier for us to catch the issue in manual reviews until we can fix the overall type safety in the future.

.github/PULL_REQUEST_TEMPLATE.md
nickmccurdy: By filtering I mean using the select commits dropdown to see all commits except the ones that copy files:

image

The problem is less about code quality, and more that it's difficult to review what changed when a file is copied unless the copy and change are done in separate commits (in which case you can just hide the copy commit in the dropdown to focus on changes).

.github/PULL_REQUEST_TEMPLATE.md
vilsbole: As you say there are multiple ways to find the author of a PR but the objective of prefixing is more to encourage short-lived branches. It's a small measure to encourage the right behaviour of small PR + quick merges + branch deletion.

If we're able to reach this objective, stories would no longer need to be reassigned because they will be completed before a reason occurs.

.github/PULL_REQUEST_TEMPLATE.md
vilsbole: Ok.
An alternative in the future would be to run linting only on changed files, which would allow us to gradually enforce better standards

.github/PULL_REQUEST_TEMPLATE.md
vilsbole: Understood. But I'm still inclined to judge a copy/paste as if it was new code and ask the reviewer to reevaluate it. Ignoring on only looking at changes of copied code would nudge the reviewer to overlook them.

(I'm just trying to stay KISS and keep conventions down to the ones that we can actually automate)

.github/PULL_REQUEST_TEMPLATE.md
nickmccurdy: Our linter already only runs on changed file. TypeScript does not support this.

.github/PULL_REQUEST_TEMPLATE.md
nickmccurdy: I'm against adding any info other than Clubhouse linking to branch names honestly

.github/PULL_REQUEST_TEMPLATE.md
nickmccurdy: That's a valid point, but you can still do that with separate commits, and the separate commits make it easier to review in isolation.

.github/PULL_REQUEST_TEMPLATE.md
nickmccurdy: I just checked and our current TSLint setup only runs on changed files. Unfortunately this isn't possible with TypeScript because the compiler needs to run with the entire project to determine types.

vilsbole: Didn't think of that though I believe that tsc can resolve the imports and find the used types.

.github/PULL_REQUEST_TEMPLATE.md
vilsbole: Ok let's try and see how it works

.github/PULL_REQUEST_TEMPLATE.md
vilsbole: I don't see the benefit and doubt this would be respected, especially if we can't enforce it. As you wish

.github/PULL_REQUEST_TEMPLATE.md
nickmccurdy: The problem is that as far as I'm aware, TypeScript needs to typecheck all imported files to understand their types, and typechecking can't be enabled or disabled per file because of the way type errors are immplemented with the typechecker. The TypeScript team seems to be against implementing this, and suggests that users should migrate their code in one fell swoop.

So ideally, we should improve as much of these unsafe type practices as possible (any, disable, overly wide types, casts, etc.) and then make our linter rules more strict (when we have time).

PR #2601, opened 2019-05-31
Feature/create new wallet

amacar:
Create new keystore / mnemonic flow


vilsbole:
closing in favour of https://github.com/MyCryptoHQ/MyCrypto/pull/2629 which is more recent.

PR #2605, opened 2019-06-03
Advanced Transaction Field tweaks / refactor

blurpesec:
### Description

Implements field-level validation for advanced transaction fields. Refactored advanced fields to conform with the format that @vilsbole used in sender-address

### DIFF
--- a/common/v2/features/SendAssets/components/validators/validators.ts
+++ b/common/v2/features/SendAssets/components/validators/validators.ts
@@ -1,0 +1,20 @@
+export function validateGasPriceField(value: string): boolean {
+  const valid = parseInt(value, 10) >= 0 && parseInt(value, 10) <= 3000;
+  return valid;
+}
+
+export function validateGasLimitField(value: string): boolean {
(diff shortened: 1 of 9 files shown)


common/v2/features/SendAssets/components/validators/validators.ts
```diff
@@ -0,0 +1,33 @@
+import { isValidHex, gasPriceValidator, gasLimitValidator } from 'v2/libs/validators';
+
+export function validateGasPriceField(value: string): string | undefined {
+  let error;
+  if (!gasPriceValidator(value)) {
+    error = 'Gas price must be a valid number.';
+  }
+  return error;
+}
```
nickmccurdy: We can simplify the syntax of these validation functions, for example:
```ts
export function validateGasPriceField(value: string) {
  if (!gasPriceValidator(value)) {
    return 'Gas price must be a valid number.';
  }
}
```

blurpesec: done

SharonManrique:
One request for next pr iteration -> Highlight the errored field in red along with the error msg

PR #2608, opened 2019-06-03
Remove fiat dropdown for MVP

Sonicrida:
### Changes

* Removes the fiat dropdown from the dashboard

### Screenshots

image


common/v2/features/Dashboard/components/WalletBreakdown.tsx
```diff
@@ -76,11 +76,6 @@ function WalletBreakdown() {
   return (
     <div className="WalletBreakdown">
       <div className="WalletBreakdown-selectWrapper">
-        <ComboBox
-          className="WalletBreakdown-selectWrapper-select"
```
nickmccurdy: Please remove the relevant SCSS

nickmccurdy:
https://app.clubhouse.io/mycrypto/story/1321

common/v2/features/Dashboard/components/WalletBreakdown.tsx
Sonicrida: this is still used for the accounts dropdown actually

nickmccurdy:
Thanks, this is better.

On a wide (1440p) screen the proportions look better:
image

But on a skinner screen it's proportionately a lot wider:
image

Do we need this extra width on smaller screens or can we make it relative to the dashboard column's width?

PR #2614, opened 2019-06-06
Update prettier and ts

vilsbole:
Update dev dependencies:
- prettier
- typescript
- tslint
- tslint-config-prettier
- tslint-microsoft-contrib
- tslint-react

1. Looking at the changes made by prettier they do work better with react
2. I added a few @ts-ignore on the old code to pass the tsc.
3. I didn't time, but out tscheck seems to run faster.


vilsbole:
Once merged in `develop` we can do the same on `gau`.

Issue #2620, opened 2019-06-07
QR codes not working in night mode (Parity Signer)


Tbaut:
### Description of Issue

QR codes need a white border to be able to be read without problems.
In night mode, QR codes look like the attached screenshot and can't be read by applications such as Parity Signer.

### How were you accessing your wallet?

Parity Signer

### Description of Your Machine

- Browser:
- Browser version:
- Operating system:
- Device:
- MyCrypto.com vs Desktop Application: same
- MyCrypto Version Number: master

###  Screenshots

![image](https://user-images.githubusercontent.com/33178835/59103813-27c26e80-8930-11e9-82a0-a4da62092dba.png)

PR #2621, opened 2019-06-07
White border for Parity Signer QR codes

Tbaut:
## Closes https://github.com/MyCryptoHQ/MyCrypto/issues/2620

🎉 🎉 🎉

## Description

This adds a white border to make sure the QR code can be scanned correctly.

## Changes

Self explanatory. 

## Steps to Test

1. Connect with Parity Signer
2. Make a transactions
3. Click on "Sign"

## Quality Assurance

### DIFF
--- a/common/components/ParityQrSigner.scss
+++ b/common/components/ParityQrSigner.scss
@@ -11,7 +11,11 @@
   margin: 0 auto 1.5em;
   border-radius: 4px;
   overflow: hidden;
-
+  
+  & img {
+    border: solid white 15px;
+  }
+  
   &.is-disabled {
     background: color(gray-lighter);
   }


nickmccurdy:
Thanks, this seems to be a good solution but I don't use Parity myself. Were you able to test it and fix the issue in dark mode?

Tbaut:
Yes I built and tested it. Testing is actually super easy on iOS or Android. Just download Parity Signer, create an account, log in on MyCrypto with this account, make a transaction to yourself (no need to have ETH as long as you send 0). Make sure MyCrypto is in night mode, click on scan and you'll see the modifies QR code.

SharonManrique:
@Tbaut Thank you for your contribution! We highly appreciate it :D

PR #2623, opened 2019-06-07
Sender Address Filtering

blurpesec:
### Description
Filtering for sender address

### Changes
Filters sender address dropdown based on the network of the asset selected. If no asset is selected, default to empty dropdown.

~*Get UX feedback before publishing*~


nickmccurdy:
We decided to fix the CORS issues later (before deploying GAU/MVP) since the images weren't added in this PR

PR #2624, opened 2019-06-07
Address Book Labeling in add-existing-account

blurpesec:
### Description
Implements address book labels.

### Changes
* Adds handling for default account labeling
* Makes UUID not tied to address

To Test:
1 ) Add multiple accounts using the same network. It should increment their names and add the labels to local cache.
2 ) Add multiple accounts using different networks. Should not increment names, but will add labels to local cache.


common/v2/features/AddAccount/components/SaveAndRedirect.tsx
```diff
@@ -19,7 +24,11 @@ function SaveAndRedirect(payload: { formData: FormData }) {
   const { displayNotification } = useContext(NotificationsContext);
   useEffect(() => {
     const network: Network | undefined = getNetworkByName(payload.formData.network);
-    if (!network || getAccountByAddress(payload.formData.account)) {
+    if (
+      !network ||
+      getAccountByAddressAndNetwork(payload.formData.account, payload.formData.network) !==
```
vilsbole: Prefer the double bang (!!) syntax to make sure you receive the account
`!!getAccountByAddressAndNetwork(...)`

common/v2/features/Dashboard/components/AccountDropdown.tsx
```diff
@@ -60,22 +61,31 @@ const SDropdown = dropdown`
...
-      key={uuid}
-      name={`account-${uuid}`}
-      checked={selected.includes(uuid)}
-      onChange={() => handleChange(uuid)}
-      label={label}
-      icon={() => <Identicon className="AccountDropdown-menu-identicon" address={address} />}
-    />
-  ));
+  accounts.map(({ uuid, address }: ExtendedAccount) => {
+    const detectedLabel: AddressBook | undefined = labels.find(
```
vilsbole: 1. Naming is important (and hard). When you right 'detected' I'm thinking of listening, scanning, watching, so imagining that you are catching something that is in movement. This is just more like a 'join' between structures. ~~Can we just use `accountLabel` | `label` instead?~~

I see that you need two variable names. How about

```
const addressCard = labels.find(...)
const addressLabel = addressCard ? addressCard.label : 'Unknown Account'
```

2. I thought we had the 'label' included in the account?
`.find()` returns only the first match, and it's my understanding that we do the address+network combination because the same address can be used on different networks. If we store the label under addressBook, we will be unable to provide a different label for address+network1 AND address+network2. And if we don't support this usecase we can simplify a lot of our data structure

3. I see you have a helper function `getLabelByAcocunt` let's use it here as well!

common/v2/features/Dashboard/components/AccountDropdown.tsx
```diff
@@ -60,22 +61,31 @@ const SDropdown = dropdown`
...
+      label => label.address.toLowerCase() === address.toLowerCase()
+    );
+    const addressLabel = !detectedLabel ? 'Unknown Account' : detectedLabel.label;
```
vilsbole: Avoid the negation in a ternary (harder to reason). It's simpler to just invert the expressions.
```
 const addressLabel = detectedLabel ? detectedLabel.label :  'Unknown Account' ;
```

common/v2/features/Dashboard/components/AccountList.tsx
```diff
@@ -60,15 +60,13 @@ function buildAccountTable(accounts: ExtendedAccount[], deleteAccount: DeleteAcc
       translateRaw('ACCOUNT_LIST_DELETE')
     ],
     body: accounts.map(account => {
+      const detectedLabel: AddressBook | undefined = getLabelByAccount(account);
```
vilsbole: i.e naming

common/v2/features/Dashboard/components/AccountList.tsx
```diff
@@ -60,15 +60,13 @@ function buildAccountTable(accounts: ExtendedAccount[], deleteAccount: DeleteAcc
...
+      const label = !detectedLabel ? 'Unknown Account' : detectedLabel.label;
```
vilsbole: i.e. negation in ternary

common/v2/features/Dashboard/components/AccountList.tsx
```diff
@@ -60,15 +60,13 @@ function buildAccountTable(accounts: ExtendedAccount[], deleteAccount: DeleteAcc
...
         // tslint:disable-next-line: jsx-key
-        <Address
-          title={`${account.label}-(${account.wallet})`}
-          address={account.address}
-          truncate={truncate}
-        />,
+        <Address title={`${label}`} address={account.address} truncate={truncate} />,
         // tslint:disable-next-line: jsx-key
         <Network color="#a682ff">{account.network}</Network>,
         // tslint:disable-next-line: jsx-key
```
vilsbole: The table body is strange, but the disable-next-line stranger (if tslint is complaining react will complain as well during runtime).

How about adding something like this:

```
body: accounts.map((account, idx) => {
 const bodyItemCount = 0;
 return [
     <Icon key={idx + bodyItemCount++} icon="star" />,
      <Address key={idx + bodyItemCount++} title={`${label}`} address={account.address} truncate={truncate} />,
   //....
 ]
}
```

common/v2/features/DevTools/DevTools.tsx
```diff
@@ -50,64 +52,68 @@ const DevTools = () => {
...
-                            onChange={handleChange}
-                            onBlur={handleBlur}
-                            value={values.address}
-                          />
-                        )}
-                      />
-                    </fieldset>
+                {({ values, handleChange, handleBlur, isSubmitting }) => {
+                  const detectedLabel: AddressBook | undefined = getLabelByAccount(values);
+                  const label = !detectedLabel ? 'Unknown Account' : detectedLabel.label;
```
vilsbole: idem

common/v2/features/DevTools/ToolsAccountList.tsx
```diff
@@ -22,12 +23,18 @@ export interface AccountListProps {
...
-      <DeleteButton onClick={() => deleteAccount(account.uuid)} icon="exit" />
-    </AccountContainer>
-  ));
+  let key = 0;
+  const list = accounts.map((account: ExtendedAccount) => {
+    key += 1;
+    const detectedLabel: AddressBook | undefined = getLabelByAccount(account);
+    const label = !detectedLabel ? 'Unknown Account' : detectedLabel.label;
+    return (
+      <AccountContainer key={key}>
```
vilsbole: the second argument of `map` is the index of the current elem. You can use this as the key.
```
const list = accounts.map((account, idx) => {
  //  ...
   <AccountContainer key={idx}>
  //...
}

File: common/v2/libs/addressbook/addressbook.ts [hunk 9]
```diff
@@ -0,0 +1,35 @@
...
+  let defaultLabel: string | undefined;
+  do {
+    const isFound: AddressBook | undefined = addressLabels.find(
+      label => label.label === `New ${networkName} Account ${iterator}`
+    );
+    defaultLabel = !isFound ? `New ${networkName} Account ${iterator}` : undefined;
+    iterator += 1;
+  } while (!defaultLabel);
+  return defaultLabel;
+};
```
vilsbole: Cool to discover the `do {} while ()` syntax but simply NO.

__Danger!__ You are depending on the value of label to be defined in a specific format that is not under the control of your function (implicit dependency). The string value or the network name can change between deploys, position in list can change (iterator) etc.  
A single incident will cause an infinite loop. It is simpler to return the value of result of `find`

common/v2/features/Dashboard/components/AccountDropdown.tsx
blurpesec: Re:
> I thought we had the 'label' included in the account?

That's what im trying to remove in this PR. I'm tying labels to a mixture of network & address so that we don't have to query for labels from 2 different places when doing lookups.

PR #2626, opened 2019-06-10
CTA buttons on dashboard

Sonicrida:
## Description
Updates call to action buttons with proper routes and links

https://app.clubhouse.io/mycrypto/story/230/cta-buttons-left-side

https://app.zeplin.io/project/5b0334f5e91e8c481645ad56/screen/5c484e95d2cd65386e5b71f1
## Changes

* Proper paths added to buttons that change pages
* Random advertisement button

### DIFF
--- a/common/v2/features/Dashboard/components/ActionTile.scss
+++ b/common/v2/features/Dashboard/components/ActionTile.scss
@@ -67,3 +67,13 @@
     }
   }
 }
+
+.wallet-icon {
+  width: 35px;
+  height: 35px;
+
+  @media (min-width: 680px) {
+    width: 60px;
+    height: 60px;
+  }
+}
--- a/common/v2/features/Dashboard/components/ActionTile.tsx
+++ b/common/v2/features/Dashboard/components/ActionTile.tsx
@@ -1,17 +1,77 @@
-import React from 'react';
+import React, { useState } from 'react';
 import { Link } from 'react-router-dom';
 import { Button, Typography } from '@mycrypto/ui';
 
 import { Action } from '../types';
 import './ActionTile.scss';
+import { AnalyticsService, ANALYTICS_CATEGORIES } from 'v2/services';
 
 type Props = Action;
 
-export default function ActionTile({ icon, title, description, link }: Props) {
+const trackAction = (analyticsMessage: string) => {
+  AnalyticsService.instance.track(ANALYTICS_CATEGORIES.DASHBOARD, analyticsMessage);
+};
+
+const ActionTile = React.memo(function ActionTileButton({
+  icon,
+  title,
+  description,
+  path,
+  url,
+  tracking
+}: Props) {
+  const [randomWallet, setRandomWallet] = useState();
+  if (Array.isArray(url)) {
+    const walletOptions = 2;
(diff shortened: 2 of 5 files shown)


SharonManrique:
@jonathansmirnoff Request Button does not route properly. ^^"

common/v2/features/Dashboard/components/ActionTile.scss
```diff
@@ -67,3 +67,13 @@
     }
   }
 }
+
+.wallet-icon {
+  width: 35px;
+  height: 35px;
+
+  @media (min-width: 680px) {
```
vilsbole: we want to have a common place for all breakpoints and media queries otherwise it becomes painful to control.
Since you are in scss, use one of the variables in `common/sass/variables/grid.scss`

common/v2/features/Dashboard/components/ActionTile.tsx
```diff
@@ -1,22 +1,84 @@
...
+            {title}
+            <span className="ActionTile-button-title-description">{description}</span>
+          </Typography>
+        </Button>
+      </a>
+    );
+  }
   return (
-    <Link to={link} className="ActionTile">
+    <Link to={String(path)} className="ActionTile" onClick={() => trackAction(String(tracking))}>
```
vilsbole: question: what can path, tracking, icon be if they are not strings? I don't understand why you want to transform a string into a string :)

common/v2/features/Dashboard/components/ActionTile.tsx
```diff
@@ -1,22 +1,84 @@
...
+const ActionTile = React.memo(function ActionTileButton({
+  icon,
+  title,
+  description,
+  path,
+  url,
+  tracking
+}: Props) {
+  const [randomWallet, setRandomWallet] = useState();
+  if (Array.isArray(url)) {
```
vilsbole: Simply **NO**.
Instead of changing our view to try to match the data object, we prefer to [malaxate](https://en.wiktionary.org/wiki/malaxate) the data.
The tiles are identical between each other (a link containing img and button) and the component should reflect that. 

From what I understand:
1. we want to randomly select between hardware wallet options 🤢
2. so we add the different options to `actions` list.
3. then we iterate on the list and display ActionTile
4. ActionTile (who is already 2 degrees removed from the initial data structure) now has to be aware how to handle 3 different renders! This is additional complexity just creates more confusion and is harmful in the long run because it creates a strong interdependency between actions in Dashboard and ActionTile.

ActionTile should only be concerned the rendering of its own props.
```
  const selectedActions = actions.map((action) => {
      // we only want to transform the hardware action item so skip the others
      if (!Array.isArray(action.url)) { return action }
      // randomly select a hardware wallet to promote
      const randomIdx = Math.floor(Math.random() * Math.floor(walletOptions))
      return {
         ...action,
         url: action.urls[randomIdx],
         tracking: action.tracking[randomIdx],
         icons: action.tracking[randomIdx],
       }
   })

  // now ActionTile always receives the same props!
```

It could even be agnostic of `tracking` and only receive a click handler. But that's another conversation.

Sonicrida:
Handled in a separate PR

PR #2631, opened 2019-06-11
Ev/default fields

vilsbole:
- connect Address and Amount to form
- include validation
- restyle slider
- minor refactors

## Quality Assurance


common/v2/types/symbols.ts
```diff
@@ -1,4 +1,1317 @@
-// @TODO:SEND add all token symbols to type
-type TSymbol = 'ETH' | 'USD' | 'generic';
+// Create union type from array
+// https://stackoverflow.com/a/45257357/2057532
+// Solution is valid until we update to typescript 3.4 where we can then use
+// const symbol = [] as const
+export type Lit = string | number | boolean | undefined | null | void | {};
+export const tuple = <T extends Lit[]>(...args: T) => args;
+
+const symbols = tuple(
```
blurpesec: This won't work as the assets / their symbols change any time we update tokens.

vilsbole: Wouldn't be more like incrementing a list with the new tokens? Or do the symbols also change?

This was just a suggestion to give a bit more confidence when using a symbol|ticker to express the idea that it's more than a simple string and that we rely on it for filtering. I pushed changes that use create the equivalent of a type alias using enum + intersection with the base structure (aka. string). With this we have one way equality checks where a TSymbol can be assigned to a string, but a string needs to be cast as TSymbol.

blurpesec: > Wouldn't be more like incrementing a list with the new tokens? Or do the symbols also change?

Yeah currently everything about the tokens config can change from pulling tokens down from ethereum-lists/tokens.

Also adding custom tokens (tokens that arent specified in the list we import) would need to be handled by that type.

vilsbole: I updated the implementation to this: https://github.com/MyCryptoHQ/MyCrypto/blob/7387450e14ab41a95238d6c950e371eaf39a95b0/common/v2/types/symbols.ts
Let me know what you think.

blurpesec: Looks good!

Issue #1741, opened 2018-05-10
Waiting on MetaMask - Nothing Happens


Azarielle:
---

If you've experienced a bug, please provide the following information:

### Description of the Issue

When the MetaMask extension is having problems connecting to the ETH network, when you select "access wallet with MetaMask" nothing happens. Would prefer if it went to the next page where it asks to connect to MetaMask and says something to tell me that MetaMask is having connection problems. I had to manually open up MetaMask to see that it was having problems in order to figure out why MyCrypto wasn't taking me to the next page to connect to MetaMask. 

### Console Logs / Screenshots

![sharex_2018-05-04_18-11-25](https://user-images.githubusercontent.com/25834495/39888767-d88c58f0-5463-11e8-9e0e-295ae6bf2bad.png)
![image](https://user-images.githubusercontent.com/25834495/39888782-e2e70782-5463-11e8-8772-cce3f67973ff.png)
Nothing happens, no warning, just sits here. 

---

tayvano:
I just encountered this issue as well (but while using Peepeth) and fixed it by switching to Ropsten and then switching back. I'm wondering if this is due to a MetaMask update, not an update on our end?

PR #2634, opened 2019-06-12
Receive flow that generates a EIP681 string and displays a QR code (CH #1741)

Sonicrida:
## Changes

* implemented leftover receive flow code
* moved the code folder up one directory and out of the dashboard folder

- [x] Routing
- [x] Analytics // this is going to be worked on after MVP, its being tracked in CH
- [x] Port over existing string generation logic from v1
- [x] Display request string in codeblock
- [x] Display request string as QRCode
- [x] Allow user to select account to receive with using a dropdown
- [x] Validate that the amount is a valid amount
- [x] Display list of possible assets to receive


SharonManrique:
If a user were to manually input an Asset that does not exist, it show show up as invalid. 
image

SharonManrique:
When an invalid amount is entered, appropriate user feedback (A simple error msg) should be given to let users know that its invalid, instead of no feedback. 
image

SharonManrique:
The copied "Payment Code" and after QR code is scanned a user is given `ethereum:0x7750540B0363634f3b29579EDd9b2E4c0Fc4C502?value=33e18` However, a user might not know how to convert this into a usable URL string to be useable.

Sonicrida:
> Payment Code copy-button is a glitchy.

Yep! We'll be fixing that in a separate PR for the UI lib since it's an issue with the copyable component.

> If a user were to manually input an Asset that does not exist, it show show up as invalid.

I wonder how this should be handled because we wouldn't want to show invalid when the user is still typing. Should it be when the ComboBox loses focus perhaps?





> The copied "Payment Code" and after QR code is scanned a user is given `ethereum:0x7750540B0363634f3b29579EDd9b2E4c0Fc4C502?value=33e18` However, a user might not know how to convert this into a usable URL string to be useable. Could we add a little help icon somewhere, to show users how to use it?

We'll be handling what to do with this in a separate PR. I think the actual functionality is still being discussed but the current way this works is fine for this PR.







> When an invalid amount is entered, appropriate user feedback (A simple error msg) should be given to let users know that its invalid, instead of no feedback.
> image

So actually this exists but it only currently exists if a correct asset is displayed and an incorrect amount is displayed. I'll add this to the asset dropdown maybe if an invalid asset is selected once focus is lost?

common/v2/features/ReceiveAssets/ReceiveAssets.tsx
```diff
@@ -0,0 +1,255 @@
...
+
+export function ReceiveAssets({ history }: RouteComponentProps<{}>) {
+  const { accounts } = useContext(AccountContext);
+  const { assets } = useContext(AssetContext);
+  const network = getNetworkByName(accounts[0].network);
+  const filteredAssets = network
+    ? assets
+        .filter(asset => asset.networkId === network.id)
+        .filter(asset => asset.type === 'base' || asset.type === 'erc20')
+    : [];
```
vilsbole: this logic is used multiple times throughout the app.
create a `getAccountAssets()` method in `StoreProvider` instead

common/v2/features/ReceiveAssets/ReceiveAssets.tsx
```diff
@@ -0,0 +1,255 @@
...
+`;
+
+const ErrorMessage = styled.span`
+  color: red;
+`;
+
+export function ReceiveAssets({ history }: RouteComponentProps<{}>) {
+  const { accounts } = useContext(AccountContext);
+  const { assets } = useContext(AssetContext);
+  const network = getNetworkByName(accounts[0].network);
```
vilsbole: Are you sure about `accounts[0].network`?

common/v2/features/ReceiveAssets/ReceiveAssets.tsx
```diff
@@ -0,0 +1,255 @@
...
+    numberIsNotNegative(+amount) && validDecimal(amount, decimal);
+
+  const validateAmount = (amount: any) => {
+    let error;
+    if (selectedAsset) {
+      const { decimal } = selectedAsset;
+      if (decimal && !isValidAmount(decimal)(amount)) {
+        error = 'Please enter a valid amount';
+      }
+    }
```
vilsbole: Prefer form validation with `Yup` https://github.com/jquense/yup

common/v2/features/ReceiveAssets/ReceiveAssets.tsx
```diff
@@ -0,0 +1,255 @@
...
+  const [requestAddress, setRequestAddress] = useState('');
+
+  const initialValues = {
+    amount: '0',
+    asset: { label: 'Ethereum', id: '7bbf42b1-9275-120b-0d0a-a788abd75ea0' },
+    chainId: network ? network.chainId : 1
+  };
+
+  const isValidAmount = (decimal: number) => (amount: string) =>
+    numberIsNotNegative(+amount) && validDecimal(amount, decimal);
```
vilsbole: Here we're in a Function Component. Every time the component re-renders, every method is redefined.
This cost can be alleviated by defining helper methods out of the component scope.
In this case since `isValidAmout` is likely to be reused, move it to `v2/utils`

common/v2/features/ReceiveAssets/ReceiveAssets.tsx
```diff
@@ -0,0 +1,255 @@
...
+  const filteredAssets = network
+    ? assets
+        .filter(asset => asset.networkId === network.id)
+        .filter(asset => asset.type === 'base' || asset.type === 'erc20')
+    : [];
+  const assetOptions = filteredAssets.map(asset => ({ label: asset.name, id: asset.uuid }));
+  const [chosenAssetName, setAssetName] = useState(assetOptions[0].label);
+  const selectedAsset = filteredAssets.find(asset => asset.name === chosenAssetName);
+
+  const [requestAddress, setRequestAddress] = useState('');
```
vilsbole: Initialise state to undefined instead of an empty string.

common/v2/features/ReceiveAssets/ReceiveAssets.tsx
```diff
@@ -0,0 +1,255 @@
...
+              <Field
+                name="recipientAddress"
+                component={({ field, form }: FieldProps) => (
+                  <AccountDropdown
+                    name={field.name}
+                    value={field.value}
+                    accounts={accounts}
+                    onSelect={(option: IExtendedAccount) => {
+                      form.setFieldValue(field.name, option);
+                      setRequestAddress(option.address);
```
vilsbole: Formik already updates it's internal state with the new value on: `form.setFieldValue(field.name, option)`;
You can access it with `values` on line 141.
You don't need to keep an extra copy in the components state.

common/v2/features/ReceiveAssets/ReceiveAssets.tsx
```diff
@@ -0,0 +1,255 @@
...
+                    <StyledSelect
+                      name="Assets"
+                      className="select-container"
+                      options={assetOptions}
+                      value={field.value}
+                      onChange={(option: Option) => {
+                        form.setFieldValue(field.name, option);
+                        if (option.label) {
+                          setAssetName(option.label);
+                        }
```
vilsbole: _idem_

common/v2/features/ReceiveAssets/ReceiveAssets.tsx
```diff
@@ -0,0 +1,255 @@
...
+                      }}
+                    />
+                  )}
+                />
+              </Asset>
+            </AssetFields>
+            {errors.amount && (
+              <ErrorMessage>{' ' + translateRaw('RECEIVE_FORM_ERROR')}</ErrorMessage>
+            )}
+            {parseFloat(amount) >= 0 &&
```
vilsbole: - Big number math for amount.
- use Formik `values` to verify required fields

common/v2/features/ReceiveAssets/ReceiveAssets.tsx
```diff
@@ -0,0 +1,255 @@
...
+                          selectedAsset.contractAddress &&
+                          selectedAsset.decimal
+                            ? buildEIP681TokenRequest(
+                                requestAddress,
+                                selectedAsset.contractAddress,
+                                network.chainId,
+                                amount,
+                                selectedAsset.decimal
+                              )
+                            : buildEIP681EtherRequest(requestAddress, chainId, amount)
```
vilsbole: Component doesn't need to be aware of this logic
Create a method in `StoreProvider` that takes `account, asset, amount` and returns a `txObject`.
That would be the best place to handle the business logic aka. knowing if it's a base asset or a token

common/v2/features/ReceiveAssets/ReceiveAssets.tsx
```diff
@@ -0,0 +1,255 @@
...
```
vilsbole: _idem_
And make sure to call the method only once.

common/v2/features/ReceiveAssets/components/AccountDropdown.tsx
```diff
@@ -0,0 +1,59 @@
...
+// Since Account summary is using Address which still has the 'copy', we must handle hover ourself.
+class AccountOption extends React.PureComponent<OptionComponentProps> {
+  public render() {
+    const { option, onSelect } = this.props;
+
+    return (
+      <>
+        <AccountSummary
+          address={option.address}
+          balance={'1000.809300'}
```
vilsbole: use account.balance

common/v2/features/ReceiveAssets/routes.ts
```diff
@@ -0,0 +1,10 @@
+import ReceiveAssets from './ReceiveAssets';
+
+export default [
+  {
+    name: 'Receive Assets',
+    path: '/receive',
```
vilsbole: paths are now centralised and defined under `v2/features/routes.ts`

common/v2/features/ReceiveAssets/components/QRCode.tsx
```diff
@@ -0,0 +1,71 @@
...
+
+interface Props {
+  data: string;
+}
+
+interface State {
+  qr?: string;
+}
+
+export default class QRCode extends React.PureComponent<Props, State> {
```
vilsbole: I suspect this is copied from the previous version.
Since this is useful for the rest of the app, please move to `v2/components`
Ideally we would refactor to remove the UNSAFE lifecycle hooks.

common/v2/features/ReceiveAssets/types.ts
```diff
@@ -0,0 +1,38 @@
...
+  gasPriceField: string; // Use only if advanced tab is open AND user has input gas price
+  nonceField: string; // Use only if user has input a manual nonce value.
+  isAdvancedTransaction: boolean; // Used to indicate whether transaction fee slider should be displayed and if Advanced Tab fields should be displayed.
+  isGasLimitManual: boolean; // Used to indicate that user has un-clicked the user-input gas-limit checkbox.
+  accountType: WalletName | undefined; // Type of wallet selected.
+  network: Network | undefined;
+  gasEstimates: GasEstimates;
+}
+
+export interface ISendState {
```
vilsbole: unused. please remove

common/v2/features/ReceiveAssets/components/AccountDropdown.tsx
```diff
@@ -0,0 +1,59 @@
...
+      )}
+    />
+  );
+}
+
+interface IAccountDropdown {
+  accounts: IExtendedAccount[];
+  name: string;
+  value: IExtendedAccount;
+  values?: ITxFields;
```
vilsbole: This is the wrong type for this form.

common/v2/features/ReceiveAssets/components/AccountDropdown.tsx
```diff
@@ -0,0 +1,59 @@
...
+
+import { translateRaw } from 'translations';
+import { ExtendedAccount as IExtendedAccount } from 'v2/services';
+import { AccountSummary, Divider, Dropdown } from 'v2/components';
+import { ITxFields } from '../types';
+
+// Option item displayed in Dropdown menu. Props are passed by react-select Select.
+// To know: Select needs to receive a class in order to attach refs https://github.com/JedWatson/react-select/issues/2459
+// Since Account summary is using Address which still has the 'copy', we must handle hover ourself.
+class AccountOption extends React.PureComponent<OptionComponentProps> {
```
vilsbole: Let's move this to `v2/components` instead of redefining it.

common/v2/features/ReceiveAssets/types.ts
```diff
@@ -0,0 +1,38 @@
+import { IAsset } from 'v2/types';
+import { Asset, assetMethod } from 'v2/services/Asset/types';
+import { ExtendedAccount as IExtendedAccount, Network } from 'v2/services';
+import { WalletName } from 'v2/config/data';
+import { GasEstimates } from 'v2/api/gas';
+
+export interface ITxFields {
```
vilsbole: This is wrong type for the receive form

common/v2/libs/formatters.ts
```diff
@@ -1,7 +1,27 @@
...
+
 export function stripHexPrefix(value: string) {
   return value.replace('0x', '');
 }
 
 export function stripHexPrefixAndLower(value: string): string {
   return stripHexPrefix(value).toLowerCase();
 }
+
+export const buildEIP681EtherRequest = (
```
vilsbole: move to `v2/services/EthService/utils`

common/v2/libs/validators.ts
```diff
@@ -134,6 +134,7 @@ export function isValidEncryptedPrivKey(privkey: string): boolean {
 
 export const validNumber = (num: number) => isFinite(num) && num >= 0;
 export const validPositiveNumber = (num: number) => validNumber(num) && num !== 0;
+export const numberIsNotNegative = (num: number) => validNumber(num) && Math.sign(num) !== -1;
```
vilsbole: Amounts need to be Big Numbers https://github.com/indutny/bn.js/

common/v2/features/ReceiveAssets/ReceiveAssets.tsx
```diff
@@ -0,0 +1,255 @@
...
+import { translateRaw } from 'translations';
+
+import QRCode from './components/QRCode';
+import AccountDropdown from './components/AccountDropdown';
+// Legacy
+import receiveIcon from 'common/assets/images/icn-receive.svg';
+
+const truncate = (children: string) => {
+  return [children.substring(0, 15), '…', children.substring(children.length - 10)].join('');
+};
```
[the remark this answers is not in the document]
Sonicrida: I don't see a truncate method in there. Want me to add it?

common/v2/features/ReceiveAssets/ReceiveAssets.tsx
```diff
@@ -0,0 +1,255 @@
...
+      if (decimal && !isValidAmount(decimal)(amount)) {
+        error = 'Please enter a valid amount';
+      }
+    }
+
+    return error;
+  };
+
+  return (
+    <ReceivePanel heading="Receive Assets" icon={receiveIcon} onBack={history.goBack}>
```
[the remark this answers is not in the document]
Sonicrida: Oh good point. I think it should go to either `/` or perhaps the Dashboard?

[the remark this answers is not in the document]
Sonicrida: I'll set this to go to `/` for now and we can change it later if necessary

common/v2/features/ReceiveAssets/components/QRCode.tsx
Sonicrida: It is copied from the previous app but I don't believe that we have any uses elsewhere for QRCodes at the moment. I think that this should be moved to a shared directory if we know it'll be used elsewhere.

common/v2/libs/formatters.ts
Sonicrida: That directory doesn't exist in this PR yet

common/v2/features/ReceiveAssets/ReceiveAssets.tsx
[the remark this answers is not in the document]
vilsbole: yes please

common/v2/features/ReceiveAssets/ReceiveAssets.tsx
Sonicrida: Yeah, I meant to default the accounts dropdown to the first selected. I changed the logic to reflect this.

common/v2/features/ReceiveAssets/ReceiveAssets.tsx
Sonicrida: How should I handle using the asset outside of the form for validation (also curious since the yup schema would need this and exists outside of the form as well if I use yup)

https://github.com/MyCryptoHQ/MyCrypto/blob/5cff9078237245171ad89c8dbcaf6fda12d5b538/common/v2/features/ReceiveAssets/ReceiveAssets.tsx#L118

common/v2/features/ReceiveAssets/ReceiveAssets.tsx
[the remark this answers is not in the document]
vilsbole: Yes that's good. Then we can let '/' handle if we go to 'Dashboard' or not

common/v2/features/ReceiveAssets/components/QRCode.tsx
vilsbole: Since it has the potential to be shared, I would prefer it under `components`.
Especially because the developer that would code the new feature may not be aware of 'ReceiveAssets' when he starts so his reflex would be to code a new one.
I think that by encouraging reusable components it creates better code (pure components with a better SOC) and the dev will be less inclined to take shortcuts and mingle everything together.

Sonicrida: Oh that makes sense! I'll do that then!

common/v2/features/ReceiveAssets/ReceiveAssets.tsx
Sonicrida: Looking at the formik examples, they don't seem to use yup unless performing validation for the entire form. I think it might be better to not use yup if validating only a single field?

common/v2/features/ReceiveAssets/components/QRCode.tsx
amacar: We actually use QRCode from old `components/ui` in v2 PaperWallet (https://github.com/MyCryptoHQ/MyCrypto/blob/gau/common/v2/components/PaperWallet.tsx#L7). Please update imports and check if everything still works (it should since it is an exact copy).

vilsbole:
### TODO

#### styles
- [ ]  fix hover on payment code. (where is the person suppose to paste the payment code?)
- [ ] border effect on asset input. (there is a black border, and the dropdown should have the same surrounding as the input area)
- [ ] instead of displaying only the 4 assets. increase size of AssetOption list it should be taller and display more.
- [ ] place amount Error under the amount field
- [ ] increase Asset field height to be the same as the Amount field.

#### logic
- [ ] Form should be invalid with an empty amount field
- [ ] Form should be valid with an amount containing a decimal value with a leading '0' or a leading '.'

common/v2/features/ReceiveAssets/ReceiveAssets.tsx
```diff
@@ -0,0 +1,284 @@
...
+        .filter(asset => asset.type === 'base' || asset.type === 'erc20')
+    : [];
+  const assetOptions = filteredAssets.map(asset => ({ label: asset.name, id: asset.uuid }));
+  const [chosenAssetName, setAssetName] = useState(assetOptions[0].label);
+  const selectedAsset = filteredAssets.find(asset => asset.name === chosenAssetName);
+
+  const initialValues = {
+    amount: '0',
+    asset: { label: 'Ethereum', id: '7bbf42b1-9275-120b-0d0a-a788abd75ea0' },
+    chainId: network ? network.chainId : 1,
```
vilsbole: chainId is not directly edited by the form. to simplify prefer using the components state with network.chainId when you need it.

common/v2/features/ReceiveAssets/ReceiveAssets.tsx
```diff
@@ -0,0 +1,284 @@
...
+        .filter(asset => asset.networkId === network.id)
+        .filter(asset => asset.type === 'base' || asset.type === 'erc20')
+    : [];
+  const assetOptions = filteredAssets.map(asset => ({ label: asset.name, id: asset.uuid }));
+  const [chosenAssetName, setAssetName] = useState(assetOptions[0].label);
+  const selectedAsset = filteredAssets.find(asset => asset.name === chosenAssetName);
+
+  const initialValues = {
+    amount: '0',
+    asset: { label: 'Ethereum', id: '7bbf42b1-9275-120b-0d0a-a788abd75ea0' },
```
vilsbole: uuid are dynamically generated at run time so you can't rely on it.
better to use a find method that selects the Ethereum asset.

PR #2639, opened 2019-06-18
ENS Resolution for To Address field.

blurpesec:
## Description

Adds ENS resolution for to-address field.

## Changes

* Extended the ethers.js class provider with ens interactions.
* Added ens resolution.
* Edited the processTx function for handling ens.

## Quality Assurance


common/v2/features/SendAssets/components/SendAssetsForm.tsx
```diff
@@ -98,18 +104,46 @@ export default function SendAssetsForm({
...
 
+          const handleFieldReset = () => {
+            setFieldValue('account', undefined);
+            setFieldValue('recipientAddress', '');
+            setFieldValue('amount', '');
+          };
+
           const setAmountFieldToAssetMax = () =>
             // @TODO get asset balance and subtract gas cost
             setFieldValue('amount', '1000');
```
SharonManrique: These functions could probably be handled by their appropriate custom component. When the component receives a value, onChange or onBlur trigger the function. 

Formik fields can handle their own state, so that we can keep FormData state as minimal as possible, and only use setState/update state when we need to pass up the transaction data. 

https://jaredpalmer.com/formik/docs/api/field#component

common/v2/features/SendAssets/SendAssets.tsx
```diff
@@ -67,15 +67,14 @@ const getInitialState = (): ISendState => {
           time: Date.now(),
           chainId: 1
         },
-        network: undefined
+        network: undefined,
+        resolvedNSAddress: '', // Address returned when attempting to resolve an ENS/RNS address.
+        isResolvingNSName: false // Used to indicate recipient-address is ENS name that is currently attempting to be resolved.
```
SharonManrique: This state should be handled seperately by the ENS name component

common/v2/features/SendAssets/SendAssets.tsx
```diff
@@ -118,15 +117,14 @@ const getInitialState = (): ISendState => {
...
       },
       isFetchingAccountValue: false, // Used to indicate looking up user's balance of currently-selected asset.
-      isResolvingNSName: false, // Used to indicate recipient-address is ENS name that is currently attempting to be resolved.
       isAddressLabelValid: false, // Used to indicate if recipient-address is found in the address book.
       isFetchingAssetPricing: false, // Used to indicate fetching CC rates for currently-selected asset.
       isEstimatingGasLimit: false, // Used to indicate that gas limit is being estimated using `eth_estimateGas` jsonrpc call.
```
SharonManrique:
```
isFetchingAccountValue,
 isAddressLabelValid,
isFetchingAssetPricing
 isEstimatingGasLimit:
```

These states should be handled by their custom component, then only setting state when they have actually validated and fetched the correct value. So that The only state we worry about in FormData is `AccountValue`, `AddressLabel`, `AssetPricing`, and `GasLimit`

PR #2650, opened 2019-06-25
re-enabled base assets in send-assets dropdown

blurpesec:
Fixes an outstanding issue with showing base-assets in send-assets dropdown.

This *does not* fix the bug where there is a red screen of death on asset selection. This should be resolved by #2639 unless i'm mistaken.

### DIFF
--- a/common/Root.tsx
+++ b/common/Root.tsx
@@ -33,7 +33,13 @@ import { NewAppReleaseModal } from 'v2/components';
 import Dashboard from 'v2/features/Dashboard';
 import DevTools from 'v2/features/DevTools';
 import PrivateRoute from 'v2/features/NoAccounts/NoAccountAuth';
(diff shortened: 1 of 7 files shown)


LuckyLuker:
https://app.clubhouse.io/mycrypto/story/2179/re-enable-base-assets-in-send-assets-dropdown

CryptoQueen628:
A few things that need to be changed:
1. Add a button for "Save"
2. The QR code should display only after all fields in the form are populated & a user clicks "Save"
3. Remove "Payment Code." Because we haven't deep linked, the string is irrelevant. 
4. The validation on amount doesn't want to accept any value that is less than 1 e.g, 0.01
5. Copy to clipboard functionality is wanky - it only works when you click directly on 'ethereum.' It spazzes out if you click or hover anywhere else & copy doesn't work. I've noticed this in other flows as well - so let's identify the component & change it globally

blurpesec:
Closed because needed significant rewrite after state rework.

PR #2657, opened 2019-07-02
ev/exchange rate polling

vilsbole:
[ch1624]

## Fetch Rates
1. Create `polling.worker` that takes an URL and Interval
2. Create `poling.service` to control worker
3. Consume `polling.service` with RateProvider
4. Use `idempotent-babel-polyfill` to avoid duplicate import

## CleanUp
1. Move providers from Root to AppProvider
2. Introduce StoreProvider to create dynamic state values
3. Clean Root component imports


Yairimedinac4621:
blockchair.com @twitter_yairi.medinac@gmail.com

PR #2659, opened 2019-07-05
ev/harmonize layout

vilsbole:
[ch1916]
- Use a single Layout inside the Switch (through PrivateRoutes). `Home` is the exception.
- Remove all Layouts from v2 components
- Flex Layout-content for full screen height
- Define all v2 routes in a single file `v2/features/routes.tsx`

---
#### Other
- include component names in the className hash generated by styled-components.


Sonicrida:
It is SUPER refreshing to see the Layout component not being a concern of the individual flows anymore. This is great work (also good consistency with interface naming updates as well for more clarity.

Consolidated routes will also make the experience of adding new features a lot better to document and manage.

I'm also a fan of the `react-responsive` usage. Good work!

vilsbole:
Thanks! It's nice to hear ;)

PR #2664, opened 2019-07-09
ev/style dashboard

vilsbole:
- correct top alignment of Dashboard columns
- correct border-radius
- Harmonise breakpoints for Dashboard 
- Better Responsive display for ActionTiles

Since we are missing designs for Tablet sizes, styles between desktop and mobile are on a best effort basis.

Screenshot 2019-07-09 at 14 06 57

Screenshot 2019-07-09 at 14 07 16

---

- clean up `features` folder by moving code that should be in `components`, `utils` and `services`


Sonicrida:
Thanks for the default route fix! This looks good!

Sonicrida:
Just to make sure this is working right, is the `Home` route supposed to be inaccessible once wallets are added? The logo currently takes me to the dashboard instead of the homepage.

vilsbole:
It's a feature made by Povio.
In `v2/routing/helpers` when going to `/` we decide wether to route to home or dashboard depending on the presence of accounts in localStorage.
It's easy enough to change in case we want another behaviour.

PR #2675, opened 2019-07-16
Onboarding notifications

PeterZlodej:
[ch1815]
[ch1817]
[ch1312]


common/v2/providers/NotificationsProvider/helpers.ts
```diff
@@ -1,3 +1,6 @@
+import { ExtendedNotification } from 'v2/services';
+import moment from 'moment';
```
vilsbole: What do you think of moving to `date-fns` instead?
Smaller bundle size + tree-shaking.
 
Screenshot 2019-07-18 at 15 02 17

amacar: Never heard of it but looks good. Since `moment` is used not only in this PR but also in some reducers that we will get rid of it, I think it would be better to do the transition in a new branch.

vilsbole: Ok.

PR #2676, opened 2019-07-17
Fix ComboBox Autocompletion

Sonicrida:
## Description
Implements `react-select` for the receive flow assets dropdown with closer styles to other inputs


Sonicrida:
[ch1963]

PR #2692, opened 2019-07-29
Re-Added base assets on send flow assets-dropdown

blurpesec:
Fixes an outstanding issue with showing base-assets in send-assets dropdown.

*Note: There appears to be something in the Analytics service tests that are broken. I haven't touched these at all in this PR.*

### DIFF
--- a/common/v2/features/SendAssets/components/SendAssetsForm.tsx
+++ b/common/v2/features/SendAssets/components/SendAssetsForm.tsx
@@ -7,7 +7,7 @@ import translate, { translateRaw } from 'translations';
 import { WhenQueryExists } from 'components/renderCbs';
 
 import { InlineErrorMsg, ENSStatus } from 'v2/components';
-import { AccountContext, getAssetByUUID } from 'v2/services/Store';
+import { AccountContext, getAssetByUUID, getBaseAssetFromAccount } from 'v2/services/Store';
 import {
   IAsset,
   Asset,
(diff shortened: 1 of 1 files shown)


vilsbole:
Left of few comments in the effort to harmonise our style and reduce the amount of cycles.
More importantly, we need this functionality multiple times throughout the app.

It would be a lot better to have something like:
```
// StoreProvider.tsx


// Since we are in a Provider we have access to all that we need.
export const getUserAssets = (): IAssets[] => {
 const tokens =  accounts.flatMap(account => account.assets)
 const baseAssets = accounts.map(account => getBaseAssetFromAccount)
 return {
   ...tokens,
   ...baseAssets
 } 
}
```

```
// SendFormAssets
import { getAccountAssets } from 'v2/services/Store'; 

const allAssets = getUserAssets(assets);
....
```

blurpesec:
> Left of few comments in the effort to harmonise our style and reduce the amount of cycles.
> More importantly, we need this functionality multiple times throughout the app.
> 
> It would be a lot better to have something like:
> 
> ```
> // StoreProvider.tsx
> 
>   state = {
>      ...,
>      userAssets: getUserAssets()
>    }
> 
>   // Since we are in a Provider we have access to all that we need.
>   const getUserAssets = (): IAssets[] => {
>     const tokens: IAssets[] =  accounts.flatMap(account => account.assets)
>     const baseAssets: IAssets[] = accounts.map(account => getBaseAssetFromAccount)
>     return {
>        ...tokens,
>        ...baseAssets
>      } 
>   }
> ```
> 
> ```
> // SendFormAssets
> import { StoreContext } from 'v2/services/Store'; 
> 
> //....
> const { userAssets: allAssets } = useContext(StoreContext);
> ....
> ```

@vilsbole I'll add this in #2695 c:

PR #2694, opened 2019-07-30
Fix error rendering when importing invalid file type that is not a json file

Sonicrida:
## Description
Fix issue where an error was displaying by default on the import page before an import was attempted

## Changes
Stops error from showing before import


Sonicrida:
[ch1827]

PR #2699, opened 2019-08-06
Fix settings import file type detection error

Sonicrida:
## Description

Fixed an error where an error message wouldn't display when a user tried to upload an incorrect filetype (like an image) as a settings file.

[ch2487]


Sonicrida:
Thanks!

PR #2700, opened 2019-08-06
Accounts table updates

Sonicrida:
## Description
Updates to the accounts table to match the designs

## Changes

* Add optional deletion column with the `deletable` prop
* added scrolling on the table body
* labels and addresses are in separate columns
* favorite functionality has been added to accounts

[ch215]

### DIFF
--- a/common/v2/components/AccountList.tsx
+++ b/common/v2/components/AccountList.tsx
@@ -1,7 +1,7 @@
 import React, { useContext } from 'react';
 import { Redirect } from 'react-router-dom';
 import styled from 'styled-components';
-import { Address, Button, CollapsibleTable, Icon, Network, Typography } from '@mycrypto/ui';
+import { Button, CollapsibleTable, Copyable, Network, Typography, Identicon } from '@mycrypto/ui';
 
 import { translateRaw } from 'translations';
 import { truncate } from 'v2/utils';
@@ -15,21 +15,57 @@ import {
 import { DashboardPanel } from './DashboardPanel';
 import './AccountList.scss';
 
+const Label = styled.span`
+  display: flex;
+  align-items: center;
+  p {
+    margin-right: 27px;
+  }
+`;
+
+interface iFavoriteProps {
+  favorited: boolean;
+}
+
+const FavoriteButton = styled(Button)`
+  span {
+    span {
+      svg {
+        path {
+          fill: ${(props: iFavoriteProps) => (props.favorited ? 'rgb(255, 209, 102)' : 'white')};
+          stroke: ${(props: iFavoriteProps) =>
+            props.favorited ? 'rgb(255, 209, 102)' : '#7b8695'};
+        }
+      }
+    }
+  }
+  align-self: flex-start;
+  margin-left: 1em;
+`;
+
 const DeleteButton = styled(Button)`
   align-self: flex-start;
   margin-left: 1em;
 `;
 
+const TableContainer = styled.div`
+  display: block;
+  max-height: 394px;
+  overflow: auto;
+`;
+
 type DeleteAccount = (uuid: string) => void;
+type UpdateAccount = (uuid: string, accountData: ExtendedAccount) => void;
 interface AccountListProps {
   className?: string;
   currentsOnly?: boolean;
+  deletable?: boolean;
 }
 
 export default function AccountList(props: AccountListProps) {
-  const { className, currentsOnly } = props;
+  const { className, currentsOnly, deletable } = props;
   const { settings } = useContext(SettingsContext);
-  const { accounts, deleteAccount } = useContext(AccountContext);
+  const { accounts, deleteAccount, updateAccount } = useContext(AccountContext);
   const currentAccounts: ExtendedAccount[] = getCurrentsFromContext(
     accounts,
     settings.dashboardAccounts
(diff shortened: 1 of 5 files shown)


common/v2/components/AccountList.tsx
```diff
@@ -46,45 +82,98 @@ export default function AccountList(props: AccountListProps) {
       actionLink="/add-account"
```
vilsbole: minor change request to harmonise routes: 
Use ROUTE_PATHS from v2/config

common/v2/components/AccountList.tsx
```diff
@@ -46,45 +82,98 @@ export default function AccountList(props: AccountListProps) {
...
       className={`AccountList ${className}`}
     >
-      <CollapsibleTable
-        breakpoint={450}
-        {...buildAccountTable(currentsOnly ? currentAccounts : accounts, deleteAccount)}
-      />
+      <TableContainer>
+        <CollapsibleTable
+          breakpoint={450}
```
vilsbole: Please use BREAK_POINTS instead

common/v2/components/AccountList.tsx
```diff
@@ -46,45 +82,98 @@ export default function AccountList(props: AccountListProps) {
...
+          translateRaw('ACCOUNT_LIST_VALUE'),
+          translateRaw('ACCOUNT_LIST_DELETE')
+        ]
+      : [
+          translateRaw('ACCOUNT_LIST_FAVOURITE'),
+          translateRaw('ACCOUNT_LIST_LABEL'),
+          translateRaw('ACCOUNT_LIST_ADDRESS'),
+          translateRaw('ACCOUNT_LIST_NETWORK'),
+          translateRaw('ACCOUNT_LIST_VALUE')
+        ],
```
vilsbole: how about we declare the head once, then add the extra label?
```
const columns = [ 
    translateRaw('ACCOUNT_LIST_FAVOURITE'),
    translateRaw('ACCOUNT_LIST_LABEL'),
    translateRaw('ACCOUNT_LIST_ADDRESS'),
    translateRaw('ACCOUNT_LIST_NETWORK'),
    translateRaw('ACCOUNT_LIST_VALUE'),
]

return {
   head: deletable ? [...columns, translateRaw('ACCOUNT_LIST_DELETE')] : columns,
    ....
}

```

common/v2/components/AccountList.tsx
```diff
@@ -46,45 +82,98 @@ export default function AccountList(props: AccountListProps) {
...
-          {account.network}
-        </Network>,
-        <Typography key={index + bodyItemCount++}>{account.balance}</Typography>,
-        <DeleteButton
-          key={index + bodyItemCount++}
-          onClick={handleAccountDelete(deleteAccount, account.uuid)}
-          icon="exit"
-        />
-      ];
+      return deletable
```
vilsbole: idem _declare once_

common/v2/components/AccountList.tsx
```diff
@@ -46,45 +82,98 @@ export default function AccountList(props: AccountListProps) {
...
+        ? [
+            <FavoriteButton
+              key={index + bodyItemCount++}
```
vilsbole: declare `key` once

common/v2/types/account.tsx
```diff
@@ -12,6 +12,7 @@ export interface Account {
   transactions: TransactionData[];
   dPath: string;
   timestamp: number;
+  favorite?: boolean;
```
vilsbole: The property should always exist and default to false

common/v2/components/AccountList.tsx
```diff
@@ -15,21 +15,57 @@ import {
...
+  max-height: 394px;
+  overflow: auto;
+`;
+
 type DeleteAccount = (uuid: string) => void;
+type UpdateAccount = (uuid: string, accountData: ExtendedAccount) => void;
 interface AccountListProps {
   className?: string;
   currentsOnly?: boolean;
+  deletable?: boolean;
```
vilsbole: In which case do we choose if we want to delete an account or not?

tayvano: I believe it depends on where the account list is being displayed. e.g. we could only make accounts deletable from the settings page, not the dashboard, or have a toggle somewhere that reveals a delete button to keep the UI clean.

vilsbole: Ah, ok!
Settings vs Dashboard seems like a simple option to start with.

Sonicrida: Yep. I already set `deletable` to true on the Settings page

common/v2/components/AccountList.tsx
Sonicrida: is the key being declared more than once?

vilsbole: Sorry for the delay to reply: 
`key={index + bodyItemCount++}` is set on every component in the `body`.
Though it would be better to have a 'row' component, let's avoid inline in incrementation and define a single key per row, something like `const rowIndex = index +...    key={rowIndex}`
I haven't tested it but I suspect `bodyItemCount++` to be entirely superfluous and that the array index is enough, so we can just do `key={index}`.

Sonicrida: Ahh that makes sense. I looked into it and `bodyItemCount` wasn't needed so I removed it and used index as suggested.

common/v2/components/AccountList.tsx
```diff
@@ -43,48 +80,76 @@ export default function AccountList(props: AccountListProps) {
...
-          onClick={handleAccountDelete(deleteAccount, account.uuid)}
-          icon="exit"
-        />
+        <Typography key={index}>{account.balance}</Typography>
       ];
+      return deletable
+        ? [
+            ...bodyContent,
+            <DeleteButton
+              key={index + bodyItemCount++}
```
vilsbole: I believe you missed this one.

common/v2/components/AccountList.tsx
```diff
@@ -43,48 +80,76 @@ export default function AccountList(props: AccountListProps) {
...
-      translateRaw('ACCOUNT_LIST_ADDRESS'),
-      translateRaw('ACCOUNT_LIST_NETWORK'),
-      translateRaw('ACCOUNT_LIST_VALUE'),
-      translateRaw('ACCOUNT_LIST_DELETE')
-    ],
+    head: deletable ? [...columns, translateRaw('ACCOUNT_LIST_DELETE')] : columns,
     body: accounts.map((account, index) => {
       const addressCard: AddressBook | undefined = getLabelByAccount(account);
       const label = addressCard ? addressCard.label : 'Unknown Account';
       let bodyItemCount = 0;
```
vilsbole: We no longer need this variable.
Please remove

PR #2701, opened 2019-08-06
Update copy

Sonicrida:
## Description
Added various copy updates and removed some hardcoded strings

## Changes

* Added a tooltip on the settings page
* removed hardcoded strings
* Switched to using "MyCrypto Settings" on the settings page

[ch2482]


common/v2/features/Settings/components/GeneralSettings.tsx
```diff
@@ -94,31 +102,48 @@ export default class GeneralSettings extends React.Component<SettingsProps> {
     }
   };
 
+  public renderTooltipContent = () => <div>Test</div>;
```
vilsbole: This seems unused.

common/v2/features/Settings/components/GeneralSettings.tsx
```diff
@@ -94,31 +102,48 @@ export default class GeneralSettings extends React.Component<SettingsProps> {
...
+
   public render() {
     const { globalSettings } = this.props;
     return (
-      <DashboardPanel heading="General Settings">
+      <DashboardPanel heading={translateRaw('SETTINGS_GENERAL_LABEL')}>
```
vilsbole: Any reason why you're using `translateRaw` instead of `translate`?
Applies to the entire component

common/v2/features/Settings/components/GeneralSettings.tsx
```diff
@@ -94,31 +102,48 @@ export default class GeneralSettings extends React.Component<SettingsProps> {
...
-          <SettingsLabel>Account Settings</SettingsLabel>
+          <SettingsLabel>
+            {translateRaw('SETTINGS_HANDLING_LABEL')}
+            <Tooltip tooltip={<span>{translateRaw('SETTINGS_TOOLTIP')}</span>}>
+              {props => (
+                <div {...props}>
+                  <SettingsTooltipIcon icon="shape" />
+                </div>
+              )}
+            </Tooltip>
```
vilsbole: 1. Tooltip hover is still jumpy. I was under the impression that Povio had corrected this.

common/v2/features/Settings/components/GeneralSettings.tsx
```diff
@@ -94,31 +102,48 @@ export default class GeneralSettings extends React.Component<SettingsProps> {
...
     return (
-      <DashboardPanel heading="General Settings">
+      <DashboardPanel heading={translateRaw('SETTINGS_GENERAL_LABEL')}>
         <Divider />
         <SettingsField>
-          <SettingsLabel>Account Settings</SettingsLabel>
+          <SettingsLabel>
+            {translateRaw('SETTINGS_HANDLING_LABEL')}
+            <Tooltip tooltip={<span>{translateRaw('SETTINGS_TOOLTIP')}</span>}>
+              {props => (
```
vilsbole: This wrapper FC seems unnecessary. If so, please remove

common/v2/features/Settings/components/GeneralSettings.tsx
```diff
@@ -94,31 +102,48 @@ export default class GeneralSettings extends React.Component<SettingsProps> {
...
+            </Tooltip>
+          </SettingsLabel>
           <SettingsControl>
             <Link to="/settings/import">
-              <SettingsButton secondary={true}>Import</SettingsButton>
+              <SettingsButton secondary={true}>
+                {translateRaw('SETTINGS_IMPORT_LABEL')}
+              </SettingsButton>
             </Link>
             <Link to="/settings/export">
```
vilsbole: Since we are touching this file, let's update the link as well to use the centralized router like in `v2/features/Layout/Header/constants`
This will help us avoid chasing broking links in the future.

common/v2/features/Settings/components/GeneralSettings.tsx
```diff
@@ -94,31 +102,48 @@ export default class GeneralSettings extends React.Component<SettingsProps> {
...
+            <Tooltip tooltip={<span>{translateRaw('SETTINGS_TOOLTIP')}</span>}>
+              {props => (
+                <div {...props}>
+                  <SettingsTooltipIcon icon="shape" />
+                </div>
+              )}
+            </Tooltip>
+          </SettingsLabel>
           <SettingsControl>
             <Link to="/settings/import">
```
vilsbole: Since we are touching this file, let's update the link as well to use the centralized router like in v2/features/Layout/Header/constants
This will help us avoid chasing broking links in the future.

common/v2/features/Settings/components/GeneralSettings.tsx
Sonicrida: They did but it's a seperate change from the ui library. I believe we just need to publish a new version. It's separate from core though.

common/v2/features/Settings/components/GeneralSettings.tsx
Sonicrida: I couldn't get it to work without it for some reason and saw the tooltip implemented by someone else in the same way elsewhere in v2. I tried to play around a little with it but didn't want to spend too much time trying to figure out why it behaves like that if it worked :P

common/v2/features/Settings/components/GeneralSettings.tsx
vilsbole: Then can we please publish the new version and update the package in core?

common/v2/features/Settings/components/GeneralSettings.tsx
vilsbole: ok, let's see if updating the package solves the issue.
The styling of the tooltip (size of font and container, + positioning of the icon which needs to be aligned with the copy) also needs to be fixed.

common/v2/features/Settings/components/GeneralSettings.tsx
vilsbole: I think you missed this line. Please change to `translate`

Sonicrida: Oh. Didn't miss, just thought that `heading` could only be a string. Looks like it can be `any` though so I will change it!

common/v2/features/Settings/components/GeneralSettings.tsx
Sonicrida: I'll get that published

PR #2702, opened 2019-08-06
Force production build to use dev environment css loaders temporarily

blurpesec:
We're going to alter the production build to use the dev-builds css loaders temporarily. 
This is to allow design team to review current state of `gau` without forcing us to figure out a solution to this bug yet since it's pretty low-priority until right before MVP release.

We're going to create a task that blocks GAU MVP release in order to fix this before MVP goes live.

This new task is here: https://app.clubhouse.io/mycrypto/story/2535/revert-https-github-com-mycryptohq-mycrypto-pull-2702-and-fix-dev-and-production-build-style-mismatch-bug

This pr closes this task: https://app.clubhouse.io/mycrypto/story/2531/create-way-for-avani-to-be-able-to-review-design-implementation

### DIFF
--- a/webpack_config/makeConfig.js
+++ b/webpack_config/makeConfig.js
@@ -79,11 +79,18 @@ module.exports = function(opts = {}) {
     rules.push(
       {
         test: /\.css$/,
-        use: [MiniCSSExtractPlugin.loader, 'css-loader']
+        include: [
+          path.resolve(config.path.src, 'vendor'),
+          path.resolve(__dirname, '../node_modules/typeface-lato')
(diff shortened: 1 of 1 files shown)


tayvano:
Can someone open another PR that keeps the MinjCss for production but adds the `inckude` parameters from develop?

It may be that there are additional styles somewhere in the folders that are being excluded in dev (via the `include` conditional 😅) but not in production.

E.g. there is scss in a `v3` folder and since be isn't specified by `jnckude`, it's not being included in development but is in production.

PR #2704, opened 2019-08-12
Move broadcast transaction to v2

PeterZlodej:
[ch2491]


common/v2/components/InputField.tsx
```diff
@@ -114,17 +134,21 @@ export class InputField extends Component<Props> {
             <CustomTextArea
               value={value}
               onChange={onChange}
+              onBlur={onBlur}
               inputError={inputError}
               onKeyUp={this.handleKeyUp}
+              placeholder={placeholder ? placeholder : ''}
```
vilsbole: Since his ternary is repeated, how about:
```
const {
  placeholder = ''
} = this.props

... placeholder={placeholder} >
```

common/v2/features/BroadcastTx/BroadcastTx.tsx
```diff
@@ -0,0 +1,161 @@
...
+  public render() {
+    const { userInput } = this.state;
+    const stateTransaction = userInput ? this.props.stateTransaction : '';
+
+    return (
+      <ExtendedContentPanel
+        heading={translateRaw('BROADCAST_TX_TITLE')}
+        description={translate('BROADCAST_TX_DESCRIPTION')}
+        centered={true}
+        maxWidth="650px"
```
vilsbole: Can we add this value in a shared const in `v2/config/constants` so we can start to apply all our panels?

common/v2/components/InputField.tsx
PeterZlodej: Makes sense, fixed.

common/v2/features/BroadcastTx/BroadcastTx.tsx
PeterZlodej: Moved to constants, unfortunately we can't just set this for all panels, since the designs are not consistent in this aspect.

PeterZlodej:
> 1. What do you think of deleting the old component as well? From the moment it is no longer available in a route we should be able to delete the code. IMHO doing it now would reduce potential confusions in the future.
> 2. It's most probably my OCD playing out, but I feel uncomfortable bringing Redux connected components into the `v2` directory. I fear it can create a bit of confusion for future devs of what the convention should be. What would you think of creating a `v1/features` folder to place `BroadcastTx.tsx` into. That way we have explanation of the architecture differences and a list of components that are connected to the store?

1. Deleted the old component, agreed it makes sense to do that as we move the features over to the v2.
2. There are already some features tied to redux in the v2. We should probably identify all of them and move them to the new v1/features folder along with newly ported components that we are working on, to at least keep them in the same place.

vilsbole:
This PR was merged with https://github.com/MyCryptoHQ/MyCrypto/pull/2722
Closing

PR #2707, opened 2019-08-13
Network fixes for deterministic wallets UI

FrederikBolding:
🎉 🎉 🎉

## Description

Fixes issues with not using the previously selected network for deterministic wallets UI when adding an existing account. 

## Changes

Changed all components using `DeterministicWallets` in the `AddAccountFlow` to get and use network details from `NetworkSelectPanel` (previous step in flow).

## Steps to Test

1. Add a new account 
2. Select a network (fx. RSK)
3. Select mnemonic phrase and login (or any other type that uses `DeterministicWallets`)
4. The wallet select UI should show correct DPath and network symbol

## Quality Assurance

### DIFF
--- a/common/v2/features/AddAccount/components/DeterministicWallets.tsx
+++ b/common/v2/features/AddAccount/components/DeterministicWallets.tsx
@@ -21,6 +21,7 @@ import radio from 'assets/images/radio.svg';
 import radioChecked from 'assets/images/radio-checked.svg';
 import { getNetworkByDPath } from 'v2/services/Store/Network/helpers';
 import { Network } from 'v2/types';
+import { StaticNetworkConfig, CustomNetworkConfig } from 'shared/types/network';
 
 function Radio({ checked }: { checked: boolean }) {
   return <img className="clickable radio-image" src={checked ? radioChecked : radio} />;
@@ -96,7 +97,7 @@ class DeterministicWalletsClass extends React.PureComponent<Props, State> {
   public render() {
     const { wallets, dPaths, onCancel } = this.props;
     const { selectedAddress, customPath, page } = this.state;
-    const network = getNetworkByDPath(this.state.currentDPath);
+    const network = getNetworkByDPath(this.state.currentDPath) || this.props.network;
 
     return (
       <div className="DW">
@@ -233,13 +234,13 @@ class DeterministicWalletsClass extends React.PureComponent<Props, State> {
     );
   }
 
-  private renderWalletRow(wallet: deterministicWalletsTypes.DeterministicWalletData, network : Network) {
+  private renderWalletRow(wallet: deterministicWalletsTypes.DeterministicWalletData, network : Network | StaticNetworkConfig | CustomNetworkConfig) {
     const { addressLabels } = this.props;
     const { selectedAddress } = this.state;
     const label = addressLabels[wallet.address.toLowerCase()];
 
     let blockExplorer;
-    if (!network.isCustom) {
+    if (!network.isCustom && network.blockExplorer) {
       blockExplorer = network.blockExplorer;
     } else {
       blockExplorer = {
@@ -286,7 +287,8 @@ class DeterministicWalletsClass extends React.PureComponent<Props, State> {
 function mapStateToProps(state: AppState): StateProps {
   return {
     addressLabels: addressBookSelectors.getAddressLabels(state),
-    wallets: state.deterministicWallets.wallets
+    wallets: state.deterministicWallets.wallets,
+    network: configSelectors.getNetworkConfig(state)
   };
 }
 
--- a/common/v2/features/AddAccount/components/DeterministicWallets.tsx
+++ b/common/v2/features/AddAccount/components/DeterministicWallets.tsx
@@ -19,6 +19,8 @@ import nextIcon from 'assets/images/next-page-button.svg';
 import prevIcon from 'assets/images/previous-page-button.svg';
 import radio from 'assets/images/radio.svg';
 import radioChecked from 'assets/images/radio-checked.svg';
+import { getNetworkByDPath } from 'v2/services/Store/Network/helpers';
+import { Network } from 'v2/types';
 
 function Radio({ checked }: { checked: boolean }) {
   return <img className="clickable radio-image" src={checked ? radioChecked : radio} />;
@@ -92,8 +94,9 @@ class DeterministicWalletsClass extends React.PureComponent<Props, State> {
   }
 
   public render() {
-    const { wallets, network, dPaths, onCancel } = this.props;
+    const { wallets, dPaths, onCancel } = this.props;
     const { selectedAddress, customPath, page } = this.state;
+    const network = getNetworkByDPath(this.state.currentDPath);
 
     return (
       <div className="DW">
@@ -137,7 +140,7 @@ class DeterministicWalletsClass extends React.PureComponent<Props, State> {
 
         <Table
           head={['#', 'Address', network.unit, translateRaw('ACTION_5')]}
-          body={wallets.map(wallet => this.renderWalletRow(wallet))}
+          body={wallets.map(wallet => this.renderWalletRow(wallet, network))}
           config={{ hiddenHeadings: ['#', translateRaw('ACTION_5')] }}
         />
 
@@ -230,8 +233,8 @@ class DeterministicWalletsClass extends React.PureComponent<Props, State> {
     );
   }
 
-  private renderWalletRow(wallet: deterministicWalletsTypes.DeterministicWalletData) {
-    const { network, addressLabels } = this.props;
+  private renderWalletRow(wallet: deterministicWalletsTypes.DeterministicWalletData, network : Network) {
+    const { addressLabels } = this.props;
     const { selectedAddress } = this.state;
     const label = addressLabels[wallet.address.toLowerCase()];
 
@@ -283,8 +286,7 @@ class DeterministicWalletsClass extends React.PureComponent<Props, State> {
 function mapStateToProps(state: AppState): StateProps {
   return {
     addressLabels: addressBookSelectors.getAddressLabels(state),
-    wallets: state.deterministicWallets.wallets,
-    network: configSelectors.getNetworkConfig(state)
+    wallets: state.deterministicWallets.wallets
   };
 }
 
--- a/common/v2/features/AddAccount/components/DeterministicWallets.tsx
+++ b/common/v2/features/AddAccount/components/DeterministicWallets.tsx
@@ -28,7 +28,7 @@ function Radio({ checked }: { checked: boolean }) {
 const WALLETS_PER_PAGE = 5;
 
 interface OwnProps {
-  network: Network;
+  network: Network | undefined;
   dPath: DPath;
   dPaths: DPath[];
   publicKey?: string;
@@ -95,7 +95,11 @@ class DeterministicWalletsClass extends React.PureComponent<Props, State> {
   public render() {
     const { wallets, network, dPaths, onCancel } = this.props;
     const { selectedAddress, customPath, page } = this.state;
-    const symbol = getBaseAssetSymbolByNetwork(network);
+    let baseAssetSymbol: string | undefined;
+    if (network) {
+      baseAssetSymbol = getBaseAssetSymbolByNetwork(network);
+    }
+    const symbol : string = baseAssetSymbol ? baseAssetSymbol : 'ETH';
 
     return (
       <div className="DW">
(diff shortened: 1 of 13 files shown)


LuckyLuker:
[ch1766]

FrederikBolding:
Still need to fix some TypeScript warnings and make implement the few changes I made to `Mnemonic.tsx` to any other wallet type using `DeterministicWallets.tsx`.

But would love some feedback, just to make sure I'm not going in the entirely wrong direction 😄

common/v2/features/AddAccount/components/DeterministicWallets.tsx
```diff
@@ -19,6 +19,9 @@ import nextIcon from 'assets/images/next-page-button.svg';
 import prevIcon from 'assets/images/previous-page-button.svg';
 import radio from 'assets/images/radio.svg';
 import radioChecked from 'assets/images/radio-checked.svg';
+import { getNetworkByDPath } from 'v2/services/Store/Network/helpers';
```
vilsbole: **convention**: Avoid direct imports from files and import from the service instead `v2/services/Store`.
FYI: we're in the middle of rewrite and want to harmonise our services. As an intermediate step we are:
1. using the explicit exports from the `index.ts` files to create a list of the functions that are used.
2. importing only from the service level to decouple the usage from service implementation and simplify the future refactor.

common/v2/features/AddAccount/components/DeterministicWallets.tsx
```diff
@@ -19,6 +19,9 @@ import nextIcon from 'assets/images/next-page-button.svg';
...
+import { Network } from 'v2/types';
+import { StaticNetworkConfig, CustomNetworkConfig } from 'shared/types/network';
```
vilsbole: **convention:** Only import from `v2`
FYI: Please only import from the `v2` directory. 
- if something is missing you are welcome to add it into the v2 dir
- in this case v2 type is actually `Network`

common/v2/features/AddAccount/components/DeterministicWallets.tsx
```diff
@@ -92,8 +95,9 @@ class DeterministicWalletsClass extends React.PureComponent<Props, State> {
   }
 
   public render() {
-    const { wallets, network, dPaths, onCancel } = this.props;
+    const { wallets, dPaths, onCancel } = this.props;
     const { selectedAddress, customPath, page } = this.state;
+    const network = getNetworkByDPath(this.state.currentDPath) || this.props.network;
```
vilsbole: Dpaths are not unique to a Network, so I don't understand why you would want to identify the network from a dPath. Please ping me on slack to talk about this.

common/v2/features/AddAccount/components/DeterministicWallets.tsx
```diff
@@ -137,7 +141,7 @@ class DeterministicWalletsClass extends React.PureComponent<Props, State> {
 
         <Table
           head={['#', 'Address', network.unit, translateRaw('ACTION_5')]}
```
vilsbole: Using `Network`, `unit` is replaced by `baseAsset` which provides the `uuid` of the networks base asset.
Once we have the asset you should be able to use the `ticker` param.

common/v2/features/AddAccount/components/DeterministicWallets.tsx
```diff
@@ -230,13 +234,13 @@ class DeterministicWalletsClass extends React.PureComponent<Props, State> {
     );
   }
 
-  private renderWalletRow(wallet: deterministicWalletsTypes.DeterministicWalletData) {
-    const { network, addressLabels } = this.props;
+  private renderWalletRow(wallet: deterministicWalletsTypes.DeterministicWalletData, network : Network | StaticNetworkConfig | CustomNetworkConfig) {
+    const { addressLabels } = this.props;
```
vilsbole: In `v2` we want to only use `Network`

common/v2/features/AddAccount/components/Mnemonic.tsx
```diff
@@ -2,11 +2,13 @@ import React, { PureComponent } from 'react';
 import { connect } from 'react-redux';
 import { mnemonicToSeed, validateMnemonic } from 'bip39';
 
+import { FormData } from 'v2/features/AddAccount/types';
```
vilsbole: **convention** Only import at the service level.
If the type needs to shared move it into `v2/types`.
In this case, I don't see any no reason to share `FormData`.
(see below)

common/v2/features/AddAccount/components/Mnemonic.tsx
```diff
@@ -16,11 +18,11 @@ import questionToolTip from 'common/assets/images/icn-question.svg';
 import './Mnemonic.scss';
 
 interface OwnProps {
+  formData: FormData;
```
vilsbole: Prefer passing the desired property. i.e `network`

common/v2/services/Store/Network/helpers.ts
```diff
@@ -11,9 +11,14 @@ import {
   NodeOptions,
   WalletName
 } from 'v2/types';
+import { makeExplorer } from 'utils/helpers';
 
 export const getAllNetworks = () => {
-  return Object.values(getCache().networks);
+  const networks = Object.values(getCache().networks);
+  networks.forEach(
+    n => (n.blockExplorer = n.blockExplorer ? makeExplorer(n.blockExplorer) : n.blockExplorer)
```
vilsbole: 1. Avoid changing an objects attributes in place. Prefer returning a copy with object destructing or `Object.assign`
```
...
return networks.map(({ blockExplorer, ...rest }) => ({
  ...rest,
   blockExplorer: blockExplorer ? makeExplorer(blockExplorer) : blockExplorer 
}));
```

2. networks blockExplorer are already statically defined, currently in `config/networks/static/reducer.ts` (change in progress). Do we really need to run makeExplorer again?

common/v2/services/Store/Network/helpers.ts
```diff
@@ -11,9 +11,14 @@ import {
   NodeOptions,
   WalletName
 } from 'v2/types';
+import { makeExplorer } from 'utils/helpers';
 
 export const getAllNetworks = () => {
-  return Object.values(getCache().networks);
+  const networks = Object.values(getCache().networks);
```
vilsbole: No direct access to cache (We're fighting to remove them).
Use the networks in the state of the `Provider`.

common/v2/services/Store/Network/helpers.ts
```diff
@@ -38,6 +43,11 @@ export const getNetworkById = (id: string): Network | undefined => {
   return networks.find((network: Network) => network.id === id);
 };
 
+export const getNetworkByDPath = (dPath: DPath): Network | undefined => {
+  const networks = getAllNetworks() || [];
+  return networks.find((network: Network) => network.dPaths.default ? network.dPaths.default.value === dPath.value : false);
+};
```
vilsbole: This will return the first match. Since multiple networks share similar dPaths, this is close to a random network selection. Please explain.

common/v2/features/AddAccount/components/DeterministicWallets.tsx
```diff
@@ -230,13 +234,13 @@ class DeterministicWalletsClass extends React.PureComponent<Props, State> {
...
     const { selectedAddress } = this.state;
     const label = addressLabels[wallet.address.toLowerCase()];
 
     let blockExplorer;
-    if (!network.isCustom) {
+    if (!network.isCustom && network.blockExplorer) {
       blockExplorer = network.blockExplorer;
     } else {
       blockExplorer = {
```
vilsbole: Ouch!
This shouldn't be done inside the component.
Prefer updating `makeExplorer` with the default value `https://ethplorer.io/address/...`

common/v2/types/network.ts
```diff
@@ -45,6 +45,7 @@ export type NetworkId =
 export interface Network {
   id: NetworkId;
   name: string;
+  unit: string;
```
vilsbole: use `baseAsset` `ticker` over `unit`

FrederikBolding:
Thanks @vilsbole - I'll look through it all.

Most importantly with regards to the DPaths. I must have misunderstood. So, the DPath field has nothing to do with networks, but must simply be left alone, correct? The network is exclusively given earlier in the flow?

tayvano:
I can elaborate a bit on the relationship between dpath and network.

Each network has a "default" dpath. In theory, every network should have one dpath and it should be consistent across the ecosystem. Unfortunately, especially in Ethereum, this isn't the case with Ledger and Trezor using different dpaths and now Ledger using an entirely new scheme of dpath. 🙄

In this way, once the network is passed correctly, we can choose which dpath is used by default and the user can manually select another one, should they choose too. You can see this behavior on the live MyCrypto.com for further clarification.

FrederikBolding:
> I can elaborate a bit on the relationship between dpath and network.
> 
> Each network has a "default" dpath. In theory, every network should have one dpath and it should be consistent across the ecosystem. Unfortunately, especially in Ethereum, this isn't the case with Ledger and Trezor using different dpaths and now Ledger using an entirely new scheme of dpath. 🙄
> 
> In this way, once the network is passed correctly, we can choose which dpath is used by default and the user can manually select another one, should they choose too. You can see this behavior on the live MyCrypto.com for further clarification.

Yeah, I figured as much by now. Thanks for the clarification though! 😄

common/v2/features/AddAccount/components/Mnemonic.tsx
FrederikBolding: I use `FormData` as it is passed to all steps in the `AddAccountFlow` and updated in `NetworkSelectPanel`. It seemed like it was intended to be used directly instead of manually adding a prop for network.

FrederikBolding: If not, please let me know.

common/v2/features/AddAccount/components/Mnemonic.tsx
FrederikBolding: Same as above.

common/v2/services/Store/Network/helpers.ts
FrederikBolding: I'm confused by your conflicting statements when reviewing this file (this one and the one above). Shouldn't I use `getNetworkByName` from this file to get network details?

The reason I had to run `makeExplorer` again is because the cache is stored in localStorage, and thus the functions `addressUrl` etc aren't available when I get the details of a network. 

Please let me know if there is another way I should be accessing network details where this is available.

common/v2/services/Store/Network/helpers.ts
FrederikBolding: See below

common/v2/features/AddAccount/components/DeterministicWallets.tsx
FrederikBolding: I'm unsure of what you mean here as it also depends on how you want me to access network details.

FrederikBolding:
I'm moving this out of draft mode as it is almost there (working and should pass tests in a few). Still waiting for further explanation from @vilsbole on some of his comments before it should be completely ready for merge (as there might be a few things to change).

common/v2/features/AddAccount/components/Mnemonic.tsx
vilsbole: It that case it's out of scope of this PR.
Thanks for pointing it out!
(...and git blame shows me I'm the culprit...since then I prefer being more explicit then sharing the form itself.)

common/v2/features/AddAccount/components/Mnemonic.tsx
vilsbole: closing because out of scope

common/v2/services/Store/Network/helpers.ts
vilsbole: Once again, networks default dPath are not unique so they cannot be used to find a network.
e.g the Homestead and POA both share the same default path.

FrederikBolding: Yeah, I'm aware. It has already been removed. Not sure why I hadn't marked it resolved.

common/v2/types/network.ts
FrederikBolding: This has also been removed, not sure why it wasn't marked as resolved.

common/v2/features/AddAccount/components/LedgerNano.tsx
```diff
@@ -150,6 +146,11 @@ class LedgerNanoSDecryptClass extends PureComponent<Props, State> {
       });
   };
 
+  private getInitialDPath() : DPath {
+    const network = getNetworkByName(this.props.formData.network);
+    return network && network.dPaths.ledgerNanoS ? network.dPaths.ledgerNanoS : this.props.dPaths[0]
+  }
+
```
vilsbole: It seems this method is defined 6 times. Please turn into a pure function and place in EthService/network/helpers and import from `v2/services/EthService`

common/v2/types/network.ts
vilsbole: I'm sorry the code was present in the changes, so I unresolved.
I may have been reading the wrong commit and apologise for the noise.

common/v2/features/AddAccount/components/DeterministicWallets.tsx
vilsbole: To access networks please use:
```
import { NetworkContext } from 'v2/services/Store';
...
const { networks } = useContext(NetworkContext
```

In which case the `blockExplorer` should be a valid object.

FrederikBolding: I messaged you on Slack with some possiblities on how to do this. The problem is that the existing components are defined as classes and not functions, thus using context isn't as easy as your example.

vilsbole:
lgtm; 
Tested with mnemonic wallet by Frederick

PR #2711, opened 2019-08-14
csp fix for gau

tayvano:
https://app.clubhouse.io/mycrypto/story/2559/can-t-open-https-mycryptobuilds-com-gau-in-safari

test fix to see if will fix safari csp issues regarding workers since safari doesn't recognize the workers-src parameter

also related to #2708


tayvano:
@blurpesec can you merge this when you get a chance so povio can continue QA on safari?

this is the fix we discussed earlier that just adds `child-src: self` to client-side CSPs. approved by harry (see clubhouse)

PR #2713, opened 2019-08-16
Added additional information to Confirm Transaction Panel

blurpesec:
Changes:
- Use baseAsset to split out asset to send and baseAsset for the confirm panel.
- Removed implicit asset tickers.

Note: This PR isn't intended to clean up Confirm Panel styles, just to make sure logic is all there. :-)

### DIFF
--- a/common/v2/features/SendAssets/components/ConfirmTransaction.scss
+++ b/common/v2/features/SendAssets/components/ConfirmTransaction.scss
@@ -56,6 +56,16 @@
           text-align: right;
         }
       }
+
+      &-data {
+        width: 30%;
+        display: block;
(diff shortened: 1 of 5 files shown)


common/v2/features/SendAssets/components/ConfirmTransaction.scss
```diff
@@ -56,6 +56,16 @@
           text-align: right;
         }
       }
+
+      &-data {
+        width: 30%;
+        display: block;
+        overflow-wrap: break-word;
+        align-items: center;
+        @media (min-width: 700px) {
```
vilsbole: use `BREAK_POINTS` from `v2/theme`

PR #2716, opened 2019-08-19
Use self-hosted CDN for asset icons

blurpesec:
~Todo: get feedback from @miagx about whether or not we can have the endpoint modified for `.svg`s instead of `.png`s~


blurpesec:
> Also from a security aspect it's important to know if the svg, png is sanitized by the cdn (aka can we trust it) or if we have to do it on the frontend as well.

Mia is going to look into sanitation. c:

PR #2720, opened 2019-08-20
ev/get account balance

vilsbole:
Add `accounts: StoreAccount[]` which contains `balances` to StoreContext.

Misc: 
- start to untangle cache initialisation
- move localCache static definitions into `v2/config`
- stop using the same network config as legacy code
- change network name to 'Homestead' instead of `Ethereum` or `ETH`
- update Assets type to include baseAsset
- move `account.network` to `account.networkId`
- create StoreAccount type
- make `accounts: StoreAccount[]` available to entire application through `StoreContext`

To know:
- Currently we only have balances only work for Homestead accounts

Next steps:
- Handle balances for other networks
- Handle Polling once worker-loader import issue is solved.
- Deploy eth-scan on test networks
- Create test accounts with tokens on test networks


vilsbole:
0. [x] use 'Ethereum' instead of 'Homestead'
1. [x] Display correct assets in WalletBreakDown
2. [x] Filter senderAccount in SendAssets flow
3. [x] Update network correctly in SendAssets flow
4. [x] Solve MM AddAccount flow by using getNetworkById
5. [x] Display correct account label (commit 1)

**bonus:**
- solve react-select forward ref bug.
- create StoreContext with StoreAccount and StoreAssets. (A single place to fetch all the data we need)
- display connected balance in SendAssetsForm Account Dropdown
- display connected balance in WalletBreakDown

PR #2721, opened 2019-08-21
Fix/travis build times

sreckosmodis:
[ch2343]
[ch2232]


vilsbole:
Yes!

If I understand correctly:
1. Since we only run the osx script we can remove all the package install from thee travis file?
2. Travis will fail faster because linting and tscheck comes before the test suit.
3. It's faster to make a commit because we the tests only run for `pre-push`
4. `yarn freezer` will actually work because the script searches for the file list with the key `frozen`

Is this correct? Did I miss anything?

sreckosmodis:
Exactly as you described.

vilsbole:
Wunderbar! I look forward to seeing it in action!

PR #2725, opened 2019-08-28
Save tx receipts to local cache

blurpesec:
## Description

Saves tx hashes to LocalCache for relevant account when. After tx is confirmed, tx receipt is saved to LC for the account using the transaction `hash` as identifier for what to replace


common/v2/services/Store/Account/AccountProvider.tsx
```diff
@@ -32,6 +34,23 @@ export class AccountProvider extends Component {
...
+      const existingTransaction = accountData.transactions.find(
+        tx => tx.hash === newTransaction.hash
+      );
+      delete newTransaction.network;
+      if (existingTransaction) {
+        const newTransactionSet = accountData.transactions.filter(
+          transaction => transaction.hash !== newTransaction.hash
+        );
+        newTransactionSet.push(newTransaction);
+        accountData.transactions = [...newTransactionSet];
```
vilsbole: prefer the declarative form:
```
   accountData.transactions = [...newTransactionSet, newTransaction];
```

NB. Instead of telling the machine "append this to the end of the list" (imperative) we write "this comes at the end" (declarative).

common/v2/services/Store/Account/AccountProvider.tsx
```diff
@@ -32,6 +34,23 @@ export class AccountProvider extends Component {
...
+      } else {
+        accountData.transactions.push(newTransaction);
```
vilsbole: idem.
```
accountData.transactions = [...accountData.transactions, newTransaction];
```

common/v2/services/Store/Account/AccountProvider.tsx
```diff
@@ -32,6 +34,23 @@ export class AccountProvider extends Component {
...
+      }
+      service.updateAccount(accountData.uuid, accountData);
+      this.getAccounts();
+    },
```
vilsbole: Since it would be better to be using Immutable objects (it eliminates a whole category of bugs), we can actually be creating a new object instead of mutating the original one:
```
addNewTransactionToAccount: (accountData, newTransaction) => {
   const { uuid, transactions } = accountData;
   const { network, ...txWithoutNetwork } = newTransaction;
   const account = {
      ....accountData,
      transactions: [
          ...transactions.filter(tx => tx.hash === newTransaction.hash),  // will return all elements if no match found
          txWithoutNetwork
      ]
   }

   service.updateAccount(uuid, account);
   this.getAccounts();
}
```

common/v2/features/SendAssets/stateFactory.tsx
```diff
@@ -144,9 +151,11 @@ const TxConfigFactory: TUseApiFactory<State> = ({ state, setState }) => {
 
   const handleSignedWeb3Tx: TStepAction = (payload: ITxReceipt | string, after) => {
     // Payload is tx hash or receipt
+    const input = typeof payload === 'string' ? { hash: payload } : fromTxReceiptObj(payload);
```
vilsbole: maybe `txReceipt` as a more explicit name than `input`?

common/v2/features/SendAssets/stateFactory.tsx
```diff
@@ -86,18 +89,22 @@ const TxConfigFactory: TUseApiFactory<State> = ({ state, setState }) => {
...
         setState((prevState: State) => ({
           ...prevState,
-          txReceipt: fromTxReceiptObj(transactionReceipt)
+          txReceipt: transactionReceipt
         }));
       })
       .catch(txHash => {
         // If rejected, data is a tx hash, not a receipt. Fetch the receipt, then save receipt for flow
-        provider.getTransactionByHash(txHash).then(transactionReceipt => {
+        provider.getTransactionByHash(txHash).then(retrievedTransactionReceipt => {
```
vilsbole: By reducing indentation levels and nesting we can increase readability.
Since `then()` will wait for the resolve of the previous return value we can chain Promises,
and you can continue chaining even after `catch`:
```
.catch(txHash => provider.getTransaction(txHash))
.then(retrievedTransactionReceipt => ...)
```

common/v2/features/SendAssets/stateFactory.tsx
```diff
@@ -86,18 +89,22 @@ const TxConfigFactory: TUseApiFactory<State> = ({ state, setState }) => {
...
+          const transactionReceipt = fromTxReceiptObj(retrievedTransactionReceipt);
+          addNewTransactionToAccount(state.txConfig.senderAccount, transactionReceipt || {});
           setState((prevState: State) => ({
             ...prevState,
-            txReceipt: fromTxReceiptObj(transactionReceipt)
+            txReceipt: transactionReceipt
```
vilsbole: FYI: There is also `.finally`
so you can join the save and state update states in a single place:
```
provider.sendRawTx(signedTx)
 .then(retrievedTxReceipt => retrievedTransactionReceipt)
 .catch(txHash => provider.getTransactionByHash(txHash))
 .finally(retrievedTxReceipt => {
    const txReceipt  = fromTxReceiptObj(retrievedTransactionReceipt);
    addNewTransactionToAccount(state.txConfig.senderAccount, txReceipt || {});
    setState((prevState: State) => ({
      ...prevState,
       txReceipt
    }));
 })
```

common/v2/features/SendAssets/components/TransactionReceipt.tsx
```diff
@@ -37,15 +38,19 @@ export default function TransactionReceipt({ txReceipt, txConfig }: IStepCompone
     );
```
vilsbole: Since we already `return` if `txReceipt` is undefined, is this ternary really necessary?

vilsbole: Since we already `return` if `txReceipt` is undefined, is this ternary really necessary?

common/v2/features/SendAssets/components/TransactionReceipt.tsx
```diff
@@ -37,15 +38,19 @@ export default function TransactionReceipt({ txReceipt, txConfig }: IStepCompone
...
     if (blockNumber === 0 && txReceipt.hash) {
       const blockNumInterval = setInterval(() => {
-        getTransactionReceiptFromHash(txReceipt.hash, provider).then(transactionReceipt => {
-          if (transactionReceipt) {
-            const transactionStatus = transactionReceipt.status === 1 ? true : false;
+        getTransactionReceiptFromHash(txReceipt.hash, provider).then(transactionOutcome => {
+          if (transactionOutcome) {
```
vilsbole: 1. Can `getTransactionReceiptFromHash` really resolve to `undefined`?
Normally a resolved promise will always return a value, therefore this conditional is unnecessary.
Otherwise to reduce the indentation level prefer the defensive form:
```
if (!transactionOutcome) return;
```

2. Since it's orthogonal to the second condition i.e `if (timestamp === 0 && blockNumber !== 0)`
It would be cleaner to have two separate `effects`

common/v2/features/SendAssets/components/TransactionReceipt.tsx
```diff
@@ -37,15 +38,19 @@ export default function TransactionReceipt({ txReceipt, txConfig }: IStepCompone
...
+            const transactionStatus = transactionOutcome.status === 1 ? true : false;
```
vilsbole: I don't understand a `status` variable that is `true` or `false`.
Maybe we can rename `status` (especially in the state) or set its value to something more explicit such as fetching, requested, pending, success, etc.

common/v2/features/SendAssets/components/TransactionReceipt.tsx
```diff
@@ -37,15 +38,19 @@ export default function TransactionReceipt({ txReceipt, txConfig }: IStepCompone
...
             setTxStatus(prevState => ({
               status: transactionStatus || prevState.status,
               known: transactionStatus !== undefined ? true : false
```
vilsbole: naming: prefer the question form for booleans e.g `isKnown`. But here I wouldn't know what isKnown.
Since it's value is always the same as `transactionStatus` I doubt we need this extra state value.

common/v2/features/SendAssets/components/TransactionReceipt.tsx
blurpesec: done.

common/v2/features/SendAssets/components/TransactionReceipt.tsx
blurpesec: done.

common/v2/services/Store/Account/AccountProvider.tsx
blurpesec: nice spot

common/v2/features/SendAssets/stateFactory.tsx
vilsbole: It's strange, the code still shows: 
Screenshot 2019-09-02 at 15 14 12

When the goal is to write:
```
provider.sendRawTx(signedTx)
 .then(retrievedTxReceipt => retrievedTransactionReceipt)
 .catch(txHash => provider.getTransactionByHash(txHash))
 .finally(retrievedTxReceipt => {
    const txReceipt  = fromTxReceiptObj(retrievedTransactionReceipt);
    addNewTransactionToAccount(state.txConfig.senderAccount, txReceipt || {});
    setState((prevState: State) => ({
      ...prevState,
       txReceipt
    }));
 })
```
Am I missing something?

common/v2/features/SendAssets/stateFactory.tsx
vilsbole: Closing in favour of comment above

common/v2/types/transaction.ts
```diff
@@ -37,3 +37,7 @@ export interface IHexStrWeb3Transaction {
   nonce: string;
   chainId: number;
 }
+
+export interface ITxReceipt {
+  [index: string]: any;
+}
```
vilsbole: This type is too general but we can complete it later

common/v2/features/SendAssets/stateFactory.tsx
blurpesec: > .finally(retrievedTxReceipt => { ...
This doesn't work with TS.

This has error: `Argument of type '(retrievedTxReceipt: ITxReceipt) => void' is not assignable to parameter of type '() => void'.`

vilsbole:
lgtm;

PR #2727, opened 2019-08-29
Fixes for get-account-balance usability

blurpesec:
There were some problems with get-account-balance usability that I wanted to fix. These are those fixes. This should get us up to near-parity with current `gau` functionality.

Changes:
1) Fixed a bug with duplicate entries when adding existing accounts.
2) Fixed an issue with base asset population when adding existing accounts.
3) Fixed Asset Dropdown duplication bug in Send-Assets flow (there still exists a mismatch bug between test accounts and add-existing accounts, but its not blocking anything -> To select an Ethereum account, make sure to click the "Ethereum" asset, NOT the "Ether" asset in dropdown).
4) Fixed Account Dropdown bug in Send-Assets flow.
5) Fixed bug with network usage for ethers.js provider initialization.
6) Fixed a bug with send-max button.

### DIFF
--- a/common/v2/features/SendAssets/components/SendAssetsForm.tsx
+++ b/common/v2/features/SendAssets/components/SendAssetsForm.tsx
@@ -3,7 +3,7 @@ import { Field, FieldProps, Form, Formik, FastField } from 'formik';
 import * as Yup from 'yup';
 import { Button, Input } from '@mycrypto/ui';
 import _ from 'lodash';
-import { BigNumber } from 'ethers/utils';
+import { BigNumber, formatEther } from 'ethers/utils';
 import BN from 'bn.js';
 
 import translate, { translateRaw } from 'translations';
@@ -165,7 +165,7 @@ export default function SendAssetsForm({
                       accountAsset => accountAsset.uuid === values.asset.uuid
                     ) || { balance: '0' }
                   ).balance
-                : getBalanceFromAccount(values.account);
+                : formatEther(getBalanceFromAccount(values.account));
               const gasPrice = values.advancedTransaction
                 ? values.gasPriceField
                 : values.gasPriceSlider;
--- a/common/v2/features/SendAssets/components/fields/AssetDropdown.tsx
+++ b/common/v2/features/SendAssets/components/fields/AssetDropdown.tsx
@@ -25,7 +25,7 @@ class AssetOption extends React.PureComponent<OptionComponentProps> {
 
 function AssetDropdown({ assets, name, value, onSelect }: Props<Asset>) {
   const filteredAssets: Asset[] = assets.filter(
-    (asset, index) => assets.indexOf(asset) >= index
+    (asset, index) => assets.map(assetObj => assetObj.uuid).indexOf(asset.uuid) >= index
   ); /* Removes duplicates */
   return (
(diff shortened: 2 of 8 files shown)


common/v2/services/EthService/network/helpers.ts
```diff
@@ -7,7 +7,7 @@ import { Network, NetworkId, NodeType, DPathFormat } from 'v2/types';
 type TValidEtherscanNetwork = 'homestead' | 'ropsten' | 'rinkeby' | 'kovan' | 'goerli';
 
 const getValidEthscanNetworkId = (id: NetworkId): TValidEtherscanNetwork =>
-  id.toLowerCase() as TValidEtherscanNetwork;
+  id === 'Ethereum' ? 'homestead' : (id.toLowerCase() as TValidEtherscanNetwork);
```
vilsbole: Forgot about that one. Good catch!

common/v2/services/Store/LocalCache/seedCache.ts
```diff
@@ -115,7 +115,7 @@ export const initNetworks = () => {
       uuid: baseAssetID,
       name: network.name,
       networkId: network.name as NetworkId,
-      ticker: networkId,
+      ticker: network.unit,
```
vilsbole: Good catch!

common/v2/features/SendAssets/components/SendAssetsForm.tsx
```diff
@@ -165,7 +165,7 @@ export default function SendAssetsForm({
                       accountAsset => accountAsset.uuid === values.asset.uuid
                     ) || { balance: '0' }
                   ).balance
-                : getBalanceFromAccount(values.account);
+                : formatEther(getBalanceFromAccount(values.account));
```
vilsbole: I like the idea of custom wrappers around ethers methods. We could start using them in future work.

common/v2/services/Store/Asset/helpers.ts
```diff
@@ -61,9 +61,9 @@ export const getAssetByContractAndNetwork = (
 export const getTotalByAsset = (assets: StoreAsset[]) =>
   assets.reduce(
     (dict, asset) => {
-      const prev = dict[asset.uuid];
+      const prev = dict[asset.name];
```
vilsbole: Since it's the `uuid` that is unique, I don't understand why we would use the `name`.
The previous implementation worked for wallet breakdown, and currently it's the only place this function is implemented.

common/v2/features/SendAssets/components/fields/AssetDropdown.tsx
```diff
@@ -25,7 +25,7 @@ class AssetOption extends React.PureComponent<OptionComponentProps> {
 
 function AssetDropdown({ assets, name, value, onSelect }: Props<Asset>) {
   const filteredAssets: Asset[] = assets.filter(
-    (asset, index) => assets.indexOf(asset) >= index
+    (asset, index) => assets.map(assetObj => assetObj.uuid).indexOf(asset.uuid) >= index
```
vilsbole: Note to self:
The extra `map` seems expensive, and I'm wondering if index order is actually what we want to use

PR #2729, opened 2019-08-29
contract selection with url params

OlivierDemeaux:
## Description

Added some code to allow a contract, address and abi to be set on the mtCrypto/contracts/interact page with a param passed in the URL. No need to search the contract in the dropdown.
ex: "myCrypto/contracts/interact?name=BCdiploma - Evidenz".
If a param is passed in the URL, the contract, address and abi will be set and the button "access" will be click.
PROBLEME: The access button crashes the localhost i'm running myCrypto on. (Even when i click on access button after i manually clicked on the contract)

## Changes

* Can now pass a contract name in the url as param to have it automatically selected on the page myCrypto/contracts/interact
ex: "myCrypto//contracts/interact?name=BCdiploma%20-%20EvidenZ"

## Steps to Test

1. yarn install
2. yarn start
3. run url "localhost/contracts/interact?name=BCdiploma%20-%20EvidenZ" (works with every contract name)
4. If localhost crashes, edit out the line 85, 86, 87 of the file : common/containers/Tabs/Contracts/components/Interact/components/InteractForm/index.tsx.
Those lines are responsibles for the click on the "access" button after that the contract name, address and abi are set.

## Quality Assurance

- [✓] The branch name is in lowercase-kebab-case with no prefix (unless it was created from Clubhouse)
- [✓] The base branch is develop or gau (no nested branches)
- [✓] This is related to a maximum of one Clubhouse story or GitHub issue
- [✓] Types are safe (avoid TypeScript/TSLint features like any and disable, instead use more specific types)
- [✓] If code is copied from existing directories, there is an explanation of why this is necesary in the description/changes, and all copying is done in separate commits to make them easy to filter out


common/containers/Tabs/Contracts/components/Interact/components/InteractForm/index.tsx
```diff
@@ -76,6 +77,37 @@ class InteractForm extends Component<Props, State> {
...
+
+  public contractSetup = () => {
+    const { contracts } = this.props;
+    // get contract name from url params
+    const contractName = queryString.parse(location.search).name;
+    if (this.isContractsValid()) {
+      contracts.map(con => {
+        const addr = con.address ? `(${con.address.substr(0, 10)}...)` : '';
+        // if contract name from Url exists, then load this contract
+        if (con.name === contractName) {
```
409H: Can you make a modification here to do the comparison as both strings in lowercase for a UX factor (ie: `MyContract` contract name won't match `mycontract` query string value)

OlivierDemeaux: Good idea! Thanks. Did just that. Still not working on a live version even tho it works on a localhost so making a bunch of console.log to understand what's happing

mwbailey:
Hi @OlivierDemeaux. I got curious when I saw you mention that strange localhost vs non-localhost bug on the MyCrypto Discord. As you've discovered, the problem is the undefined `abi` field in the `Contract` class in `common/libs/contracts/index.ts`, which you fixed by assigning it a value in the `Interact` component.

Turns out the reason it was crashing the development build but not the production build wasn't because of a localhost issue. The development build (when you run `yarn start`) uses Babel to compile/transpile the Typescript, and the production build (e.g., `yarn build`) uses the official Typescript compiler (tsc). I assume the reason they set it up this way is that Babel has faster build times and React Hot Loader compatibility, but tsc is needed to actually do type-checking.

Babel (correctly, according to the proposed [spec](https://github.com/tc39/proposal-class-fields#fields-without-initializers-are-set-to-undefined)) transpiles the uninitialized `abi` field as though it were `abi = undefined`. This results in the crash in development builds. The Typescript compiler currently ignores uninitialized field declarations, so the `abi` property does not appear in the emitted Javascript code at all. That's why the production build doesn't crash when iterating through the properties in `Contract.getFunctions()`.

Unless I'm missing something, it seems like the `public abi: any;` line in the `Contract` class can just be deleted. The production build has been running without it with no apparent problems. I don't see anywhere in the code where that `abi` field is used. (Note: there's a separate `NetworkContract` interface with an `abi` field that is needed.) I think it may be an unnecessary leftover from a refactor of the Contract code about 2 years ago around release tag 0.0.6.

common/components/ui/Dropdown.tsx
```diff
@@ -17,15 +18,15 @@ export default class Dropdown extends React.Component<Props> {
...
-
-  public componentDidUpdate(_: Props, prevState: State) {
-    const { value } = this.props;
-    const { selectedOption } = prevState;
-
-    if (selectedOption !== value) {
-      this.setState({ selectedOption: value });
-    }
-  }
+  //
```
Sonicrida: It would be preferable to remove commented out code.

common/containers/Tabs/Contracts/components/Interact/components/InteractExplorer/index.tsx
```diff
@@ -73,19 +74,91 @@ class InteractExplorerClass extends Component<Props, State> {
...
+      if (readButton) {
+        readButton.click();
+      }
+    }
+  };
+
   public componentDidMount() {
     this.props.setAsContractInteraction();
     this.props.resetTransactionRequested();
+    setTimeout(
```
Sonicrida: It would be a good idea to make sure that this timeout is cleared in `componentWillUnmount`

common/containers/Tabs/Contracts/components/Interact/components/InteractForm/index.tsx
```diff
@@ -76,6 +77,48 @@ class InteractForm extends Component<Props, State> {
     return contracts && contracts.length;
   };
 
+  public componentDidMount = () => {
+    setTimeout(
```
Sonicrida: It would be a good idea to make sure that this timeout is also cleared in `componentWillUnmount`

common/containers/Tabs/Contracts/components/Interact/components/InteractForm/index.tsx
```diff
@@ -76,6 +77,48 @@ class InteractForm extends Component<Props, State> {
...
+      function() {
+        this.contractSetup();
+        const item = document.getElementById('accessButton');
```
blurpesec: It would be nice to have this use the method `AccessContract` instead of relying on something that could change in the future like the ID.

common/containers/Tabs/Contracts/components/Interact/components/InteractExplorer/index.tsx
```diff
@@ -73,19 +74,91 @@ class InteractExplorerClass extends Component<Props, State> {
...
+            inputs: {
+              ...this.state.inputs,
+              [element.name]: value
+            }
+          });
+        }
+      }
+    });
+    if (count > 0) {
+      const itemToScrollTo = document.getElementById('contractRead');
```
blurpesec: Can you change this to use the react methods for reading contract instead of using js to click a button that may change at some point in the future ? :-)

common/containers/Tabs/Contracts/components/Interact/components/InteractExplorer/index.tsx
```diff
@@ -73,19 +74,91 @@ class InteractExplorerClass extends Component<Props, State> {
...
+      if (itemToScrollTo) {
+        itemToScrollTo.scrollIntoView();
+      }
+      const readButton = document.getElementById('readButton');
```
blurpesec: > Can you change this to use the react methods for button click instead of using js to click a button that may change at some point in the future ? :-)

Same as above ^

common/containers/Tabs/Contracts/components/Interact/components/InteractForm/index.tsx
OlivierDemeaux: Sorry for the late response, i was travelling! Thanks for taking the time to review the code. What do you mean by timeout is cleared ?

common/containers/Tabs/Contracts/components/Interact/components/InteractForm/index.tsx
OlivierDemeaux: Thanks for the great point! I changed it

common/containers/Tabs/Contracts/components/Interact/components/InteractForm/index.tsx
OlivierDemeaux: Oooh i got it. Sorry for the delay. Thanks for the good feedback

common/containers/Tabs/Contracts/components/Interact/components/InteractExplorer/index.tsx
OlivierDemeaux: Did it. Thanks

common/containers/Tabs/Contracts/components/Interact/components/InteractExplorer/index.tsx
OlivierDemeaux: Hey ! I changed it

common/containers/Tabs/Contracts/components/Interact/components/InteractExplorer/index.tsx
OlivierDemeaux: Took my way longer than i like to admit it but i changed it. Take a look and let me know if that's good. Thanks a lot !

common/containers/Tabs/Contracts/components/Interact/components/InteractExplorer/index.tsx
OlivierDemeaux: Hi @blurpesec ! Can i get a review for my code and your opinions on it? Hope we can merge!

common/components/ui/Dropdown.tsx
OlivierDemeaux: done

common/containers/Tabs/Contracts/components/Interact/components/InteractForm/index.tsx
```diff
@@ -76,6 +80,51 @@ class InteractForm extends Component<Props, State> {
...
+    );
+  };
+
+  public componentWillUnmount = () => {
+    if (this.timer) {
+      clearTimeout(this.timer);
+    }
+  };
+
+  public getParamsFromUrl = () => {
```
blurpesec: This function seems to be repeated 3x. Is there any way that you can extract this out, or use it in a higher order component and pass the data down as props?

common/containers/Tabs/Contracts/components/Interact/components/InteractExplorer/index.tsx
```diff
@@ -67,25 +68,102 @@ class InteractExplorerClass extends Component<Props, State> {
...
   };
 
+  public setParams = () => {
+    const params = this.getParamsFromUrl();
+    let count = 0;
+    Object.keys(params).map((index: any) => {
+      const item = params[index];
+      if (index.substr(0, 5) === 'param') {
+        count++;
+        const element: any = document.getElementById(index);
```
blurpesec: Can you explain what this method is attempting to do and why this `getElementById` is needed?

OlivierDemeaux: Just grab the params from the url and set them in the function's input fields according to their respective numbers. It needs getElementById in order to get the input and set the value to that input

common/containers/Tabs/Contracts/components/Interact/components/InteractForm/index.tsx
OlivierDemeaux: Took me longer than i want to admit but I managed to extract it! Tell me what you think about it!

common/containers/Tabs/Contracts/components/Interact/components/InteractExplorer/index.tsx
OlivierDemeaux: Hi @blurpesec. Any news?

VinceBCD:
Hi guys! From our point of view, it's a must have for adoption. If you have time I can give you a demo of our product to show you how crucial it is (and not only for ourselves.)
With this feature, the general public can, in an application context, verify information stored on the blockchain with a single click. This PR is opened for 3 months now, how to move forward ?

FrederikBolding:
Closing in favor of #2920

PR #2732, opened 2019-09-03
Misc design fixes 2

Sonicrida:
## More design tweaks and updates

......

## Changes

* Fixed ad rendering issue in all browsers
* Fixed table column display
* Add view details CTA to wallet breakdown
* Fixed overflowing layout on certain pages

[ch2766]


common/v2/features/Dashboard/components/WalletBreakdown/WalletBreakdownView.tsx
```diff
@@ -230,6 +235,9 @@ export default function WalletBreakdownView({
           ))}
         </BreakDownBalanceList>
         <BalanceTotalWrapper>
+          <ViewDetailsLink onClick={toggleShowChart}>
+            {translate('WALLET_BREAKDOWN_MORE')}
```
amacar: View details link is already present but visible only when Other Tokens are visible. If you want to change the behavior please remove existing code (https://github.com/MyCryptoHQ/MyCrypto/blob/misc-design-fixes-2/common/v2/features/Dashboard/components/WalletBreakdown/WalletBreakdownView.tsx#L226) because right now this button could be rendered twice.

Screenshot:
image

PR #2735, opened 2019-09-07
ev/balance fetching for other networks

vilsbole:
[2857] https://app.clubhouse.io/mycrypto/story/2857/implement-balance-fetch-for-non-eth-networks
[2858] https://app.clubhouse.io/mycrypto/story/2858/implement-polling-for-balance-fetching

- Add balance fetching for all networks.
- Implement naive polling for balances.
- Limit Store state changes to when the balance is actually changed.


common/v2/services/EthService/network/providerHandler.ts
```diff
@@ -38,6 +52,7 @@ export class ProviderHandler {
         data: this.requests.getTokenBalance(address, token).params[0].data
       })
       .then(data => ERC20.balanceOf.decodeOutput(data))
+      .then(({ balance }) => balance)
```
amacar: Am I missing something or this line leads to balance being undefined in next `.then()` call when `balance` is destructured again.

Since getTokenBalance is using the same calls as getRawTokenBalance it might be good to reuse it.
```
/* Tested */
  public getTokenBalance(address: string, token: Asset): Promise<string> {
    return this.getRawTokenBalance(address, token).then(balance => {
      if (token.decimal) {
        return baseToConvertedUnit(balance, token.decimal);
      }
      return baseToConvertedUnit(balance, 18);
    });
  }
```

common/v2/services/useInterval.ts
```diff
@@ -0,0 +1,28 @@
+import { useEffect, useRef } from 'react';
```
amacar: Isn't this more like a helper function and not a service?

vilsbole: Indeed!
Moved to `v2/utils`

common/v2/services/EthService/network/providerHandler.ts
vilsbole: Done

vilsbole:
> I am getting undefined errors occasionally when running this branch (attached image below). It would be great if you could check where lies the problem. Otherwise, good work, I don't have any significant complaints.
> 
> error

As per slack exchange, can not reproduce. Will address in case it surfaces again:
https://mycryptohq.slack.com/archives/GFYJ8G9L7/p1568118026040700?thread_ts=1567867681.028100&cid=GFYJ8G9L7

PR #2737, opened 2019-09-10
ev/file organisation

vilsbole:
[ch2339] https://app.clubhouse.io/mycrypto/story/2339/migrate-remaining-v2-providers-to-v2-services
[ch ..... ] 

- move remaining providers into services
- activate Drawer for language selection


LuckyLuker:
[ch2963]

PR #2740, opened 2019-09-12
Display fiat value in AccountList

FrederikBolding:
🎉 🎉 🎉

## Description

Displayed fiat value in AccountList used on the Dashboard and Settings page

## Changes

- Added `totalFiat` method to `StoreProvider`
- Displayed fiat value in AccountList using `totalFiat`

## Steps to Test

View either the Dashboard or Settings page 😄 

## Quality Assurance

### DIFF
--- a/common/v2/components/AccountList.tsx
+++ b/common/v2/components/AccountList.tsx
@@ -65,8 +65,8 @@ interface AccountListProps {
 
 export default function AccountList(props: AccountListProps) {
   const { className, currentsOnly, deletable, favoritable, footerAction, footerActionLink } = props;
-  const { currentAccounts } = useContext(StoreContext);
-  const { accounts, deleteAccount, updateAccount } = useContext(AccountContext);
+  const { currentAccounts, accounts } = useContext(StoreContext);
+  const { deleteAccount, updateAccount } = useContext(AccountContext);
   const shouldRedirect = accounts === undefined || accounts === null || accounts.length === 0;
   if (shouldRedirect) {
     return <Redirect to="/no-accounts" />;
@@ -98,7 +98,7 @@ export default function AccountList(props: AccountListProps) {
 }
 
 function buildAccountTable(
-  accounts: ExtendedAccount[],
+  accounts: StoreAccount[],
   deleteAccount: DeleteAccount,
   updateAccount: UpdateAccount,
   deletable?: boolean,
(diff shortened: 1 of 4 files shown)


common/v2/services/Store/StoreProvider.tsx
```diff
@@ -61,6 +64,13 @@ export const StoreProvider = ({ children }: { children: React.ReactNode }) => {
       selectedAssets.filter((asset: StoreAsset) => asset.type !== 'base'),
     totals: (selectedAccounts = state.accounts) =>
       Object.values(getTotalByAsset(state.assets(selectedAccounts))),
+    totalFiat: (selectedAccounts = state.accounts) => {
+      return (getRate: (ticker: TTicker) => number | undefined) => {
+        return state.totals(selectedAccounts).reduce((sum, asset) => {
+          return (sum += convertToFiat(asset.balance, getRate(asset.ticker as TTicker)));
+        }, 0);
+      };
+    },
```
vilsbole: How about using implicit return to reduce indentation? 
```
(selectedAccounts = state.accounts) => (
  getRate: (ticker: TTicker) => number | undefined
) =>
  state
    .totals(selectedAccounts)
    .reduce(
      (sum, asset) => (sum += convertToFiat(asset.balance, getRate(asset.ticker as TTicker))),
      0
    );
```

common/v2/components/AccountList.tsx
```diff
@@ -133,7 +127,10 @@ function buildAccountTable(
         <Network key={index} color="#a682ff">
           {account.networkId}
         </Network>,
-        <Typography key={index}>{getBalanceFromAccount(account)}</Typography>
+        <Typography key={index}>
+          {Fiats[0].symbol}
+          {total.toFixed(2)}
+        </Typography>
```
vilsbole: 1. Use `v2/components/currency` for the display.
2. I find it troublesome to do `Fiats[0]`. How about changing the export to an object and access it as `Fiats[USER_FIAT]` (with a better name of course).

NB. It seems that we don't need to have fiats in the localstorage. We should be able to remove it from initialisation all together. I'll ask in the slack

common/v2/config/cacheData.ts
```diff
@@ -3,26 +3,35 @@ import { Asset, Contract, NetworkId } from 'v2/types';
...
 };
 export const GBP = {
   code: 'GBP',
-  name: 'British Pounds'
+  name: 'British Pounds',
+  symbol: '£' as TSymbol
 };
 
-export const Fiats: Fiat[] = [USD, EUR, GBP];
+export const Fiats: FiatObject = { USD: USD, EUR: EUR, GBP: GBP };
```
vilsbole: Travis tslint complains here: 
`expected property shorthand`

prefer: { USD, EUR, GBP }

FrederikBolding: My bad, there you go! 😄

PR #2745, opened 2019-09-18
Fixed a few issues with new test assets

FrederikBolding:
🎉 🎉 🎉

## Description

Fixed a few issues with new test assets as discussed with @blurpesec on Slack.

## Changes

* Fixed contract address comparison
* Renamed some testnet assets/tokens

## Steps to Test

View the dashboard, and see that issues with weird assets are gone! 😄 

## Quality Assurance

### DIFF
--- a/common/v2/config/tokens/kovan.json
+++ b/common/v2/config/tokens/kovan.json
@@ -15,13 +15,13 @@
   },
   {
     "address": "0x4c38cdc08f1260f5c4b21685654393bb1e66a858",
-    "symbol": "DAI",
+    "symbol": "KovanDAI",
     "decimal": 18,
     "name": "DAI Stablecoin Kovan"
   },
   {
     "address": "0xac94ea989f6955c67200dd67f0101e1865a560ea",
-    "symbol": "MKR",
+    "symbol": "KovanMKR",
     "decimal": 18,
     "name": "Maker Kovan"
   }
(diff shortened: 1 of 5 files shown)


common/v2/config/tokens/kovan.json
```diff
@@ -15,13 +15,13 @@
   },
   {
     "address": "0x4c38cdc08f1260f5c4b21685654393bb1e66a858",
-    "symbol": "DAI",
+    "symbol": "KovanDAI",
```
vilsbole: @FrederikBolding  @blurpesec  Just want to double check about this.
From what I gather, we change the token symbol to distinguish between each token when we display the totals in Walletbreakdown.
Taking this path would mean we need to create a unique symbol.
And that the conflict will still arise for custom tokens etc.
And we can't get balances for tokens on other networks (not sure this is even possible today).
Shouldn't we modify the `totals` function instead and take the network into account as well ?

FrederikBolding: Currently we rely too much on the symbols being unique, this is indeed a temporary solution. 

We assume they are unique when looking up rates, giving wrong rates for total balances, but we also assume they are unique when initializing cache here: https://github.com/MyCryptoHQ/MyCrypto/blob/281717a62c8ae71561242b184c65f1c870da9225/common/v2/config/cacheData.ts#L60

If you want, a more permanent solution to at least one of the problems in changing how we initialize the assets into cache?

FrederikBolding: As discussed with @blurpesec on Slack - having unique symbols for testnet tokens makes it much less confusing for users as well.

FrederikBolding:
Agreed. This is a very temporary solution. Created a card in Clubhouse to figure out a more permanent solution.

PR #2746, opened 2019-09-18
Added support for prefixing currency symbols

FrederikBolding:
🎉 🎉 🎉

## Description

Added support for prefixing currency symbols.
I.e. USD should be prefixed to the value as "$10" instead of "10$".

## Changes

* Added prefix prop to Currency component
* Added prefix field to Fiat interface
* Used prefix prop in AccountList

## Steps to Test

View dashboard and see that $ is used correctly as prefix instead of postfix.

## Quality Assurance

### DIFF
--- a/common/v2/components/AccountList.tsx
+++ b/common/v2/components/AccountList.tsx
@@ -138,6 +138,7 @@ function buildAccountTable(
           key={index}
           amount={total.toString()}
           symbol={Fiats[settings.fiatCurrency].symbol}
+          prefix={Fiats[settings.fiatCurrency].prefix}
           decimals={2}
         />
       ];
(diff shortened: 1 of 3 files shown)


FrederikBolding:
[ch3029]

FrederikBolding:
> How about we also update the code in the `AccountList` PR to use this change.
> Or would you prefer doing a separate PR?

I'm not sure what you mean? I changed the `AccountList` to use it in this PR?

vilsbole:
Terribly sorry!
I had the memory of an access by index in the previous PR.
Merging now.

FrederikBolding:
No problem, thanks! 😄

PR #2748, opened 2019-09-19
Fix StoreProvider Bugs

blurpesec:
### Changes
- [x] Fixed a bug with mismatched StoreProvider accounts/LocalCache accounts where re-renders didn't occur on deletions of accounts.
- [ ] Fix bug with newly-added accounts not being added correctly. Won't display on dashboard till page refresh.

### DIFF
--- a/common/v2/services/Store/StoreProvider.tsx
+++ b/common/v2/services/Store/StoreProvider.tsx
@@ -1,4 +1,4 @@
-import React, { useState, useContext, useMemo, createContext } from 'react';
+import React, { useState, useContext, useMemo, createContext, useEffect } from 'react';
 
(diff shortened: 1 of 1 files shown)


common/v2/services/Store/StoreProvider.tsx
```diff
@@ -40,6 +40,9 @@ export const StoreProvider = ({ children }: { children: React.ReactNode }) => {
     networks
   ]);
   const [accounts, setAccounts] = useState(storeAccounts);
+  useEffect(() => {
+    setAccounts(storeAccounts);
+  }, [storeAccounts]);
```
vilsbole: This creates an infinite loop.
Please remove!

PR #2749, opened 2019-09-19
Fix undefined file hash

### LINKED ISSUE #2563: JS asset file name hashes are 'undefined'
### Description of Issue

The url's of the various JS bundles contain the word `undefined`, which would seem to indicate whatever content hash is added there for cache busting isn't properly set. This might cause launches of new code to load old assets if they have them cached, especially since there are some hefty cache-control headers in place right now.

Example of one such asset: https://mycrypto.com/badBrowserCheckA.undefined.js

Mrtenz:
## Closes #2563 
🎉 🎉 🎉

## Description

`npm_package_gitHead` is not defined when using `yarn`, resulting in undefined file hashes on the production build. This PR replaces it with the Webpack build hash.

## Changes

* Replaced `npm_package_gitHead` with Webpack's hash.

## Steps to Test

1. `yarn build`
2. Check if filenames in `dist/prod` include file hashes.

## Quality Assurance

### DIFF
--- a/webpack_config/makeConfig.js
+++ b/webpack_config/makeConfig.js
@@ -26,7 +26,6 @@ const DEFAULT_OPTIONS = {
 module.exports = function(opts = {}) {
   const options = Object.assign({}, DEFAULT_OPTIONS, opts);
   const isDownloadable = options.isHTMLBuild || options.isElectronBuild;
-  const commitHash = process.env.npm_package_gitHead;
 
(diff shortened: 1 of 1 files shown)


vilsbole:
lgtm;
@wtzb can we merge into develop then merge into gau?

wtzb:
@vilsbole sounds good to me 👍

PR #2750, opened 2019-09-19
Refactor browser compatibility check

Mrtenz:
## Description

I refactored the browser checks to use Modernizr instead of our own implementation.

## Changes

* Removed old browser checks
* Implemented `webpack-modernizr-loader` for small Modernizr builds
* Added new browser check using Modernizr

## Steps to Test

1. Disable one of the required features (e.g. local storage)
2. Open the website
3. Check if the bad browser overlay shows up

## Quality Assurance


Mrtenz:
[ch1844]

common/badBrowserCheck.ts
```diff
@@ -0,0 +1,31 @@
...
+ * @return {boolean}
+ */
+const isSupported = (feature: keyof typeof Modernizr): boolean => {
+  if (!Modernizr[feature]) {
+    const element = document.getElementsByClassName('BadBrowser')[0];
+    element.className += ' is-open';
+
+    if (/iPhone|iPad|iPod|Android/i.test(navigator.userAgent)) {
+      element.className += ' is-mobile';
+    }
```
vilsbole: Avoid nested conditionals
```
if (Mondernizr[feature]) return true;
const element = ....

```

common/badBrowserCheck.ts
```diff
@@ -0,0 +1,31 @@
...
+
+    console.log(`Feature '${feature}' not supported.`);
```
vilsbole: Use `console.debug` or `.warn` instead

common/badBrowserCheck.ts
```diff
@@ -0,0 +1,31 @@
...
+
+    return false;
+  }
+
+  return true;
+};
+
+document.addEventListener('DOMContentLoaded', () => {
```
vilsbole: Is it possible to run the check before?
eg. do we really need to wait for React to load.

common/badBrowserCheck.ts
vilsbole: Have you looked into `Modernizr.addTest`? https://modernizr.com/docs/
It seems we could replace this check with a `ismobile` test

Mrtenz: AFAIK this is only for visual changes on the bad browser overlay. Does it really make sense to use Modernizr for that?

common/badBrowserCheck.ts
vilsbole: ’m suspicious of the hidden dependency.
We add a class if a feature is missing, and somewhere in the code there is a listener of the class that will render a view.
Instead of using a class to make the switch maybe we could be more declarative and append to the DOM right away?
This way we don’t need to wait for the rest of the DOM to be loaded (edite

common/badBrowserCheck.ts
vilsbole: I doubt it. We should be able to accomplish the same using responsive css

PR #2754, opened 2019-09-24
Fixed network selection issues

FrederikBolding:
🎉 🎉 🎉

## Description

Fixed a few issues in relation to `NetworkSelectDropdown` and `NetworkSelectPanel`.
See Clubhouse card or changes below!

## Changes

- Used `Dropdown` component for `NetworkSelectDropdown`, allows for searching etc.
- Disabled next button when an invalid network is selected
- Automatic network selection when using MetaMask

### Dropdown
1. Go to add account
2. Add any account other than metamask
3. See that you can search and choose networks

### Invalid networks
1. Go to add account
2. Add any account other than metamask
3. See that you can't continue with an invalid network

### Metamask
1. Go to add account
2. Add metamask account
3. See that the correct network is automatically chosen and the account is added

## Quality Assurance

### DIFF
--- a/common/v2/components/NetworkSelectDropdown.tsx
+++ b/common/v2/components/NetworkSelectDropdown.tsx
@@ -1,15 +1,38 @@
 import React, { useContext, useEffect } from 'react';
-import { ComboBox } from '@mycrypto/ui';
 
+import styled from 'styled-components';
 import { translate } from 'translations';
 import { NetworkContext, isWalletFormatSupportedOnNetwork } from 'v2/services/Store';
 import { NetworkId, WalletId } from 'v2/types';
 import { DEFAULT_NETWORK } from 'v2/config';
+import { Typography, Dropdown } from 'v2/components';
+import { OptionComponentProps } from 'react-select';
 
 interface Props {
   network: string | undefined;
   accountType?: WalletId;
   onChange(network: NetworkId): void;
+}
+
+const SContainer = styled('div')`
+  display: flex;
+  flex-direction: row;
+  padding: 16px 15px 16px 15px;
+
+  &:hover {
+    background-color: var(--color-gray-lighter);
+  }
+`;
+
+class NetworkOption extends React.PureComponent<OptionComponentProps> {
+  public render() {
+    const { option, onSelect } = this.props;
+    return (
+      <SContainer onClick={() => onSelect!(option, null)}>
+        <Typography value={option.label} />
+      </SContainer>
+    );
+  }
 }
 
 function NetworkSelectDropdown({ network, accountType, onChange }: Props) {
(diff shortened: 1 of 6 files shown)


FrederikBolding:
[ch2356]

common/v2/components/NetworkSelectDropdown.tsx
```diff
@@ -1,17 +1,48 @@
 import React, { useContext, useEffect } from 'react';
-import { ComboBox } from '@mycrypto/ui';
 
+import styled from 'styled-components';
 import { translate } from 'translations';
 import { NetworkContext, isWalletFormatSupportedOnNetwork } from 'v2/services/Store';
 import { NetworkId, WalletId } from 'v2/types';
 import { DEFAULT_NETWORK } from 'v2/config';
+import { Typography, Dropdown } from 'v2/components';
+import { OptionComponentProps } from 'react-select';
```
vilsbole: Please respect order of imports

```
import { any } from  'npm_module'

import { any } from 'v1'
import { any } from 'v2'
import { any } from 'relative import'

```

common/v2/components/NetworkSelectDropdown.tsx
```diff
@@ -32,12 +63,17 @@ function NetworkSelectDropdown({ network, accountType, onChange }: Props) {
...
-      <ComboBox
-        value={network}
-        items={new Set(validNetworks.sort())}
-        placeholder={DEFAULT_NETWORK}
-        onChange={({ target: { value } }) => onChange(value as NetworkId)}
-      />
+      <DropdownContainer>
+        <Dropdown
+          value={{ label: network }}
+          options={validNetworks.sort().map(n => ({ value: n, label: n }))}
```
vilsbole: From what I understand this forces us to use the `name` as a `NetworkId`.
How about changing `validNetworks` to return an object `{ label: n.name, value: n }`
It should allow us on L71 to do `onChange(value.id)`

common/v2/services/WalletService/web3/web3.ts
```diff
@@ -46,24 +45,18 @@ export const initWeb3Node = async () => {
...
       action.type === configNodesSelectedTypes.ConfigNodesSelectedActions.CHANGE_SUCCEEDED &&
         action.payload.nodeId === 'web3'
     }*/
 
-    const web3Node: any | null = await getWeb3Node();
-    if (!web3Node) {
-      throw Error('Web3 node config not found!');
-    }
-    const network = web3Node.network;
-
```
vilsbole: Great job at identifying and solving this bug!

FrederikBolding: Thanks! 😄

common/v2/components/NetworkSelectDropdown.tsx
FrederikBolding: Good call, thanks!

PR #2755, opened 2019-09-24
Fix Bug with Address Label Addition

blurpesec:
###### Description
There was a bug with adding a new label to an address in the settings page. This PR fixes that.

###### Changes
- Added Network Field
- Altered the function finding address labels to use context instead of reading directly from LC.


vilsbole:
Account selection in Dashboard appears to be broken

Screenshot 2019-09-26 at 12 05 04

PR #2771, opened 2019-09-29
Fix sendmax button

blurpesec:
## Description

Fixes the sendmax button when user is trying to send tokens.

## Changes

* Added handling for erc20 tokens in the sendmax button logic.
* Added logic for Send New Transaction button in receipt panel

### DIFF
--- a/common/v2/features/SendAssets/components/SignTransactionWallets/Hardware.tsx
+++ b/common/v2/features/SendAssets/components/SignTransactionWallets/Hardware.tsx
@@ -7,6 +7,7 @@ import { InlineErrorMsg } from 'v2/components/ErrorMessages';
 
 import { ITxObject, ISignedTx } from '../../types';
 import './Hardware.scss';
+import { WalletName } from 'v2/types/walletId';
 
 export interface IDestructuredDPath {
   dpath: string;
@@ -100,7 +101,7 @@ export default function HardwareSignTransaction({
   return (
     <div className="SignTransaction-panel">
       <div className="SignTransactionHardware-title">
-        Sign the Transaction with your {senderAccount.wallet} Wallet
+        {`Sign the Transaction with your ${WalletName[senderAccount.wallet]} Wallet`}
       </div>
       <div className="SignTransactionHardware-instructions">{signerDescription}</div>
       <div className="SignTransactionHardware-content">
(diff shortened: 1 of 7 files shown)


common/v2/features/SendAssets/components/SignTransactionWallets/Hardware.tsx
```diff
@@ -100,7 +101,7 @@ export default function HardwareSignTransaction({
   return (
     <div className="SignTransaction-panel">
       <div className="SignTransactionHardware-title">
-        Sign the Transaction with your {senderAccount.wallet} Wallet
+        {`Sign the Transaction with your ${WalletName[senderAccount.wallet]} Wallet`}
```
vilsbole: We want to keep all the configuration in a single place.
Prefer:

 ```
import { WALLETS_CONFIG } from 'v2/config`

...WALLETS_CONFIG[senderAccount.wallet].name
```

File: common/v2/types/walletId.ts [hunk 2]
```diff
@@ -9,3 +9,15 @@ export enum WalletId {
   MNEMONIC_PHRASE = 'MNEMONIC_PHRASE',
   VIEW_ONLY = 'VIEW_ONLY'
 }
+
+export enum WalletName {
```
vilsbole: This is already defined with `WALLETS_CONFIG[walletId].name`

common/v2/features/SendAssets/components/TransactionReceipt.tsx
```diff
@@ -173,9 +173,11 @@ export default function TransactionReceipt({ txReceipt, txConfig }: IStepCompone
       <Link to="/dashboard">
         <Button className="TransactionReceipt-back">Back to Dashboard</Button>
       </Link>
-      <Button secondary={true} className="TransactionReceipt-another">
-        Send Another Transaction
-      </Button>
+      <Link to="/send" onClick={() => window.location.reload()}>
```
vilsbole: Why the `.reload()` ?! This will reset the entire app...

blurpesec: Is there a better way to handle restarting the flow? If you don't add the reload, it doesn't actually route anywhere because you're already at the same location

PR #2773, opened 2019-09-30
Added filter for removed accounts in dashboard accounts list

blurpesec:
#### Description

Gets rid of the "Showing 15 out of 14 accounts" bug with account dropdown on dashboard. Also truncates addresses and adds labels.

#### Changes

- Filter dashboardAccounts from settings when there is no account the uuid is assigned to.
- Truncates addresses in account dropdown.
- Adds labels to account dropdown.

### DIFF
--- a/common/v2/features/Dashboard/components/WalletBreakdown/AccountDropdown.tsx
+++ b/common/v2/features/Dashboard/components/WalletBreakdown/AccountDropdown.tsx
@@ -4,7 +4,7 @@ import styled, { StyledFunction } from 'styled-components';
 
 import { translateRaw } from 'translations';
 import { Checkbox } from 'v2/components';
-import { useOnClickOutside } from 'v2/utils';
+import { useOnClickOutside, truncate } from 'v2/utils';
 import { getLabelByAccount, AddressBookContext } from 'v2/services/Store';
 import { COLORS } from 'v2/theme';
 import { ExtendedAccount, ExtendedAddressBook } from 'v2/types';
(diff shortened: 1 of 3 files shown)


common/v2/features/Dashboard/components/WalletBreakdown/AccountDropdown.tsx
```diff
@@ -113,7 +113,13 @@ const AccountDropdown = ({ accounts = [], selected = [], onSubmit }: AccountDrop
...
   // Only update our draft if the prop changed.
   // https://reactjs.org/docs/hooks-effect.html#tip-optimizing-performance-by-skipping-effects
-  useEffect(() => setDraftSelected(selected), [selected]);
+  useEffect(() => {
+    // Filters out dashboard account uuids that have been deleted.
+    const filteredSelected = selected.filter(selectedAccountUUID => {
+      return accounts.find(account => account.uuid === selectedAccountUUID) ? true : false;
+    });
+    setDraftSelected(filteredSelected);
+  }, [selected, accounts]);
```
vilsbole: Yes, it needs to be triggered on accounts update.

common/v2/features/Dashboard/components/WalletBreakdown/AccountDropdown.tsx
```diff
@@ -113,7 +113,13 @@ const AccountDropdown = ({ accounts = [], selected = [], onSubmit }: AccountDrop
 
   // Only update our draft if the prop changed.
   // https://reactjs.org/docs/hooks-effect.html#tip-optimizing-performance-by-skipping-effects
-  useEffect(() => setDraftSelected(selected), [selected]);
+  useEffect(() => {
+    // Filters out dashboard account uuids that have been deleted.
+    const filteredSelected = selected.filter(selectedAccountUUID => {
```
vilsbole: This is hidding a more important bug.
- AccountDropdown is passed `accounts` & `selected`.
- it displays the length of both.
- `selected` corresponds to `Store/SettingsProvider` and when we delete an account we don't update our `settings`.

To solve the bug we could either enforce the coupling between settings and accounts or change the data structure. 
In any case I would prefer to keep the bug visible then to hide it behind a fix.

PR #2779, opened 2019-10-01
Update UI in Core

Sonicrida:
Updates UI to version 0.18.3


vilsbole:
@Sonicrida Nice!
Though looking at the build, it seems to remove the copy icon. 
Can you take a look.

Sonicrida:
Fixed!

PR #2782, opened 2019-10-01
return to top of layout on route change

vilsbole:
[ch3150](https://app.clubhouse.io/mycrypto/story/3150/scroll-to-top-on-route-change)

important:
'scrollTo' will fail if the main div has an overflow y or x (?!)
so we remove it from the #app container


Sonicrida:
This looks good. Thank you for linking the resource in your comments as well!

PR #2784, opened 2019-10-01
Save rates to LC for persistence

blurpesec:


common/v2/services/RatesProvider.tsx
```diff
@@ -42,7 +55,7 @@ export function RatesProvider({ children }: { children: React.ReactNode }) {
 
     worker.start();
     return terminateWorker; // make sure we terminate the previous worker on teardown.
-  }, [rawAccounts]); // only update if an account has been added or removed from LocalStorage.
+  }, [rawAccounts, Object.keys(rates)]); // only update if an account has been added or removed from LocalStorage.
```
vilsbole: `rawAccounts` should be enough to trigger the update of rates.

PR #2786, opened 2019-10-02
Added logic to handle different web3 providers/injectors

409H:
## Description

* Shows the logo of the web3 provider/injector so it is not static set to MetaMask on the "Connect and Unlock with xxx" screen (the screen after "How would you like to access your assets?")
  * Supports; MetaMask and Trust

Clubhouse story 3137

## Changes

* Added translation constants for Trust Wallet
* Modified `common/v2/components/WalletUnlock/Web3Provider.tsx` to look at the `window.web3.currentProvider` to determine the provider to show the correct name and icon
* Added `TRUST` as a `WalletID` which is just a duplicate of `WalletID: METAMASK`

## Steps to Test

1. Steps
2. to
3. test

## Quality Assurance


common/v2/components/WalletUnlock/Web3Provider.tsx
```diff
@@ -90,6 +88,48 @@ class Web3ProviderDecrypt extends Component<Props, State> {
       this.setState({ ...this.state, web3Unlocked: false });
     }
   }
+
+  private getWeb3Footer(strWeb3Name: string) {
+    switch (strWeb3Name.toUpperCase()) {
+      default:
+      case 'METAMASK':
```
vilsbole: We try to avoid 'magic' strings.
Prefer using the enum and `WalletId.METAMASK`

common/v2/components/WalletUnlock/Web3Provider.tsx
```diff
@@ -36,37 +44,27 @@ class Web3ProviderDecrypt extends Component<Props, State> {
     const { web3Unlocked } = this.state;
```
vilsbole: You can use destructuring here and shorten the future accesses to state.
```
const { web3Unlocked, web3ProviderSettings: provider } = this.state;

....translate(`ADD_ACCOUNT_${provider.name}_TITLE`)
```

common/v2/components/WalletUnlock/Web3Provider.tsx
```diff
@@ -90,6 +88,48 @@ class Web3ProviderDecrypt extends Component<Props, State> {
...
+                content={translate(
+                  `ADD_ACCOUNT_${this.state.web3ProviderSettings.name}_FOOTER_LINK`
+                )}
+                href="https://chrome.google.com/webstore/detail/metamask/nkbihfbeogaeaoehlefnkodbefgpgknn?hl=en"
+              />
+            </div>
+            <div>
+              <NewTabLink
+                content={translate('ADD_ACCOUNT_METAMASK_HELP')}
+                href={`${KB_URL}/how-to/migrating/moving-from-mycrypto-to-metamask`}
```
vilsbole: I know it's copied from the previous, but it's a good time to start using `WALLETS_CONFIG` here.
```
import { WALLETS_CONFIG } from 'v2/config'
const WalletConfig = WALLETS_CONFIG[provider.name]
...
href={WalletConfig.helpLink}
```

common/v2/features/SendAssets/components/SignTransaction.tsx
```diff
@@ -18,6 +18,7 @@ type SigningComponents = {
 const SigningComponents: SigningComponents = {
   [WalletId.PRIVATE_KEY]: SignTransactionPrivateKey,
   [WalletId.METAMASK]: SignTransactionMetaMask,
+  [WalletId.TRUST]: SignTransactionMetaMask,
```
vilsbole: How about renaming the file to something like `SignTransactionWeb3Provider`?

common/v2/components/WalletUnlock/Web3Provider.tsx
```diff
@@ -90,6 +88,48 @@ class Web3ProviderDecrypt extends Component<Props, State> {
...
+  }
+
+  private getWeb3Provider() {
+    if (window.web3) {
+      switch (true) {
+        default:
+        case window.web3.currentProvider.isMetaMask:
+          return { name: 'METAMASK', icon: MetamaskSVG };
+        case window.web3.currentProvider.isTrust:
+          return { name: 'TRUST', icon: TrustWalletSVG };
```
vilsbole: Let's since the icon and the name are already defined in `WALLETS_CONFIG` let's use it for here as well.

```
case window.web3.currentProvider.isTrust:
          return WALLETS_CONFIG[WalletId.TRUST]
```

common/v2/utils/web3.ts
```diff
@@ -20,6 +20,11 @@ const WEB3_CONFIGS: {
     lid: 'X_METAMASK',
     icon: MetamaskIcon
   },
+  InpageBridge: {
+    // MetaMask Mobile has the web3.currentProvider.constructor.name as InpageBridge
+    lid: 'X_METAMASK',
+    icon: MetamaskIcon
+  },
```
vilsbole: Good to know!

common/v2/services/WalletService/walletService.ts
```diff
@@ -27,6 +27,7 @@ import { unlockWeb3 } from './web3';
 export const WalletFactory = (walletId: WalletId): WalletService | any => {
   switch (walletId) {
     case WalletId.METAMASK:
+    case WalletId.TRUST:
```
vilsbole: It's strange that we don't add the case for the other web3 wallets here as well.

PR #2787, opened 2019-10-03
Add account deletion confirmation to table on desktop

Sonicrida:
## Description

* Added a table row overlay for account deletion confirmation step

~Will prob need to be cleaned up a bit later~

[ch2593]

### Mobile Overlay
Screenshot 2019-11-14 at 20 41 17

### Desktop Overlay
Screenshot 2019-11-14 at 20 47 02

### DIFF
--- a/common/v2/components/AccountList.tsx
+++ b/common/v2/components/AccountList.tsx
@@ -91,7 +91,6 @@ const TableContainer = styled.div`
 `;
 
 const TableOverlay = styled.div`
-  max-width: 674px;
   height: 67px;
   display: flex;
   flex-direction: row;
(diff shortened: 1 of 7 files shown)


vilsbole:
1. Desktop view layover has an unwanted right-margin.
Screenshot 2019-10-03 at 15 50 53

2. No confirmation step for mobile

common/v2/components/AccountList.tsx
```diff
@@ -170,6 +214,34 @@ function buildAccountTable(
 
   return {
     head: deletable ? [...columns, translateRaw('ACCOUNT_LIST_DELETE')] : columns,
+    overlay:
+      overlayRows && overlayRows[0] !== undefined ? (
+        <TableOverlay>
```
vilsbole: It seems that this code is duplicated in AccountList and AddressBook.
It seems like the right moment to create a `v2/component` for the overlay.

Sonicrida: Done!

PR #2788, opened 2019-10-03
Implement 3box into GAU (CH-3065)

MrLuit:
Story details: https://app.clubhouse.io/mycrypto/story/3065


MrLuit:
3box addresses to test:

- 0x6b44862103b8a45f5ec701b69ec28a5d6d304950
- 0xbaebb7d18f8b16b0a970fda91f1efa626d67423e
- 0xe2d128323cf7560a6e7a82726d7b425aedc7a556

notes:
- no styling done yet
- more information from 3box can be implemented (right now it only has name, image, and location)
- data from 3box is not validated, strange behaviour can be expected for really long names and such

common/v2/features/Settings/components/AddressBook.tsx
```diff
@@ -1,14 +1,8 @@
 import React, { useState } from 'react';
 import styled from 'styled-components';
-import { Icon, Copyable, Identicon, Button } from '@mycrypto/ui';
+import { Icon, Copyable, Typography, Button } from '@mycrypto/ui';
```
vilsbole: Prefer Typography from `common/v2/components/`.
The ratios in ui are messed up

common/v2/components/Profile/Profile.tsx
```diff
@@ -0,0 +1,73 @@
...
+interface Props {
+  address: string;
+  label?: string;
+}
+
+interface State {
+  profile: ThreeBoxProfile;
+}
+
+export default class Profile extends Component<Props, State> {
```
vilsbole: Prefer Function components over class components.
This facilitates reusing hooks.

common/v2/components/Profile/Profile.tsx
```diff
@@ -0,0 +1,73 @@
...
+  }
+
+  public componentDidMount() {
+    const { address } = this.props;
+    get3BoxProfile(address).then(result => {
+      this.setState({
+        profile: result
+      });
+    });
+  }
```
vilsbole: When calling setState after an async call, make sure to cancel the request when the component un-mounts.
Otherwise we occasionally receive a cannot setState on unmounted Component error.
https://www.robinwieruch.de/react-warning-cant-call-setstate-on-an-unmounted-component
In the app we currently handle this case with common/v2/vendor/react-use/usePromise.ts, or any other https://github.com/streamich/react-use that is deemed useful.

See the use of `usePromise` in https://github.com/MyCryptoHQ/MyCrypto/pull/2947#discussion_r366251954

common/v2/services/3Box/api.ts
```diff
@@ -0,0 +1,19 @@
...
+    validateStatus: () => true
+  });
+
+  if (result.status === 200) {
+    const { data } = result;
+    return new Profile(address, data);
+  } else {
+    return false;
+  }
+};
```
vilsbole: Prefer using `common/v2/services/ApiService/ApiService.ts`
We want to use a single instance of axios for all our ajax calls. 
It will become useful when we implement offline handling and retries.
(happy to discuss about this choice)

yarn.lock
```diff
@@ -1829,6 +1829,11 @@
   dependencies:
     "@types/react" "*"
 
+"@types/react-inlinesvg@^0.8.1":
+  version "0.8.1"
+  resolved "https://registry.yarnpkg.com/@types/react-inlinesvg/-/react-inlinesvg-0.8.1.tgz#c3b9ab683c356a81869b3294dbd3e4f0f191aa4a"
+  integrity sha512-pONmeCNqot4diXLhxb+jajDWX57rH/JmMu9H4NQWNMl5WUHAHmcrOfDfB05IgYBfRUX1Cvyjl9EZzKHxSxZpAA==
+
```
vilsbole: `package.json` is unchanged this seems to be an error.

PR #2789, opened 2019-10-04
Make homepage full width

vilsbole:
Fullwidth layout for Homepage


common/v2/features/Home/Home.tsx
```diff
@@ -38,20 +37,7 @@ const Section = styled.section`
...
-    background-color: #fff;
-  }
-`;
-
-const BottomSection = styled(Section)`
-  @media (max-width: ${SCREEN_SM}) {
-    flex-direction: column;
-    align-items: center;
-  }
-`;
```
amacar: This section needs `flex-direction: column` to display ok on smaller screens.

without:
image

with:
image

vilsbole: Fixed. Thanks!

PR #2797, opened 2019-10-10
Remove incompleted links for beta

vilsbole:
Deactivate incomplete links:
- Contract interaction
- Swap


blurpesec:
Only removes from Nav bar. Doesn't remove links from action buttons on dashboard.

PR #2801, opened 2019-10-14
Fixed inactivity timer

409H:
## Description

The inactivity timer view will show the correct minutes of inactivity instead of static "5 minutes"

### DIFF
--- a/common/v2/features/ScreenLock/ScreenLockLocking.tsx
+++ b/common/v2/features/ScreenLock/ScreenLockLocking.tsx
@@ -1,13 +1,12 @@
-import React, { Component, useContext } from 'react';
+import React, { useContext } from 'react';
+import moment from 'moment';
 import { Button } from '@mycrypto/ui';
 import styled from 'styled-components';
 
 import translate, { translateRaw } from 'translations';
 import { AnalyticsService, ANALYTICS_CATEGORIES } from 'v2/services/ApiService';
 
-import {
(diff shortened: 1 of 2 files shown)


common/v2/features/ScreenLock/ScreenLockLocking.tsx
```diff
@@ -86,7 +93,7 @@ export default class ScreenLockLocking extends Component<LockScreenProps> {
         <ContentWrapper>
           <Title>{translate('SCREEN_LOCK_LOCKING_HEADING')}</Title>
           <Description>
-            {translate('SCREEN_LOCK_LOCKING_DESCRIPTION')}{' '}
+            {translate(`SCREEN_LOCK_LOCKING_DESCRIPTION`, { $inactive_time: ((settings.inactivityTimer/1000)/60).toString() })}{' '}
```
vilsbole: use `moment.duration().format` or equivalent: https://stackoverflow.com/questions/13262621/how-do-i-use-format-on-a-moment-js-duration

vilsbole:
Is this still WIP

FrederikBolding:
Fixed the formatting issue, rebased and changed the component to be functional so we could use `SettingsContext`.

Tested and it works. Should be ready for review!

FrederikBolding:
Btw, this one should be working. For some reason the AWS CodeBuild thing failed initially, but not when retrying.

PR #2807, opened 2019-10-15
Fixed isNaN asset values in receiveAssets flow

blurpesec:
### Description

Replaces AccountsContext with StoreContext in order to import balances correctly.


common/v2/components/AccountDropdown.tsx
```diff
@@ -19,14 +19,12 @@ interface IAccountDropdownProps {
 
 function AccountDropdown({ accounts, name, value, onSelect }: IAccountDropdownProps) {
   const { getAccountLabel } = useContext(AddressBookContext);
-  const relevantAccounts: StoreAccount[] = accounts
-    .filter(account => account.wallet !== WalletId.VIEW_ONLY)
```
vilsbole: ~~Shouldn't we keep the filter to avoid a VIEW_ONLY (an account that can't sign a message) from being selected in a send or sign flow?~~

vilsbole: ~~Or since we share it with receive, maybe adding a boolean prop to the component to choose if we filter or not?~~

vilsbole: Filtering the accounts before passing them is even cleaner. Nice!

PR #2808, opened 2019-10-15
Added no-accounts redirect to ReceiveAssets Panel

blurpesec:
### Description
Added no-accounts redirection to the ReceiveAssets Panel when there are no accounts present.

### DIFF
--- a/common/v2/features/ReceiveAssets/ReceiveAssets.tsx
+++ b/common/v2/features/ReceiveAssets/ReceiveAssets.tsx
@@ -2,7 +2,7 @@ import React, { useContext, useState } from 'react';
 import { Formik, Form, Field, FieldProps, FormikProps } from 'formik';
 import noop from 'lodash/noop';
 import { Copyable, Heading, Input, Tooltip } from '@mycrypto/ui';
-import { withRouter, RouteComponentProps } from 'react-router-dom';
+import { withRouter, RouteComponentProps, Redirect } from 'react-router-dom';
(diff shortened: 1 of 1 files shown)


common/v2/features/ReceiveAssets/ReceiveAssets.tsx
```diff
@@ -116,6 +116,11 @@ const ErrorMessage = styled.span`
 
 export function ReceiveAssets({ history }: RouteComponentProps<{}>) {
   const { accounts } = useContext(AccountContext);
+  const shouldRedirect = accounts === undefined || accounts === null || accounts.length === 0;
```
amacar: Can be shortened to: `const shouldRedirect = !accounts || !accounts.length;`

common/v2/features/ReceiveAssets/ReceiveAssets.tsx
```diff
@@ -116,6 +116,11 @@ const ErrorMessage = styled.span`
...
+  if (shouldRedirect) {
+    return <Redirect to="/no-accounts" />;
```
amacar: Prefer path from https://github.com/MyCryptoHQ/MyCrypto/blob/gau/common/v2/config/routePaths.ts#L125. For example ROUTE_PATHS.NO_ACCOUNTS.path instead of hardcoded one.

blurpesec:
Closing in favor of #2813

PR #2809, opened 2019-10-15
Fixed back button in Send Assets panel

FrederikBolding:
## Description

Fixed back button in send assets panel


common/v2/features/SendAssets/SendAssets.tsx
```diff
@@ -63,9 +65,11 @@ function SendAssets() {
   const goToPrevStep = () => setStep(Math.max(0, step - 1));
   const goToFirstStep = () => setStep(0);
 
+  const goBack = () => (step === 0 ? history.push('/') : goToPrevStep());
```
amacar: Prefer path from https://github.com/MyCryptoHQ/MyCrypto/blob/gau/common/v2/config/routePaths.ts#L125. For example `ROUTE_PATHS.ROOT.path` or `ROUTE_PATHS.DASHBOARD.path` instead of hardcoded path.

FrederikBolding: Thanks! I just did the same as in ReceiveAssets, but I see your point.

Changed it in both of them.

amacar: Awesome, thanks. We are trying to use paths from config for future code; unfortunately there is still a lot of old code which should be changed eventually.

PR #2810, opened 2019-10-15
Fixed no accounts redirect after adding an account

FrederikBolding:
## Description

Fixed no accounts redirect after adding an account


FrederikBolding:
Closed in favor of #2813

PR #2817, opened 2019-10-18
Refactor Webpack configuration and scripts

Mrtenz:
- Removed old scripts (freezer, Jenkins related stuff)
- Removed `derivation-checker` configuration
- Updated all Webpack scripts to use `webpack-merge` with base development and production configurations
- Updated most Webpack related dependencies


Mrtenz:
[ch3254]

yarn.lock
```diff
@@ -1769,6 +2149,162 @@
   resolved "https://registry.yarnpkg.com/@types/zxcvbn/-/zxcvbn-4.4.0.tgz#fbc1d941cc6d9d37d18405c513ba6b294f89b609"
   integrity sha512-GQLOT+SN20a+AI51y3fAimhyTF4Y0RG+YP3gf91OibIZ7CJmPFgoZi+ZR5a+vRbS01LbQosITWum4ATmJ1Z6Pg==
 
+"@webassemblyjs/ast@1.8.5":
```
vilsbole: Would you know why we're installing `webassemblyjs?`

Mrtenz:
```
=> Found "@webassemblyjs/ast@1.8.5"
info Reasons this module exists
   - "webpack" depends on it
   - Hoisted from "webpack#@webassemblyjs#ast"
   - Hoisted from "webpack#@webassemblyjs#wasm-edit#@webassemblyjs#ast"
   - Hoisted from "webpack#@webassemblyjs#wasm-parser#@webassemblyjs#ast"
   - Hoisted from "webpack#@webassemblyjs#helper-module-context#@webassemblyjs#ast"
[...]
```

I guess it's part of some WASM bundling feature of Webpack.

FrederikBolding:
Before:

| start | build | tscheck | test |
|-------|-------|---------|------|
| 104   | 108   | 16      | 120  |
| 8     | 113   | 13      | 30   |
| 5     | 97    | 13      | 28   |
| 5     | 96    | 13      | 29   |

After:

| start | build | tscheck | test |
|-------|-------|---------|------|
| 68    | 89    | 12      | 64   |
| 7     | 59    | 13      | 28   |
| 6     | 44    | 12      | 28   |
| 5     | 44    | 12      | 28   |

Did a few benchmarking tests. In general, they both seem to use caching very well (after multiple runs). But the newer configuration is still definitely faster, which is clearest to see when running builds.

I didn't test stability, but that is probably something we should look at as well! The previous configuration had crashing issues (at least for me). But that might also be resolved when we remove more legacy code 😄

Mrtenz:
@vilsbole AFAIK we don't use Jenkins, we can still build for all platforms using macOS. Drawbridge doesn't use it either.

@FrederikBolding Thanks for the benchmarks! Crashing issues may be resolved just by updating the dependencies, but I didn't specifically test for this.

FrederikBolding:
@Mrtenz Not sure whether updating dependencies will help or not. The crashes seemed memory-leaky and usually happens if you rebuild a bunch of times while developing a feature. @blurpesec and I discussed this at the meeting last Tuesday.

yarn.lock
vilsbole: 👍 ok

PR #2821, opened 2019-10-22
Rewrote parts of README

blurpesec:


README.md
```diff
@@ -17,11 +15,11 @@ You can find how to do this on our [support document](https://support.mycrypto.c
 * Python 2.7.X\*\*\*
```
vilsbole: Do we still need Node and Python after removing the derivation path checker?

README.md
```diff
@@ -46,154 +44,3 @@ To disable it you can your settings in chrome: chrome://flags/#allow-insecure-lo
...
-```
-
-#### Integration Tests:
-
-```bash
-# run integration tests with Jest
-yarn test:int
-```
-
-#### Address Derivation Checker:
```
vilsbole: Open question if we really want to remove the derivation checker?

README.md
```diff
@@ -46,154 +44,3 @@ To disable it you can your settings in chrome: chrome://flags/#allow-insecure-lo
...
-├── spec - Jest unit tests, mirror's common's structure
-├── static - Files that don't get compiled, just moved to build
-└── webpack_config - Webpack configuration
-```
-
-## Typescript Resources:
-- https://basarat.gitbooks.io/typescript/
-- https://blog.mariusschulz.com/series/typescript-evolution
-
-## More information is available on the [Wiki Pages](https://github.com/MyCryptoHQ/MyCrypto/wiki)
```
vilsbole: The links to the wiki page seem useful to keep

README.md
tayvano: Yes!

Well

Remove it from this repo.

The only thing in this repo should be a super simple function that outputs a testfile with like 10k private keys and their derived addresses using the methods and libs we use in our website/app

We can then use that file as input in another repo.

tayvano: Honestly most issues occur with a 1/128 probability so 10k is serious overkill. 2k would be super safe, 1k would be safe. 200 would probablistically get 1-2 wrong addresses and be safe.

tayvano: Also, realistically we should run this test like once a quarter or whenever libs are updated or derivation logic is changed (which should be at least that rare)

README.md
blurpesec: Im going to leave this in for now until we do an analysis of our dependencies to make sure we don't use node-gyp anywhere, since that was what used python 2.7.x

README.md
blurpesec: Will keep in, but will probably need to update them.

README.md
vilsbole: I created a card to move the test into another repo
https://app.clubhouse.io/mycrypto/story/3489/create-test-for-derivation-checker

PR #2822, opened 2019-10-22
Update EGEM Network

r2clan:
Quick update to egem network for url.


vilsbole:
@r2clan Thanks for your PR! Could you just email `support@mycrypto.com` from the `egem.io` domain so we can confirm the changes?

PR #2827, opened 2019-10-27
Fixed missing variable on Web3 install translation

FrederikBolding:
## Description

Fixed missing variable on Web3 install translation

### DIFF
--- a/common/v2/components/WalletUnlock/Web3ProviderInstall.tsx
+++ b/common/v2/components/WalletUnlock/Web3ProviderInstall.tsx
@@ -16,7 +16,9 @@ import './Web3ProviderInstall.scss';
 function InstallTrunk() {
   return (
     <div className="Panel">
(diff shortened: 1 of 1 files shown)


common/v2/components/WalletUnlock/Web3ProviderInstall.tsx
```diff
@@ -16,7 +16,9 @@ import './Web3ProviderInstall.scss';
 function InstallTrunk() {
   return (
     <div className="Panel">
-      <div className="Panel-title">{translate('ADD_ACCOUNT_WEB3_INSTALL_TITLE')}</div>
+      <div className="Panel-title">
+        {translate('ADD_ACCOUNT_WEB3_INSTALL_TITLE', { $walletId: 'Trunk' })}
```
Mrtenz: Shouldn't this be "Trust"?

FrederikBolding: Ah, my bad. But the problem is that there are two choices. I'll ask in Slack.

PR #2828, opened 2019-10-27
Updated currency formatting

FrederikBolding:
## Description

- Updated currency formatting to include commas in larger numbers
- Used currency formatting in WalletBreakdown
- Used fiat setting for WalletBreakdown

### DIFF
--- a/common/v2/components/Currency.tsx
+++ b/common/v2/components/Currency.tsx
@@ -15,12 +15,25 @@ interface Props {
   decimals?: number;
   icon?: boolean;
   prefix?: boolean;
+  useTypography?: boolean;
 }
 
-function Currency({ amount, symbol, decimals = 5, icon = false, prefix = false, ...props }: Props) {
+function ConditionalTypography({ condition, children }: { condition: boolean; children?: any }) {
+  return condition ? <Typography>{children}</Typography> : children;
+}
+
+function Currency({
+  amount,
+  symbol,
+  decimals = 5,
+  icon = false,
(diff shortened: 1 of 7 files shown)


common/v2/components/Currency.tsx
```diff
@@ -15,12 +15,25 @@ interface Props {
...
   icon?: boolean;
   prefix?: boolean;
+  useTypography?: boolean;
 }
 
-function Currency({ amount, symbol, decimals = 5, icon = false, prefix = false, ...props }: Props) {
+function ConditionalTypography({ condition, children }: { condition: boolean; children?: any }) {
+  return condition ? <Typography>{children}</Typography> : children;
+}
+
```
vilsbole: Adding a prop `useTypography` creates a strong dependance between the 2 components (it means the caller knows the implementation of the callee).  

Ideally we want every displayed text to be wrapped with a `Typography` component which would handle all our font scaling. The component scales poorly so it's wrapped with `v2/components/Typography`.

Prefer making the desired style modifications / flags to `v2/components/Typography`.

FrederikBolding: So removing `PanelFigureValue` and making `Typography` and thereby `Currency` support props that allow for similar styling?

vilsbole: Yes that is the right direction!

FrederikBolding: Pushed changes, let me know what you think!

PR #2832, opened 2019-10-29
Show chosen asset on AccountDropdown in Send flow

FrederikBolding:
## Description
- Show chosen asset on AccountDropdown in Send flow
- Only show assets that can actually be sent
- Fixed formatting (on previous stuff, oops)


FrederikBolding:
[ch3540]

FrederikBolding:
Fixed! Do we want to delete the `v2/utils/symbolIcon.ts` then?

blurpesec:
> Fixed! Do we want to delete the `v2/utils/symbolIcon.ts` then?

I'd say so.

PR #2841, opened 2019-11-04
Refactor token update script

Mrtenz:
Token script refactor to add deterministic UUIDs to each token, based on chain ID and address.


Mrtenz:
[ch3620]

PR #2850, opened 2019-11-11
Use Cache instead of LocalCache

FrederikBolding:
## Description

- Use `Cache` implementation instead of `LocalCache`
- Use asset uuids generated by parsing library


FrederikBolding:
[ch3727]

FrederikBolding:
Ready for review! - Might need to be squashed before being merged in.

PR #2851, opened 2019-11-12
Add Unlock Protocol detection and notification

blurpesec:
### Description
I wanted to test erc721 interaction and decided to try to do Unlock Protocol detection, and ended up doing the entire integration for determining if the user has one of our Unlock tokens. In order to test it, i added a notification thanking the user for donating.

ToDo: 
- [x] Remove the notification and overwrite an ad slot with a thank-you ad c:

With unlock token:
image

Without unlock token:
image

### DIFF
--- a/common/v2/config/constants.ts
+++ b/common/v2/config/constants.ts
@@ -13,3 +13,5 @@ export const DEFAULT_NETWORK_FOR_FALLBACK = 'ropsten';
 export const DEFAULT_NETWORK: NetworkId = 'Ethereum';
 
 export const DEFAULT_ASSET_DECIMAL = 18;
+
+export const MYCRYPTO_UNLOCK_CONTRACT_ADDRESS = '0x14e81196e60b128527db03d40bdba00710777805';
--- a/common/v2/services/Store/BalanceService.tsx
+++ b/common/v2/services/Store/BalanceService.tsx
@@ -3,7 +3,7 @@ import partition from 'lodash/partition';
 import { bigNumberify, BigNumber } from 'ethers/utils';
 import { default as BN } from 'bignumber.js';
 
-import { ETHSCAN_NETWORKS } from 'v2/config';
+import { ETHSCAN_NETWORKS, MYCRYPTO_UNLOCK_CONTRACT_ADDRESS } from 'v2/config';
 import { TAddress, StoreAccount, StoreAsset, Asset, NodeConfig, Network } from 'v2/types';
 import { ProviderHandler } from 'v2/services/EthService';
 import { FallbackProvider } from 'ethers/providers';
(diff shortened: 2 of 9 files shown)


common/v2/services/Store/BalanceService.tsx
```diff
@@ -136,3 +136,22 @@ export const getAllTokensBalancesOfAccount = async (account: StoreAccount, asset
...
+  } catch (err) {
+    throw new Error(err);
+  }
+};
+
+export const getAccountsUnlockVIPStatus = async (accounts: StoreAccount[]) => {
+  return getAccountsTokenBalance(accounts, MYCRYPTO_UNLOCK_CONTRACT_ADDRESS)
+    .then(unlockStatusBalanceMap =>
+      Object.keys(unlockStatusBalanceMap).filter(accountAddress =>
+        unlockStatusBalanceMap[accountAddress].isGreaterThan(new BN(0))
```
vilsbole: 1. The method name is `...status`. Best to either
    - rename the function to return all the accounts with a valid balance
    - return the `status`: 'pending' | 'valid' etc
    - return a `boolean` if there is an account that is valid.
2. Since we only want to compare the value how about:
   ```
      Object.values(unlockStatusBalanceMap).filter(balance => balance.isGreaterThan(new BN(0)))
   ```
3. Let's add a comment explaining why anything > 0 is valid.

common/v2/services/Store/StoreProvider.tsx
```diff
@@ -51,6 +54,15 @@ export const StoreProvider = ({ children }: { children: React.ReactNode }) => {
...
         if (isArrayEqual(accounts, accountsWithBalances)) return;
+        const accountsToCheck = accounts
+          .filter(account => account.networkId === 'Ethereum')
+          .filter(account => account.wallet !== WalletId.VIEW_ONLY);
+        getAccountsUnlockVIPStatus(accountsToCheck).then((unlockAccounts: string[]) => {
+          if (!unlockAccounts || unlockAccounts.length === 0) {
+            return;
+          }
+          setIsUnlockVerified(true);
+        });
```
vilsbole: Avoid nested `then`'s. It's the equivalent of [callback hell](http://callbackhell.com/)
Prefer Promise chaining:
```
   getAccountsAssetsBalancess(accounts)
       .then(accountsWithBalances => {
           if (isArrayEqual(accounts, accountsWithBalances)) return;
           setAccounts(accountsWithBalances);
           return accountsToCheck = accounts
              .filter(account => account.networkId === 'Ethereum')
              .filter(account => account.wallet !== WalletId.VIEW_ONLY);
       })
       .then(getAccountsUnlockVIPStatus)
       .then(setIsUnlockVerified)
```
or use async/await

vilsbole:
Changed the title to use the active voice
/s Added Add

common/v2/config/constants.ts
```diff
@@ -13,3 +13,5 @@ export const DEFAULT_NETWORK_FOR_FALLBACK = 'ropsten';
 export const DEFAULT_NETWORK: NetworkId = 'Ethereum';
 
 export const DEFAULT_ASSET_DECIMAL = 18;
+
+export const MYCRYPTO_UNLOCK_CONTRACT_ADDRESS = '0x14e81196e60b128527db03d40bdba00710777805';
```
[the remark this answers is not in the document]
blurpesec: We may need to have a static `UNLOCK_PROTOCOL_CONTRACT_ADDRESS` in the future, if we decide to do native integration of unlock token buying, so i figured `MYCRYPTO_UNLOCK` would be better, because it more describes our current contract address instance.

common/v2/services/Store/BalanceService.tsx
blurpesec: 2. In the future we may want the addresses for some reason or another, so i extracted this into a function that returns success `boolean`

PR #2854, opened 2019-11-13
Add initial implementation of tx history table

blurpesec:
[ch1675] 
### Description

Re-enable recent txs sent using mycrypto and handle styling bugs / ux bugs / issues associated with this.
- [x] Re-enable table outside of dev mode.
- [x] No-Txs view.
- [x] Don't show pending tx grouping if there are no pending txs. Do the same with completed tx groupings.
- [x] Handle polling for updates to pending txs.
- [x] Add icons for tx history entries.
- [x] Remove polling from txReceipt component and make it fetch status globally ? or separate it into it's own standalone component?

image


common/v2/features/Dashboard/Dashboard.tsx
```diff
@@ -39,7 +39,7 @@ const EmptyTile = styled.div`
 `;
 
 export default function Dashboard() {
-  const { isActive: isDevToolsActive, displayRecentTransactionList } = useDevTools();
```
vilsbole: Please remove it from DevTools as well.
It's good to avoid leaving dead code lying around.

vilsbole:
Changed title to use active voice
/s Added Add

common/v2/services/Store/StoreProvider.tsx
```diff
@@ -59,6 +70,67 @@ export const StoreProvider = ({ children }: { children: React.ReactNode }) => {
...
+    setPendingTransactions(
+      accounts
+        .map(account =>
+          account.transactions.map(tx => {
+            return { ...tx, network: account.network };
+          })
+        )
+        .flatMap(tx => tx)
+        .filter((txObject: ITxReceipt) => txObject.stage === ITxStatus.PENDING) || []
+    );
```
vilsbole: 1. Here you have a great opportunity to extract a pure function. eg. `setPendingTransactions(pendingTxs)`
By keeping the pure functions separate from side effects it facilitates testing (of the pure function) and facilitates maintainability. 

2. `.map().flatmap(tx => tx)`  is equivalent to `flatmap()`.

For example:
```
  const txIsPending = ({ stage }) => stage === ITxStatus.PENDING
  const pendingTxs = accounts
     .filter(txIsPending) // reduces the size of the list to append network to.
     .flatMap(({transactions: txs, network}) => txs.map(tx => ({...tx, network}))

  
   setPendingTransactions(pendingTxs || [])

```

common/v2/services/Store/StoreProvider.tsx
```diff
@@ -59,6 +70,67 @@ export const StoreProvider = ({ children }: { children: React.ReactNode }) => {
...
+            return;
+          }
+
+          // Get block tx success/fail and timestamp for block number, then overwrite existing tx in account.
+          Promise.all([
+            getTxStatus(provider, receipt.hash),
+            getTimestampFromBlockNum(receipt.blockNumber, provider)
+          ]).then(values => {
+            const txStatus = values[0];
+            const txTimestamp = values[1];
```
vilsbole: Prefer Array destructuring over access by index
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment

common/v2/services/Store/StoreProvider.tsx
```diff
@@ -59,6 +70,67 @@ export const StoreProvider = ({ children }: { children: React.ReactNode }) => {
...
+        const provider = new ProviderHandler(network);
+
+        provider.getTransactionByHash(pendingTransactionObj.hash).then(transactionReceipt => {
+          if (!transactionReceipt) {
+            return;
+          }
+          const receipt = fromTxReceiptObj(transactionReceipt);
+          if (!receipt) {
+            return;
+          }
```
vilsbole: Why would `fromTxReceiptObj` fail?
I thought it was a pure function behaving as a transducer

common/v2/services/Store/StoreProvider.tsx
```diff
@@ -59,6 +70,67 @@ export const StoreProvider = ({ children }: { children: React.ReactNode }) => {
...
+          // Get block tx success/fail and timestamp for block number, then overwrite existing tx in account.
+          Promise.all([
+            getTxStatus(provider, receipt.hash),
+            getTimestampFromBlockNum(receipt.blockNumber, provider)
+          ]).then(values => {
+            const txStatus = values[0];
+            const txTimestamp = values[1];
+            if (!txStatus || !txTimestamp) {
+              return;
+            }
```
vilsbole: Since we are inside a Promise.resolve, both values will exist

common/v2/services/Store/StoreProvider.tsx
```diff
@@ -59,6 +70,67 @@ export const StoreProvider = ({ children }: { children: React.ReactNode }) => {
...
+  }, [accounts]);
+
+  // A pending transaction is detected.
+  useEffect(() => {
+    if (pendingTransactions.length <= 0) {
+      return;
+    }
+    // This interval is used to poll for status of txs.
+    const blockNumInterval = setInterval(() => {
+      pendingTransactions.forEach((pendingTransactionObj: any) => {
```
vilsbole: This is a good opportunity to apply the aforementioned coding principles such as succinct naming, destructuring, separation of concerns. To go further you can have fun with a Promise returning [pipe](https://ramdajs.com/0.19.1/docs/#pipeP)!

common/v2/features/Dashboard/components/RecentTransactionList.tsx
```diff
@@ -1,79 +1,98 @@
...
-        {' '}
-        {/* TODO - fix this.*/}
-        <img src={newWindowIcon} alt="View more information about this transaction" />
-      </Link>
-    ]);
+    collection.map(
+      ({ timestamp, label, hash, stage, from, to, amount, asset, network }: ITxReceipt) => [
+        <TransactionLabel
+          key={0}
+          image="https://placehold.it/45x45"
```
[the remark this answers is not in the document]
blurpesec: This will be replaced by src when i get them from Avani.

common/v2/features/Dashboard/components/RecentTransactionList.tsx
```diff
@@ -1,79 +1,98 @@
...
 import TransactionLabel from './TransactionLabel';
 import './RecentTransactionList.scss';
 
 import newWindowIcon from 'common/assets/images/icn-new-window.svg';
-import { truncate } from 'v2/utils';
-import { ExtendedAccount, AddressBook } from 'v2/types';
+import { truncate, convertToFiat } from 'v2/utils';
+import { ExtendedAccount, ITxReceipt, TTicker } from 'v2/types';
+import { RatesContext, AddressBookContext, getLabelByAddressAndNetwork } from 'v2/services';
+import { ITxStatus } from 'v2/components/TransactionFlow/TransactionReceipt';
```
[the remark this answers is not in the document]
blurpesec: Nice catch.

common/v2/services/Store/StoreProvider.tsx
blurpesec: If its unable to fetch the network given tx chain id, then it returns undefined.

common/v2/features/Dashboard/components/RecentTransactionList.tsx
```diff
@@ -1,79 +1,98 @@
...
+        <Amount
+          key={3}
+          assetValue={`${parseFloat(amount).toFixed(6)} ${asset.ticker}`}
+          fiatValue={`$${convertToFiat(
+            parseFloat(amount),
+            getRate(asset.ticker as TTicker)
+          ).toFixed(2)}
+      `}
+        />,
+        <NewTabLink key={4} href={`https://etherscan.io/tx/${hash}`}>
```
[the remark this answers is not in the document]
blurpesec: Fixing this involved re-introducing blockExplorer and tokenExplorer variables to networks in LocalStorage.

common/v2/features/Dashboard/components/RecentTransactionList.tsx
```diff
@@ -85,11 +104,15 @@ export default function RecentTransactionList({
...
       heading="Recent Transactions"
-      headingRight="Export"
-      actionLink="/dashboard/recent-transactions"
+      //headingRight="Export"
+      //actionLink="/dashboard/recent-transactions"
       className={`RecentTransactionsList ${className}`}
     >
-      <CollapsibleTable breakpoint={1000} {...recentTransactionsTable} />
+      {filteredGroups.length >= 1 ? (
+        <CollapsibleTable breakpoint={1000} {...recentTransactionsTable} />
```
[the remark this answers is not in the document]
blurpesec: @vilsbole. Is this intended to be red? https://github.com/MyCryptoHQ/MyCrypto/blob/master/common/v2/components/CollapsibleTable.tsx#L113 It messes up the mojo of the collapsible table in mobile. Im going to go ahead and remove it for now until i find a reason why it needs to remain in lol.

image

common/v2/services/Store/StoreProvider.tsx
blurpesec: Values return undefined if the data is unavailable or lookup otherwise fails.

common/v2/services/EthService/utils/providerWrappers.ts
```diff
@@ -38,3 +40,10 @@ export const getTransactionReceiptFromHash = async (
     .catch(_ => {
       return undefined;
     });
+
+export const getTxStatus = (providerInstance: ProviderHandler, txHash: string) =>
+  getTransactionReceiptFromHash(txHash, providerInstance).then(({ status }: any) => {
+    // Status will return 0 or 1 which is used to indicate success or failure of tx
+    if (![0, 1].includes(status)) return;
```
vilsbole: What does it mean when `status` is undefined or different from 0 and 1? 
Can we include it in the comment plz?

common/v2/features/Dashboard/components/RecentTransactionList.tsx
```diff
@@ -1,79 +1,206 @@
...
+export const getTxIcon = (type: ITxType) => {
+  switch (type) {
+    case ITxType.TRANSFER:
+      return transfer;
+    case ITxType.INBOUND:
+      return inbound;
+    case ITxType.OUTBOUND:
+      return outbound;
+  }
+};
```
vilsbole: In a effort to simplify the code, how about we create a single config object for TxType:
```
{
   [ITxType.TRANSFER]: {
      color: ...,
      icon:  require(...),
      text: <Translate const>  
   }
}
```
So we can reduce the amount of switches by simply calling `TxType[ITxType.TRANSFER].icon`

common/v2/features/Dashboard/components/TransactionLabel.tsx
```diff
@@ -19,15 +26,39 @@ const capitalize = (word: string): string =>
...
+const getColor = (txStatus: ITxStatus) => {
+  switch (txStatus) {
+    case ITxStatus.FAILED:
+      return '#F05424';
+    case ITxStatus.SUCCESS:
+      return '#75b433';
+    case ITxStatus.PENDING:
+      return '#424242';
+  }
+};
```
vilsbole: Prefer shared config instead.

common/v2/features/Dashboard/components/RecentTransactionList.tsx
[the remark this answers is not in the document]
vilsbole: I'm sorry I don't know what the designs says. I think you're right removing it.

common/v2/features/Dashboard/components/RecentTransactionList.tsx
```diff
@@ -1,79 +1,206 @@
...
+    case ITxType.TRANSFER:
+      return translateRaw('RECENT_TX_LIST_LABEL_TRANSFERRED', {
+        $ticker: asset.ticker || 'Unknown'
+      });
+    case ITxType.OUTBOUND:
+      return translateRaw('RECENT_TX_LIST_LABEL_SENT', { $ticker: asset.ticker || 'Unknown' });
+    case ITxType.INBOUND:
+      return translateRaw('RECENT_TX_LIST_LABEL_RECEIVED', { $ticker: asset.ticker || 'Unknown' });
+  }
+};
```
vilsbole: ie. Prefer config object.

PR #2862, opened 2019-11-18
Add toasts

FrederikBolding:
## Description

- Added Toasts


FrederikBolding:
[ch3117]

FrederikBolding:
This one is ready for review! Currently have only added one toast, not sure if we need to add more. Last I checked we didn't have a list of toasts to be implemented yet.

FrederikBolding:
Another thing is that I'm not sure whether or not we want to store toasts in LS.

common/v2/features/index.ts
```diff
@@ -28,3 +28,4 @@ export { SignAndVerifyMessage } from './SignAndVerifyMessage';
 export { NotificationsProvider } from './NotificationsPanel';
 export { BroadcastTransactionFlow } from './BroadcastTransaction';
 export { SwapAssetsFlow } from './SwapAssets';
+export { ToastProvider } from './Toasts';
```
vilsbole: So this is tricky. Toasts can be triggered by any part of the app and by placing it in features we run the risk of increasing circular dependencies.
I'm open to opinions, but believe it would be better to move it into either services or components.

common/v2/features/Toasts/ToastProvider.tsx
```diff
@@ -0,0 +1,42 @@
+import React, { Component, createContext } from 'react';
+import toast from 'toasted-notes';
+
+import { default as ToastComponent } from './components/Toast';
+import { toastConfigs } from './constants';
+
+export interface ProviderState {
```
vilsbole: Is there a reason to export the interface ?

FrederikBolding:
> ! The integrity sha of every package in `yarn.lock` has been deleted.
> Please revert the changes.

Oh! That's not intentional, will fix 😄

common/v2/features/index.ts
FrederikBolding: Yeah, I thought about this as well. But shouldn't we be moving Notifications as well then? Maybe moving both of them to service?

common/v2/features/Toasts/ToastProvider.tsx
FrederikBolding: Don't think so - It's copied from `NotificationsProvider`, I can remove if not needed!

FrederikBolding:
@vilsbole Moved the Provider to `services` and the component to `components`. _Now_ I'm getting a circular dependency issue 😆 Any ideas?

common/v2/features/Toasts/ToastProvider.tsx
vilsbole: Please do. Having a clean interface for the module simplifies maintainability.

FrederikBolding: Already done 😄

PR #2869, opened 2019-11-24
Separate localstorage from application state

vilsbole:
__Warning__

On first load the app will ~~preserve your existing LS~~ migrate your existing LS to the latest verison!

In this PR `uuid` have been added to multiple db items. 
It is possible that some queries fail when you try to use an item without an uuid.
We can write a `migration` to solve this.

### Changes
1. Remove all LS access from Models and `helpers`
2. Models talk only to `DataProvider`
3. `DataProvider` has 4 responsibilities:
   - keep a reference of `appState`. This is our single source of truth.
   - decouples `state` and db formats (i.e marshall)
   - determines when to sync `state` and LS
   - provide a `ModelFactory`
5. Solve all `cannot setState inside unmounted component' (imported some `react-use` inside vendor).
4. Decouple Schema from SeedData. Now we can `toggle` seed accounts.
6. Create DevTools links to
   - Reset DB
   - add seed accounts
   - remove seed accounts
7. Introducing the notion of migrations. We preserve a history of changes to our schema and can migrate from one to the other.

### Usage
1. Open app in Private mode and open `DevTools`
2. Click 'Reset'
3. Play with adding and removing DevAccounts

### Tested
- [x] It can add an account in DataProvider
- [x] It includes new accounts in favorites by default
- [ ] It can add transaction to Account
- [x] It throttles writing to LS (1min)
- [x] It creates a backup `onBeforeUnload`
- [x] It preserves previous LS
- [ ] It can export a backup
- [ ] It can import a backup

### Know issues
- [ ] Fix NotificationsProvider (On refresh the Dashboard Notifications will display the last notification)
- [ ] Fix ScreenLockProvider (Need to reactivate lock function in order to fix)
- [ ] For the moment the only way to reset the DB is through our `DevTools` and through 'Settings'. Manual deletion inside the chrome devtools will fail. The app is listening to the `onBeforeUnload` event and updates the DB at that moment. So the emptied DB will be replenished with the appState at that moment. 
- [ ] Contract and Swap dropdowns trigger a `use forwardRef` error.

### Later
- [ ] Merge StoreProvider into DataProvider. At least for StoreAccounts
- [ ] For each Model we can use the `use<ModelContext>` pattern to reduce imports.


vilsbole:
Code is updated. Now there is a listener on LS that will initialise the seed whenever the remove event is triggered. (Console > Application > local storage > use the delete key)
So LS will be recreated if ever it is deleted.
To hydrate the app on initialisation I plan on using a Provider that talks to LS.
This should be able to replace both DataCache and DataManager

common/v2/services/Store/Asset/helpers.ts
```diff
@@ -1,19 +1,15 @@
...
 
-export const getNewDefaultAssetTemplateByNetwork = (network: Network): Asset => {
-  const baseAssetOfNetwork: Asset | undefined = getAssetByUUID(network.baseAsset);
+export const getNewDefaultAssetTemplateByNetwork = (assets: Asset[]) => (
+  network: Network
+): Asset => {
+  const baseAssetOfNetwork: Asset | undefined = getAssetByUUID(assets)(network.baseAsset);
   if (!baseAssetOfNetwork) {
     return {
       uuid: generateUUID(),
```
blurpesec: Removing the seedCache file exposes an issue with this function that results in base assets being created here having non-deterministic uuids.

vilsbole: Good catch. Done!

common/v2/database/v1.0.0/generateDefaultValues.ts
```diff
@@ -0,0 +1,185 @@
...
+  AssetLegacy,
+  LSKeys
+} from 'v2/types';
+
+import { SeedData, StoreAction } from './types';
+import { toArray, toObject, add } from './helpers';
+/* Transducers */
+const addNetworks = add(LSKeys.NETWORKS)((networks: SeedData) => {
+  const formatNetwork = (n: NetworkLegacy): Network => {
+    const baseAssetUuid = generateUUID();
```
blurpesec: This should be deterministic.

common/v2/database/v1.0.0/generateDefaultValues.ts
```diff
@@ -0,0 +1,185 @@
...
+    toArray,
+    R.map(formatAsset),
+    R.reduce((acc, curr) => ({ ...acc, [curr.uuid]: curr }), {}),
+    R.mergeRight(store.assets) // Ensure we return an object with existing assets as well
+  )(store.networks);
+});
+
+const addFiatsToAssets = add(LSKeys.ASSETS)((fiats: Fiat[], store: LocalStorage) => {
+  const formatFiat = ({ code, name }: Fiat): ExtendedAsset => ({
+    uuid: generateUUID(),
```
blurpesec: Should we make these deterministic as well? Maybe generate uuid from the fiat code? Those are unique afaik.

common/v2/features/Settings/components/GeneralSettings.tsx
```diff
@@ -124,6 +124,7 @@ export default class GeneralSettings extends React.Component<SettingsProps> {
             <Link to={ROUTE_PATHS.SETTINGS_EXPORT.path}>
               <SettingsButton secondary={true}>{translate('SETTINGS_EXPORT_LABEL')}</SettingsButton>
             </Link>
+            ]
```
blurpesec: Remove dis.

common/v2/features/Settings/Settings.tsx
```diff
@@ -44,10 +44,15 @@ const SettingsTabs = styled(TabsNav)`
 `;
 
 function renderAccountPanel() {
+  const { accounts } = useContext(StoreContext);
   return (
     <FlippablePanel>
       {({ flipped }) =>
-        flipped ? <p>Add Account</p> : <AccountList deletable={true} copyable={true} />
+        flipped ? (
+          <p>Add Account</p>
```
blurpesec: We probably have a translation string for this somewhere :dancer:

common/v2/services/Store/DataManager/DataProvider.tsx
```diff
@@ -0,0 +1,104 @@
...
+   * Manage sync between appState an db
+   */
+  const syncDb = useCallback(() => {
+    console.debug('Updating LocalStorage');
+    setDb(deMarshallState(appState));
+  }, [appState]);
+  // By default we sync no more than once a minute
+  useThrottleFn(syncDb, 60000, [appState]);
+  // In any case we sync before the tab is closed
+  useEvent('beforeunload', syncDb);
```
blurpesec: Is this the only trigger that can possibly happen? Does this trigger when all of the following?:
1) Closes the tab
2) Closes the window
3) Kills the browser process using cmd
4) Hard shuts down the computer

common/v2/services/Store/DataManager/actions.ts
```diff
@@ -0,0 +1,107 @@
...
+    });
+  };
+
+  const destroyStorage = () => {
+    console.debug('Destroy Storage');
+  };
+
+  const importStorage = (data: string) => {
+    const d = JSON.parse(data);
+    // @TODO: perfom version and validity check.
```
blurpesec: Seems to be a problem with importing previous backups. It shows an error in the settings/import, but still imports i think?

May be due to this todo

blurpesec:
Some notes i came up with while testing features:
1) Dev account on hard reset of db doesn’t have uuid (showing `undefined`)
2) UUIDs in dev accounts (`common/v2/database/seed/accounts.ts`) aren’t accurate since we started using deterministic uuids.
3) Base asset uuids don’t appear to be deterministic anymore. This is breaking rates for Ether, and other base assets.
4) Exporting settings file ends up exporting what is currently in the localstorage, not the current state of the app.

common/v2/database/v1.0.0/generateDefaultValues.ts
```diff
@@ -0,0 +1,185 @@
...
+      name: a.name,
+      decimal: a.decimal,
+      ticker: (a.symbol as unknown) as TSymbol,
+      networkId: id,
+      contractAddress: a.address,
+      type: 'erc20'
+    });
+
+    // From { ETH: { tokens: [ {<tokens>} ] }}
+    // to   { <asset_uuid>: {<asset>} }
```
blurpesec: I like these conversion comments, makes reading the code much easier.

common/v2/database/v1.0.0/generateDefaultValues.ts
vilsbole: changed to `generateAssetUUID(n.chainId)`

common/v2/database/v1.0.0/generateDefaultValues.ts
vilsbole: changed to `generateAssetUUID(code, name)`

common/v2/features/Settings/components/GeneralSettings.tsx
vilsbole: Oops! Nice catch

common/v2/features/Settings/Settings.tsx
vilsbole: I would prefer to change this in a separate translations PR.
If I start I would want to translate the entire view, so I prefer to keep it out of scope atm.

common/v2/services/Store/DataManager/DataProvider.tsx
vilsbole: So according to the Page Lifecycle best practices, I learn that using `beforeunload` is a bad practice: https://developers.google.com/web/updates/2018/07/page-lifecycle-api

Updated the code for `visibilitychange` event and a check for `document.hidden`
https://developer.mozilla.org/en-US/docs/Web/API/Document/visibilityState

This will execute the JS when "document is either a background tab or part of a minimised window, or the OS screen lock is active."

For the other possibilities, I would prefer to address this PR.
Since we're syncing every minute, and that most operations can be either recovered or reproduced by the user I'm confortable with the possibility of missing a sync atm.
If you think the user could perform an important operation within the 60sec, we can maybe reduce the sync time?

vilsbole:
> 1. Dev account on hard reset of db doesn’t have uuid (showing `undefined`)
A 'reset' through the ourDevTools or the devTools will produce a config object with empty accounts.
A priori solved.

> 2. UUIDs in dev accounts (`common/v2/database/seed/accounts.ts`) aren’t accurate since we started using deterministic uuids.
Solved using deterministic uuids

> 3. Base asset uuids don’t appear to be deterministic anymore. This is breaking rates for Ether, and other base assets.
Rates had other issues. The mount/unmount pattern I used would only run once.
Solved with refactor of RatesProvider.

> 4. Exporting settings file ends up exporting what is currently in the localstorage, not the current state of the app.
WIP

common/v2/services/Store/DataManager/actions.ts
vilsbole: With the latest change to `isValidImport` we check that the imports `version` and `key` corresponds to the accepted db. This means that we fail immediately if the file is invalid.

For a better UX i would suggest we start using to https://github.com/epoberezkin/ajv to validate the schema. If we choose to do so we can start generating our types from the schema definition https://github.com/bcherny/json-schema-to-typescript, produce more specific errors and eventually be able to migrate between the different schemas. 
ie: https://spin.atomicobject.com/2018/03/26/typescript-data-validation/

Since this change would requires modifying all our data types I prefer doing in another chore.

LuckyLuker:
[ch3769]

PR #2870, opened 2019-11-25
Remove LOVE token

wtzb:
Note: not on prod, but in 1.7.4 release candidate.

One of those lovely contracts that mess up our interface by returning insane values :eyes: 

image

From the [contract](https://etherscan.io/address/0x5a276Aeb77bCfDAc8Ac6f31BBC7416AE1A85eEF2#code):

```
function balanceOf(address _owner) public view returns (uint256 balance) {
      return MAX_UINT256;
}
```

### DIFF
--- a/common/config/tokens/eth.json
+++ b/common/config/tokens/eth.json
@@ -5278,13 +5278,6 @@
     "uuid": "9f563cf7-f883-5459-845e-5021a2f58629"
   },
   {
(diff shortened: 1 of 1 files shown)


FrederikBolding:
Problem with simply removing it, is that it will be automatically re-added when we update the token list.

wtzb:
@FrederikBolding using our parse-eth-tokens tool it's possible to exclude certain addresses from being added, but it's not a very permanent solution.

PR #2872, opened 2019-11-25
Make the contentpanel wider and see how it goes

tayvano:
Partially addresses [CH3900] - Adjust global panel to show full address

https://app.clubhouse.io/mycrypto/story/3900/adjust-global-panels-to-view-full-address

### DIFF
--- a/common/v2/components/ContentPanel.tsx
+++ b/common/v2/components/ContentPanel.tsx
@@ -89,7 +89,7 @@ export default function ContentPanel({
   description,
   children,
   className = '',
-  width = '550px',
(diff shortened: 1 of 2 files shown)


tayvano:
playing "see if it breaks and adjust if needed" with this one so that we can organically check for issues rather than having to manually check down the road

tayvano:
2019-11-25 at 3 00 32 PM

This *should* leave enough space even when we add the identicon next to the receipt address field

LuckyLuker:
[ch3099]

PR #2873, opened 2019-11-26
DevMode modal [ch3557]

LinFeiLong:
https://app.clubhouse.io/mycrypto/story/3557/devmode-modal

### DIFF
--- a/common/v2/features/DevTools/DevTools.tsx
+++ b/common/v2/features/DevTools/DevTools.tsx
@@ -107,7 +107,6 @@ const DevTools = () => {
                           )}
                         />
                       </fieldset>
-                      Current dev-mode only features
                       <br />
                       Current dev-mode only features
                       <ul>
--- a/common/v2/features/DevTools/DevTools.tsx
+++ b/common/v2/features/DevTools/DevTools.tsx
@@ -11,14 +11,6 @@ import { Account, AddressBook, ExtendedAccount, WalletId } from 'v2/types';
 import ToolsNotifications from './ToolsNotifications';
 import ToolsAccountList from './ToolsAccountList';
 
-const DevToolsContainer = styled.div`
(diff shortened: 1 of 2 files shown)


FrederikBolding:
Can we remove one of these? https://github.com/MyCryptoHQ/MyCrypto/blob/b53e46498b465327bcebfc7d15d4d52e9faaef05/common/v2/features/DevTools/DevTools.tsx#L110-L112

FrederikBolding:
Also, it isn't really a modal anymore, but if everyone is cool with that, then I dont see any issues! 😄

common/v2/features/DevTools/DevTools.tsx
```diff
@@ -139,24 +136,24 @@ const DevToolsToggle = () => {
       onClick={toggleDevTools}
       style={{
         position: 'fixed',
-        bottom: 0,
-        right: 0,
+        top: 0,
+        left: 0,
         zIndex: 99,
-        height: '5rem'
+        width: 112
```
FrederikBolding: Do we usually handle width like this?

LinFeiLong: Honestly, I don't know how we do.
I don't like `rem` since its value is relative to the parent and because you never know the actual size.
@vilsbole any advice on this ?

vilsbole: imho `rem` is useful for responsive design when you care about scaling your components according to screen size. 
It this case, since it's just a dev tool and we don't have a responsive grid today, I would opt for the simplest solution. Whatever you think is best @FrederikBolding

FrederikBolding: @vilsbole If you don't have a problem with it, I don't either. Especially since its a dev tool.

PR #2877, opened 2019-11-27
Removed more legacy send flow

FrederikBolding:


FrederikBolding:
Actually a continuation of #2875

PR #2885, opened 2019-12-03
Removed legacy libs and things dependant on them

FrederikBolding:
- Libs
- UI elements depending on libs
- Fixed imports on existing tests


FrederikBolding:
@vilsbole No, I haven't looked at them much. There might be some stuff, but we'll have to write new tests for that then? 😄

vilsbole:
@FrederikBolding I looked at `decrypt.spec.ts` and it would be good to keep the tests and point them the appropriate methods that have been copied to v2.
The same for all the other spec files [if methods exists in v2 ? change import to v2 : delete spec]

FrederikBolding:
@vilsbole Sure! I'll go through them and see which should be kept 👍

FrederikBolding:
@vilsbole Saved most of the tests!! 👍

PR #2890, opened 2019-12-05
Improve non-Metamask web3 account ux 

blurpesec:
[ch3863]

### Description
- Finished up copy changes for ch3863.
- Added handling for non-Metamask web3 designation in AddAccount flow (this is handled in later flows already). This is done by defaulting unknown web3 wallet types to `WEB3` instead of `METAMASK`


LinFeiLong:
Hm I don't understand why there are both `Improve non-Metamask web3 account ux` and `copy changes` in the same PR.
Was it intended ?

blurpesec:
> Hm I don't understand why there are both `Improve non-Metamask web3 account ux` and `copy changes` in the same PR.
> Was it intended ?

Some of the copy changes for Web3 stuff relies on the non-metamask web3 account ux being more accurate, which is what is included in this pr.

PR #2892, opened 2019-12-06
Make header draggable for Electron

kunstefix:
[ch4065]


common/v2/features/Layout/Header/Header.tsx
```diff
@@ -46,6 +46,7 @@ const HeaderTop = styled.div`
   height: 77px;
   padding: 0 ${MIN_CONTENT_PADDING};
   border-bottom: 1px solid #3e546d;
+  -webkit-app-region: drag;
```
LinFeiLong: It's working perfectly

vilsbole: Any suggestions how we separate Electron specific code? 
Also the feature is important for Electron but the code line is inconspicuous for a web-dev.
Can we either add comment to explain the feature, or add a test to make sure it's not removed?

PR #2897, opened 2019-12-10
Fix Electron not running

LinFeiLong:
Fix Electron crash after this commit ([#2895)](https://github.com/MyCryptoHQ/MyCrypto/pull/2895)


FrederikBolding:
I think we should try to use the v2 methods. When is the crash happening?

FrederikBolding:
Also, I don't see the difference between the `v2` methods and the old ones.

LinFeiLong:
It's not really a crash, it's a warning/alert window.
And the Electron App won't launch.
I tried to fix it. But I don't really understand why it is not working.
Looks like the electron app cant find some functions.

LinFeiLong:
Closing since it is fixed on [2898](https://github.com/MyCryptoHQ/MyCrypto/pull/2898)

PR #2899, opened 2019-12-10
Fernandlime/ch4044/update dev dependencies

LinFeiLong:
Update of most of the devDependencies.


LinFeiLong:
@Mrtenz I encourage you to check what I did since I removed some packages that you added.
If you want me to put those packages back for any reason. I'm open for negociations :)

PR #2903, opened 2019-12-10
Fernandlime/ch4044/update electron

LinFeiLong:
Bump **electron** from 2.0.8 to 6.1.5


electron-app/main/menu.ts
```diff
@@ -1,88 +1,92 @@
...
+//       ...(HELP_MENU.submenu as MenuItemConstructorOptions[]),
+//       {
+//         label: 'Speech',
+//         submenu: [{ role: 'startspeaking' }, { role: 'stopspeaking' }]
+//       }
+//     ]
+//   });
+// } else {
+//   MENU.push(HELP_MENU);
+// }
```
blurpesec: Is there a reason to keep all of this commented stuff?

LinFeiLong: I was wondering what it was used for.
But it definitely looks like dead code.
If It is no commented, it messes up.

Would you suggest to remove this code without any thought ?

vilsbole: Can you check what it was for? 
I suspect it was the side menu of the previous designs

PR #2905, opened 2019-12-10
Merge legacy and v2 types

FrederikBolding:
- [x] Delete unused types
- [x] Use `@types` if possible
-  [x] Move remaining types to `v2/types/vendor`


FrederikBolding:
[ch4096]

FrederikBolding:
@vilsbole Seems that there were only a few unused types - also they don't seem to be in `v2/types` already. Do you want me to move the existing ones to ``v2/types?`

PR #2907, opened 2019-12-11
Klemenkunstek/ch4136/update electron

kunstefix:
Update Electron version to 7.1.4


LinFeiLong:
@FrederikBolding @Mrtenz 
I disabled the pre-build for linux, since it doesn't work for now.
The linux build works on MacOs though 
Please could you try to run `yarn dev:electron` and check if it runs, since you both work on linux ?
Also I do not expect you to build `yarn build:electron:linux` successfully on your local machine.

PR #2908, opened 2019-12-11
Added check to make sure ENS resolved address is valid

FrederikBolding:
0x0000000000000000000000000000000000000000 was already considered an invalid address.

Just made sure we checked it when doing ENS lookups


FrederikBolding:
[ch3967]

FrederikBolding:
Agreed, done! 😄

PR #2913, opened 2019-12-13
Implement desktop app analytics

wtzb:
This PR correctly implements desktop app analytics, and makes the Windows build not report an URL that would identify the user (which is why desktop app analytics have not been up before this).

What Windows would report: `file:///C:/Users/NameOfUser/AppData/Local/Programs/MyCrypto/resources/app.asar/index.html#/generate`

What Linux/Mac would report: `localhost:3000/generate`

This PR makes all platforms report a more generic url, like `desktop.app/account`, and actually send them to our analytics platform.


common/v2/services/Analytics/Analytics.ts
```diff
@@ -41,9 +41,11 @@ export default class AnalyticsService {
 
     const analyticsId = isDesktop() ? ANALYTICS_ID_DESKTOP : ANALYTICS_ID_SITE;
 
+    const url = isDesktop() ? 'https://desktop.app' + pathHash.substr(1) : pageUrl;
```
vilsbole: 1. Why do we need a different url for desktop? Instead the analyticsId sufficient to distinguish between both?
2. We want to avoid magic strings spread throughout the app.
Please create a constant for the url `desktop.app`
3. `'https://desktop.app' + pathHash.substr(1)` is an odd url.  Wouldn't you want a slash between the two?: `https://desktop.app/${pathHash.substr(1)}`
3. When I run `dev:electron` on `:legacy` it seems that we don't use the hashRouter and when I navigate to `localhost:3000/generate` `window.location.hash` rightfully returns an empty string. What am I missing?

wtzb: 1. Since the desktop app uses Electron which is based on Chromium, it would skew the Browsers analytics in Matomo as it doesn't see Electron as something separate, and would count Chrome browser visits and desktop app visits as the same browser.

2. :+1: 

3. `pathHash.substr(1)` returns a value that already contains a slash.

4. It indeed returns an empty string on development, but in a build or on production it does return a value. I think this is fine, as analytics don't really need to be reported in dev, but would there be a better solution for this?

In build:
image

What Matomo would report with the current setup:
image

vilsbole: Oh dear, it means we have a specific router Electron + Production. :(
Can you add the production flag to the ternary and a comment for the next dev who will work on this explaining the env differences ?

wtzb: 1. Ah, I see what you mean. We need a different url for desktop as there is an issue with electron on Windows where `window.location.href` would leak the user's account name on their computer, as it would return a value like `file:///C:/Users/vbox (the machine username)/AppData/Local/Programs/MyCrypto/resources/app.asar/index.html#/account`, instead of `file:///opt/mycrypto/app/resources/app.asar/index.html#/account` which is what Linux would report, and is not identifiable. This makes `window.location.href` fine for use for web, like we do, but not for desktop.

3. I brought this up in the tread on Slack, but perhaps we could instead use a specific Matomo ID just for analytics so it doesn't clutter the production analytics?

PR #2914, opened 2019-12-13
Remove unused code from 'shared' and 'electron-app' folders

kunstefix:


FrederikBolding:
Hmm, do we not want to use the updater at least in the future?

LinFeiLong:
> Hmm, do we not want to use the updater at least in the future?

If it is not used in a close future, I would definitely prefer to get rid of it.
For two reasons:
First, it makes noise and the code should reflect what the app is capable of.
Second, if we need to re-implement it. We'll do it either buy recovering the code in the history or by re-implementing our self.

So I see no reasons to keep it.

PR #2915, opened 2019-12-16
Deploy contracts feature updates

amacar:
Deploy Contracts feature updates:
- styling
- removed redux

Part of:
[ch2490]


common/v2/utils/transaction.ts
```diff
@@ -26,7 +26,7 @@ export function fromTxReceiptObj(txReceipt: ITxReceipt): ITxReceipt | undefined
             contractAsset.decimal || DEFAULT_ASSET_DECIMAL
           )
         : fromWei(Wei(hexWeiToString(txReceipt.value._hex)), 'ether').toString(),
-      to: contractAsset ? ERC20.transfer.decodeInput(txReceipt.data)._to : txReceipt.to,
+      to: contractAsset ? ERC20.transfer.decodeInput(txReceipt.data)._to : txReceipt.to || '0x',
```
vilsbole: It seems dangerous to default to '0x' for all transaction formatting.
Can you please explain?

common/v2/features/DeployContracts/stateFactory.tsx
```diff
@@ -0,0 +1,143 @@
...
+
+import { DeployContractsState } from './types';
+import { makeTxConfigFromTransaction, constructGasCallProps } from './helpers';
+
+const deployContractsInitialState = {
+  account: undefined,
+  rawTransaction: {
+    gasPrice: '0xee6b2800',
+    gasLimit: 21000,
+    nonce: 0
```
vilsbole: Avoid magic strings.
Prefer a const in config/gasPrice

common/v2/features/DeployContracts/stateFactory.tsx
```diff
@@ -0,0 +1,143 @@
...
+      account
+    }));
+  };
+
+  const handleTxSigned = async (signResponse: any, after: () => void) => {
+    const { account, txConfig } = state;
+
+    if (!account) {
+      return;
+    }
```
vilsbole: To facilitate readability prefer to return on the same line. 
```
if (!account) return;
```

common/v2/features/DeployContracts/helpers.ts
```diff
@@ -0,0 +1,73 @@
...
+  accounts.filter(acc => acc.networkId === networkId && WALLET_STEPS[acc.wallet]);
+
+export const makeTxConfigFromTransaction = (
+  rawTransaction: ITxObject,
+  account: StoreAccount,
+  amount: string
+): ITxConfig => {
+  const { gasPrice, gasLimit, nonce, data, to, value } = rawTransaction;
+  const { address, network } = account;
+  const baseAsset = getAssetByUUID(network.baseAsset)!;
```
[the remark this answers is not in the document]
amacar: Sure, that was caused after the last merged PR.

common/v2/features/DeployContracts/components/Deploy.tsx
```diff
@@ -0,0 +1,205 @@
...
+  const handleGasPriceChange = (val: string) => {
+    if (val.length) {
+      val = hexWeiToString(inputGasPriceToHex(val));
+      val = addHexPrefix(new BN(val).toString(16));
+    }
+    handleGasSelectorChange({ gasPrice: val } as ITxConfig);
+  };
+
+  const formatGasPrice = () =>
+    gasPrice.length ? baseToConvertedUnit(hexToString(gasPrice), 9) : gasPrice;
```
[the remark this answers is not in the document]
amacar: Agree

common/v2/features/DeployContracts/components/Deploy.tsx
```diff
@@ -0,0 +1,205 @@
...
+
+  const updateGasCallProps = (acc: StoreAccount, byteCodeVal: string) => {
+    if (!acc) return;
+
+    setGasCallProps(constructGasCallProps(byteCodeVal, acc));
+  };
+
+  const debouncedUpdateGasCall = useRef(
+    debounce((acc: StoreAccount, byteCodeVal: string) => updateGasCallProps(acc, byteCodeVal), 700)
+  );
```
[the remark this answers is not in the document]
amacar: Thanks, code looks cleaner now.

common/v2/utils/transaction.ts
amacar: Usually, we have `to` parameter when sending a transaction but when we deploy a contract there is no `to` parameter so I just used a creation address (0x) as a placeholder. We can change it to some other address. Please tell me to which. Thanks

vilsbole: Since we don't need the `to` param for a contract deployment it would be better to make it explicit and remove it from the object.
If the view requires a `to` field, we can disable the field when displaying a deployment receipt.
Would this work for you?

amacar: The only problem is in transaction list in dashboard. If saved transaction doesn't have `to` field, the app then crashes when trying to display transaction list. I can probably do some check there and just show blank address for contracts deplyoments.

vilsbole: Sounds good! It would also be more reassuring for the user than seeing a tx to 0x in his history.

amacar: Fixed

PR #2920, opened 2019-12-18
Prefill interact with contracts data from URL

amacar:
[ch3421]
Below listed params should be used:
- network,
- address,
- function,
- input (for each input in format `param name:param value`)

Test url: [https://mycryptobuilds.com/pr/2920/#/interact-with-contracts?network=Ropsten&address=0xbb4AAaF8cAA1A575B43E7673e5b155C1c5A8BC13&function=etherBalances&input=addresses:[%220xCb1E63be3e892126eED1345831720759435c9066%22,%220x82D69476357A03415E92B5780C89e5E9e972Ce75%22]&input=secondParam:secondValue&input=thirdP:thirdV](https://mycryptobuilds.com/pr/2920/#/interact-with-contracts?network=Ropsten&address=0xbb4AAaF8cAA1A575B43E7673e5b155C1c5A8BC13&function=etherBalances&input=addresses:[%220xCb1E63be3e892126eED1345831720759435c9066%22,%220x82D69476357A03415E92B5780C89e5E9e972Ce75%22]&input=secondParam:secondValue&input=thirdP:thirdV)

### DIFF
--- a/common/v2/features/InteractWithContracts/components/Interact.tsx
+++ b/common/v2/features/InteractWithContracts/components/Interact.tsx
@@ -1,5 +1,7 @@
 import React, { useEffect, useState } from 'react';
+import { withRouter, RouteComponentProps } from 'react-router-dom';
 import styled from 'styled-components';
+import * as qs from 'query-string';
 import { Identicon } from '@mycrypto/ui';
 
 import { NetworkSelectDropdown, InputField, Dropdown, InlineErrorMsg, Button } from 'v2/components';
@@ -7,6 +9,7 @@ import { Contract, StoreAccount, ITxConfig, ExtendedContract, Network } from 'v2
 import { COLORS, BREAK_POINTS } from 'v2/theme';
 import { translateRaw } from 'v2/translations';
 import { isValidETHAddress, isCreationAddress } from 'v2/services/EthService/validators';
+import { getNetworkById } from 'v2/services';
 
 import ContractDropdownOption from './ContractDropdownOption';
 import ContractDropdownValue from './ContractDropdownValue';
@@ -148,7 +151,9 @@ interface Props {
   handleDeleteContract(contractUuid: string): void;
 }
 
-export default function Interact(props: Props) {
+type CombinedProps = RouteComponentProps<{}> & Props;
+
+function Interact(props: CombinedProps) {
   const {
     network,
     contractAddress,
@@ -176,8 +181,14 @@ export default function Interact(props: Props) {
     handleDeleteContract
   } = props;
 
-  const [error, setError] = useState(undefined);
+  const [error, setError] = useState<string | undefined>(undefined);
+  const [areFieldsPopulatedFromUrl, setAreFieldsPopulatedFromUrl] = useState(false);
+  const [wasAbiEditedManually, setWasAbiEditedManually] = useState(false);
   const [wasContractInteracted, setWasContractInteracted] = useState(false);
+  const [interactionDataFromURL, setInteractionDataFromURL] = useState<any>({});
+  const { network: networkIdFromUrl, address: addressFromUrl } = qs.parse(props.location.search);
+  const networkAndAddressMatchURL =
+    network.id === networkIdFromUrl && contractAddress === addressFromUrl;
 
   useEffect(() => {
     updateNetworkContractOptions();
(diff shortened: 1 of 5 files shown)


vilsbole:
Tagging @409H Just to double check that parsing the url with `qs` is sufficient.

amacar:
> I've tested the usability of this and it looks/feels awesome! Good job :-D.
> 
> One suggestion: If the user is on a read-only interaction where they've pre-filled all of the fields using url-parameters, is it possible to automatically trigger the "Read" button? That would provide a slight improvement to the UX, just not having to click that button.
> 
> Another thing, we want to update the default contracts that beta has to include all of the ones from `legacy`. Do you think this would be an appropriate pr to include that in since it's a small change, or should we add it to a new task?

I added triggering Read button if all inputs are prefilled from URL.

Regarding the change of default contracts, feel free to push to this branch.

VinceBCD:
Congratulation about the UX, it's great.
How to deal with unamed parameters (of mappings for example, displayed as Input#0 / Input#1)?

amacar:
> Congratulation about the UX, it's great.
> How to deal with unamed parameters (of mappings for example, displayed as Input#0 / Input#1)?

There is noted in specification that if parameters are unnamed then are inputs mapped by index.
The same is done here. One example: https://mycryptobuilds.com/pr/2920/#/interact-with-contracts?network=Ethereum&address=0x986EE2B944c42D017F52Af21c4c69B84DBeA35d8&function=allowance&input=0:0x986EE2B944c42D017F52Af21c4c69B84DBeA35d8&input=1:0x986EE2B944c42D017F52Af21c4c69B84DBeA35d8

common/v2/features/InteractWithContracts/components/Interact.tsx
```diff
@@ -187,9 +198,32 @@ export default function Interact(props: Props) {
...
+    if (areFieldsPopulatedFromUrl && networkAndAddressMatchURL && !wasAbiEditedManually) {
+      submitInteract();
+      const { function: functionFromUrl, input } = qs.parse(props.location.search);
+
+      const inputsArray: string[] = !input ? [] : Array.isArray(input) ? input : [input];
+
+      const inputsFromUrl = inputsArray.map(i => ({
+        name: i.includes(':') ? i.substr(0, i.indexOf(':')) : '',
+        value: i.includes(':') ? i.substr(i.indexOf(':') + 1, i.length) : ''
+      }));
```
vilsbole: We should extract and test this logic along with L189 & 191.
Maybe a function that takes `props.location.search` and returns an object that is sanitised and can directly be used by the form?

common/v2/features/InteractWithContracts/components/Interact.tsx
```diff
@@ -220,6 +254,30 @@ export default function Interact(props: Props) {
     }
   };
 
+  useEffect(() => {
+    if (getNetworkById(networkIdFromUrl)) {
```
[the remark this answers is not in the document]
amacar: Done

common/v2/features/InteractWithContracts/components/Interact.tsx
amacar: Made some tests. Name of the contract could be anything (also containing special chars - `<>()'`...) so I don't think we can do much to sanitise the query params. Parsed params are only used in the equality checks and shown in react components as inputs, so there shouldn't be security issues because React handles XSS attack for us.

vilsbole: From what I understand the flow for `input` is:
`url -> location.search (react-router) -> parse (query-string) -> setInteractionDataFromURL --> then passing interactionDataFromURL as prop for GeneratedInteractionForm`

It appears that attacker controlled props are not escaped in React:
- https://stackoverflow.com/a/51852579/2057532
- https://medium.com/javascript-security/avoiding-xss-in-react-is-still-hard-d2b5c7ad9412

Though 'query-string' appears to handle escape attempts correctly.
Merging now to move forward.
Tagging @409H to see if this introduces a xss vulnerability: 

https://mycryptobuilds.com/pr/2920/#/interact-with-contracts?network=Ethereum&address=0x986EE2B944c42D017F52Af21c4c69B84DBeA35d8&function=allowance&input=0:0x986EE2B944c42D017F52Af21c4c69B84DBeA35d8&input=1:0x986EE2B944c42D017F52Af21c4c69B84DBeA35d8

PR #2921, opened 2019-12-19
Fix 'export not found' warnings

LinFeiLong:
Added 'commonjs' to the babel.config.js to fix the false positive warnings


FrederikBolding:
@Mrtenz I seem to remember these happened due to some TypeScript issues?

PR #2924, opened 2019-12-19
Return missing properties from MnemonicWallet

kunstefix:
- add missing properties to be returned from MnemonicWallet
- read dPath from sender account instead from state

### DIFF
--- a/common/v2/components/SignTransactionWallets/Mnemonic.tsx
+++ b/common/v2/components/SignTransactionWallets/Mnemonic.tsx
@@ -125,7 +125,8 @@ export default class SignTransactionMnemonic extends Component<
   private async getPublicKey() {
     this.setState({ isSigning: true });
     try {
-      const wallet = await ethers.Wallet.fromMnemonic(this.state.phrase, this.state.selectedDPath);
+      const { senderAccount } = this.props;
+      const wallet = await ethers.Wallet.fromMnemonic(this.state.phrase, senderAccount.dPath);
       const checkSumAddress = utils.getAddress(wallet.address);
       this.checkPublicKeyMatchesCache(checkSumAddress);
     } catch (err) {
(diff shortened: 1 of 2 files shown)


LinFeiLong:
@kunstefix For me the behavior is still the same, I'm having a blank screen juste after this one
Screenshot 2019-12-19 at 13 43 48

kunstefix:
@LinFeiLong  Thanks for noticing. I found that this bug is not present only in electron, but also on webapp. So I fixed it, but didn't test in production build (it worked in development). I will try to add another commit here to fix production as well.

kunstefix:
@LinFeiLong  Looks like maOS production build fails because of another issue caused by Node library
image

LinFeiLong:
if you check the ticket detail on top

Bump electron from 3.1.0 to 4.0.0      8f25c10

may have caused the issue

LinFeiLong:
@kunstefix I read your comments ahead again.
Are you sure, the issue is only in production build ?
If so the only thing that can may cause the issue is how we build it.
what do you think ?

LinFeiLong:
@vilsbole So I checked what you said about ripemd160. It definitely helped 
https://github.com/electron/electron/issues/16195

LinFeiLong:
Screenshot 2020-01-08 at 12 52 28

LinFeiLong:
Screenshot 2020-01-08 at 08 39 10

LinFeiLong:
@kunstefix 

> add missing properties to be returned from MnemonicWallet
> read dPath from sender account instead from state

Tell me if it is still relevant to keep this

kunstefix:
@LinFeiLong This was a fix to another bug related to Mnemonic, so it should be merged.

kunstefix:
@vilsbole @LinFeiLong  now that node problems are fixed we should still merge this one. It fixes this undocumented bug for adding account with mnemonic: 
ezgif com-video-to-gif

kunstefix:
> I was wondering where we set `state.selectedDPath` and the setter appears to have been removed.
> Though looking at:
> 
> ```
>  const signerWallet = await ethers.Wallet.fromMnemonic(
>       this.state.phrase,
>       this.state.selectedDPath
>     );
> ```
> 
> It is called in another place in the component.
> Should we remove the state attribute and update the code here as well?

Yes, i removed it.

PR #2929, opened 2019-12-20
Update legacy tokenlist

wtzb:


vilsbole:
Should these tokens be aded to beta as well?
Should we include them in ethereum-lists?

wtzb:
@vilsbole these tokens should be added to beta as well, and https://github.com/MyCryptoHQ/parse-eth-tokens can be used for this, which directly pulls from ethereum-lists. We don't have to add them to the list ourselves.

PR #2930, opened 2019-12-20
Replace UUID library to fix Electron blank screen

kunstefix:
MyCrypto is now using:
MyCryptoHQ/uuid-by-string
   |_MyCryptoHQ/js-md5
   |_MyCryptoHQ/js-sha1

Removing the eval() causing a blank screen on electron

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -8693,15 +8693,13 @@ js-levenshtein@^1.1.3:
   resolved "https://registry.yarnpkg.com/js-levenshtein/-/js-levenshtein-1.1.6.tgz#c6cee58eb3550372df8deb85fad5ce66ce01d59d"
   integrity sha512-X2BB11YZtrRqY4EnQcLX5Rh373zbK4alC1FW7D7MBhL2gtcC17cTnr6DmfHZeS0s2rTHjUTMMHfG7gO8SSdw+g==
 
-js-md5@^0.7.3:
+js-md5@MyCryptoHQ/js-md5#b300da8:
   version "0.7.3"
-  resolved "https://registry.yarnpkg.com/js-md5/-/js-md5-0.7.3.tgz#b4f2fbb0b327455f598d6727e38ec272cd09c3f2"
-  integrity sha512-ZC41vPSTLKGwIRjqDh8DfXoCrdQIyBgspJVPXHBGu4nZlAEvG3nf+jO9avM9RmLiGakg7vz974ms99nEV0tmTQ==
+  resolved "https://codeload.github.com/MyCryptoHQ/js-md5/tar.gz/b300da833e5e2c83dae186359b72c69e1711acd2"
 
-js-sha1@^0.6.0:
+js-sha1@MyCryptoHQ/js-sha1#7474a64:
   version "0.6.0"
-  resolved "https://registry.yarnpkg.com/js-sha1/-/js-sha1-0.6.0.tgz#adbee10f0e8e18aa07cdea807cf08e9183dbc7f9"
-  integrity sha512-01gwBFreYydzmU9BmZxpVk6svJJHrVxEN3IOiGl6VO93bVKYETJ0sIth6DASI6mIFdt7NmfX9UiByRzsYHGU9w==
+  resolved "https://codeload.github.com/MyCryptoHQ/js-sha1/tar.gz/7474a6478649ff9a784e565a71c2c97ab18ef634"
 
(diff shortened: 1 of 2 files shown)


blurpesec:
We weren't using the part of `uuid-by-string` lib that uses js-md5, so i just pulled the `uuid-by-string` lib into the codebase and removed the stuff we didn't need.

LinFeiLong:
I'm good for now on this task, I can finish it on my own.
Thank you for the team effort guys. 🤗

common/v2/utils/uuidUtils.ts
```diff
@@ -0,0 +1,107 @@
+import sha1 from 'js-sha1';
```
vilsbole: When we copy code from another repo let's:
- place it in the `vendor` directory and add a commented link to source file so we know where it comes from.
- also copy the spec files whenever possible.

common/v2/utils/index.ts
```diff
@@ -25,3 +25,4 @@ export {
 export { makeBlob } from './blob';
 export { default as consoleAdvertisement } from './consoleAdvertisement';
 export { fromTxReceiptObj } from './transaction';
+export * from './uuidUtils';
```
vilsbole: Please favor explicit exports for the module.
It facilitates refactors by having an explicit list of the methods that are used.

yarn.lock
```diff
@@ -14494,7 +14478,7 @@ utils-merge@1.0.1:
   resolved "https://registry.yarnpkg.com/utils-merge/-/utils-merge-1.0.1.tgz#9f95710f50a267947b2ccc124741c1028427e713"
   integrity sha1-n5VxD1CiZ5R7LMwSR0HBAoQn5xM=
 
-uuid-by-string@3.0.2, uuid-by-string@^3.0.1:
+uuid-by-string@^3.0.1:
```
vilsbole: It seems that another dependency also relies on `uuid-by-string`

LinFeiLong:
@vilsbole, though @kunstefix approach by rewriting the package is relevant. I'll definitely prefer to do use modified forks.

For 3 reasons:
- less code modifications
- we still know where the code is coming from
- it's easier to maintain

LinFeiLong:
@409H 
I'm using 3 forks now :
git://github.com/LinFeiLong/uuid-by-string.git
|__ git://github.com/LinFeiLong/js-md5.git
|__ git://github.com/LinFeiLong/js-sha1.git

Please, could you check them and make a clone of those on the MYC if it is ok for you ?

LinFeiLong:
@vilsbole hm I after some thoughts, I didn't see the point of removing all the stuff we didn't need in this package (uuid-by-string).
Reducing the size of the package, but it is minor change.

Also we mentioned the `parse-eth-tokens` dependecy which uses `uuid-by-string` but won't affect security because it is only used in the script. 
For consistency I don't mind changing `parse-eth-token` to use our fork of `uuid-by-string` if you prefer so.

409H:
> @409H
> I'm using 3 forks now :
> git://github.com/LinFeiLong/uuid-by-string.git
> |__ git://github.com/LinFeiLong/js-md5.git
> |__ git://github.com/LinFeiLong/js-sha1.git
> 
> Please, could you check them and make a clone of those on the MYC if it is ok for you ?

Looks good. Can we version lock the dependancies at https://github.com/LinFeiLong/uuid-by-string/commit/04b8d63496deb16a8700812768088d2e4b42011b#diff-b9cfc7f2cdf78a7f4b91a753d10865a2R41

Thanks for removing the `eval()` on the packages

vilsbole:
Following https://mycryptohq.slack.com/archives/G8G1S4TG9/p1578356396118600
Let's use the forks under the @Mycrypto npm namespace.

409H:
> Following https://mycryptohq.slack.com/archives/G8G1S4TG9/p1578356396118600
> Let's use the forks under the @Mycrypto npm namespace.

I think we need to issue the patch again to remove the `eval()` @LinFeiLong

LinFeiLong:
I had to revert https://github.com/MyCryptoHQ/MyCrypto/pull/2945, since it is making a blankScreen too.

LinFeiLong:
> Following https://mycryptohq.slack.com/archives/G8G1S4TG9/p1578356396118600
> Let's use the forks under the @Mycrypto npm namespace.

LinFeiLong:
Closed accidentely

PR #2933, opened 2019-12-22
Added proxy contract to transaction data correctly

FrederikBolding:
Part of fixing swap

### DIFF
--- a/common/v2/features/SwapAssets/helpers.ts
+++ b/common/v2/features/SwapAssets/helpers.ts
@@ -3,6 +3,7 @@ import BN from 'bn.js';
 import { addHexPrefix } from 'ethereumjs-util';
 
 import { Asset, StoreAccount, WalletId, ITxConfig, SigningComponents } from 'v2/types';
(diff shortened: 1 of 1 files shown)


FrederikBolding:
@PeterZlodej @amacar Did this initially work? Seems that the `spender` field isn't returned by the dex.ag API.

amacar:
> @PeterZlodej @amacar Did this initially work? Seems that the `spender` field isn't returned by the dex.ag API.

As far as I remember it was working without a problem, I also tested swapping Aragon For Bat two months ago. This is the txid: https://etherscan.io/tx/0xb421d6710bc3aeda145dafe56136f88b9ea20cc050db57227b69d9be1d377113
Sent from this account: https://etherscan.io/address/0x82d69476357a03415e92b5780c89e5e9e972ce75#tokentxns through this proxy contract: https://etherscan.io/address/0xa65440c4cc83d70b44cf244a0da5373aca16a9cb

FrederikBolding:
Part of fixing swap - someone check that I understood correctly!

PR #2934, opened 2019-12-22
Fixed some Ledger issues

FrederikBolding:
- Fixed translations
- Disabled ethereumjs-tx EIP-155 support for Ledger


FrederikBolding:
Was supposed to be a draft PR, oops. WIP!

FrederikBolding:
I think the conclusion to this is that Ledger doesn't handle EIP-155 correctly. Thus, I've disabled ethereumjs-tx checks for EIP-155 for Ledger.

Still need to fix DPaths not entirely correct for Ledger Live.

FrederikBolding:
DPaths is a bigger task and should be handled later. Ready for review!

common/v2/services/WalletService/deterministic/ledger.ts
```diff
@@ -45,6 +45,8 @@ export class LedgerWallet extends HardwareWallet {
   }
 
   public async signRawTransaction(t: EthTx): Promise<Buffer> {
+    // Disable EIP155 in Ethereumjs-tx since it conflicts with Ledger
+    t = new EthTx(t, { chain: t.getChainId(), hardfork: 'tangerineWhistle' });
```
vilsbole: For best practices sake please avoid reassigning a functions arguments.
https://spin.atomicobject.com/2011/04/10/javascript-don-t-reassign-your-function-arguments/

Since objects are passed by reference in JS, the previous implementation of `signRawTransaction` which modifies an arguments params is particularly dangerous and cause some nasty bugs.

common/v2/services/EthService/utils/makeTransaction.ts
```diff
@@ -10,7 +10,14 @@ import { hexEncodeQuantity } from './hexEncode';
 
 export const makeTransaction = (
   t: Partial<Tx> | Partial<ITransaction> | Partial<IHexStrTransaction> | Buffer | string
-) => new Tx(t);
+) => {
+  if (typeof t === 'object' && t.hasOwnProperty('chainId')) {
```
vilsbole: If you wish, we now have include `ramda` in the project.
It provides a `R.prop` that also handles if the object is valid. https://github.com/ramda/ramda/tree/v0.26.1/source/prop.js

common/v2/services/WalletService/deterministic/ledger.ts
FrederikBolding: Wasn't aware, thanks 😄

FrederikBolding:
@vilsbole Fixed! Ready for another review 😄

PR #2936, opened 2019-12-24
Use selected language from local storage, removed unused translations

amacar:
[ch3665]
[ch964]


common/v2/translations/translateRaw.ts
```diff
@@ -54,8 +56,9 @@ export function getTranslators() {
 }
 
 export function translateRaw(key: string, variables?: { [name: string]: string }): string {
-  // redux store isn't initialized in time which throws errors, instead we get the language selection from localstorage
-  const language = fallbackLanguage;
+  // we cannot access language from settings context, instead we get the language selection from localstorage
+  const { settings } = getCacheRaw();
+  const language = settings.language || fallbackLanguage;
```
FrederikBolding: We're trying to eliminate direct access to localstorage, such that we access through contexts exclusively. For this instance, we'll need to figure something out and make a decision.

Afaik this change was also initially delayed as the context is unavailable here.

@blurpesec @vilsbole

amacar: It would still be beneficial to merge translations files as soon as possible because it is hard to follow all new changes.

FrederikBolding: Agreed, but we need to figure out how we want to access the language setting.

amacar: I think that we can deal with it after this massive PR that will change LS access is about to be merged: https://github.com/MyCryptoHQ/MyCrypto/pull/2869

vilsbole: Answered https://github.com/MyCryptoHQ/MyCrypto/pull/2936#pullrequestreview-339282577

PR #2940, opened 2019-12-29
Added automatic scan for tokens after adding a new account

FrederikBolding:
- Added automatic scan for tokens after adding a new account


FrederikBolding:
[ch4118]

PR #2941, opened 2019-12-29
Fixed small styling issues

FrederikBolding:
- Fixed `ViewOnly` styling issues
- Fixed button overlap on `SafeT`


FrederikBolding:
[ch4310]

FrederikBolding:
Not sure if there is more to do with `ViewOnly` styling. I couldn't find a Figma design for it.

vilsbole:
Next steps would be to standardised the display of the error message as well as the breakpoint where the title wraps. In the absence of an appropriate design rule this is good enough.

PR #2948, opened 2020-01-08
Web3 improvements and fixes

FrederikBolding:
- Fixed web3 styling and copy issues
- Redid web3 provider detection
- Supported `ethereum` injected object
- Delayed computation of `HAS_WEB3_PROVIDER` to allow for late injection


FrederikBolding:
[ch3724]

LinFeiLong:
IMG_5524
The logo is missing on iOS chrome/safari

LinFeiLong:
The icon is fixed thx @FrederikBolding

common/v2/components/WalletUnlock/Web3ProviderInstall.tsx
```diff
@@ -25,34 +25,44 @@ function InstallTrunk() {
...
               <img src={AppStoreBadgeIMG} />
             </NewTabLink>
-            <NewTabLink href="https://play.google.com/store/apps/details?id=com.wallet.crypto.trustapp">
+            <NewTabLink
+              className="download-option"
+              href="https://play.google.com/store/apps/details?id=com.wallet.crypto.trustapp"
+            >
               <img src={GooglePlayBadgeIMG} />
             </NewTabLink>
           </div>
```
vilsbole: Can we refactor this please to use `config/wallets.ts` ?
- having the download links in the config file can helps us be more careful about url changes
we can add a param like:
```
  install: {
    getItLink: ?
    google: 'https://play.google.com/<app_id>'
  }
```
- icons are already available in the config file (so we update them in a single location)
- the translation labels as well (e.g lid)
- once we extract the data into the config file we can reduce the `tsx` by iterating on the list of valid wallets.

common/v2/config/wallets.ts
```diff
@@ -203,3 +207,6 @@ export const INSECURE_WALLETS: WalletSubType = filterObjectOfObjects(WALLETS_CON
 export const HARDWARE_WALLETS: WalletSubType = filterObjectOfObjects(WALLETS_CONFIG)(
   ({ type }: { type: WalletType }) => type === WalletType.HARDWARE
 );
+export const WEB3_WALLETS: WalletSubType = filterObjectOfObjects(WALLETS_CONFIG)(
+  ({ type }: { type: WalletType }) => type === WalletType.WEB3
+);
```
vilsbole: Nice 👍

common/v2/utils/environment.ts
```diff
@@ -2,4 +2,5 @@ export const IS_DEV: boolean = process.env.NODE_ENV === 'development';
 
 export const IS_PROD: boolean = process.env.NODE_ENV === 'production';
 
-export const HAS_WEB3_PROVIDER: boolean = window && 'web3' in window;
+export const HAS_WEB3_PROVIDER = (): boolean =>
+  window && ('web3' in window || 'ethereum' in window);
```
vilsbole: Since this is a function now, maybe we should change the syntax to camelCase or any other convention we decide on to not be surprised by `HAS_WEB3_PROVIDER()`. As you see fit

common/v2/utils/web3.ts
```diff
@@ -14,61 +12,46 @@ interface Web3ProviderInfo {
   walletId: WalletId;
 }
 
-const WEB3_CONFIGS: {
```
vilsbole: From what I see this object is only called within `config/wallets`.
Let's simplify and merge this into the config object.
That way we will also be able to merge  `getWeb3ProviderInfo` and `getWeb3Config`

common/v2/services/EthService/nodes/web3/node.ts
```diff
@@ -78,7 +78,9 @@ export async function setupWeb3Node() {
   const { ethereum } = window as any;
```
vilsbole: - To clean up as much as we can before going out of beta, can you delete  the uncalled methods `ensureWeb3NodeStillAvailable` & `isWeb3NodeAvailable` please?
- ```
      if (ethereum) {
        // Overwrite the legacy Web3 with the newer version.	
	       if ((window as any).Web3) {
           (window as any).web3 = new (window as any).Web3(ethereum);
         }
         //....
  ```
Nested conditionals start to smell and the repeated access to window as well. How about we use `HAS_WEB3_PROVIDER` or siblings for the control flow?

common/v2/components/WalletUnlock/Web3ProviderInstall.tsx
FrederikBolding: I'm not sure I agree with this. Since we show Coinbase Wallet etc, which we don't have as a wallet in `config/wallets.ts``. We shouldn't be adding extra wallets just to store their download links imo.

FrederikBolding: Unless we want to support Coinbase Wallet with an icon etc in WalletList

common/v2/services/EthService/nodes/web3/node.ts
FrederikBolding: Removed the methods. What do you prefer for the nested conditionals? The whole thing is quite hacky either way.

vilsbole: True. Let's leave it for now then.

common/v2/components/WalletUnlock/Web3ProviderInstall.tsx
vilsbole: It seems like we already do: L50 ``
I find it useful to have a single place to discover all the wallets we support as well as a single make changes.

common/v2/components/WalletList/WalletList.tsx
```diff
@@ -99,7 +99,9 @@ export default class WalletList extends PureComponent<Props> {
         )}
         <WalletsContainer>
           {validWallets.map((wallet: IStory) => {
-            const walletInfo = WALLETS_CONFIG[wallet.name];
+            const walletInfo =
+              wallet.name === WalletId.WEB3 ? getWeb3Config() : WALLETS_CONFIG[wallet.name];
```
vilsbole: We don't have any TypeGuards on `name` so WALLETS_CONFIG[wallet.name] may return `undefined`. Maybe use `wallet.type` and `wallet.id` here?
```
    wallet.type === WalletType.WEB3 ? getWeb3Config() : WALLETS_CONFIG[wallet.id];
```

FrederikBolding: `wallet` is an `IStory` which by definion has the `name` property with type `WalletId`. They're defined in https://github.com/MyCryptoHQ/MyCrypto/blob/822e1b899bda559c48e8b1da72d1e5fc138bc99d/common/v2/features/AddAccount/stories.tsx

I made the type of `wallets` clearer.

PR #2949, opened 2020-01-08
Fix Chart Hover [ch4108]

GuillaumeRx:
- Improved fix on the Chart for the Clubhouse [ch4108] card


common/v2/features/Dashboard/components/WalletBreakdown/BreakdownChart.tsx
```diff
@@ -113,6 +113,14 @@ export default class BreakdownChart extends Component<BreakdownChartProps> {
     this.props.setSelectedAssetIndex(index);
   };
 
+  public handleMouseLeave = (_: any) => {
+    if (this.isChartAnimating) {
+      return;
+    }
```
vilsbole: To facilitate readability prefer defending on a single line.
`if (this.isChartAnimating) return;`

PR #2952, opened 2020-01-13
Additional analytics implementations

wtzb:
[ch3372]

This PR adds the following analytics to the GAU codebase:

### Tokenscanner
- Track 'Scan' button click
- Track '+Add' button click
- Track when custom token is added
- Track when adding custom token is cancelled

### Add Account
- Track which wallet option was clicked
- If web3, track which provider or none
- Track selected network

### Wallet breakdown
- Track if no accounts are added yet

...

...

...

This PR is WIP.


vilsbole:
How about merging this PR now before it becomes too stale?

wtzb:
@vilsbole in favor of that :+1:

common/v2/components/WalletUnlock/Web3Provider.tsx
```diff
@@ -44,6 +45,11 @@ class Web3ProviderDecrypt extends Component<Props & ISettingsContext & INetworkC
   public render() {
     const { web3Unlocked, web3ProviderSettings: provider } = this.state;
     const isDefault = provider.id === WalletId.WEB3;
+    AnalyticsService.instance.track(
```
vilsbole, line 48: 1. With React a component can re-renders every time a state variable of a parent changes. We have little control on how many times a component will re-render. This means that in the same session, `.track` can be called multiple times.
2. To avoid this the `.track` call should be made within a hook (on first render) or a lifecycle event.
3. As a general rule, all side effects should be made within a hook. Since `.track` makes an external api call it's a side-effect https://en.wikipedia.org/wiki/Side_effect_(computer_science)

The alternative would be to place the call on click instead of component render.

common/v2/components/WalletList/WalletList.tsx
```diff
@@ -105,7 +110,7 @@ export const WalletList = ({ wallets, onSelect, showHeader, calculateMargin }: P
               icon={walletInfo.icon}
               description={translateRaw(walletInfo.description)}
               margin={calculateMargin && calculateMargin(index)}
-              onClick={() => onSelect(wallet.name)}
+              onClick={() => onWalletClick(wallet.name)}
```
vilsbole, line 113: Prefer naming the event handler by what it does instead of when https://medium.com/@vdsabev/simple-naming-tips-for-event-handlers-aeba8cf338ba
e.g `on` is reserved for the prop.

common/v2/components/WalletUnlock/Web3ProviderInstall.tsx
```diff
@@ -72,6 +73,10 @@ function InstallMetaMask() {
 }
 
 function Web3ProviderInstall() {
+  useEffect(() => {
+    AnalyticsService.instance.track(ANALYTICS_CATEGORIES.ADD_WEB3_ACCOUNT, `No provider detected`);
+  }, [])
```
vilsbole, line 78: How about a custom hook for analytics? Something we can use like:
```
const { track } = useAnalytics()
useFirstMount(() => { track(ANALYTICS_CATEGORIES.ADD_WEB3_ACCOUNT, 'No provider detected' )})
```

common/v2/features/NoAccounts/NoAccounts.tsx
```diff
@@ -94,23 +95,33 @@ const ButtonGroup = styled.div`
...
-      <Header>{translate('NO_ACCOUNTS_HEADER')}</Header>
-      <Description>{translate('NO_ACCOUNTS_DESCRIPTION')}</Description>
-      <ButtonGroup>
-        <Link to={ROUTE_PATHS.ADD_ACCOUNT.path}>
-          <WhiteButtonFirst secondary={true}>{translate('ADD_EXISTING_ACCOUNT')}</WhiteButtonFirst>
-        </Link>
-        <Link to={ROUTE_PATHS.SETTINGS_IMPORT.path}>
-          <PrimaryButton secondary={true}>{translate('SETTINGS_IMPORT_HEADING')}</PrimaryButton>
+export const NoAccounts = () => {
+  useEffect(() => {
```
vilsbole, line 99: We have `useFirstMount` available under `v2/vendor`

common/v2/components/WalletUnlock/Web3ProviderInstall.tsx
wtzb: This sounds like a solid idea - we would be able to replace all current calls with this approach, right?

common/v2/components/WalletUnlock/Web3Provider.tsx
wtzb: As hooks aren't available in classes, would `componentDidMount` suffice? Or would it be better to place this call on click instead?

common/v2/components/WalletList/WalletList.tsx
wtzb: Sound :+1:

common/v2/features/NoAccounts/NoAccounts.tsx
wtzb: How would I use this, exactly?

common/v2/components/WalletUnlock/Web3Provider.tsx
wtzb: Clear! Changing this to be on click.

common/v2/components/WalletUnlock/Web3ProviderInstall.tsx
wtzb: As Povio will be picking up the remaining analytics tasks after this PR is merged, do you think it would be wise to assign this as a separate task to them? Preferably, I would like to have this PR merged ASAP, so Povio can start working on the rest :+1:

vilsbole: Ok, I'll merge now then

PR #2956, opened 2020-01-14
Change webpack target for Electron production build

kunstefix:
Previous target value `electron-renderer` instructs webpack to use NodeTargetPlugin which should be only used if budle is running in Node.js environment.
Since we have `nodeIntegration` disabled, the bundle target should be set to `web`.

### DIFF
--- a/webpack_config/electron.production.js
+++ b/webpack_config/electron.production.js
@@ -20,7 +20,7 @@ const main = merge.smart(electronMain, {
 });
 
 const render = merge.smart(electronRender, {
-  target: 'electron-renderer',
+  target: 'web',
 
   output: {
     path: path.join(config.path.output, 'electron-js'),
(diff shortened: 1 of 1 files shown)


LinFeiLong:
@kunstefix is there any consequence to this ?
Could you add a documentation source/link ? pls

LinFeiLong:
Tested locally and with the macOS artifact, it is good for me.

kunstefix:
> @kunstefix is there any consequence to this ?
> Could you add a documentation source/link ? pls

Webpack docs: https://webpack.js.org/configuration/target/
So far i did not notice or find consequences in app behaviour, error logs or documentation.

webpack_config/electron.production.js
```diff
@@ -20,7 +20,7 @@ const main = merge.smart(electronMain, {
 });
 
 const render = merge.smart(electronRender, {
-  target: 'electron-renderer',
+  target: 'web',
```
vilsbole: Can we please add a comment as documentation for our future selfs.
It's important to know that this is related to `nodeIntegration`

vilsbole: Thanks!

PR #2958, opened 2020-01-14
change block explorer for ETC

tayvano:
needs to be tested to ensure both transaction and address links work


gitr0n1n:
Hey guys. Can you check that the "confirm sent transaction" popup links when you send ETC are updated? I believe they're still routing to gastoken/emerald payment. Thank you.

gitr0n1n:
Hey guys. Can you check that the "confirm sent transaction" popup links when you send ETC are updated? I believe they're still routing to Gastracker and emerald payment. Thank you.

tayvano:
We need to do a couple more things before it goes live. (Security checks, update legacy branch, etc.) At best, it may be live this Thursday. At worst, next Thursday.

PR #2959, opened 2020-01-14
Update styled components

Mrtenz:
`styled-components` v5 has some pretty significant performance improvements: https://medium.com/styled-components/announcing-styled-components-v5-beast-mode-389747abd987


package.json
```diff
@@ -49,6 +49,7 @@
     "react-copy-to-clipboard": "5.0.1",
     "react-dom": "16.8.6",
     "react-hot-loader": "4.12.15",
+    "react-is": "16.12.0",
```
vilsbole: Is this intentional. The installation guide of `styled-components` doesn't mention `react-is`

Mrtenz: It's a peer dependency of styled-components and is mentioned here: https://github.com/styled-components/styled-components#upgrading-from-v4

PR #2963, opened 2020-01-15
Update ENS validation

MrLuit:
- Changed minimum ENS name length from 7 to 3 characters

### DIFF
--- a/common/v2/services/EthService/ens/validators.ts
+++ b/common/v2/services/EthService/ens/validators.ts
@@ -26,7 +26,7 @@ export function getValidTLDsForChain(chainId: number): ITLDCollection {
 export function isValidENSName(str: string) {
   try {
     return (
-      str.length > 6 && !str.includes('.') && normalise(str) !== '' && str.substring(0, 2) !== '0x'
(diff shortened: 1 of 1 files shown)


common/v2/services/EthService/ens/validators.ts
```diff
@@ -26,7 +26,7 @@ export function getValidTLDsForChain(chainId: number): ITLDCollection {
 export function isValidENSName(str: string) {
   try {
     return (
-      str.length > 6 && !str.includes('.') && normalise(str) !== '' && str.substring(0, 2) !== '0x'
+      str.length >= 3 && str.includes('.') && normalise(str) !== '' && str.substring(0, 2) !== '0x'
```
vilsbole: - Is there a specific reason to choose a length of 3?
- Can you add a unit test so we can document the intended behaviour.

MrLuit: The minimum amount of characters for an ENS name is 3 (https://medium.com/the-ethereum-name-service/announcing-the-ens-3-6-character-auction-31bfee7a4eab).

I'll add a test unit 👍

MrLuit:
Just realised my mistake, it's actually supposed to be `> 6` because `str` also includes `.eth` which adds 4 characters.

PR #2964, opened 2020-01-15
Move registerProtocol method from preload, remove empty preload

kunstefix:

### DIFF
--- a/electron-app/main/index.ts
+++ b/electron-app/main/index.ts
@@ -2,6 +2,9 @@ import '@babel/polyfill';
 import { app } from 'electron';
 import { registerServer } from 'shared/enclave/server';
 import getWindow from './window';
+import { registerProtocol } from 'shared/enclave/preload';
+
+registerProtocol(); // Needs be called before 'ready' event is emitted
 
 // Quit application when all windows are closed
 app.on('window-all-closed', () => {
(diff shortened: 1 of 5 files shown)


LinFeiLong:
Interesting, what is it doing ?

kunstefix:
> Interesting, what is it doing ?

Solving story [4334](https://app.clubhouse.io/mycrypto/story/4334/cannot-read-property-registerschemesasprivileged-of-undefined)

LinFeiLong:
Do you feel it needs to be replaced ?
Or we just need to get rid of it ?

kunstefix:
> Do you feel it needs to be replaced ?
> Or we just need to get rid of it ?

The fix is to move method call out of preload. After that the preload is empty so it can be removed.

LinFeiLong:
Ok perfect

electron-app/main/index.ts
```diff
@@ -20,7 +23,7 @@ app.on('activate', () => {
 
 // Create main BrowserWindow when electron is ready
 app.on('ready', () => {
-  getWindow();
+  getWindow().webContents.openDevTools();
```
LinFeiLong: I don't mind keeping it for the moment.
I'll add a comment though.

PR #2980, opened 2020-01-22
Fix appState not syncing correctly to LS

amacar:
[ch4478]


common/v2/vendor/react-use/useLocalStorage.ts
```diff
@@ -1,7 +1,6 @@
-import { useEffect, useState } from 'react';
+import { noOp } from 'v2/utils';
```
vilsbole: Since this is no longer a hook, lets simplify the code.
We can turn merge it with `useDatabase` and turn it into a DatabaseService.

common/v2/services/Store/DataManager/DataProvider.tsx
```diff
@@ -53,17 +53,20 @@ export const DataProvider: React.FC = ({ children }) => {
...
-  }, [appState]);
+  const syncDb = useCallback(
+    state => {
+      console.debug('Updating LocalStorage');
+      updateDb(deMarshallState(state));
+    },
+    [appState]
+  );
   // By default we sync no more than once a minute
   useThrottleFn(syncDb, 60000, [appState]);
```
vilsbole: As discussed let's reduce the the throttle to 30000

amacar: Done

common/v2/vendor/react-use/useLocalStorage.ts
amacar: Hook useLocalStorage merged with useDatabase transformed into DatabaseService.

common/v2/services/Store/DataManager/DataProvider.tsx
```diff
@@ -53,17 +53,20 @@ export const DataProvider: React.FC = ({ children }) => {
...
-  const syncDb = useCallback(() => {
-    console.debug('Updating LocalStorage');
-    updateDb(deMarshallState(appState));
-  }, [appState]);
+  const syncDb = useCallback(
+    state => {
+      console.debug('Updating LocalStorage');
+      updateDb(deMarshallState(state));
+    },
+    [appState]
```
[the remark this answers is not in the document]
amacar: Callback and deps removed.

common/v2/services/Store/DataManager/index.ts
```diff
@@ -1,2 +1,2 @@
 export { IDataContext, DataProvider, DataContext } from './DataProvider';
-export { useDatabase } from './useDatabase';
+export { DatabaseService } from './DatabaseService';
```
vilsbole: This is only used within the `DataProvider`.
Let's keep a clean api for the module and remove the export.

amacar:
After updates also: [ch4320]

- reduce sync time (maybe by using multiple keys or with indexeddb?) => reduced to 30s
- review and test sync event (browser close? computer shut down?) => browser close is ok, computer shutdown can't be detected => but we are saving every 30s or on appState change
- sync only if unsaved changes. => done by useThrottleFn

PR #2984, opened 2020-01-22
Use uuid as uniq prop when summing asset balances

vilsbole:
[ch4173](https://app.clubhouse.io/mycrypto/story/4173/tusd-balance-is-wrong)

Now that we have uuid for tokens we should use them when we sum the users asset balances. 
ref: https://mycryptohq.slack.com/archives/G8G1S4TG9/p1579654425069200

- use `uuid` when summing asset balances
- add tests
- use `uuid` as `key` when displaying token list


FrederikBolding:
I think that issue is present in beta without this change as well.

blurpesec:
> I think that issue is present in beta without this change as well.

My beta doesnt :3
image

FrederikBolding:
nvm! confused it with something else.

PR #2995, opened 2020-01-28
Fix styling for Send: Sign TX and TxBroadcasted [ch4024]

GuillaumeRx:
* Added Asset Icon next to Sent amount

* Created LinkOut Component for TxID

* Added color props to Button

* Various Style corrections

Story details: https://app.clubhouse.io/mycrypto/story/4024


common/v2/components/LinkOut.tsx
```diff
@@ -0,0 +1,43 @@
...
+interface Props {
+  text: string;
+  link: string;
+  truncate?(text: string): string;
+}
+
+const LinkOut = ({ text, link, truncate }: Props) => {
+  const content = truncate ? truncate(text) : text;
+  return (
+    <LinkOutWrapper href={link} target="_blank">
```
vilsbole: Every `target="_blank">` should be accompanied with `noreferrerr`  & `nooopener`
https://developers.google.com/web/tools/lighthouse/audits/noopener

I'm for merging all our different link components into a single one.
We can continue discussing on slack

common/v2/components/Button.tsx
```diff
@@ -4,15 +4,21 @@ import { Button } from '@mycrypto/ui';
...
+    props.fullwidth === true &&
+    css`
+      width: 100%;
+      margin-top: 1rem;
+    `}
+  ${props =>
+    props.color &&
+    css`
+      color: ${props.color};
+    `}
```
vilsbole: For a single property it's shorter to write:
```
color: ${props => props.color ? props.color : inherit }
```

vilsbole:
FYI: 
- Prefer rebase over merge. It's more respectful for the other devs (master is the source of truth), and generates a more readable history.

common/v2/components/LinkOut.tsx
GuillaumeRx: Thanks for the info, gonna add this asap !

Let's discuss ont this.

vilsbole:
Update is missing `noopenner` atrr.
Merging in favour of refactoring of link components

GuillaumeRx:
According to : https://developers.google.com/web/tools/lighthouse/audits/noopener

> rel="noreferrer" attribute has the same effect [as rel="noopener"], but also prevents the Referer header from being sent to the new page. See Link type "noreferrer".

Should I add it also ?

vilsbole:
Sorry my mistake. Good call!

PR #2997, opened 2020-01-30
Update legacy tokenlist

wtzb:


vilsbole:
How are the updates from legacy run?
Does it use out token list?

wtzb:
@vilsbole they are ran via https://github.com/MyCryptoHQ/parse-eth-tokens

PR #2999, opened 2020-01-31
Swap input sanity checks

409H:
🎉 🎉 🎉

## Description

Added some input sanity validation to reduce the will-be-error responses from dexag and added logic to show a more logical error message.

* If the user clears one of the inputs, clear all swap inputs and don't send a request to dexag that will result in a 400 bad request
* If the user inputs a send/receive input of `0`, show a better error message and don't send a request to dexag that will result in a 400 bad request
* If the user inputs a send/receive input of `-n`, show a better error message and don't send a request to dexag that will result in a 400 bad request

We are now doing sanity checks to make sure we _should_ send a request to DEXAG without it intentionally resulting in a 400 bad request.

image

## Changes

* Sanity checks on input with the swap input event handlers

## Quality Assurance

### DIFF
--- a/common/v2/features/SwapAssets/stateFactory.tsx
+++ b/common/v2/features/SwapAssets/stateFactory.tsx
@@ -100,6 +100,24 @@ const SwapFlowFactory: TUseStateReducerFactory<SwapState> = ({ state, setState }
       return;
     }
 
+    if(value.length === 0) {
+      setState((prevState: SwapState) => ({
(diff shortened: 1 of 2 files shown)


409H:
Pushed branch to `MyCryptoHQ` namespace instead as I'm having issues with tests after rebasing... New PR: https://github.com/MyCryptoHQ/MyCrypto/pull/3011

PR #3000, opened 2020-01-31
added wallet type to account list

blurpesec:
## Description

Added the wallet type to AccountList in Dashboard

image

### DIFF
--- a/common/v2/components/AccountList.tsx
+++ b/common/v2/components/AccountList.tsx
@@ -3,7 +3,7 @@ import styled, { css } from 'styled-components';
 import { Button, Identicon } from '@mycrypto/ui';
 
 import { translateRaw } from 'v2/translations';
-import { ROUTE_PATHS, Fiats } from 'v2/config';
+import { ROUTE_PATHS, Fiats, WALLETS_CONFIG } from 'v2/config';
 import {
   EthAddress,
   CollapsibleTable,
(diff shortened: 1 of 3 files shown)


common/v2/database/v1.0.0/generateDefaultValues.spec.ts
```diff
@@ -36,7 +36,7 @@ describe('Schema', () => {
 
     it('adds Nodes to each Network', () => {
       const nodes = toArray(defaultData[LSKeys.NETWORKS]).flatMap(n => n.nodes);
-      expect(nodes.length).toBe(43);
+      expect(nodes.length).toBe(44);
```
vilsbole: This is surprising.
I don't understand how this PR can modify the node check.

common/v2/components/AccountList.tsx
```diff
@@ -30,9 +30,34 @@ import { TUuid } from 'v2/types/uuid';
...
     font-weight: bold;
-  })
+  }
+`;
+
+const WalletTypeLabelContainer = styled.div`
+  display: block;
+`;
+
+const WalletTypeLabel = styled.p`
```
vilsbole: Identicon and content are misaligned.
This is solved if you remove margin-bottom of the paragraph tag.

common/v2/components/AccountList.tsx
```diff
@@ -30,9 +30,34 @@ import { TUuid } from 'v2/types/uuid';
...
+  background: #b7bfc6;
```
vilsbole: Use COLORS

common/v2/components/AccountList.tsx
```diff
@@ -30,9 +30,34 @@ import { TUuid } from 'v2/types/uuid';
...
+  display: inline-block;
+  text-align: center;
+  border-radius: 600px;
+  font-size: 0.8rem;
+  padding: 3px 6px;
+  color: ${COLORS.WHITE};
+  @media (min-width: ${BREAK_POINTS.SCREEN_SM}) {
+    font-size: 0.6rem;
+  }
```
vilsbole: There is a font jump when changing screen size.
IMHO, you can set the font-size once using `em`.

common/v2/components/AccountList.tsx
```diff
@@ -228,7 +253,12 @@ function buildAccountTable(
       const bodyContent = [
         <Label key={index}>
           <SIdenticon address={account.address} />
-          <Typography value={label} />
+          <LabelWithWallet>
+            <Typography value={label} />
+            <WalletTypeLabelContainer>
```
vilsbole: This container div seems unnecessary, but we can address it later.

common/v2/database/v1.0.0/generateDefaultValues.spec.ts
vilsbole: Solved through #3003

common/v2/components/AccountList.tsx
blurpesec: We don't need a styled div, but we do seem to need the div still so that it doesn't stretch to fill the whole length.

PR #3006, opened 2020-02-01
Error Handling in case of React Rendering errors

FrederikBolding:
Handled React rendering errors by showing a bar at the top to report the error and redirecting to show the homepage.

- [x] Redirect properly
- [x] Styling


FrederikBolding:
[ch4563]

common/AppRouter.tsx
```diff
@@ -39,14 +40,17 @@ export const AppRouter = () => {
...
-                    <Switch>
-                      <Route path={ROUTE_PATHS.ROOT.path} component={Home} exact={true} />
-                      <Route path={ROUTE_PATHS.HOME.path} component={Home} exact={true} />
-                      {APP_ROUTES.filter(route => !route.seperateLayout).map((config, idx) => (
-                        <PrivateRoute key={idx} {...config} />
-                      ))}
-                      <Route component={PageNotFound} />
-                    </Switch>
+                    {shouldShowError() && <Home />}
+                    {!shouldShowError() && (
```
blurpesec: Lines 43 & 44 can be *terned* into a ternary ;)

blurpesec: Or maybe, it may be a good idea to instead of just showing <Home /> under whatever the endpoint the user attempts to navigate to, maybe redirect to `/home`.

FrederikBolding: Yeah, I'm beginning to think we should just redirect on error instead, that way we don't lock down the app.

PR #3014, opened 2020-02-03
Fix broadcast transaction

amacar:
Bump `ethereumjs-util` to support EIP-155 when validating a signature. Due to version bump sha3 export was removed => we changed it with keccak because methods are the same (https://github.com/ethereumjs/ethereumjs-util/commit/9396416913125711e526a05591e3ce8471c3a528#diff-168726dbe96b3ce427e7fedce31bb0bcL251)

Due to the bump of `ethereumjs-tx` in one of PRs (https://github.com/MyCryptoHQ/MyCrypto/commit/472395f3b2f7d9c3a18720c9dfd73c707adfe333) we added chainId param when parsing signed Tx, swap nonexisting params with newer ones (_chainId = getChainId(), from = getSenderAddress())


vilsbole:
When I head to https://mycryptobuilds.com/pr/3014/#/broadcast-transaction and focus in the field the console displays the following error:

Screenshot 2020-02-05 at 15 03 18

LuckyLuker:
[ch4791]

common/v2/features/SendAssets/stateFactory.tsx
```diff
@@ -175,8 +176,8 @@ const TxConfigFactory: TUseStateReducerFactory<State> = ({ state, setState }) =>
         ? {
             ...state.txConfig,
             hash: payload,
-            to: state.txConfig.senderAccount.address,
-            from: state.txConfig.receiverAddress
+            to: state.txConfig.receiverAddress,
+            from: state.txConfig.senderAccount.address
```
blurpesec: Nice catch :)

common/v2/features/BroadcastTransaction/components/ConfirmTransaction.tsx
```diff
@@ -1,6 +1,8 @@
 import React, { useState, useContext } from 'react';
 import styled from 'styled-components';
 import BN from 'bn.js';
+import { Transaction } from 'ethereumjs-tx';
+import { addHexPrefix } from 'ethereumjs-util';
```
blurpesec: Is this ConfirmTransaction component specific to Broadcast Transaction flow? Can it be folded in using the one from `v2/components/TransactionFlow` ?

If not, do the changes made in this Confirm component need to also be implemented in `v2/components/TransactionFlow/ConfirmTransaction` ?

common/v2/services/EthService/utils/units.ts
```diff
@@ -115,7 +115,7 @@ const totalTxFeeToString = (gasPriceEther: string, gasLimit: string): string =>
   parseFloat(fromWei(totalTxFeeToWei(gasPriceEther, gasLimit), 'ether')).toFixed(6);
 
 const totalTxFeeToWei = (gasPriceEther: string, gasLimit: string): Wei =>
-  new BN(parseInt(gasPriceEther, 10) * parseInt(gasLimit, 10));
+  new BN(gasPriceEther).mul(new BN(gasLimit));
```
blurpesec: Nice catch :D

common/v2/services/EthService/utils/signing.ts
```diff
@@ -41,7 +41,7 @@ export function verifySignedMessage({ address, msg, sig, version }: ISignedMessa
   }
   //TODO: explain what's going on here
   sigb[64] = sigb[64] === 0 || sigb[64] === 1 ? sigb[64] + 27 : sigb[64];
-  const hash = version === '2' ? hashPersonalMessage(toBuffer(msg)) : sha3(msg);
+  const hash = version === '2' ? hashPersonalMessage(toBuffer(msg)) : keccak(msg);
```
blurpesec: Will this allow us to still verify messages signed with `sha3` instead of `keccak`? If this was a recent change, we'll want to support that backwards compatibility, if it's not too much effort.

amacar: They just remove an export in newer version, functionality is still the same.
image

common/v2/features/BroadcastTransaction/components/ConfirmTransaction.tsx
amacar: It was made before the state factory pattern was introduced; it is a wrapper with some logic around `ConfirmTransaction` from TransactionFlow. I can probably move some functions to the `BroadcastFlow` file and use pure `ConfirmTransaction` component.

common/v2/services/EthService/utils/signing.ts
blurpesec: aye. gotcha. thanks c:

common/v2/features/BroadcastTransaction/components/ConfirmTransaction.tsx
amacar: I reused the `GeneralStepper` component, moved the logic from custom `ConfirmTransaction` component to `stateFactory` and get rid of it.

common/v2/features/BroadcastTransaction/stateFactory.tsx
```diff
@@ -0,0 +1,148 @@
...
+  const { displayToast, toastTemplates } = useContext(ToastContext);
+
+  const handleNetworkChanged = (network: NetworkId) => {
+    setState((prevState: State) => ({
+      ...prevState,
+      network
+    }));
+  };
+
+  const handleSendClicked = (cb: any) => {
```
blurpesec: Except the first line, ` const signedTx = hexlify(payload);` (because the user-input transaction should already be in hex-format), you could extract the below out to a function and use it instead of re-implementing it: https://github.com/MyCryptoHQ/MyCrypto/blob/master/common/v2/features/SendAssets/stateFactory.tsx#L129-L175

A cursory glance tells me you may have also missed ERC20 transfers vs ETH transactions in your implementation.

common/v2/features/BroadcastTransaction/stateFactory.tsx
```diff
@@ -0,0 +1,148 @@
...
+    setState((prevState: State) => ({
+      ...prevState,
+      networkSelectError: '',
+      txConfig
+    }));
+
+    cb();
+  };
+
+  const handleSignedTxChanged = (signedTransaction: string) => {
```
blurpesec: it might be better to have this as an on-submit trigger and just add validation to input field (isHexString && isCorrectNumberOfCharacters). This would enable you to do the above comment as well ^.

amacar:
> 1. This QR code probably shouldn't be here. Theres also nothing indicating what it's for

This feature was ported directly from old mycrypto page (QR code is there). I can remove it if you agree; I also don't know what is the point of it.
image

common/v2/features/BroadcastTransaction/stateFactory.tsx
amacar: Extracted code into another util function. Reusing this function also fixes second screenshot.
>2. There seems to be a problem resolving address book entry on Confirm Panel:

common/v2/features/BroadcastTransaction/stateFactory.tsx
amacar: That makes sense, moved changeHandler from stateFactory to component together with signature validation.

PR #3018, opened 2020-02-04
Lock screen on demand

kunstefix:
- call lock screen from menu icon
- update label
- remove obsolete AccountScreen and UnlockScreen components


vilsbole:
code is good.
UX of locked state should be reviewed by product team

PR #3022, opened 2020-02-06
Better ENS Validation

409H:
[CH4517]

🎉 🎉 🎉

## Description

* Fixed bug that caused the user to be able to sign a transaction to a non-intended address if the ENS name was invalid (description in comment on ch4517)
* Adjusted logic to allow for more ENS TLDs so we can send to `*.argent.xyz` addresses
* Adjusted logic to check if a value is valid ENS name
   * Remove `0x` startswith condition because you can have a name like `0xharry.eth`
   * Remove strlen condition because short names released and we should be relying on onchain data. Too tight of a restriction for read-only interaction.

## Changes

* Added new TLD to the TLD dictionary
* Modified sanity checks

## Quality Assurance

### DIFF
--- a/common/v2/features/SendAssets/components/SendAssetsForm.tsx
+++ b/common/v2/features/SendAssets/components/SendAssetsForm.tsx
@@ -179,7 +179,9 @@ export default function SendAssetsForm({ txConfig, onComplete }: IStepComponentP
         .test(
           'check-eth-address',
           translateRaw('TO_FIELD_ERROR'),
-          value => isValidETHAddress(value) || isValidENSName(value)
+          value =>
+            isValidETHAddress(value) ||
+            (isValidENSName(value) && UnstoppableResolution.isValidDomain(value))
         )
         // @ts-ignore Hack as Formik doesn't officially support warnings
         // tslint:disable-next-line
--- a/common/v2/features/SendAssets/components/SendAssetsForm.tsx
+++ b/common/v2/features/SendAssets/components/SendAssetsForm.tsx
@@ -48,7 +48,8 @@ import {
   isValidPositiveNumber,
   isTransactionFeeHigh,
   isChecksumAddress,
-  isBurnAddress
+  isBurnAddress,
+  isValidENSName
 } from 'v2/services/EthService';
 import UnstoppableResolution from 'v2/services/UnstoppableService';
 import { fetchGasPriceEstimates, getGasEstimate } from 'v2/services/ApiService';
@@ -178,7 +179,7 @@ export default function SendAssetsForm({ txConfig, onComplete }: IStepComponentP
         .test(
           'check-eth-address',
           translateRaw('TO_FIELD_ERROR'),
-          value => isValidETHAddress(value) || UnstoppableResolution.isValidDomain(value)
+          value => isValidETHAddress(value) || isValidENSName(value)
         )
         // @ts-ignore Hack as Formik doesn't officially support warnings
         // tslint:disable-next-line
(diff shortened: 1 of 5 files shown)


FrederikBolding:
Will look at this today or tomorrow!

409H:
Apart from the following UI bugs and bug out-of-scope from this PR, from commit https://github.com/MyCryptoHQ/MyCrypto/pull/3022/commits/d23d5610f324d3b46f98fe18c5be5bf82a136e4c it seems to work as intended

* You can only send ETH to an ENS/UD domain - else you get "<Asset> is not supported!" error
* On view load, the error message "Could not resolve the domain" is present - should not be there
* If you input a 0x address, you still have the "Resolved address: <address>" note - it should not be there

I think we need tickets on updating the UD implementation/handling... UI element behaviour has changed since I started this PR

common/v2/services/EthService/ens/__tests__/validators.spec.ts
```diff
@@ -4,9 +4,7 @@ describe('isValidENSName', () => {
   const valid = ['bob.eth', 'edouard.eth'];
 
   const invalid = [
-    'et.eth', // second level domain must be longer than 2
```
vilsbole: Looking at https://opensea.io/ens it seems that 2 char ens names are still invalid.
Why do we remove this test?

common/v2/features/SendAssets/components/SendAssetsForm.tsx
```diff
@@ -178,7 +179,7 @@ export default function SendAssetsForm({ txConfig, onComplete }: IStepComponentP
         .test(
           'check-eth-address',
           translateRaw('TO_FIELD_ERROR'),
-          value => isValidETHAddress(value) || UnstoppableResolution.isValidDomain(value)
+          value => isValidETHAddress(value) || isValidENSName(value)
```
vilsbole: Do we really cover the same validation surface as UnstoppableResolution?

common/v2/services/EthService/ens/__tests__/validators.spec.ts
FrederikBolding: See discussion here: https://mycryptohq.slack.com/archives/G8G1S4TG9/p1581453596290900

If you feel the other way around, I'm open to changing it.

common/v2/features/SendAssets/components/SendAssetsForm.tsx
FrederikBolding: You're right, we should probably combine the efforts.

common/v2/services/EthService/ens/__tests__/validators.spec.ts
409H: I suggested the edit here. The restrictions (in this case, 2 char) should only be enforced on the UI when registering a name - not on readonly lookup, as we should rely on the onchain registery response instead of implementing our own rulesets

common/v2/features/SendAssets/components/SendAssetsForm.tsx
```diff
@@ -310,6 +313,11 @@ export default function SendAssetsForm({ txConfig, onComplete }: IStepComponentP
                 value: unstoppableAddress
               });
             } catch (err) {
+              // Force the field value to error so that isValidAddress is triggered!
+              setFieldValue('address', {
+                ...values.address,
```
vilsbole: This is strange. It looks like we're spreading the address which will result in `{ 0: '0', x: 'x',  ...etc }`

FrederikBolding: Address is an object which contains `display` and `value`.

PR #3029, opened 2020-02-11
Swap: Adjust receive amount on the confirm view to include MyCrypto's 2.5% Fee [ch4515]

GuillaumeRx:
### Description

- Added 2.5% on base asset amount when entering target amount in desired asset.

- Removed 2.5% of base asset amount when calculating the target amount of desired asset.

[Story details](https://app.clubhouse.io/mycrypto/story/4515)


GuillaumeRx:
Going back to initial solution since Michael one seems to bomb out on certain cases

common/v2/utils/convert.ts
```diff
@@ -20,3 +20,23 @@ export const convertToFiat = (userViewableBalance: number, rate: number = 1): nu
 
 export const weiToFloat = (wei: BigNumber, decimal?: number): number =>
   parseFloat(fromTokenBase(new BN(wei.toString()), decimal || DEFAULT_ASSET_DECIMAL));
+
+export const withCommission = ({
+  amount,
+  rate,
+  substract
+}: {
+  amount: number;
```
vilsbole: SHOULD be a `ethers.js` BigNumber

common/v2/utils/convert.ts
```diff
@@ -20,3 +20,23 @@ export const convertToFiat = (userViewableBalance: number, rate: number = 1): nu
...
+  rate: number;
+  substract?: boolean;
+}): number => {
+  const amountBN = parseEther(amount.toString());
+  const commission = substract ? (100 - rate) / 100 : (100 + rate) / 100;
+  const splitCommisson = commission.toString().split('.');
+  const decimals = splitCommisson.length > 1 ? splitCommisson[1].length : 0;
+  const commissionDivisor = Math.pow(10, decimals);
+  const commissionBN = bigNumberify(Math.round(commission * commissionDivisor));
```
vilsbole: A general comment explaining the operation in plain english is welcome.

common/v2/features/SwapAssets/stateFactory.tsx
```diff
@@ -131,7 +136,10 @@ const SwapFlowFactory: TUseStateReducerFactory<SwapState> = ({ state, setState }
       setState((prevState: SwapState) => ({
         ...prevState,
         isCalculatingFromAmount: false,
-        fromAmount: (Number(value) * price).toString(),
+        fromAmount: withCommission({
+          amount: Number(value) * price,
```
vilsbole: Danger! See below.

common/v2/utils/convert.spec.ts
```diff
@@ -51,3 +52,54 @@ describe('it converts balance to fiat', () => {
...
+
+describe('it Remove / Add commission from amount', () => {
+  it('remove commission from decimal amount', () => {
+    const expected = 195;
+    const amount = 200;
+    const converted = withCommission({
+      amount,
+      rate: MYC_DEXAG_COMMISSION_RATE,
+      substract: true
+    });
```
vilsbole: These tests will break when the RATE changes.
Now that the utility accepts a dynamic rate, we can use it to calculate the expected value and make our tests more robust.  Maybe with another math library in devDeps?

common/v2/features/SwapAssets/stateFactory.tsx
```diff
@@ -188,7 +196,11 @@ const SwapFlowFactory: TUseStateReducerFactory<SwapState> = ({ state, setState }
       setState((prevState: SwapState) => ({
         ...prevState,
         isCalculatingToAmount: false,
-        toAmount: (Number(value) * price).toString(),
+        toAmount: withCommission({
+          amount: Number(value) * price,
```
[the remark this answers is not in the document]
GuillaumeRx: Yes of course ! Could be a good Idea to rename it since we're going to use it for something else than fiat ? Any ideas ?

common/v2/utils/convert.ts
GuillaumeRx: it's going to change when I will be using convertToFiat :)

common/v2/features/SwapAssets/stateFactory.tsx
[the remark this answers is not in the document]
vilsbole: how about just `convert` which takes an asset and a conversion rate?

common/v2/features/SwapAssets/stateFactory.tsx
```diff
@@ -188,7 +197,11 @@ const SwapFlowFactory: TUseStateReducerFactory<SwapState> = ({ state, setState }
       setState((prevState: SwapState) => ({
         ...prevState,
         isCalculatingToAmount: false,
-        toAmount: (Number(value) * price).toString(),
+        toAmount: withCommission({
+          amount: convert(Number(value), price),
```
vilsbole: 😍

PR #3030, opened 2020-02-11
Enabled signed tx in confirm panel

blurpesec:
### Description
Ran into an issue of there being no signed tx field in the ConfirmTransaction panel for hardware wallets.

### Steps to Test:
1) Navigate to https://mycryptobuilds.com/pr/3030/ and add a hardware wallet address.
2) Attempt to send a transaction.
3) After signing the transaction, on the confirm transaction step, click the "more details" section and see if a signed transaction field appears. does it?

### DIFF
--- a/common/v2/features/SendAssets/stateFactory.tsx
+++ b/common/v2/features/SendAssets/stateFactory.tsx
@@ -30,6 +30,7 @@ import { ProviderHandler } from 'v2/services/EthService';
 
 import { TStepAction } from './types';
 import { processFormDataToTx, decodeTransaction } from './helpers';
+import { bufferToHex } from 'ethereumjs-util';
 
 const txConfigInitialState = {
   tx: {
@@ -125,7 +126,14 @@ const TxConfigFactory: TUseStateReducerFactory<State> = ({ state, setState }) =>
       .finally(cb);
   };
 
-  const handleSignedTx: TStepAction = (payload: ISignedTx, cb) => {
+  const handleSignedTx: TStepAction = (payload: any, cb) => {
+    let signedTx = payload;
+    // Used when signedTx is a buffer instead of a string.
+    // Hardware wallets return a buffer.
+    if (typeof signedTx === 'object') {
+      signedTx = bufferToHex(signedTx);
+    }
+
     const decodedTx = decodeTransaction(payload);
     const networkDetected = getNetworkByChainId(decodedTx.chainId, networks);
     const contractAsset = getAssetByContractAndNetwork(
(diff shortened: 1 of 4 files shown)


common/v2/features/SendAssets/stateFactory.tsx
```diff
@@ -194,7 +202,8 @@ const TxConfigFactory: TUseStateReducerFactory<State> = ({ state, setState }) =>
 
   const txFactoryState = {
     txConfig: state.txConfig,
-    txReceipt: state.txReceipt
+    txReceipt: state.txReceipt,
+    signedTx: state.signedTx
   };
```
vilsbole: It's surprising that we can modify the return value without updating a TS declaration.

common/v2/features/SendAssets/stateFactory.tsx
```diff
@@ -125,7 +126,14 @@ const TxConfigFactory: TUseStateReducerFactory<State> = ({ state, setState }) =>
       .finally(cb);
   };
 
-  const handleSignedTx: TStepAction = (payload: ISignedTx, cb) => {
+  const handleSignedTx: TStepAction = (payload: any, cb) => {
```
vilsbole: Please avoid breaking type-safety.
Prefer updating `ISignedTx` or adding an option to the argument `payload: ISignedTx | <whateveryouneed>`

common/v2/features/SendAssets/stateFactory.tsx
```diff
@@ -125,7 +126,14 @@ const TxConfigFactory: TUseStateReducerFactory<State> = ({ state, setState }) =>
...
+    let signedTx = payload;
+    // Used when signedTx is a buffer instead of a string.
+    // Hardware wallets return a buffer.
+    if (typeof signedTx === 'object') {
+      signedTx = bufferToHex(signedTx);
```
vilsbole: 1. Can we find the `ethers.js` alternative? It would be nice to remove the duplicate lib.
2. Can we move the conversion into the wallet logic (`signRawTransaction` instead of the SendAssetsFlow) so the interface can . always manipulate the same type?

blurpesec: 1) Done.
2) It's actually always a buffer, so i swapped to using ethers.js's `Arrayish` type.

common/v2/features/SendAssets/stateFactory.tsx
blurpesec: Done c:

common/v2/features/SendAssets/stateFactory.tsx
blurpesec: Yeet.

PR #3032, opened 2020-02-12
Settings Form Validation

FrederikBolding:
- Added form validation to `AddToAddressBook` component on Settings page
- Changed styling of import box validation
- Fixed styling inconsistencies with `AddToAddressBook`

### DIFF
--- a/common/v2/components/NetworkSelectDropdown.tsx
+++ b/common/v2/components/NetworkSelectDropdown.tsx
@@ -11,13 +11,14 @@ import { Typography, Dropdown } from 'v2/components';
 interface Props {
   network: string | undefined;
   accountType?: WalletId;
+  className?: string;
   onChange(network: NetworkId): void;
 }
 
 const DropdownContainer = styled('div')`
   .is-open > .Select-control > .Select-multi-value-wrapper > .Select-input:only-child {
     transform: translateY(0%);
-    padding: 16px 15px 16px 15px;
+    padding: 12px 12px;
     position: inherit;
   }
 `;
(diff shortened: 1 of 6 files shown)


FrederikBolding:
[ch5074]

common/v2/components/NetworkSelectDropdown.tsx
```diff
@@ -61,7 +62,7 @@ function NetworkSelectDropdown({ network, accountType, onChange }: Props) {
     .map(n => ({ label: n.name, value: n }));
 
   return (
-    <div>
+    <div className={className}>
```
vilsbole: How about: 
```
function NetworkSelectDropdown({ network, accountType, onChange, ...props }: React.FC<Props>) {
	//...
	<div {...props}>
    //...
```

common/v2/components/NetworkSelectDropdown.tsx
```diff
@@ -11,21 +11,22 @@ import { Typography, Dropdown } from 'v2/components';
...
+    padding: 12px 12px;
     position: inherit;
   }
 `;
 
 const SContainer = styled('div')`
   display: flex;
   flex-direction: row;
-  padding: 16px 15px 16px 15px;
+  padding: 12px 12px;
```
vilsbole: 1. Since they are identical prefer shorthand: `padding: 12px;`
2. Do we have a theme/SPACING that is equivalent?

FrederikBolding: We have `FONT_SIZE.XS` but not sure how much sense that makes 😄

vilsbole:
Should we also include ens lookup when we add a contact?

PR #3038, opened 2020-02-14
Added sorting in the AccountList table

blurpesec:
### Description
The goal of this pr is to introduce sorting to the AccountList table. I removed the sorting for AccountList from the AbstractTable, and use a switcher for the sorting functions.

### DIFF
--- a/common/v2/components/AccountList.tsx
+++ b/common/v2/components/AccountList.tsx
@@ -26,6 +26,7 @@ import { DashboardPanel } from './DashboardPanel';
 import { RatesContext } from 'v2/services';
 import { default as Currency } from './Currency';
 import { TUuid } from 'v2/types/uuid';
+import IconArrow from './IconArrow';
 
 const Label = styled.span`
   display: flex;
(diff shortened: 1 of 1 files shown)


common/v2/components/AccountList.tsx
```diff
@@ -214,22 +285,70 @@ function buildAccountTable(
...
+      <IconArrow isFlipped={getColumnSortDirection('ACCOUNT_LIST_VALUE')} />
     </HeaderAlignment>,
     <HeaderAlignment key={'ACCOUNT_LIST_DELETE'} align="center">
       {translateRaw('ACCOUNT_LIST_DELETE')}
     </HeaderAlignment>
   ];
 
+  const getFullTableData = accounts
+    .map((account, index) => {
+      const addressCard: ExtendedAddressBook | undefined = getLabelByAccount(account, addressBook);
```
vilsbole: How about including the `ExtendedAddressBook` directly inside `StoreAccount.label` that way the component wouldn't need to know about our data structure.

vilsbole:
Feel free to merge as you see fit :)

common/v2/components/AccountList.tsx
blurpesec: Lets do this in a different pr when we go to change StoreAccount c:

PR #3045, opened 2020-02-14
ev/combine account types

vilsbole:
Within the app, an account should always have an uuid.
We replace set `ExtendedAccount` as the default type declaration and rename to a shorter `IAccount`.
We use Omit to define the case of an account that is about to created and name it `IRawAccount`


blurpesec:
Rebase, then merge ? c:

PR #3047, opened 2020-02-18
Small bug fixes

FrederikBolding:
- Fixes issue with crashing when using DevTools following recent change
- Fixes issue where inputting symbols in send asset flow would crash
- Fixes issue where you would be redirected even when suppressing errors

### DIFF
--- a/common/v2/features/ErrorHandling/ErrorProvider.tsx
+++ b/common/v2/features/ErrorHandling/ErrorProvider.tsx
@@ -46,16 +46,18 @@ class ErrorProvider extends Component<RouteComponentProps<{}>> {
   }
 
   public componentDidCatch(error: Error) {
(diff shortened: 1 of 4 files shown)


common/v2/features/ErrorHandling/ErrorProvider.tsx
```diff
@@ -46,16 +46,18 @@ class ErrorProvider extends Component<RouteComponentProps<{}>> {
...
     console.error(error);
-    this.props.history.replace(ROUTE_PATHS.HOME.path);
-    this.setState({
-      error: { error, path }
-    });
-    clearTimeout(errorTimer);
-    errorTimer = setTimeout(() => {
-      this.setState({ error: undefined });
-    }, ERROR_TIMEOUT_MS);
+    if (!this.state.suppressErrors) {
```
vilsbole: ❤️

PR #3050, opened 2020-02-18
Add Storybook

Mrtenz:
This PR adds Storybook with a pretty basic configuration. It has support for:

* `styled-components` themes, with a theme switcher
* Figma design integration, even though the support for it seems to be not great
* Accessibility testing
* Basic component documentation

There are a few simple examples for how to use it, found in `common/v2/components/AmountInput.stories.tsx` and `common/v2/components/Banner.stories.tsx`.

To run it, use:
```
$ yarn storybook
```


common/v2/components/Banner.stories.tsx
```diff
@@ -0,0 +1,22 @@
...
+import { Banner } from './Banner';
+import { BannerType } from '../types';
+
+export default { title: 'Banner' };
+
+export const announcement = () => (
+  <Banner value="I'm an announcement banner!" type={BannerType.ANNOUNCEMENT} />
+);
+
+// Uncomment this for Figma support:
```
FrederikBolding: Should this be included?

Mrtenz: @vilsbole asked for Figma support, so this is just an example of how to use it.

vilsbole: It would be nice to see the react component side-by-side figma for comparing.

PR #3051, opened 2020-02-19
Private Wallet Tags

blurpesec:
_Note: This feature is switched off for now until we bang out some copy to let users know that something will deanonymize them, and logic to remove their private account tag._

### Description
This would allow a user to track which accounts are "private". The user can decide what that privacy looks like to them.

1) A normal user, for instance, usually receives ether from an exchange. To everybody else, this crypto looks like it's coming from an exchange. It's "externally" private. A government can still figure out "who" you are because they can subpoena Coinbase or Kraken or Gemini. Then they can build a cluster of addresses that are "you". This is not actually truly-private, but it's the level of privacy that most people want because other /normal/ people won't be able to tell that the money is attached to the rest of their cluster of accounts.

2) A hardcore privacy user can designate empty accounts as "private" on their MyCrypto dashboard, and then fund those accounts only with ETH/DAI from their chosen privacy tools (Mixers, Aztec's zkDai, etc). These funds will be truly private. Unlinkable to the rest of their addresses unless the hardcore privacy user makes a transaction to another account that some service is able to identify.

For both of these types of users, they can self-identify "private" accounts to their client-side MyCrypto, and we can help them to manage their private funds. When sending assets on MyCrypto, if they attempt to send from a private account to a non-private account in their cluster (or vice-versa), we can inform them that this will "unmask" their account, and ask them if they want to unmark their account as private. We can also prompt them the same way when they're sending to unknown accounts.

### ToDos for this pr:
- [x] Hide this feature behind feature flag and turn it off before merge.
- [x] Tooltip in accountlist where user indicates private account.

### ToDos for Full feature release:
- [ ] Create inline messages to explain things to users.
- [ ] Talk to support team about creating docs to explain this feature well.
- [ ] Add logic to revert private tags for accounts.

### Steps to Test:
1) Clone locally. Checkout branch `mh/private-tags`
2) Navigate to `/common/v2/config/isActiveFeature.ts` then turn on the feature "PRIVATE_TAGS" by switching it from `false` to `true`.
3) To add a private tag, go to settings page and click the checkbox in the Account List **Private** column next to the account you want to label as private. Make sure this correctly appends the label in the **Label** column.
4) If you go to the Dashboard, this should show up there too.

### Screenshots:
AccountList:
image

Settings Panel:

image


common/v2/components/AccountList.tsx
```diff
@@ -349,8 +393,28 @@ function buildAccountTable(
...
+      // Includes only the delete column, excludes the privacy tag column
+      case 'true-false':
+        return [...columnList.slice(0, columnList.length - 2), columnList[columnList.length - 1]];
+      // Includes only the privacy tag column, excludes the delete column
+      case 'false-true':
+        return columnList.slice(0, columnList.length - 1);
+      // Includes both delete && privacy tag column
+      case 'true-true':
+        return columnList;
+    }
```
FrederikBolding: There has got to be a better way to do this, right?

common/v2/components/AccountList.tsx
```diff
@@ -413,30 +480,47 @@ function buildAccountTable(
...
-            ...bodyContent
-          ]
-        : bodyContent;
+
+      if (privacyTriggerEnabled) {
+        bodyContent = [
+          ...bodyContent,
+          <PrivacyCheckBox
+            key={index}
+            name={'Privacy Stoof?'}
```
FrederikBolding: Rename 😆

common/v2/components/AccountList.tsx
```diff
@@ -275,21 +311,23 @@ const getSortingFunction = (sortKey: ISortTypes): TSortFunction => {
...
 
-function buildAccountTable(
+const buildAccountTable = (
   accounts: StoreAccount[],
   deleteAccount: (a: IAccount) => void,
   updateAccount: (u: TUuid, a: IAccount) => void,
   deletable?: boolean,
   favoritable?: boolean,
   copyable?: boolean,
+  privacyTriggerEnabled?: boolean,
```
FrederikBolding: Perhaps we could call this something else? Calling it a trigger seems confusing?

common/v2/features/Dashboard/Dashboard.tsx
```diff
@@ -45,6 +45,7 @@ export default function Dashboard() {
             className="Dashboard-mobile-modifiedPanel"
             copyable={true}
             dashboard={true}
+            privacyTriggerEnabled={false}
```
FrederikBolding: Is this necessary with your default value being `false`?

common/v2/features/Dashboard/Dashboard.tsx
```diff
@@ -79,6 +80,7 @@ export default function Dashboard() {
                 className="Dashboard-desktop-modifiedPanel"
                 copyable={true}
                 dashboard={true}
+                privacyTriggerEnabled={false}
```
FrederikBolding: Is this necessary with your default value being `false`?

common/v2/components/AccountList.tsx
blurpesec: `privacyTriggerEnabled` => `privacyCheckboxEnabled` ?

common/v2/features/Dashboard/Dashboard.tsx
blurpesec: nahhhh

common/v2/features/Dashboard/Dashboard.tsx
blurpesec: nah

common/v2/components/AccountList.tsx
blurpesec: i think it reads easier, which is why i used it during dev. but not a big deal to have it just be
```
case (deletePresent && privacyPresent):
case (deletePresent && !privacyPresent):
case (!deletePresent && privacyPresent):
case (!deletePresent && !privacyPresent):
```

common/v2/components/AccountList.tsx
```diff
@@ -398,19 +398,21 @@ const buildAccountTable = (
...
+    // Includes only the delete column, excludes the privacy tag column
+    else if (deletePresent && !privacyPresent) {
+      return [...columnList.slice(0, columnList.length - 2), columnList[columnList.length - 1]];
+    }
+    // Includes only the privacy tag column, excludes the delete column
+    else if (!deletePresent && privacyPresent) {
+      return columnList.slice(0, columnList.length - 1);
+    }
+    // Includes both delete && privacy tag column
+    else if (deletePresent && privacyPresent) {
```
FrederikBolding: Can this last one just be an else?

PR #3024, opened 2020-02-07
ev/walletconnect integration

vilsbole:
Replaces #2911 

### Steps 
1. Add a WallectConnect (WC) account to your dashboard.
2. Send a transaction from your account.

You should be able to go through the flow smoothly, and also have correct error handling when you reject a connection, a transaction, or modify the selected account or network within the mobile app.

---
Suggested WC mobile apps :
- Argent
- Metamask
- Trust

PR #3053, opened 2020-02-19
Resubmit transaction with higher gas

amacar:
[ch3820]


common/v2/components/TransactionFlow/TransactionReceipt.tsx
```diff
@@ -90,21 +96,24 @@ export default function TransactionReceipt({
...
+  const senderAccountLabel = senderContact ? senderContact.label : translateRaw('NO_LABEL');
 
   const localTimestamp = new Date(Math.floor(timestamp * 1000)).toLocaleString();
   const assetAmount = displayTxReceipt.amount || txConfig.amount;
   const assetTicker = 'asset' in displayTxReceipt ? displayTxReceipt.asset.ticker : 'ETH';
   const assetForRateFetch = 'asset' in displayTxReceipt ? displayTxReceipt.asset : undefined;
 
   const txUrl = displayTxReceipt.network.blockExplorer.txUrl(displayTxReceipt.hash);
+  const shouldRenderPendingBtn =
+    pendingButton && txStatus === ITxStatus.PENDING && !isWeb3Wallet(senderAccount.wallet);
```
blurpesec: We'll want to also disallow WALLETCONNECT wallet type

amacar: WalletConnect's wallet type is also WEB3 (from PR: https://github.com/MyCryptoHQ/MyCrypto/pull/3024/files#diff-29db5304b05c462e07ad1bd0dab2df7dR211) and it should be covered automatically after PR is merged due to using `isWeb3Wallet` function which checks if `wallet.type === WEB3`.

PR #3063, opened 2020-02-25
Added support for hiding tokenlist values

FrederikBolding:
- View Dashboard and see that Token value is hidden in the TokenList (left hand side)


FrederikBolding:
[ch5379]

PR #3065, opened 2020-02-26
Generalized Tx Receipt Panel

blurpesec:
## Description

For DeFiZap flow we're going to make Tx Receipt panel more customized so we can mix & match what is displayed so it's more useful. This pr sets that up.


common/v2/components/TransactionFlow/displays/FromToAccount.tsx
```diff
@@ -3,14 +3,18 @@ import styled from 'styled-components';
...
+export interface IAddressAndLabel {
+  address: string;
+  label: string | undefined;
+}
+
 interface Props {
-  fromAccount: StoreAccount;
-  toAccount: StoreAccount;
+  fromAddressAndLabel: IAddressAndLabel;
+  toAddressAndLabel: IAddressAndLabel;
```
vilsbole: IMHO we don't need to be verbose, especially since this is contained to this component.
Reading the type should be sufficient to know what it's composed of. This allows us to have a variable name that is more fluid to read.

```
interface Props = {
 from: Pick<StoreAccount, 'address' | 'label'>, 
 to: Pick<StoreAccount, 'address' | 'label'>
}
```

common/v2/components/TransactionFlow/displays/FromToAccount.tsx
```diff
@@ -55,7 +59,7 @@ const LabelWrapper = styled.div`
 `;
 
 export default function FromToAccount(props: Props) {
-  const { fromAccount, toAccount } = props;
+  const { fromAddressAndLabel, toAddressAndLabel } = props;
```
vilsbole: Is there a reason not to destructure the prop during the argument declaration?
```
function FromToAccount({ from, to }: Props) {}
```

It removes a LOC, but mainly it makes it really simple to know what props are used (you don't need to search for a `props.`)

common/v2/types/transactionFlow.ts
```diff
@@ -79,8 +80,15 @@ export interface IReceiverAddress {
...
   SUCCESS = 'SUCCESS',
   FAILED = 'FAILED',
   PENDING = 'PENDING'
 }
+
+export enum ITxType {
+  STANDARD = 'STANDARD',
+  BROADCAST_TX = 'BROADCAST_TX',
+  SWAP = 'SWAP'
+}
```
vilsbole: Is there a reason for `BROADCAST` to have a suffix and not the others?

common/v2/components/TransactionFlow/TransactionReceipt.scss
```diff
@@ -59,6 +72,21 @@
...
+          margin-right: 10px;
+          @media (max-width: 550px) {
+            display: none;
+          }
+        }
+        .Amount .Amount-fiat{
+          color: #B5BFC7;
+          font-size: 16px;
+        }
+      }
```
vilsbole: If you really must use scss, please create a new className.
This method of nesting name declarations is really difficult to maintain.

PR #3066, opened 2020-02-26
Added dashboard tooltips

FrederikBolding:
Added dashboard tooltips for account selection, token panel and account list.


FrederikBolding:
[ch5142]

jspence425:
The tooltip next to the account selector has incorrect copy. 

It should say: "Use this Account Selector to display the balance of a single account, or the combined balances of any accounts you have added."

ss (2020-02-26 at 01 23 53)

The other two tooltips in the Token area and the Accounts area look good.

jspence425:
Thumbs up all around! 👍

PR #3069, opened 2020-02-26
Small styling fix to Dropdown component

FrederikBolding:
Fixed weird styling offset caused by `react-select`
- Applied to `Dropdown` component
- Needs testing to make sure it doesn't break any dropdowns


FrederikBolding:
I used the same fix I came up with when doing it for `AssetDropdown`

amacar:
This PR created another issue with dropdown :(

image
image

PR #3072, opened 2020-02-27
[legacy] remove contracts that are old af, add new addresses that are more used

tayvano:
@wtzb I don't know if we can test this due to the issue we had last time with the build process? but maybe check with the linter at the very least to make sure I didn't screw up any \"


wtzb:
@tayvano updated some ENS addresses per https://docs.ens.domains/ens-migration/technical-description. Testing can be done through `yarn build:electron:{os}`.

blurpesec:
The EvidenZ contract is used by a company that frequently messages us on Discord for contracts tab development / feature requests. Would not recommend removing

tayvano:
@blurpesec it looks like it was removed but it actually wasn't, the order just changed.

vilsbole:
@vilsbole Need to cherry-pick to `master`

PR #3080, opened 2020-02-28
Exlcude 0x aka radar-relay from Dex.ag lookups

vilsbole:
Exclude 0x pool from Swap per https://mycryptohq.slack.com/archives/G8G1S4TG9/p1582830405054300

[This is submitted now while looking into our Dex implementation]

### Test
- Create a Swap
- Check network tab and see that radar-relay is discluded.


vilsbole:
Should we update to v2?

blurpesec:
> Should we update to v2?

Ill reach out to their team to see what's involved in the update

PR #3088, opened 2020-03-04
Vertical Stepper UI for Swap/MTX

FrederikBolding:
- Added general `VerticalStepper`
- Added `ConfirmSwapMultiTx` for use in MTX implementation of Swap

It is not currently in use in the codebase, so see Storybook to test 😄


FrederikBolding:
[ch5420]

PR #3095, opened 2020-03-08
(Legacy) Added Banner for user to acknowledge verifying builds

409H:
[CH5455]

🎉 🎉 🎉

## Description

* Adds a banner to the Electron view to prompt users to verify the builds
   * Fetches the version number from `config/data.tsx` (which is from the `packageJson.version`) and compares that string with the one in LocalStorage from when they last accepted

## Changes

* Added new component `ElectronBuildVerified`

## Quality Assurance

### DIFF
--- a/common/components/ElectronBuildVerified/index.tsx
+++ b/common/components/ElectronBuildVerified/index.tsx
@@ -4,68 +4,76 @@ import { VERSION } from 'config';
 import './Banner.scss';
 
 interface BannerState {
-    displayBanner: boolean;
+  displayBanner: boolean;
 }
 
 interface BannerProps {
-    versionNow: string;
+  versionNow: string;
 }
 
-const STORAGE_NAME = "ElectronBuildVerified";
+const STORAGE_NAME = 'ElectronBuildVerified';
 
 export default class ElectronBuildVerified extends Component<BannerProps, BannerState> {
-    public state: State = {
-        displayBanner: this.isBannerPrompted()
-    };
+  public state: State = {
+    displayBanner: this.isBannerPrompted()
+  };
 
-    constructor(props: BannerProps) {
-        super(props);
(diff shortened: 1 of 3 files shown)


common/components/ElectronBuildVerified/index.tsx
```diff
@@ -0,0 +1,77 @@
...
+                            onChange={(e) => this.CheckedBox(e)}
+                        />{' '}
+                        <p>
+                            I have{' '}
+                            <a href="https://support.mycrypto.com/staying-safe/verifying-authenticity-of-desktop-app">verified that this build ({VERSION})</a>{' '}
+                            is signed by MyCrypto and I understand the risks of not verifying the build.
+                        </p>
+                    </div>)
+                :
+                    (<></>)
```
Mrtenz: Instead of using a ternary operator, you can do something like
```ts
if (displayBanner) {
  return (...);
}

return null;
```

common/components/ElectronBuildVerified/index.tsx
```diff
@@ -0,0 +1,77 @@
...
+
+interface BannerState {
+    displayBanner: boolean;
+}
+
+interface BannerProps {
+    versionNow: string;
+}
+
+const StorageName = "ElectronBuildVerified";
```
Mrtenz: `StorageName` -> `STORAGE_NAME`

common/components/ElectronBuildVerified/index.tsx
```diff
@@ -0,0 +1,77 @@
...
+                }
+            } catch {
+                //
+            }
+        } 
+
+        return true;
+    }
+
+    private CheckedBox = (event: React.ChangeEvent<HTMLInputElement>) => {
```
Mrtenz: `CheckedBox` -> `handleCheck` or something

PR #3101, opened 2020-03-11
Sign Message: Implement loading state for button

lubejmatej:
Closes [ch5467]


vilsbole:
> I don't see the spinner when signing a message with Mnemonic. Haven't tested keystore.

For beta we stopped support for electron features. This means that we no longer support keystore and mnemonic signing. Since the code lgtm and we may eventually support then in the future, I'm merging just to close this PR.

PR #3103, opened 2020-03-12
ev/multi tx switch

vilsbole:
Create custom `useTxMulti` hook that:
1. receives an array of partial transaction objects `{to, value, data, gasPrice }`
2. completes the missing information `{ nonce, gasLimit, from }`
3. after receiving a `walletResponse` it can send the transaction, updates the `status` when the tx is confirmed, and includes the blocks timestamp with `minedAt`.
4. moves to the next transaction in the array.

Integrate the hook into `SwapAssetFlow`

### Test
1. Create a TOKEN -> ETH or TOKEN -> TOKEN swap.
2. Go through the flow with different wallets.


common/v2/services/ApiService/Dex/Dex.ts
```diff
@@ -135,3 +157,43 @@ export default class DexService {
...
+    }
+  ];
+
+  // First 4 bytes of the hash of "fee()" for the sighash selector
+  const funcHash = ethers.utils.hexDataSlice(ethers.utils.id('approve(address,uint256)'), 0, 4);
+  // [spender, amount]. Spender is the contract the user needs to authorize.
+  // It's value is also available in the API response obj `data.metadata.input.spender`
+  const params = [DEXAG_MYC_HANDLER_CONTRACT, value];
+  const bytes = abi.encode(inputs, params).substr(2);
+  const data = `${funcHash}${bytes}`;
```
blurpesec: We have methods for doing abi encoding/decoding already that looks like this:
```
const data = ERC20.approve.encodeInput({ _spender: DEXAG_MYC_HANDLER_CONTRACT, _value: value })
```

common/v2/services/ApiService/Dex/Dex.ts
```diff
@@ -1,10 +1,15 @@
...
 import { default as ApiService } from '../ApiService';
-import { TSymbol } from 'v2/types';
-import { DEXAG_PROXY_CONTRACT } from 'v2/config';
+import { TSymbol, ITxObject } from 'v2/types';
+import { DEXAG_MYC_TRADE_CONTRACT, DEXAG_MYC_HANDLER_CONTRACT } from 'v2/config';
 
-const DEX_BASE_URL = 'https://api.dex.ag/';
+import { DexTrade } from './types';
+
+const DEX_BASE_URL = 'https://api-v2.dex.ag/';
```
blurpesec: Good candidate for `v2/config/constants.ts`?

common/v2/components/SignTransactionWallets/Web3.tsx
```diff
@@ -67,7 +67,7 @@ class SignTransactionWeb3 extends Component<ISignComponentProps & INetworkContex
     }
   }
 
-  public componentWillMount() {
+  public UNSAFE_componentWillMount() {
```
blurpesec: does this change functionality at all?

common/v2/utils/useTxMulti/reducer.ts
```diff
@@ -0,0 +1,134 @@
...
+    }
+    case ActionTypes.HALT_FLOW: {
+      return { ...state, canYield: false };
+    }
+    case ActionTypes.RESET: {
+      return initialState;
+    }
+    default:
+      return state;
+  }
```
blurpesec: Can you add the logic to add these TXs to LS `account.transactions` array on `PENDING | SUCCESS | FAILED` ?

common/v2/components/SignTransactionWallets/Web3.tsx
vilsbole: It's been deprecated so to avoid the warning in the console we need to prefix with `UNSAFE_`

common/v2/services/ApiService/Dex/Dex.ts
vilsbole: I was hesitant about this. Since it's an ApiService so only used once I figured it was good leave inside the service itself.

common/v2/utils/useTxMulti/reducer.ts
vilsbole: Hmm, after looking into it do you mind if we tackle this later?
Currently we're storing way too much information within the account.transactions which has a `senderAccount` that seems to be full blow `StoreAccount`, and it seems good to tackle it in one blow.

Replace all our send transactions with this hook, have a single place to `addNewTranscation`, and handle the polling when the transactions are updated.

blurpesec: > Currently we're storing way too much information within the account.transactions which has a `senderAccount`

I believe this `senderAccount` only exists while the tx is pending, then it's overwritten with the tx receipt when it's done.

> Currently we're storing way too much information within the account.transactions

But agree. 

> after looking into it do you mind if we tackle this later?

So no, i don't mind.

PR #3105, opened 2020-03-12
Revert "Vertical Stepper UI for Swap/MTX (#3088)"

blurpesec:
Reverts the vertical stepper ui for swap/mtx commit to test if this is the build error.


vilsbole:
Solved with yarn update

PR #3112, opened 2020-03-14
Updated eth-scan

FrederikBolding:
- Updated eth-scan
- Fixed issues in BalanceService following update
- Made sure to return `BigNumber.js`
- Added new method for fetching balances
- Fixed issues with CI caching

### DIFF
--- a/common/v2/services/Store/BalanceService.tsx
+++ b/common/v2/services/Store/BalanceService.tsx
@@ -92,15 +92,15 @@ const getTokenBalances = (
   provider: ProviderHandler,
   address: TAddress,
   tokens: StoreAsset[]
-): BalanceMap => {
-  return toBigNumberJS(
-    tokens.reduce(async (balances, token) => {
-      return {
+): Promise<BalanceMap> => {
+  return tokens
+    .reduce<Promise<EthScanBalanceMap>>(async (balances, token) => {
+      return Promise.resolve({
         ...balances,
         [token.contractAddress as TAddress]: await provider.getRawTokenBalance(address, token)
-      };
-    }, {})
-  );
+      });
+    }, Promise.resolve<EthScanBalanceMap>({}))
+    .then(toBigNumberJS);
 };
 
 const getAccountAssetsBalancesWithJsonRPC = async (
--- a/common/v2/services/Store/BalanceService.tsx
+++ b/common/v2/services/Store/BalanceService.tsx
@@ -1,7 +1,8 @@
 import {
   getEtherBalances,
   getTokensBalance,
-  getTokenBalances as getTokenBalancesFromEthScan
+  getTokenBalances as getTokenBalancesFromEthScan,
+  getTokensBalances
 } from '@mycrypto/eth-scan';
 import partition from 'lodash/partition';
 import { default as BN } from 'bignumber.js';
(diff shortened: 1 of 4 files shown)


common/v2/services/Store/BalanceService.tsx
```diff
@@ -1,27 +1,44 @@
...
+import { BigNumber as EthScanBN } from '@ethersproject/bignumber';
 
 import { ETHSCAN_NETWORKS, MYCRYPTO_UNLOCK_CONTRACT_ADDRESS } from 'v2/config';
-import { TAddress, StoreAccount, StoreAsset, Asset, NodeConfig, Network } from 'v2/types';
+import { TAddress, StoreAccount, StoreAsset, Asset, Network } from 'v2/types';
 import { ProviderHandler } from 'v2/services/EthService';
 
-export interface BalanceMap {
-  [key: string]: BN | BigNumber;
+export interface BalanceMap<T = BN> {
```
Mrtenz: BalanceMap is exported from eth-scan

FrederikBolding: Simplified my use of `BalanceMap` 😄

common/v2/services/Store/BalanceService.tsx
```diff
@@ -82,12 +93,14 @@ const getTokenBalances = (
...
-    };
-  }, {});
+  return toBigNumberJS(
+    tokens.reduce(async (balances, token) => {
+      return {
+        ...balances,
+        [token.contractAddress as TAddress]: await provider.getRawTokenBalance(address, token)
+      };
+    }, {})
+  );
```
Mrtenz: The `Promise` here is ignored, so this would throw an error if `provider.getRawTokenBalance` fails. You can solve this by reducing a `Promise` and returning a `Promise<BalanceMap>` from this function instead, for example:

https://github.com/MyCryptoHQ/eth-scan/blob/master/src/utils/batch.ts#L34-L36

common/v2/services/Store/BalanceService.tsx
```diff
@@ -123,24 +141,41 @@ export const getAccountsAssetsBalances = async (accounts: StoreAccount[]) => {
...
 
 export const getAllTokensBalancesOfAccount = async (account: StoreAccount, assets: Asset[]) => {
-  const scanner = getScanner(account.network.nodes[0]);
+  const provider = account.network.nodes[0];
   const assetsInNetwork = assets.filter(x => x.networkId === account.network.id);
   const assetAddresses = getAssetAddresses(assetsInNetwork) as string[];
 
   try {
-    return scanner.getTokensBalance(account.address, assetAddresses);
+    return getTokensBalance(provider, account.address, assetAddresses).then(toBigNumberJS);
```
Mrtenz: This can possibly be simplified by making the function not async and directly returning the promise instead. The try-catch block wouldn't be necessary (and it seems kind of redundant right now too tbh).

common/v2/services/Store/BalanceService.tsx
FrederikBolding: Could you verify my changes?

common/v2/services/Store/BalanceService.tsx
FrederikBolding: I would prefer to keep as much of it the same as possible as this isn't a priority 😄

Mrtenz: Well, the only difference would be removing the `async` and `try-catch` block. The function would behave exactly the same

common/v2/services/Store/BalanceService.tsx
```diff
@@ -81,13 +92,15 @@ const getTokenBalances = (
...
-  return tokens.reduce(async (balances, token) => {
-    return {
-      ...balances,
-      [token.contractAddress as TAddress]: await provider.getRawTokenBalance(address, token)
-    };
-  }, {});
+): Promise<BalanceMap> => {
+  return tokens
+    .reduce<Promise<EthScanBalanceMap>>(async (balances, token) => {
+      return Promise.resolve({
```
Mrtenz: I don't think you need the `Promise.resolve` here (since the function is already a `Promise`. Other than that it looks good to me.

common/v2/services/Store/BalanceService.tsx
FrederikBolding: How come the try-catch isn't needed?

Mrtenz:
```ts
const foo = async (): Promise<any> => {
  throw new Error('bar');
};
```
Is the same as
```ts
const foo = (): Promise<any> => {
  return Promise.reject(new Error('bar'));
};
```

PR #3114, opened 2020-03-16
Fix coingecko logo alignment

MrLuit:
## Issue
image

## Changes
Before:
image

After: 
image


bowensanders:
The new text size makes the alignment look right.

This may not need a real code review as it was a single line of CSS and it fixes the problem

PR #3115, opened 2020-03-17
Sign message with WalletConnect

amacar:
[ch5342]

### DIFF
--- a/common/v2/components/WalletList/WalletButton.tsx
+++ b/common/v2/components/WalletList/WalletButton.tsx
@@ -1,8 +1,10 @@
 import React from 'react';
-import styled from 'styled-components';
+import styled, { keyframes } from 'styled-components';
+import { Tooltip } from '@mycrypto/ui';
 
 import { WalletId } from 'v2/types';
 import { BREAK_POINTS, COLORS } from 'v2/theme';
+import translate from 'v2/translations';
 
 const { SCREEN_SM, SCREEN_XS } = BREAK_POINTS;
 const { WHITE } = COLORS;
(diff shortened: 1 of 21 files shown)


common/v2/components/WalletList/WalletButton.tsx
```diff
@@ -78,26 +77,29 @@ const WalletLabel = styled.div`
...
-        <WalletLabel>{name}</WalletLabel>
-      </WalletButtonWrapper>
-    );
-  }
-
-  private handleClick = () => {
-    this.props.onClick(this.props.walletType);
-  };
-}
+export const WalletButton = (props: Props) => {
```
vilsbole: Can we add a `story` for this please?

common/v2/services/WalletService/walletconnect/useWalletConnect.ts
```diff
@@ -86,7 +87,7 @@ export function useWalletConnect() {
       } else {
         return service
           .sendTx(tx)
-          .then((txHash: ITxData) => {
+          .then((txHash: string) => {
```
vilsbole: We now have a `ITxSigned` and `ITxHash` `Brand` types and type guards to deal with a response from a wallet. It's a slight improvement over using a simple `string`.
https://github.com/MyCryptoHQ/MyCrypto/blob/85c493da74a334806eb476bbc391074e6566aff3/common/v2/utils/typeGuards.ts#L8

amacar: Changed to ITxHash, thanks for the info!

common/v2/components/WalletList/WalletButton.tsx
amacar: Added story with two states (enabled/disabled).

PR #3117, opened 2020-03-17
Added Unlock Token Expiration

blurpesec:
### Description
Added detection of Unlock Token Expiration date. Added in unlock abi as well so we can call the functions from it :-D ?

### DIFF
--- a/common/v2/services/EthService/network/unlockProtocolProvider.ts
+++ b/common/v2/services/EthService/network/unlockProtocolProvider.ts
@@ -1,0 +1,15 @@
+import { ProviderHandler } from './providerHandler';
+import { UnlockToken } from '../contracts';
+
+export class UnlockProtocolHandler extends ProviderHandler {
+  /* Unlock Token Methods */
+  public fetchUnlockKeyExpiration(address: string, lockAddress: string): Promise<number> {
+    return this.client
+      .call({
+        to: this.requests.getUnlockKeyExpirationTime(address, lockAddress).params[0].to,
+        data: this.requests.getUnlockKeyExpirationTime(address, lockAddress).params[0].data
+      })
+      .then(data => UnlockToken.keyExpirationTimestampFor.decodeOutput(data))
+      .then(({ timestamp }) => parseInt(timestamp, 10));
+  }
+}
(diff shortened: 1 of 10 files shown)


common/v2/services/Store/StoreProvider.tsx
```diff
@@ -134,6 +138,36 @@ export const StoreProvider: React.FC = ({ children }) => {
...
+    Promise.all(
+      membershipLookups.map(membershipLookupObj =>
+        unlockProvider.fetchUnlockKeyExpiration(
+          membershipLookupObj.account,
+          membershipLookupObj.lockAddress
+        )
+      )
+    )
+      .then(data => {
+        setMembershipExpiration(Math.max(...data));
```
FrederikBolding: Can we save all expirations and then just extract the max for showing? Not sure if we might need them at some point, also saved all memberships across accounts.

common/v2/services/Store/StoreProvider.tsx
```diff
@@ -272,3 +307,7 @@ export const StoreProvider: React.FC = ({ children }) => {
 
   return <StoreContext.Provider value={state}>{children}</StoreContext.Provider>;
 };
+
+// const determineOldestKey = (data: string[][]) {
```
FrederikBolding: ?

common/v2/services/EthService/network/unlockProtocolProvider.ts
```diff
@@ -0,0 +1,15 @@
+import { ProviderHandler } from './providerHandler';
+import { UnlockToken } from '../contracts';
+
+export class UnlockProtocolHandler extends ProviderHandler {
+  /* Unlock Token Methods */
+  public fetchUnlockKeyExpiration(address: string, lockAddress: string): Promise<number> {
+    return this.client
+      .call({
+        to: this.requests.getUnlockKeyExpirationTime(address, lockAddress).params[0].to,
```
FrederikBolding: Is there a reason this method call is done twice and stored?

common/v2/services/Store/StoreProvider.tsx
blurpesec: I couldn't think of a reason for why we'd need a list of them, but if you think it'll be useful i can add it?

FrederikBolding: Yeah it could possibly be, let's just do it 😄

PR #3130, opened 2020-03-24
Fixes for dropdown component and delete account clipping

amacar:
Fixes for: 
- Settings: Delete account clipping [ch5536]
- Global: searchable dropdown component dowshifts text [ch5538]
- Mobile Send: Trying to paste address into recipient field can erase the address [ch5529]

### DIFF
--- a/common/v2/components/Dropdown.tsx
+++ b/common/v2/components/Dropdown.tsx
@@ -57,12 +57,6 @@ const DropdownContainer = styled('div')`
     padding: 16px 15px 16px 15px;
     position: inherit;
   }
-
-  .is-open > .Select-control > .Select-multi-value-wrapper > .Select-input:only-child {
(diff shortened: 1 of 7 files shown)


common/v2/components/Dropdown.tsx
```diff
@@ -57,12 +67,6 @@ const DropdownContainer = styled('div')`
     padding: 16px 15px 16px 15px;
     position: inherit;
   }
-
-  .is-open > .Select-control > .Select-multi-value-wrapper > .Select-input:only-child {
```
FrederikBolding: How come we don't need this anymore?

amacar: This is the cause of downshifting because we need to downshift only when there is value in dropbox and not every time when the dropdown is open.

PR #3132, opened 2020-03-24
Replace download desktop app link

MrLuit:
Replace download desktop app link [ch5555]


vilsbole:
Specs clarified in ch5555. Merging

PR #3152, opened 2020-03-31
Some bug fixes / Improvements

blurpesec:
### Description:
- Adds tooltip informational icon to dashboard accounts table for View-Only wallet types.
- Renames View Only wallet type to View-Only
- Adds max-height to the account table so that it doesn't stretch forever.
- Fixed copy mismatch in settings page with "Add Account" vs "Add Address"


GuillaumeRx:
Capture d’écran 2020-04-02 à 15 18 30
Not a bug introduiced by your PR but maybe you can correct it in the same time ?

PR #3155, opened 2020-04-01
[Legacy] Add survey message to desktop app

wtzb:
[ch5585]

image


vilsbole:
Why is the link url visible? Can't we do something like [survey](https://example.com)?

wtzb:
@vilsbole this has been decided as discussed in CH card 5585. Basically, because we want to show the full URL as people might be hesitant to click links.

PR #3159, opened 2020-04-03
Handle nonce calculation in UseEffect

lubejmatej:
Closes [ch5745]
Not able to send ETH if picked by default

### DIFF
--- a/common/v2/features/SendAssets/components/SendAssetsForm.tsx
+++ b/common/v2/features/SendAssets/components/SendAssetsForm.tsx
@@ -1,4 +1,4 @@
-import React, { useContext, useState } from 'react';
+import React, { useContext, useEffect, useState } from 'react';
 import { Field, FieldProps, Form, Formik, FastField } from 'formik';
 import * as Yup from 'yup';
 import { Button } from '@mycrypto/ui';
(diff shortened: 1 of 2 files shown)


common/v2/features/SendAssets/components/SendAssetsForm.tsx
```diff
@@ -255,6 +255,19 @@ export default function SendAssetsForm({ txConfig, onComplete }: IStepComponentP
...
+              return;
+            }
+            setIsEstimatingNonce(true);
+            (async () => {
+              const nonce: number = await getNonce(values.network, values.account.address);
+              setFieldValue('nonceField', nonce.toString());
+              setIsEstimatingNonce(false);
+            })();
+          }, [values.network, values.account]);
+
```
vilsbole, line 270: Danger: According to the docs effects should only be declared at the top level function:
https://reactjs.org/docs/hooks-rules.html

lubejmatej: HOC is a top level function. Setting values in parent component with setState would trigger infinitive loop. But anyway this was fixed by https://github.com/MyCryptoHQ/MyCrypto/pull/3161. So I will just close this PR

lubejmatej:
Already fixed in https://github.com/MyCryptoHQ/MyCrypto/pull/3161

Issue #3160, opened 2020-04-03
Beta doesn't resolve domain names


JohnnyJumper:
### Description of Issue
The resolution of ens and unstoppable domains is not working correctly on the beta version of mycrypto. Seems like the code for selecting the address from the address-book intervened with the resolution code. 

### What exactly were you doing when you encountered this issue?
- Went on https://beta.mycrypto.com/send
- Tried to send some ETH to brad.crypto, brad.zil, brad.eth. Instead of resolving them and continuing with the flow, I ended up with an empty dropdown. When I close the dropdown everything clears like I never typed anything. 

### Expectations
- Once the address book dropdown is closed and if nothing was selected from it the resolution code should trigger which will attempt to resolve the domain (.crypto, .eth, .zil) and in case of success placed the address in the label. In case of error should show users the reasons like (domain is not registered, not configured, doesn't have the currency you are trying to send)
 
### How were you accessing your wallet?
MetaMask
### Description of Your Machine

- Browser: Chrome 
- Browser version: Version 80.0.3987.149 (Official Build) (64-bit)
- Operating system: macOS Mojave version 10.14.5
- Device: MacBook Pro (13-inch)
- MyCrypto.com vs Desktop Application? I went to beta.mycrypto.com
- MyCrypto Version Number: 1.7.3 (took this number from bottom left of the website near the word Disclaimer)

### Console Logs / Screenshots
<img width="650" alt="Screen Shot 2020-04-03 at 2 25 36 PM" src="https://user-images.githubusercontent.com/16522024/78406225-0e716980-75b7-11ea-90d4-4b29a497789a.png">
<img width="1080" alt="Screen Shot 2020-04-03 at 2 25 44 PM" src="https://user-images.githubusercontent.com/16522024/78406230-0fa29680-75b7-11ea-9203-7bc1cd0ba658.png">
...

blurpesec:
Thanks for pointing this out Johnny :-)

PR #3161, opened 2020-04-03
Fix default asset ens bug in send flow

blurpesec:
Fixes: #3160 

We added a change to set default asset as ETH, but doesn't set default network too (which is conducted on asset selection). We should check to make sure this doesn't break `gasPrice` lookup as well.

### DIFF
--- a/common/v2/features/SendAssets/components/SendAssetsForm.tsx
+++ b/common/v2/features/SendAssets/components/SendAssetsForm.tsx
@@ -59,7 +59,8 @@ import {
   GAS_LIMIT_UPPER_BOUND,
   GAS_PRICE_GWEI_LOWER_BOUND,
   GAS_PRICE_GWEI_UPPER_BOUND,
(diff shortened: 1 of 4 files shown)


JohnnyJumper:
Can you add the following changes as well to allow .crypto resolution? 

### v2/services/UnstoppableService/index.ts
Screen Shot 2020-04-03 at 3 47 16 PM
### package.json
Screen Shot 2020-04-03 at 3 47 23 PM

blurpesec:
> @blurpesec Can you add the following changes as well to allow .crypto resolution?
> 
> It should allow you to resolve the brad.crypto domain and have an actual address instead of "CNS namespace is down" error message
> 
> ### v2/services/UnstoppableService/index.ts
> Screen Shot 2020-04-03 at 3 47 16 PM
> 
> ### package.json
> Screen Shot 2020-04-03 at 3 47 23 PM

done :-)

PR #3162, opened 2020-04-04
Implement meta tags for Twitter card

MrLuit:
Implement meta tags for Twitter card [ch5735]

### DIFF
--- a/common/index.html
+++ b/common/index.html
@@ -15,12 +15,12 @@
     <meta property="og:description" content="<%= htmlWebpackPlugin.options.appDescription %>" />
     <meta property="og:site_name" content="<%= htmlWebpackPlugin.options.title %>" />
     <meta property="og:type" content="<%= htmlWebpackPlugin.options.type %>" />
-    <meta property="og:image" content="/common/assets/images/link-preview.png" />
+    <meta property="og:image" content="<%= htmlWebpackPlugin.options.image %>" />
     <meta name="twitter:title" content="<%= htmlWebpackPlugin.options.title %>" />
     <meta name="twitter:card" content="summary_large_image" />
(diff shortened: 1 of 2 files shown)


vilsbole:
Looking at https://cards-dev.twitter.com/validator
It seems that the image is not displayed. Can you double check?

Screenshot 2020-04-06 at 3 10 43 PM

common/index.html
```diff
@@ -15,12 +15,12 @@
     <meta property="og:description" content="<%= htmlWebpackPlugin.options.appDescription %>" />
     <meta property="og:site_name" content="<%= htmlWebpackPlugin.options.title %>" />
     <meta property="og:type" content="<%= htmlWebpackPlugin.options.type %>" />
-    <meta property="og:image" content="/common/assets/images/link-preview.png" />
+    <meta property="og:image" content="<%= htmlWebpackPlugin.options.image %>" />
```
vilsbole, line 18: It seems that the path in `config.img` points to `images/link-preview` which is no longer valid. Can you update it to `../common/assets/images/link-preview.png`

Can you use the [validator](https://cards-dev.twitter.com/validator) to test please?
It may be that your previous tests work only because of twitters caching.

MrLuit: Unfortunately I found out that both `og:image` and `twitter:image` require an absolute URL to work. Furthermore, it [seems that link-preview.png is copied to root](https://github.com/MyCryptoHQ/MyCrypto/blob/master/webpack_config/common.js#L181). This is confirmed here: https://beta.mycrypto.com/link-preview.png

I'm afraid I'll just have to hardcode `https://beta.mycrypto.com/link-preview.png` in the webpack config.js

vilsbole: Good enough. We need to keep in mind to change it when we move to root domain.

PR #3163, opened 2020-04-04
Mobile Settings: Fix position of submenu for large screens

MrLuit:
Mobile Settings Fix position of submenu for [ch5714]


vilsbole:
The submenu is still partially hidden by the header.

Screenshot 2020-04-07 at 2 35 30 PM

PR #3164, opened 2020-04-04
Membership: Create Expired State CTA

MrLuit:
Story details: https://app.clubhouse.io/mycrypto/story/5567


common/v2/services/Store/StoreProvider.tsx
```diff
@@ -125,12 +125,18 @@ export const StoreProvider: React.FC = ({ children }) => {
...
+    if (!memberships) {
+      return MembershipState.ERROR;
+    } else if (Object.values(memberships).length === 0) {
+      return MembershipState.NOTMEMBER;
+    } else {
+      const currentTime = Math.round(Date.now() / 1000);
+      if (membershipExpiration.some(expirationTime => expirationTime > currentTime)) {
+        return MembershipState.MEMBER;
+      } else {
+        return MembershipState.EXPIREDMEMBER;
```
FrederikBolding, line 137: @blurpesec Will this ever be possible? I seem to remember you had some remarks regarding that we didn't detect expired memberships?

MrLuit: From what I saw in the smart contract, the `keyExpirationTimestampFor(address)` function does return the correct uint256 expiration timestamp. This could be faked by changing the device time, but that's a calculated risk

FrederikBolding: @MrLuit But I thought that didn't work when the actual lock was expired? Since we don't own the lock anymore then.

common/v2/features/Dashboard/components/MembershipPanel.tsx
```diff
@@ -54,29 +59,42 @@ const SButton = styled(Button)`
...
   const { membershipState, memberships, membershipExpiration } = useContext(StoreContext);
 
   const isMember = membershipState === MembershipState.MEMBER;
+  const isExpired = membershipState === MembershipState.EXPIREDMEMBER;
   const allMemberships = memberships ? R.uniq(R.flatten(memberships.map(m => m.memberships))) : [];
   const membership =
     allMemberships.length > 0 ? allMemberships[allMemberships.length - 1] : undefined;
 
-  const icon = membership ? MEMBERSHIP_CONFIG[membership].icon : defaultIcon;
+  const icon = (() => {
```
FrederikBolding, line 76: Is there a reason this is a function?

common/v2/features/Dashboard/components/MembershipPanel.tsx
```diff
@@ -54,29 +59,42 @@ const SButton = styled(Button)`
...
+    <SDashboardPanel isMemberOrExpired={isMember || isExpired} padChildren={true}>
       <Wrapper>
         <ImageWrapper>
-          <Icon isMember={isMember} src={icon} />
+          <Icon isMemberOrExpired={isMember || isExpired} src={icon} />
         </ImageWrapper>
         <TextWrapper isMember={isMember}>
-          <Header as="div">{translateRaw('MEMBERSHIP')}</Header>
+          {membershipState === MembershipState.EXPIREDMEMBER ? (
+            <Header as="div">{translateRaw('MEMBERSHIP_EXPIRED')}</Header>
```
FrederikBolding, line 94: Can we just do the ternery inside the `Header´ to reduce code duplication?

common/v2/services/Store/StoreProvider.tsx
```diff
@@ -134,12 +134,18 @@ export const StoreProvider: React.FC = ({ children }) => {
...
```
FrederikBolding, line 146: Can we just call this `EXPIRED´?

common/v2/features/Dashboard/components/MembershipPanel.tsx
FrederikBolding: Actually should be fine!

PR #3171, opened 2020-04-08
Make all strings translatable throughout the app

lubejmatej:
Closes [ch5586]


common/v2/features/Dashboard/components/RecentTransactionList.tsx
```diff
@@ -200,15 +200,21 @@ export default function RecentTransactionList({ accountsList, className = '' }:
   const filteredGroups = groups.filter(group => group.entries.length !== 0);
 
   const recentTransactionsTable = {
-    head: ['Date', 'From Address', 'To Address', 'Amount', 'View More'],
+    head: [
+      translateRaw('RECENT_TRANSACTIONS_DATE') || 'Date',
+      translateRaw('RECENT_TRANSACTIONS_FROM_ADDRESS'),
+      translateRaw('RECENT_TRANSACTIONS_TO_ADDRESS'),
+      translateRaw('RECENT_TRANSACTIONS_TO_AMOUNT'),
+      translateRaw('RECENT_TRANSACTIONS_VIEW_MORE') || 'View More'
```
blurpesec, line 208: Why do we need the `|| 'View More'` here?

common/v2/features/Dashboard/components/RecentTransactionList.tsx
```diff
@@ -200,15 +200,21 @@ export default function RecentTransactionList({ accountsList, className = '' }:
   const filteredGroups = groups.filter(group => group.entries.length !== 0);
 
   const recentTransactionsTable = {
-    head: ['Date', 'From Address', 'To Address', 'Amount', 'View More'],
+    head: [
+      translateRaw('RECENT_TRANSACTIONS_DATE') || 'Date',
```
blurpesec, line 204: Why do we need the ` || 'Date'` here?

common/v2/features/Settings/components/GeneralSettings.tsx
```diff
@@ -73,17 +73,17 @@ interface SettingsProps {
...
-  { name: '5 Minutes', value: '300000' },
-  { name: '10 Minutes', value: '600000' },
-  { name: '15 Minutes', value: '900000' },
-  { name: '30 Minutes', value: '1800000' },
-  { name: '45 Minutes', value: '2700000' },
-  { name: '1 Hours', value: '3600000' },
-  { name: '3 Hours', value: '10800000' },
-  { name: '6 Hours', value: '21600000' },
-  { name: '12 Hours', value: '43200000' }
+  { name: `1 ${translateRaw('ELAPSED_TIME_MINUTE')}`, value: '60000' },
```
blurpesec, line 76: Could probably pass these numbers into the translation strings using vars. Is there a reason we wouldn't do that?

common/v2/features/Dashboard/components/RecentTransactionList.tsx
lubejmatej: Well in case, there is no translation available I wanted to default to some string. Because, those columns are used in sortable columns, and in hidden columns and providing empty string, would break the table. But I am wondering, if this is necessary since. Since it will default to english translation anyway. Should I remove it?

common/v2/features/Dashboard/components/RecentTransactionList.tsx
lubejmatej: https://github.com/MyCryptoHQ/MyCrypto/pull/3171#discussion_r417842123

common/v2/features/Dashboard/components/RecentTransactionList.tsx
blurpesec: > Since it will default to english translation anyway. Should I remove it?

I'd say so c:

node-scripts/translations/translations-extract.ts
```diff
@@ -0,0 +1,121 @@
+/* tslint:disable:no-console */
```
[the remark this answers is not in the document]
lubejmatej: Sure

PR #3173, opened 2020-04-08
Reset scroll on page change

GuillaumeRx:
### What it changes
**Links when clicked now reset page scroll to top**

### What I did
- Removed `overflow` css rule on Router container (_This is what resolved the bug_)
- Updated scrollToTop method for the newest recommended one on [React Training](https://reacttraining.com/react-router/web/guides/scroll-restoration)
- Made the DevTool `fixed` so we can still use it

### Steps to test
Just scroll and change page via a button or navigation.

### Extra
Verify that the new integration of DevTool is okay for you since we couldn’t keep it as it was before due to the necessity of adding the `overflow` rule to the container to keep the "shape" of the app.

[Story details](https://app.clubhouse.io/mycrypto/story/5782)

### DIFF
--- a/common/Root.tsx
+++ b/common/Root.tsx
@@ -17,6 +17,7 @@ const AppProvidersInnerContainer = styled.div`
 `;
 const AppRouterContainer = styled.div`
   flex: 1;
+  max-width: 100vw;
   max-height: 100vh;
 `;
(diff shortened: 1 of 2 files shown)


bowensanders:
This seems to be working great here, thanks!

PR #3175, opened 2020-04-08
Added Protected TX for members

FrederikBolding:
- Added Protected TX for members


FrederikBolding:
[ch5435]

FrederikBolding:
So, do you need anything from me @blurpesec ?

PR #3176, opened 2020-04-09
Fixed an issue with initial value of selected asset on send flow

FrederikBolding:
- Fixed an issue with initial value of selected asset


FrederikBolding:
Not sure if this has a card yet, was one of the bugs noticed yesterday.

PR #3177, opened 2020-04-09
Remove data from localstorage

amacar:
Remove default networks/assets/contracts from local storage. This decreases the local storage size for approx 750kb. [ch5669]


blurpesec:
@amacar I can't seem to import my previously-exported settings to test

amacar:
> @amacar I can't seem to import my previously-exported settings to test

This must be due to our primitive `importStorage` function which only checks for version and equality of the keys. In this PR I added a new key (`networkNodes`) to the local storage that is why import is not working. I tried to add a migration but then the function will fail again due to different versions. For this test, you can try manually adding a missing key in the previously-exported settings (`"networkNodes":{}`), but eventually, we would need to implement a proper migration script, which will take care of migration through multiple versions and allowing to import previously-exported settings even with different keys and versions.

common/v2/utils/generateUUID.tsx
```diff
@@ -24,6 +24,9 @@ export const generateAssetUUID = (chainId: string | number, address?: string): T
     ? (getUuid(`${chainId.toString()}-${toChecksumAddress(address)}`) as TUuid)
     : (getUuid(chainId.toString()) as TUuid);
 
-export const generateContractUUID = (abi: string) => getUuid(abi) as TUuid;
+export const generateContractUUID = (id: NetworkId, address: string, abi: string) =>
+  address
+    ? (getUuid(`${id}-${toChecksumAddress(address)}-${abi}`) as TUuid)
+    : (getUuid(`${id}-${abi}`) as TUuid);
```
blurpesec, line 30: We should remove the abi part of this to stick with the uuid convention we follow with assets. We should do the same with user accounts if we aren't already doing this, as well.

common/v2/database/data/networks.ts
```diff
@@ -77,7 +77,7 @@ export const NETWORKS_CONFIG: NetworkConfig = {
       origin: 'https://kovan.etherscan.io'
     }),
     tokens: require('./tokens/kov.json'),
-    contracts: require('./contracts/ropsten.json'),
+    contracts: require('./contracts/kovan.json'),
```
blurpesec, line 80: nice!

blurpesec:
> > @amacar I can't seem to import my previously-exported settings to test
> 
> This must be due to our primitive `importStorage` function which only checks for version and equality of the keys. In this PR I added a new key (`networkNodes`) to the local storage that is why import is not working. I tried to add a migration but then the function will fail again due to different versions. For this test, you can try manually adding a missing key in the previously-exported settings (`"networkNodes":{}`), but eventually, we would need to implement a proper migration script, which will take care of migration through multiple versions and allowing to import previously-exported settings even with different keys and versions.

regarding the settings import, howabout this?:
```const isValidImport = (importedCache: string, localStorage: string) => {
  try {
    const parsedImport = JSON.parse(importedCache);
    const parsedLocalStorage = JSON.parse(localStorage);
    if (parsedImport.version !== parsedLocalStorage.version) {
      throw new Error(
        `Outdated version detected. Cannot import ${parsedImport.version} into ${parsedLocalStorage.version}`
      );
    }
    const oldKeys = Object.keys(parsedImport);
    const newKeys = Object.keys(parsedLocalStorage);
    return oldKeys.every(key => newKeys.includes(key))
  } catch (error) {
    console.debug(error);
    return false;
  }
};
```

As long as the keys from the old are present in new, it returns `true` which allows for merging when adding new keys to the LS obj ?
I just tested locally and it seems to work fine. It won't fix validation when changing the contents of settings schema, but will allow for adding new keys to obj :3

common/v2/utils/generateUUID.tsx
amacar: I removed a network object from each account. The settings-export filesize decreased a lot :) Nice catch!

For generating UUIDs I removed abi from contracts generating, created a similar method for accounts and apply those methods (for assets/account/contract) throughout the project.

common/v2/database/generateDefaultValues.ts
```diff
@@ -15,16 +15,22 @@ import {
...
+
 /* Transducers */
 const addNetworks = add(LSKeys.NETWORKS)((networks: SeedData) => {
   const formatNetwork = (n: NetworkLegacy): Network => {
     const baseAssetUuid = generateAssetUUID(n.chainId);
-    const nodes = NODES_CONFIG[n.id] || NODES_CONFIG[n.name as NetworkId] || [];
+    const nodes: NodeOptions[] = NODES_CONFIG[n.id] || [];
+    if (n.nodes) {
+      nodes.push(...n.nodes); // add custom nodes
+    }
```
blurpesec, line 33: Couldn't this just be: 
```
const nodes: NodeOptions[] = [...(NODES_CONFIG[n.id] || []), ...n.nodes];
```
?

common/v2/services/Store/DataManager/utils.ts
```diff
@@ -11,25 +11,59 @@ import {
...
 type ArrayToObj = <K extends string | number>(
   k: K
 ) => <V extends any[]>(arr: V) => Record<K, ValuesType<V>>;
-const arrayToObj: ArrayToObj = key => arr =>
+const arrayToObj: ArrayToObj = (key) => (arr) =>
   arr.reduce((acc, curr) => ({ ...acc, [curr[key]]: curr }), {});
-const memoizeObjToArray = R.memoizeWith(createObjHash, objToArray);
-const memoizeArrayToObj = (key: string) => R.memoizeWith(createArrHash, arrayToObj(key));
+
+const mergeConfigWithLocalStorage = (
```
blurpesec, line 31: If possible can we mock & test this? :3

common/v2/services/Store/DataManager/utils.ts
```diff
@@ -56,30 +92,43 @@ export function marshallState(ls: LocalStorage): DataStore {
   };
 }
 
+const constructNetworkNodes = (networks: DataStore[LSKeys.NETWORKS]) => {
```
blurpesec, line 95: If possible can we mock & test this as well? :3

common/v2/database/generateDefaultValues.ts
amacar: This will fail since `n.nodes` can be undefined, but I slightly changed it to:
```
const nodes: NodeOptions[] = [...(NODES_CONFIG[n.id] || []), ...(n.nodes || [])];
```

common/v2/services/Store/DataManager/reducer.spec.ts
```diff
@@ -9,7 +9,7 @@ describe('AppStateReducer', () => {
       const account = { address: '0x0', uuid: 'fakeUUID' } as IAccount;
       const prevState = { [LSKeys.ACCOUNTS]: [] };
       const payload = {
-        model: LSKeys.ACCOUNTS,
+        model: LSKeys.ACCOUNTS as DSKeys,
```
vilsbole, line 12: Can we remove `LSKeys` completly?

vilsbole:
code lgtm

PR #3184, opened 2020-04-10
Harmonize settings title

vilsbole:
Make settings page title style correspond to 'Your Dashboard'

reference: https://www.figma.com/file/BY0SWc75teEUZzws8JdgLMpy/MyCrypto-GAU-Master?node-id=1522%3A84934

### Before
Screenshot 2020-04-10 at 3 45 10 PM

### After
Screenshot 2020-04-10 at 3 44 32 PM

### DIFF
--- a/common/AppRoutes.tsx
+++ b/common/AppRoutes.tsx
@@ -24,7 +24,6 @@ const layoutConfig = (path: string, isMobile: boolean): LayoutConfig => {
         fullW: true,
         bgColor: '#fff'
       };
-    case ROUTE_PATHS.SETTINGS.path:
     case ROUTE_PATHS.DASHBOARD.path:
       return {
         centered: true,
@@ -33,7 +32,7 @@ const layoutConfig = (path: string, isMobile: boolean): LayoutConfig => {
     case ROUTE_PATHS.SETTINGS.path:
       return {
         centered: true,
-        ...(isMobile && { paddingV: '0px' })
+        paddingV: isMobile ? '0px' : SPACING.MD
       };
     default:
       return {
--- a/common/AppRoutes.tsx
+++ b/common/AppRoutes.tsx
@@ -24,6 +24,7 @@ const layoutConfig = (path: string, isMobile: boolean): LayoutConfig => {
         fullW: true,
         bgColor: '#fff'
       };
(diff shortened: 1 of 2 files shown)


common/v2/features/Settings/Settings.tsx
```diff
@@ -11,15 +11,18 @@ import { AddressBookPanel, AddToAddressBook, GeneralSettings, DangerZone } from
...
-  margin-bottom: 22px;
-  color: #163150;
+  margin-bottom: 24px;
+  font-weight: bold;
+  margin-top: 0;
 `;
 
 const SettingsHeadingIcon = styled.img`
-  margin-right: 12px;
+  margin-right: 24px;
```
blurpesec, line 23: should these use `${SPACING}` variables?

amacar:
Is the alignment of the icon and the title intentional? Gear icon looks a little off.

image

common/AppRoutes.tsx
```diff
@@ -24,6 +24,7 @@ const layoutConfig = (path: string, isMobile: boolean): LayoutConfig => {
         fullW: true,
         bgColor: '#fff'
       };
+    case ROUTE_PATHS.SETTINGS.path:
```
amacar, line 27: There is the same case below (`case ROUTE_PATHS.SETTINGS.path:`) which is unreachable after adding this line.
Link to the case: https://github.com/MyCryptoHQ/MyCrypto/pull/3184/files#diff-58901591c0a8f6591a7b3c4d5b1c1904R33

common/v2/features/Settings/Settings.tsx
vilsbole: Well we don't have 16px increments yet and I hope this style will change when they see it, so I'm fine leaving them inline atm.

vilsbole:
> Is the alignment of the icon and the title intentional? Gear icon looks a little off.
> 
> image

Strangely it's what is indicating in the Figma files. It seems to be aligned with the inner circle of the icon. And since it's on several screens I suspect it's intentional. I'm expecting a design change when they see this.

common/AppRoutes.tsx
vilsbole: 😱 Good catch!

PR #3188, opened 2020-04-12
Dashboard design fixes

GuillaumeRx:
### What is it ?

Add padding to Tables so the content is align with the Table Heading and top right button

Balance breakdown under chart now take full width and align with CoinGecko logo

Address [CH5829](https://app.clubhouse.io/mycrypto/story/5829) and [CH5831](https://app.clubhouse.io/mycrypto/story/5831)

### DIFF
--- a/common/v2/components/Table.tsx
+++ b/common/v2/components/Table.tsx
@@ -16,6 +16,7 @@ import { Theme } from '@mycrypto/ui';
 import * as R from 'ramda';
 
 import { noOp } from 'v2/utils';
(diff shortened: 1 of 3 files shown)


vilsbole:
Screenshot 2020-04-13 at 1 47 57 PM


I think we would want to:
- right align the text in the right column.
- correct copy: 
  - "Of Your Funds" -> lower case
  - "Value in US Dollars" -> USD

Can you check with design plz?

PR #3191, opened 2020-04-13
Fixed slowdown when disabling MetaMask

FrederikBolding:
Fixed an obscure bug where disabling the MetaMask extension would cause massive amounts of requests to our API, re-renders of the site etc.

When polling for balances we only update the balance state in case there is changes. Due to the bug with node configuration, the request would always fail, and therefore there would always be a difference between the current balance and the new balance. This would trigger an infinite loop of requests to ethscan.


FrederikBolding:
[ch5835]

PR #3197, opened 2020-04-15
Fixed gas issues with Send Assets

FrederikBolding:
- Reduced renders and gas estimation calls
- Fixed gas estimates not being fetched initially
- Fixed a small styling issue

### DIFF
--- a/common/v2/features/SendAssets/components/SendAssetsForm.tsx
+++ b/common/v2/features/SendAssets/components/SendAssetsForm.tsx
@@ -274,7 +274,7 @@ const SendAssetsForm = ({ txConfig, onComplete }: IStepComponentProps) => {
         onSubmit={fields => {
           onComplete(fields);
         }}
-        render={({ errors, setFieldValue, setFieldTouched, touched, values, handleChange }) => {
+        render={({ errors, setFieldValue, setFieldTouched, touched, values }) => {
           if (getProTxValue(['setMainTransactionFormCallback'])) {
             getProTxValue(['setMainTransactionFormCallback'])(() => ({
               isValid: isFormValid,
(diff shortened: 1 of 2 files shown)


FrederikBolding:
[ch5836] [ch5837]

common/v2/features/SendAssets/components/fields/GasPriceSlider.tsx
```diff
@@ -59,7 +70,13 @@ export default class SimpleGas extends Component<Props> {
               <div className="GasPriceSlider-slider">
                 <SliderWithTooltip
                   {...field}
-                  onChange={e => form.setFieldValue('gasPriceSlider', e)}
+                  onChange={e => {
+                    this.setState({ gasPrice: e.toString() });
+                  }}
```
vilsbole, line 75: Since this component is in a form I believe we can make a controlled component and remove the inner state. It can receive the value from the parent, and update the parents state with the event handler.

common/v2/features/SendAssets/components/SendAssetsForm.tsx
```diff
@@ -274,14 +279,24 @@ const SendAssetsForm = ({ txConfig, onComplete }: IStepComponentProps) => {
...
 
+          // Set gas estimates if default asset is selected
+          useEffectOnce(() => {
+            if (!isEmpty(values.asset)) {
+              fetchGasPriceEstimates(values.network).then(data => {
+                setFieldValue('gasEstimates', data);
+                setFieldValue('gasPriceSlider', data.fast);
+              });
+            }
+          });
```
vilsbole, line 298: DANGER! Hooks should only be called at the top level. This is an unsupported behaviour https://reactjs.org/docs/hooks-rules.html#only-call-hooks-at-the-top-level

common/v2/features/SendAssets/components/fields/GasPriceSlider.tsx
FrederikBolding: I'm not sure I understand this. We want the inner state to reduce render calls in the form, as everytime we update the form it re-renders?

PR #3204, opened 2020-04-17
Scan tokens ux improvements

blurpesec:
[ch5875]

### Description
After adding an account, scan tokens only on the account being added. Change `scanTokens()` to use `model.updateAll` in provider instead of `model.update` to reduce overall "glitchiness".

##### Steps to Test:
1) Add accounts back to back with network tab open. After each add, you should only see a couple of new network requests instead of hundreds ;)
2) With 10+ accounts on your dashboard, click the `Scan Tokens` button with network tab open. This should result in 100-200 new network calls as opposed to the previous ~500-700 that were being conducted.

### DIFF
--- a/common/v2/services/Store/Account/AccountProvider.tsx
+++ b/common/v2/services/Store/Account/AccountProvider.tsx
@@ -26,6 +26,7 @@ export interface IAccountContext {
   addNewTransactionToAccount(account: IAccount, transaction: ITxReceipt): void;
   getAccountByAddressAndNetworkName(address: string, network: string): IAccount | undefined;
   updateAccountAssets(account: StoreAccount, assets: Asset[]): Promise<void>;
+  updateAllAccountsAssets(accounts: StoreAccount[], assets: Asset[]): Promise<void>;
   updateAccountsBalances(toUpate: IAccount[]): void;
   toggleAccountPrivacy(uuid: TUuid): void;
 }
@@ -50,7 +51,7 @@ export const AccountProvider: React.FC = ({ children }) => {
       const newAccountData = {
         ...accountData,
         transactions: [
-          ...accountData.transactions.filter(tx => tx.hash !== newTransaction.hash),
+          ...accountData.transactions.filter((tx) => tx.hash !== newTransaction.hash),
           newTxWithoutNetwork
         ]
       };
@@ -64,12 +65,12 @@ export const AccountProvider: React.FC = ({ children }) => {
         ([_, value]) => !value.isZero()
       );
 
-      const existingAccount = accounts.find(x => x.uuid === storeAccount.uuid);
+      const existingAccount = accounts.find((x) => x.uuid === storeAccount.uuid);
 
       if (existingAccount) {
         const newAssets: AssetBalanceObject[] = positiveAssetBalances.reduce(
           (tempAssets: AssetBalanceObject[], [contractAddress, balance]: [string, BigNumber]) => {
-            const tempAsset = assets.find(x => x.contractAddress === contractAddress);
+            const tempAsset = assets.find((x) => x.contractAddress === contractAddress);
             if (tempAsset) {
               tempAssets.push({
                 uuid: tempAsset.uuid,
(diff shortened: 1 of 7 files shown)


blurpesec:
[ch5875]

FrederikBolding:
Would prefer that @vilsbole takes a look at this as well. Looks good to me though.

common/v2/services/Store/Account/AccountProvider.tsx
```diff
@@ -86,14 +87,55 @@ export const AccountProvider: React.FC = ({ children }) => {
...
+                tempAssets: AssetBalanceObject[],
+                [contractAddress, balance]: [string, BigNumber]
+              ) => {
+                const tempAsset = assets.find((x) => x.contractAddress === contractAddress);
+                if (tempAsset) {
+                  tempAssets.push({
+                    uuid: tempAsset.uuid,
+                    balance: balance.toString(10),
+                    mtime: Date.now()
+                  });
```
vilsbole, line 113: Avoid mutating an existing object.

common/v2/services/Store/Account/AccountProvider.tsx
```diff
@@ -86,14 +87,55 @@ export const AccountProvider: React.FC = ({ children }) => {
...
+            const newAssets: AssetBalanceObject[] = positiveAssetBalances.reduce(
+              (
+                tempAssets: AssetBalanceObject[],
+                [contractAddress, balance]: [string, BigNumber]
+              ) => {
+                const tempAsset = assets.find((x) => x.contractAddress === contractAddress);
+                if (tempAsset) {
+                  tempAssets.push({
+                    uuid: tempAsset.uuid,
+                    balance: balance.toString(10),
```
vilsbole, line 111: In memory it's better for the balance to be stored as a BigNumber.
It's only when it's saved to localstorage that we need to coerce to string.
And that's the responsibility of demarshalling

common/v2/services/Store/Account/AccountProvider.tsx
```diff
@@ -86,14 +87,55 @@ export const AccountProvider: React.FC = ({ children }) => {
         state.updateAccount(existingAccount.uuid, existingAccount);
       }
     },
-    updateAccountsBalances: toUpdate => {
+    updateAllAccountsAssets: async (storeAccounts, assets) =>
```
vilsbole, line 90: Can you try refactoring using lenses? Basically this is what we want to do:
```
const people = [{
  name: 'Terry',
  cats: ['Korin', 'Sweep', 'Caterina']
}, {
  name: 'Trevor',
  cats: ['Jazz']
}]
const cats = R.lensProp('cats')
const uppercaseCats = R.map(R.over(cats, R.map(R.toUpper)))
uppercaseCats(people)
> [{
    name: 'Terry',
    cats: ['KORIN', 'SWEEP', 'CATERINA']
  }, {
    name: 'Trevor',
    cats: ['JAZZ']
  }]
```
https://itnext.io/a-beginners-guide-to-ramda-part-2-lenses-62bdd3993598

package.json
```diff
@@ -154,7 +154,7 @@
     "fork-ts-checker-webpack-plugin": "3.1.1",
     "friendly-errors-webpack-plugin": "1.7.0",
     "html-webpack-plugin": "3.2.0",
-    "husky": "4.0.9",
+    "husky": "4.0.10",
```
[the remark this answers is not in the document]
blurpesec: bumped because it was throwing errors on `git push` ill push it to latest

common/v2/services/Store/Account/AccountProvider.tsx
blurpesec: I've been scratching my head over how to do this and realize that there may be something wrong with implementing it this way. Currently `updateAllAccountsAssets` doesn't take in `storeAccounts` and update `StoreProvider`'s `storeAccounts`, which is what I think this change would be helpful for. It currently takes in `storeAccounts` and updates `AccountProvider`'s `accounts` which `StoreProvider` is watching for changes to (in a `useMemo`), then updates `storeAccounts`. 

I didn't change that functionality at all, i just changed how many times we update `AccountProvider`'s `accounts` by swapping from `model.update` to `model.updateAll`, and swapping from just scanning all tokens to scanning tokens for 1 account in relevant places.

I think this change would be a good one to make, but it'd be a larger change and should probably be it's own task.

common/v2/services/Store/Account/AccountProvider.tsx
```diff
@@ -86,14 +87,55 @@ export const AccountProvider: React.FC = ({ children }) => {
...
+      Promise.all(
+        storeAccounts.map(async (storeAccount) => {
+          // Find all tokens with a positive balance for given account, and add those tokens to the assets array of the account
+          const assetBalances = await getAllTokensBalancesOfAccount(storeAccount, assets);
+          const positiveAssetBalances = Object.entries(assetBalances).filter(
+            ([_, value]) => !value.isZero()
+          );
+
+          const existingAccount = accounts.find((x) => x.uuid === storeAccount.uuid);
```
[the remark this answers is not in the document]
blurpesec: Reference this: https://github.com/MyCryptoHQ/MyCrypto/pull/3204#discussion_r411717926

common/v2/services/Store/Account/AccountProvider.tsx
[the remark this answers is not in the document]
blurpesec: Reference this: https://github.com/MyCryptoHQ/MyCrypto/pull/3204#discussion_r411717926

common/v2/services/Store/Account/AccountProvider.tsx
blurpesec: This is how it had been implemented in AccountProvider all-along

PR #3210, opened 2020-04-20
Fixed bug in Add Contact (settings panel)

blurpesec:
### Description
Fixed a bug resulting in white screen when add-address button is clicked in address book panel of settings page.

### Steps to Test
1) Navigate to mycryptobuilds.com/pr/3210/#/settings and scroll down till you're at the bottom of the Address Book Panel. Click (add address). Did it white screen? good!


common/v2/features/Settings/components/AddToAddressBook.tsx
```diff
@@ -90,7 +92,7 @@ export default function AddToAddressBook({ toggleFlipped, createAddressBooks }:
           label: '',
           address: '',
           notes: '',
-          network: ''
+          network: 'Ethereum'
```
amacar, line 95: Prefer `DEFAULT_NETWORK` from config :)

FrederikBolding: True 😄

PR #3211, opened 2020-04-20
Membership Improvements

FrederikBolding:
- Updated contract addresses
- Added membership TX to tx history
- Overhauled membership/expiry fetching


FrederikBolding:
Don't merge until I revert the price change, but good to go for review!

FrederikBolding:
[ch5853]

common/v2/services/Store/StoreProvider.tsx
```diff
@@ -135,14 +136,22 @@ export const StoreProvider: React.FC = ({ children }) => {
 
   const [memberships, setMemberships] = useState<MembershipStatus[] | undefined>([]);
 
+  const membershipExpiration = memberships
```
blurpesec, line 139: Can you change this var name so its more indicative of what it is?

PR #3212, opened 2020-04-20
Add mycrypto donation addr to address book

blurpesec:
### Description

Default will have mycrypto tip jar in the address book:

image

### DIFF
--- a/common/v2/database/seed/addressBook.ts
+++ b/common/v2/database/seed/addressBook.ts
@@ -14,3 +14,12 @@ export const addressBook: { [key in string]: AddressBook } = {
     network: 'Ethereum'
   }
 };
+
+export const defaultAddressBook: { [key in string]: AddressBook } = {
+  'a1acf1f2-0380-5bd6-90c3-2b4a0974a6fe': {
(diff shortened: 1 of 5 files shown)


blurpesec:
[ch5888]

common/v2/database/seed/addressBook.ts
```diff
@@ -14,3 +14,12 @@ export const addressBook: { [key in string]: AddressBook } = {
...
 };
+
+export const defaultAddressBook: { [key in string]: AddressBook } = {
+  'a1acf1f2-0380-5bd6-90c3-2b4a0974a6fe': {
+    label: 'MyCrypto Tip Jar',
+    address: '0x4bbeEB066eD09B7AEd07bF39EEe0460DFa261520',
+    notes: 'Toss us a coin!',
+    network: 'Ethereum'
+  }
+};
```
vilsbole, line 25: Can you move this into `/data` dir please?
I would like to reserve `/seed` to be reserved for fake data used for developers and use `/data` for defaults.

blurpesec: will dew.

PR #3213, opened 2020-04-21
Standardize the types for ITxReceipt and other transaction-related types

lubejmatej:
Closes [ch5652]


lubejmatej:
This PR can potentially break, Send Assets flow and Broadcast transaction. I did test the flow. But I am unsure if there are any side effects.
We could also revert, the flows back, and only use unified types.

vilsbole:
Tested send eth successfully. Merging.

PR #3221, opened 2020-04-24
Added txType to analytics when account is added to LS

blurpesec:
### Description
This allows us to record _only_ the type of tx (Standard, DefiZap, Swap, Contract Interaction, Contract Deploy, Membership Purchase, or Unknown if missing on the flow) being saved to account object so that we can know what features people are using MyCrypto for.

### DIFF
--- a/common/v2/services/Store/Account/AccountProvider.tsx
+++ b/common/v2/services/Store/Account/AccountProvider.tsx
@@ -56,7 +56,7 @@ export const AccountProvider: React.FC = ({ children }) => {
         'stage' in newTxWithoutNetwork &&
         [ITxStatus.SUCCESS, ITxStatus.FAILED].includes(newTxWithoutNetwork.stage)
       ) {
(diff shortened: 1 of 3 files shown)


common/v2/services/Store/Account/AccountProvider.tsx
```diff
@@ -11,12 +11,15 @@ import {
...
-  TUuid
+  TUuid,
+  ITxStatus,
+  ITxType
 } from 'v2/types';
 import { DataContext } from '../DataManager';
 import { SettingsContext } from '../Settings';
 import { getAccountByAddressAndNetworkName } from './helpers';
 import { getAllTokensBalancesOfAccount } from '../BalanceService';
+import { AnalyticsService, ANALYTICS_CATEGORIES } from 'v2/services/ApiService/Analytics';
```
FrederikBolding, line 22: Import ordering

PR #3225, opened 2020-04-25
add Fantom FTM network & staking contract

dfadden424:


vilsbole:
Please update the PR and get to green so we can review.

PR #3229, opened 2020-04-28
Update Trezor libraries

### LINKED ISSUE #3200: upgrade trezor-connect dependency to latest
Please upgrade the trezor-connect[1] dependency of your application to
latest (8.1.1 atm), otherwise your users won't be able to enter the
passphrase on the device.

Thanks!

[1] https://www.npmjs.com/package/trezor-connect

Mrtenz:
Closes #3200.


FrederikBolding:
This does currently not solve the issue - we're still investigating.

leshik:
@FrederikBolding Trezor changed the defaults, now passphrase is asked on the host by default, thus apps should be updated to allow this input. The old behavior can be restored but requires using `trezorctl` CLI.

blurpesec:
This worked fine for adding acc + sending tx on web. Linux electron build didn't work. Can't connect in add-account flow due to CSP i think?
image

FrederikBolding:
[ch6004]

shared/enclave/preload.ts
```diff
@@ -4,6 +4,6 @@ import { PROTOCOL_NAME } from 'shared/enclave/utils';
 export function registerProtocol() {
   // Whitelist custom protocol
   protocol.registerSchemesAsPrivileged([
-    { scheme: PROTOCOL_NAME, privileges: { standard: true, secure: true } }
+    { scheme: PROTOCOL_NAME, privileges: { standard: true, secure: true, bypassCSP: true } }
```
vilsbole, line 7: This is way too permissive even in electron: https://www.contextis.com/us/blog/basic-electron-framework-exploitation
How about we add the required trezor domain to our CSP instead?

FrederikBolding: This is exclusively for the protocol that accesses the enclave. Is it still an issue then?

Mrtenz: Wouldn't it be possible to add the protocol itself to CSP? e.g. `connect-src eth-enclave:;`?

FrederikBolding: Maybe - but we only want it to be set for Electron.

blurpesec:
> This worked fine for adding acc + sending tx on web. Linux electron build didn't work. Can't connect in add-account flow due to CSP i think?
> image

I was able to add account now and send tx:
image

wtzb:
Was able to add a Ropsten Trezor account. However, I got stuck on the following screen after confirming a Ropsten transaction on my Trezor:

image

All the necessary Trezor popups showed up successfully, and I was able to confirm the transaction on my device. However, after confirming everything, the MyCrypto screen didn't continue.

Console:
image

It seems like a Ropsten (or non-Ether chain) issue. Sending ETH works just fine.

image

PR #3231, opened 2020-04-30
Reduced token scanning after DefiZap TX goes through

FrederikBolding:
Changed so we only scan the account in question instead of all accounts.


FrederikBolding:
Small change - but needs a bit of testing. Haven't had time to go through it yet.

vilsbole:
When I add an account on beta I get 1 call to `tokens?contractAddress...` vs 2 calls with this change.

FrederikBolding:
This seems unrelated to my change @vilsbole ?

PR #3243, opened 2020-05-03
Fixed white screen in case of errors with keystore

FrederikBolding:
As per Bowens bug document


FrederikBolding:
[ch5786]

bowensanders:
I tested the branch locally and it no longer errors out to whitepage on incorrect password! 👍

PR #3246, opened 2020-05-04
Github Actions - mycryptobuilds migration

miagx:
[CH5801]


vilsbole:
@miagx Do we need both of these jobs? They seem both used to publish to S3.
```
- name: S3 Sync - Sha


and 

- name: Mycrypto-bot trigger

```

PR #3250, opened 2020-05-05
Update AccountDropdown for really long account names, add account nam…

lubejmatej:
Closes [ch5905]
Update balance account with with account name.
Trim Account names in Account dropdown.

### DIFF
--- a/common/v2/features/Dashboard/components/WalletBreakdown/AccountDropdown.tsx
+++ b/common/v2/features/Dashboard/components/WalletBreakdown/AccountDropdown.tsx
@@ -4,7 +4,7 @@ import styled, { StyledFunction } from 'styled-components';
 
 import { translateRaw } from 'v2/translations';
 import { Checkbox } from 'v2/components';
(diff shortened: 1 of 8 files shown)


common/v2/features/Dashboard/components/WalletBreakdown/AccountDropdown.tsx
```diff
@@ -94,8 +95,18 @@ const SCheckbox = styled(Checkbox)`
...
+  > label {
+    min-width: 20px;
+  }
+
+  img {
+    min-width: 30px;
+  }
 `;
 
+const trimAccountLabel = (label: string) => (label.length > 65 ? `${label.substr(0, 65)}…` : label);
```
vilsbole, line 108: Please
- create utility function which accepts a dynamic length (to encourage reuse).
- add simple unit test (respects length, handles undefined, includes ellipsis) to facilitate maintenance

PS: We're going to start focusing on increasing our code and case coverages. Abstracting utility functions and uniting testing when possible is a good place to start.

PR #3256, opened 2020-05-06
Remove zero balances 

blurpesec:
[ch6039]

### Description
Filters out zeroed balances in account asset balance updates, to remove assets that users used to have but don't anymore.

### DIFF
--- a/common/v2/services/Store/BalanceService.tsx
+++ b/common/v2/services/Store/BalanceService.tsx
@@ -140,14 +140,14 @@ export const getAccountsAssetsBalances = async (accounts: StoreAccount[]) => {
     ].map((p) => p.catch((e) => console.debug(e))) // convert Promise.all ie. into allSettled https://dev.to/vitalets/what-s-wrong-with-promise-allsettled-and-promise-any-5e6o
   );
 
-  const filterNonZeroBN = (n: BigNumber) => !bigify(n.toString()).eq(0);
+  const filterZeroBN = (n: BigNumber) => !bigify(n.toString()).eq(0);
 
   const filteredUpdatedAccounts = updatedAccounts.map((updatedAccount) => ({
     ...updatedAccount,
     assets:
       (updatedAccount &&
         updatedAccount.assets &&
-        updatedAccount.assets.filter(({ balance }) => !filterNonZeroBN(balance))) ||
+        updatedAccount.assets.filter(({ balance }) => !filterZeroBN(balance))) ||
       []
   }));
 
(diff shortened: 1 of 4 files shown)


common/v2/services/Store/BalanceService.tsx
```diff
@@ -7,12 +7,13 @@ import {
 } from '@mycrypto/eth-scan';
 import partition from 'lodash/partition';
 import { default as BN } from 'bignumber.js';
-import { bigNumberify } from 'ethers/utils';
+import { bigNumberify, BigNumber } from 'ethers/utils';
```
vilsbole, line 10: `bigify` uses `bignumber.js` and ethers version is a modified one. 
Let's create a type alias that we control. Like `TBN` ?
That way we have better control on the TS.

common/v2/services/Store/BalanceService.tsx
```diff
@@ -132,14 +133,25 @@ export const getAccountsAssetsBalances = async (accounts: StoreAccount[]) => {
...
-  const accountBalances = await Promise.all(
+  const updatedAccounts = await Promise.all(
     [
       ...ethScanCompatibleAccounts.map(getAccountAssetsBalancesWithEthScan),
       ...jsonRPCAccounts.map(getAccountAssetsBalancesWithJsonRPC)
     ].map((p) => p.catch((e) => console.debug(e))) // convert Promise.all ie. into allSettled https://dev.to/vitalets/what-s-wrong-with-promise-allsettled-and-promise-any-5e6o
   );
 
-  return accountBalances;
+  const filterZeroBN = (n: BigNumber) => !bigify(n.toString()).eq(0);
```
vilsbole, line 143: Use `isZero()` from bignumber https://mikemcl.github.io/bignumber.js/#isZ

common/v2/services/Store/StoreProvider.tsx
```diff
@@ -326,7 +326,9 @@ export const StoreProvider: React.FC = ({ children }) => {
       return sortBy(prop('ticker'), uniq);
     },
     assets: (selectedAccounts = state.accounts) =>
-      selectedAccounts.flatMap((account: StoreAccount) => account.assets),
+      selectedAccounts
+        .flatMap((account: StoreAccount) => account.assets)
+        .filter((asset) => !asset.balance.eq(0)),
```
vilsbole, line 331: This is only hides them from the view.
Wouldn't we want to delete them from storage instead?

common/v2/services/Store/BalanceService.tsx
blurpesec: We ended up not needing `bigify` in this because ethers.js has the `.isZero()` method already.
> Let's create a type alias that we control. Like TBN ?
That way we have better control on the TS.

Should this be it's own task, since it'll likely end up needing to be distributed throughout the `services/Store/` & `utils/` folders?

PR #3266, opened 2020-05-09
Reduce bundle size further

FrederikBolding:
- Reduced imports of recharts
- Used already required crypto libs for `AES` and `SHA256` (was already in our bundle through other dependencies)
- Removed unused dependency `wallet-address-validator` and relevant code


common/v2/services/EthService/validators.ts
```diff
@@ -155,10 +154,6 @@ export function isValidETHAddress(address: string): boolean {
 export const isCreationAddress = (address: string): boolean =>
   address === '0x0' || address === CREATION_ADDRESS;
 
-export function isValidBTCAddress(address: string): boolean {
```
vilsbole, line 158: 😆

common/v2/features/ScreenLock/ScreenLockProvider.tsx
```diff
@@ -47,6 +48,18 @@ class ScreenLockProvider extends Component<
...
+    const decryptedBytes = aes.decrypt(utils.hex.toBytes(data));
+    return utils.utf8.fromBytes(decryptedBytes);
+  };
+
+  public encrypt = (data: string, key: Uint8Array) => {
+    const aes = new ModeOfOperation.ctr(key);
+    const encryptedBytes = aes.encrypt(utils.utf8.toBytes(data));
+    return utils.hex.fromBytes(encryptedBytes);
+  };
+
```
vilsbole, line 62: Definitely utility functions.
Extract from component and test please.
Add a comment, name or scope (could stay in the same in file) so we know that's it's only used for our password handling and not chain related.

vilsbole:
Tagging @lubejmatej just FYI.

PR #3269, opened 2020-05-11
Resize AssetIcon

blurpesec:
### Description
Changes the width of AssetIcon to be the same as the height, to make the images square.

Old:
image

New:
image


GuillaumeRx:
DAI icon looks weird now, maybe making the width fixed and the height auto should fix the problem and not loose proportion of the icons that are not square

blurpesec:
> DAI icon looks weird now, maybe making the width fixed and the height auto should fix the problem and not loose proportion of the icons that are not square

Seems fine to me? if you really look hard its like a tiny bit fuzzy, but thats the only thing i see?
image

GuillaumeRx:
If you compare with the screenshot of the old version you can see that it width compressed it, the old version respected the ratio of the icons (that's what the `height: auto` is doing, flex don't care about ratio if you don't use this trick)

PR #3270, opened 2020-05-11
Add Currency Selection

blurpesec:
# Description
Adds currency selection to settings page and updates `getRate` && `getAssetRate` methods to pull those rates and display on dashboard, and throughout the app :-)

To start, added support for USD, EUR, GBP, TRY (Turkish Lira), CNY (Chinese Yuan), INR (Indian Rupee), and RUB (Russian Rubles). I defaulted to showing the asset symbols in front of the numbers due to it seeming that the symbol being a prefix/suffix depends on the locality of the user, not necessarily homogeneous per currency.

Here's the currency selection:
image

Here's an example of balance breakdown:
image


LuckyLuker:
[ch177]

common/v2/components/TransactionFlow/TxReceipt.tsx
```diff
@@ -346,7 +352,10 @@ export const TxReceiptUI = ({
             <AssetIcon uuid={asset.uuid} size={'24px'} />
             <Amount
               assetValue={`${parseFloat(assetAmount()).toFixed(6)} ${assetTicker()}`}
-              fiatValue={`$${convertToFiat(parseFloat(assetAmount()), assetRate()).toFixed(2)}
+              fiatValue={`${Fiats[settings.fiatCurrency].symbol}${convertToFiat(
```
FrederikBolding, line 355: Can we do a helper function to find the fiat instead this kind of lookup?

(Same for other files)

common/v2/config/fiats.ts
```diff
@@ -13,12 +13,38 @@ export const USD = {
...
-  symbol: '€' as TSymbol
+  symbol: '€' as TSymbol,
+  prefix: true
 };
 export const GBP = {
   code: 'GBP',
   name: 'British Pounds',
-  symbol: '£' as TSymbol
+  symbol: '£' as TSymbol,
+  prefix: true
```
FrederikBolding, line 23: Not sure if prefixing by default is the best option? I think I'm used to seeing it suffixed unless it's USD?

common/v2/features/SendAssets/components/SendAssetsForm.tsx
```diff
@@ -512,9 +515,9 @@ const SendAssetsForm = ({ txConfig, onComplete }: IStepComponentProps) => {
                       values.advancedTransaction ? values.gasPriceField : values.gasPriceSlider
                     }
                     fiatAsset={{
-                      fiat: 'USD',
+                      fiat: Fiats[settings.fiatCurrency].code || 'USD',
```
FrederikBolding, line 518: Can this ever be undefined?

common/v2/features/SendAssets/components/SendAssetsForm.tsx
```diff
@@ -512,9 +515,9 @@ const SendAssetsForm = ({ txConfig, onComplete }: IStepComponentProps) => {
...
                       rate: (getAssetRate(baseAsset || undefined) || 0).toString(),
-                      symbol: '$'
+                      symbol: Fiats[settings.fiatCurrency].symbol || '$'
```
FrederikBolding, line 520: Can this ever be undefined?

common/v2/components/TransactionFlow/TxReceipt.tsx
blurpesec: sure sure. that'd be good to help with `fiatCurrency of undefined` errs too :)

common/v2/features/SendAssets/components/SendAssetsForm.tsx
blurpesec: I suppose not :-0

common/v2/features/SendAssets/components/SendAssetsForm.tsx
blurpesec: I suppose not :-0

common/v2/config/fiats.ts
blurpesec: It seems to be based on locality of the user, not on the currency symbol itself. But if it's not actually a symbol, but a string of letters instead you're supposed to post-fix. We don't really have a way to implement the localities though.. soooo?

PR #3273, opened 2020-05-12
Find ways to make app load faster

lubejmatej:
Closes [ch5920]

I did some improvement, but the load time are still far from perfect. But it seems like every thing I try, the load time only increases.


lubejmatej:
Before merging this one, a loading component should be replaced with placed in `<Suspense fallback=""`  tag.

lubejmatej:
https://mycryptoapi.com/api/v1/assets.json zipping, or at least minify this JSON file, could also improve loading times.

webpack_config/common.js
```diff
@@ -58,18 +88,22 @@ module.exports = {
   module: {
     rules: [
       /**
-       * TypeScript files
+       * TypeScript files without stories
        */
       {
-        test: /\.tsx?$/,
+        test: /(?!.*\.stories\.tsx?$).*\.tsx?$/,
```
vilsbole, line 94: 👍

webpack_config/common.js
```diff
@@ -40,16 +40,46 @@ module.exports = {
...
+        },
+        common: {
+          enforce: true,
+          name: 'common.bundle',
+          test: /[\\/]common[\\/]/,
+          minChunks: 2,
+          reuseExistingChunk: true,
+          priority: 10
+        },
+        vendorIndividual: {
```
vilsbole, line 66: Can we find a better name?
'Individual' is not self evident. What is it suppose to represent?

webpack_config/production.js
```diff
@@ -79,6 +67,33 @@ module.exports = merge.smart(common, {
...
+      }
+    }),
+
+    new CompressionPlugin({
+      filename: '[path].gz',
+      algorithm: 'gzip',
+      test: /\.js$|\.css$|\.html$/,
+      threshold: 10240,
+      minRatio: 0.8
+    }),
```
vilsbole, line 95: Does it make a difference if the gzip is done by the build process instead of the server?

common/v2/routing/routes.tsx
```diff
@@ -1,37 +1,144 @@
...
+import { ROUTE_PATHS, IS_ACTIVE_FEATURE } from 'v2/config';
+import { Home, Dashboard, NoAccounts } from 'v2/features';
 import { requiresDesktopApp } from './helpers';
 
+const CreateWallet = lazy(() =>
+  import(
+    /* webpackChunkName: "CreateWallet" */
+    'v2/features/CreateWallet/CreateWallet'
+  )
+);
```
vilsbole, line 14: 1. Can we use aliases eg. `@features` and index files for the imports?
2. Can we create these dynamically? Something like:
```
const COMPONENTS = {
 AddAccountFlow: '@features/AddAccount'
}

const lazify = (v, k) => lazy(() => import(`${v}))
const lazyComponents = mapObjIndexed(lazify, COMPONENTS)
```
3. You mentioned that during experimentation the bundle size would increase. Could it be because of each of these includes have a lot of shared components between them that are included in each bundle? Would it be better to create larger groups like (AddAccount + SendAssets + SwapAssets)?

common/v2/routing/routes.tsx
```diff
@@ -1,37 +1,144 @@
...
+    /* webpackChunkName: "Mnemonic" */
+    'v2/features/CreateWallet/Mnemonic/Mnemonic'
+  )
+);
+const Keystore = lazy(() =>
+  import(
+    /* webpackChunkName: "Keystore" */
+    'v2/features/CreateWallet/Keystore/Keystore'
+  )
+);
```
vilsbole, line 38: FYI: These (and their dependencies are electron only.

lubejmatej: Ok, they will be lazy loaded anyway, in case the route is hit.

common/v2/routing/routes.tsx
lubejmatej: 1. Yeah, this was already rebased;
2. Not really, because it needs to be statically available. Also just tried it.
3. Yeah, this is my next step, decreasing main bundle size, and joining components together, so that each separate page bundle gets bigger. But this would also mean, that reexporting some components would not be a wise thing to do, since webpack, can't decide, if they can be joined in separate page bundle and main bundle.

webpack_config/common.js
lubejmatej: Well, this was already implemented, and I didn't want to break existing code. It's for some bigger packages to be seperated from node_modules

webpack_config/production.js
lubejmatej: Not really, I am just testing locally with this, just so Lighthouse does not complain, that files are not zipped. And also to give me more clearer picture, what production build size would look like.

lubejmatej:
image
This are the latest results. First Contentful Paint went from 4.5s(beta.mycrypto.com) -> 1.2s

lubejmatej:
@vilsbole I just pushed the prefetch flags again. Basically the rest of page chunks should be auto prefetched, after app is already loaded, so it wouldn't affect the load times. But it would increase speed, after moving to other pages of application.

Prefetching works with a plugin that disables CSP i.e.(https://chrome.google.com/webstore/detail/disable-content-security/ieelmcmcagommplceebfedjlakkhpden/related?hl=en), but when trying to add domain to the meta tag "Content-Security-Policy" it seems it didn't work. Maybe I was doing something wrong.

lubejmatej:
@vilsbole the only way I make the prefetching works is setting `"default-src 'self'`, which is far from ideal, because it makes page less secure. I would suggest pausing the feature for prefetching until https://bugs.chromium.org/p/chromium/issues/detail?id=801561 is resolved.

vilsbole:
Ok, then let's pause the prefetch, keep `default-src 'none'` and merge these changes. We'll be able to activate pre-fetch once the bug is fixed.

PR #3276, opened 2020-05-12
Replace ContractProvider with custom hook

vilsbole:
Proposal to replace the intermediate contexts with custom custom hooks.
- It implements the same api, so there are only TS changes to the callers.
- It's simple to test (demo included)
- It reduces the dependency chain of in `AppProviders`.
- It simplifies testing since the caller components only need to mock `DataProvider`.
- I expect that removing nested providers will help reduce scripting time on first load.


src/services/Store/Contract/useContracts.spec.tsx
```diff
@@ -0,0 +1,61 @@
...
+  const wrapper: React.FC = ({ children }) => (
+    <DataContext.Provider value={({ contracts, createActions } as any) as IDataContext}>
+      {' '}
+      {children}
+    </DataContext.Provider>
+  );
+  return renderHook(() => useContracts(), { wrapper });
+};
+
+describe('useContract', () => {
```
blurpesec, line 23: Can you add some more unit tests for createContract , createContractWithId , and deleteContract ?

PR #3278, opened 2020-05-12
Set table max-height

blurpesec:
### Description
Sets table max-height and overflow so that they're not infinitely long.


src/components/WalletUnlock/DeterministicWallets.tsx
```diff
@@ -448,7 +448,6 @@ export function DeterministicWalletsClass({
         head={['#', translateRaw('ADDRESS'), symbol, translateRaw('ACTION_5')]}
         body={wallets.map((wallet) => renderWalletRow(wallet, network, symbol))}
         config={{
-          hiddenHeadings: ['#', translateRaw('ACTION_5')],
```
FrederikBolding, line 451: Can we remove these some other way then?

src/features/Dashboard/components/WalletBreakdown/BalancesDetailView.tsx
```diff
@@ -289,7 +289,7 @@ export default function BalancesDetailView({
           </BalancesOnlyTotal>
         }
       >
-        <CollapsibleTable {...balancesTable} />
+        <FixedSizeCollapsibleTable {...balancesTable} maxHeight={'650px'} />
```
FrederikBolding, line 292: No need to set max height to default.

FrederikBolding:
Well okay then..

PR #3281, opened 2020-05-13
DeFiZap: Design Improvements and Rebranding (Zapper)

lubejmatej:
Closes [ch5978]


GuillaumeRx:
Can you please rebase your branch in order to get DefaultAsset bug resolution

src/features/DeFiZap/components/DashboardZapCTA.tsx
```diff
@@ -28,11 +40,31 @@ const DashboardZapCTA = ({ className }: any) => {
...
+                  inline={true}
+                  fontSize={FONT_SIZE.BASE}
+                  fontColor={COLORS.BLUE_BRIGHT}
+                  link={getKBHelpArticle(KB_HELP_ARTICLE.HOW_TO_USE_ZAPPER)}
+                  text={translateRaw('READ_MORE')}
+                />
+              )
+            }}
+          />
+        </SubHeaderText>
```
GuillaumeRx, line 60: You can use Markdown in translation file to define the link

src/features/DeFiZap/components/ZapCard.tsx
```diff
@@ -139,14 +149,15 @@ const ZapCardRiskProfile = styled('div')`
...
+const ZapCardButton = styled(Button)`
   @media (min-width: ${BREAK_POINTS.SCREEN_XS}) and (max-width: ${BREAK_POINTS.SCREEN_MD}) {
     width: 100%;
   }
   display: flex;
+  flex: 1 auto;
   justify-content: center;
   align-items: center;
-  width: ${(props) => props.width};
+  width: 100%;
```
GuillaumeRx, line 160: Copy don’t respect designs, `add` button should be smaller than `withdraw` button

lubejmatej: Yeah, I realize, but in the task description, it says, they should be 50%/50%. https://app.clubhouse.io/mycrypto/story/5978/defizap-design-improvements-and-rebranding-zapper

src/features/DeFiZap/components/DashboardZapCTA.tsx
lubejmatej: I think this is a better way, hardcoding links in translations file is not the best way, and this way the link could also be customized more easily. For example, changing color, adding icon, and so on.

src/features/DeFiZap/components/ZapCard.tsx
GuillaumeRx: You’re right ! All good then :)

src/features/DeFiZap/components/DashboardZapCTA.tsx
lubejmatej: Hardcoding this in translation files, it would require changing all languages links, this way only one line needs to change.

GuillaumeRx: Yes it adds more flexibility but won’t follow global style, we're trying to uniform the style to have less style rules declarations within small components

It don't even needs to be hardcoded, you can keep the variables but define the link as `[$text]($link)` in translation file so no need to change all languages links :)

PR #3287, opened 2020-05-15
Implement WebHID and WebUSB with U2F fallback

Mrtenz:
This implements WebHID and WebUSB, with fallback to U2F on error. I also updated all Ledger packages to the latest version.


src/services/WalletService/deterministic/ledger.ts
```diff
@@ -131,8 +134,24 @@ export class LedgerWallet extends HardwareWallet {
...
+    }
+
+    if (await TransportWebUSB.isSupported()) {
+      return TransportWebUSB.create();
+    }
+  } catch {
+    // Fallback to U2F
+  }
+
+  return TransportU2F.create();
```
vilsbole, line 150: Shouldn't this be in the catch block?

src/types/vendor/ledger/hw-transport-webhid.d.ts
```diff
@@ -0,0 +1,73 @@
+declare module '@ledgerhq/hw-transport-webhid' {
```
vilsbole, line 1: Please add comment with library version number and source link

src/types/vendor/ledger/hw-transport-webusb.d.ts
```diff
@@ -0,0 +1,73 @@
+declare module '@ledgerhq/hw-transport-webusb' {
```
vilsbole, line 1: Please add comment with library version number and source link

PR #3312, opened 2020-05-26
Setup Testcafe

vilsbole:
Setup `yarn test:e2e` to run Testcafe suite.


lubejmatej:
Closes [ch6216].
In order for the e2e tests to works, you should spin up local dev server.
And fill `yarn test:e2e:set:env` with valid mnemonic phrase, that should contain at least one account.

PR #3314, opened 2020-05-27
Added Telegram link in footer

409H:
[CH6189]

🎉 🎉 🎉

## Description

* Added option to add an svg icon to the SocialMediaLink part as the fontpack we use does not support everything

Screenshot 2020-05-27 at 22 08 07

Screenshot 2020-05-27 at 22 06 08

## Changes

* Changes to the `Link` interface to have a new property `icon` to specify an svg

## Quality Assurance

- [ ] If your code adds new text to the UI, the added text is [translatable](https://github.com/MyCryptoHQ/MyCrypto/wiki/Contributing---Translatable-strings)


src/config/links.ts
```diff
@@ -2,9 +2,12 @@ import { translateRaw } from '@translations';
 import { CRYPTOSCAMDB } from './data';
 import { TURL } from '@types';
 
+import SocialTelegramIcon from '../assets/images/social-icons/social-telegram-white.svg';
+
 interface Link {
   link: string;
   text: string;
+  icon: null | string;
```
FrederikBolding, line 10: Would prefer `icon?: string`.

Then it can be `undefined` and thus not necessary to declare `null` every time below.

src/features/Layout/Footer/components/SocialsAndLegal.tsx
```diff
@@ -7,11 +7,33 @@ import { DisclaimerModal, NewTabLink } from '@components';
...
+  if (icon) {
+    return (
+      <NewTabLink
+        className="SocialMediaLink"
+        key={link}
+        href={link}
+        aria-label={text}
+        onClick={() => trackSocialIconClicked(text, trackSocial)}
+      >
+        <img src={icon} width="18.41px" height="18px" style={{ verticalAlign: 'top' }} />
```
blurpesec, line 32: Can this be `width="18px" height="18px"` ?

409H: Yeah - I was going to do that, but the others are 18.406/18.41px so thought I'd do the same sizing

src/config/links.ts
409H: I just pushed up changes

src/features/Layout/Footer/components/SocialsAndLegal.tsx
409H: I just pushed up changes

PR #3320, opened 2020-06-02
Revert ethers imports

lubejmatej:
Revert ethers imports, don't merge


blurpesec:
@lubejmatej this is g2g with just the first commit :_)

PR #3289, opened 2020-05-16
Nansen Provider for PTX

FrederikBolding:
- Changed to use Nansen as the provider for PTX UI
- Improved membership PTX flow

PR #3327, opened 2020-06-11
Protected tx UI fixes

vilsbole:
[[ch6124]](https://app.clubhouse.io/mycrypto/story/6124/protected-transactions-qa-review-changes)

ProtectedTx
- Harmonize font-weight in sidebar
- Center main and sidebar
- Activate `has-side-panel` class

Further changes may only be addressed after merging #3289 Nansen PR


blurpesec:
oops. just noticed this on the bottom of the panels: 
image

vilsbole:
Are you referring to the bottom line on the grey space between the Panels?
Currently the side panel is a child of ContentPanel which the shadow belongs to, so there is no simple way of removing it without recoding the entire panel

PR #3328, opened 2020-06-15
Set node version to lts

vilsbole:
Node use pair notation to specify it's lts version.
By using `12.x` we ensure we use the latest patches as well.


FrederikBolding:
Is this something we want to pursue? If so, please rebase @vilsbole

PR #3332, opened 2020-06-19
Slightly change handling for acc balance lookups

blurpesec:
### Description
Changes handling for acc balance lookups so that it better handles errors :3.

### What changed
Basically, it'll catch failed batch calls from ethscan and just feed back and empty object so that it doesn't fail out of the entire token scan from 1 failed lookup.

Also introduced the batch-size variable to limit size of lookups. This should decrease timeouts & errors but the tradeoff is it increases # of lookups.

Going to see what I can do to tweak load on our infra a little bit


blurpesec:
[ch6300]

PR #3338, opened 2020-06-26
Use checksum addresses for copyable address components

blurpesec:
### Description
Change address components to use checksum addresses.


blurpesec:
[ch6280]

PR #3341, opened 2020-06-30
Screen Lock Rework

FrederikBolding:


FrederikBolding:
[ch6091]

FrederikBolding:
Needs thorough testing before being merged.

src/features/ScreenLock/ScreenLockProvider.tsx
```diff
@@ -190,17 +222,18 @@ class ScreenLockProvider extends Component<
     this.setState({ locking: false, locked: true });
     document.title = translateRaw('SCREEN_LOCK_TAB_TITLE_LOCKED');
 
+    const isOutsideLock = (path: string) =>
+      !path.includes('screen-lock') && path !== ROUTE_PATHS.SETTINGS_IMPORT.path;
+
```
blurpesec, line 227: So this makes the lock screen occur everywhere in the app instead of just dashboard/no-accounts pages?

FrederikBolding: Yep! We allowed to show a bunch of pages that were invalid given that no accounts are loaded when locked.

PR #3350, opened 2020-07-06
PTX refactor, fixes and tests

FrederikBolding:
All around fixes to PTX + some refactoring.
- Undo functionality etc.
- Fixed issues with existing E2E tests
- Implemented small E2E test for send (PTX tests should be included when we have support for feature flags)


src/features/SendAssets/SendAssets.tsx
```diff
@@ -39,7 +41,7 @@ function SendAssets() {
       component: IS_ACTIVE_FEATURE.PROTECT_TX ? SendAssetsFormWithProtectTx : SendAssetsForm,
       props: (({ txConfig }) => ({ txConfig }))(txFactoryState),
       actions: (payload: IFormikFields, cb: any) => {
-        if (getProTxValue(['state', 'protectTxEnabled'])) {
+        if (getProTxValue(['state', 'protectTxEnabled']) && !isMyCryptoMember) {
```
vilsbole, line 44: How about we change this syntax and access the values directly from the context?
```
const { PTxEnabled } = useContext(PTxContext)
```

src/features/ProtectTransaction/ProtectTxProvider.tsx
```diff
@@ -8,10 +8,11 @@ import {
   GetTokenTxResponse
 } from '@services/ApiService';
 import { AssetContext, getAssetByUUID, StoreContext } from '@services/Store';
+import { NansenService, NansenServiceEntry } from '@services/ApiService/Nansen';
 import { useScreenSize } from '@utils';
+import { WALLETS_CONFIG } from '@config';
 
-import { WALLETS_CONFIG } from '../../config';
-import { NansenService, NansenServiceEntry } from '@services/ApiService/Nansen';
+import { IS_ACTIVE_FEATURE } from '../../config/isActiveFeature';
```
vilsbole, line 15: Import from @config please

src/features/SendAssets/components/SendAssetsForm.tsx
```diff
@@ -333,11 +338,7 @@ const SendAssetsForm = ({ txConfig, onComplete }: IStepComponentProps) => {
   const userAccountEthAsset = userAssets.find((a) => a.uuid === ETHUUID);
 
   return (
-    <div
-      className={`SendAssetsForm ${
-        getProTxValue(['state', 'mainComponentDisabled']) ? 'SendAssetsForm-disabled' : ''
```
vilsbole, line 338: Did you check if we can remove these class definitions as well?

src/features/ProtectTransaction/ProtectTxProvider.tsx
FrederikBolding: It's imported directly to prevent circular dependency issues.

PR #3351, opened 2020-07-06
Added trigger for save settings notification

blurpesec:
[ch6221]
### Description
Added trigger for the save settings notification and edited the interval it appears on

### DIFF
--- a/src/features/AddAccount/AddAccountFlow.tsx
+++ b/src/features/AddAccount/AddAccountFlow.tsx
@@ -90,10 +90,10 @@ const AddAccountFlow = withRouter(({ history, match }) => {
       .filter((a) => a !== undefined) as StoreAccount[];
 
     if (newAccounts.length !== 0) {
+      if (accountData.length > 1) displayNotification(NotificationTemplates.saveSettings);
       const handleAddition =
-        accountData.length === 1
+        newAccounts.length === 1
           ? () => {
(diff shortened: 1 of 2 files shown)


src/features/AddAccount/AddAccountFlow.tsx
```diff
@@ -93,6 +93,7 @@ const AddAccountFlow = withRouter(({ history, match }) => {
       const handleAddition =
         accountData.length === 1
           ? () => {
+              if (accountData.length > 1) displayNotification(NotificationTemplates.saveSettings);
```
GuillaumeRx, line 96: If we test for the length to be equal to 1, why do we test to see if it's greater than one after ?

blurpesec: > If we test for the length to be equal to 1, why do we test to see if it's greater than one after ?

We want to display the notification to save settings after the user adds their second account because that's the point where it starts to get annoying to re-add your accounts.

blurpesec: nice catch xD

GuillaumeRx:
The notification is not triggered for me no matter how much accounts I add :(

src/features/AddAccount/AddAccountFlow.tsx
GuillaumeRx: I think it should be greater or equal to 1 ;)

PR #3362, opened 2020-07-14
Change settings file name

FrederikBolding:


FrederikBolding:
[ch6383]

src/database/versions.ts
```diff
@@ -32,7 +35,7 @@ export const getPreviousDBConfig = () => dbVersions[dbHistory[1]];
 
 export const getExportFileName = () => {
   const currentDb = getCurrentDBConfig();
-  return `${currentDb.main}_${currentDb.version}.json`;
+  return `${SETTINGS_FILENAME}_${moment().format('YYYY-MM-DD')}_${currentDb.version}.json`;
```
vilsbole, line 38: Unit test plz.
To do so it may be simpler to pass the `dbConfig` and the date as arguments.
It may seem small, but it's these small tests that allow us to catch accidental mistakes such as setting a separator that doesn't have cross OS support, making a typo in the extension name etc.

PR #3370, opened 2020-07-18
Harmonise asset types

vilsbole:
[[6438]](https://app.clubhouse.io/mycrypto/story/6438/harmonize-asset-code-to-ticker)

- [x] We were using both `ticker` and `symbol` to represent the asset code eg. `ETH`
Since `code` is also used for languages, and `symbol` represent a currency symbol eg, `$`, harmonise asset code with `ticker`.
- [x] Harmonize `IAsset` and `SwapAsset`
- [x] Use `ticker` instead of `code` for Fiat 
- [x] Create TFiatTicker type and type guard
- [x] Remove `symbol` from `Asset`
- [x] Introduce `TFiatTicker`
- [x] Use `TFiatTicker` in `ISettings`


__tests__/send.test.js
```diff
@@ -74,7 +74,7 @@ test('Should be able to send ETH on Ropsten', async (t) => {
   await t.expect(getAllByText(FIXTURE_SEND_CONTACT)).ok();
 
   // Send TX
-  await t.click(getByText(getTransValueByKey('CONFIRM_AND_SEND')));
+  await t.click(getByText(new RegExp(getTransValueByKey('CONFIRM_AND_SEND'), 'i')));
```
blurpesec, line 77: its not immediately clear to me what the purpose of this change is.

vilsbole: It seems that sometimes the text is rendered differently so we use a regex pattern to get the values like in the videos `getByText(/confirm and send/i)`.
Since the search value is a variable, the only way i found how to do dynamic regex patterns in JS is with `new RegExp()`

PR #3371, opened 2020-07-20
Remove old landing page

GuillaumeRx:
_**REVIEW BUT DON'T MERGE WITHOUT THE NEW LANDING OUT**_

### Summary
* Remove Home
* Remove No Account
* Edit Root route to be Dashboard
* Fallback to Add Account instead of No Account when user don't have a wallet
* Add Import setting option to Add account footer
* Edit e2e tests to handle new redirections

### Steps to Test

1. Try to access app without any account : You should be redirected to `add-account`
2. Try using the app without account: You should be redirected to `add-account`
3. `/home` should fallback to a 404
4. `/` should land on `Dashboard`
5. on `add-account` you should find a link to import settings at the bottom

[Story Details](https://app.clubhouse.io/mycrypto/story/6082/)


GuillaumeRx:
Homepage and NoAccount where using styled-components, no scss files to clean 🧹

blurpesec:
> * Fallback to Add Account instead of No Account when user don't have a wallet

Can we add a trigger for when this is a No-Account redirect to display a notif saying "you don't have any accounts added to mycrypto yet, please add one or import a settings file if you have one" or something like that

GuillaumeRx:
> > * Fallback to Add Account instead of No Account when user don't have a wallet
> 
> Can we add a trigger for when this is a No-Account redirect to display a notif saying "you don't have any accounts added to mycrypto yet, please add one or import a settings file if you have one" or something like that at the top of the `/add-account` page?

Yes :) Gonna add it today

vilsbole (approved):

PR #3372, opened 2020-07-20
Allow scroll for mobile

GuillaumeRx:
### Fix bug with mobile view 🦄

### DIFF
--- a/src/types/global.d.ts
+++ b/src/types/global.d.ts
@@ -14,3 +14,39 @@ interface CustomWindow extends Window {
   setFeatureFlag?(key: keyof IIS_ACTIVE_FEATURE, value: boolean): void;
   resetFeatureFlags?(): void;
 }
+
+// Polyfill for ResizeObserver since type is missing from TS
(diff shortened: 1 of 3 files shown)


src/types/global.d.ts
```diff
@@ -14,3 +14,39 @@ interface CustomWindow extends Window {
   setFeatureFlag?(key: keyof IIS_ACTIVE_FEATURE, value: boolean): void;
   resetFeatureFlags?(): void;
 }
+
+// Polyfill for ResizeObserver since type is missing from TS
+interface DOMRectReadOnly {
```
vilsbole, line 19: Add url ref to typings

GuillaumeRx:
The `requestAnimationFrame` fix for the `resizeObserver` error significantly impact the performance of the tricks but it's the only way to fix it so the E2E test can run. This error is not an important error by the way

ref : https://stackoverflow.com/questions/49384120/resizeobserver-loop-limit-exceeded

PR #3397, opened 2020-07-29
Fix PTX fee mismatch

FrederikBolding:


FrederikBolding:
[ch6516]

src/services/EthService/utils/transactions.ts
```diff
@@ -50,10 +50,16 @@ export const appendGasLimit = (network: Network) => async (
...
+    const gasLimit = await (async () => {
+      // Respect gas limit if present
+      if (tx.gasLimit) {
+        return tx.gasLimit;
+      }
+      return await getGasEstimate(network, tx)
+        .then(hexToNumber)
+        .then((n: number) => Math.round(n * 1.2))
+        .then((n: number) => inputGasLimitToHex(n.toString()));
+    })();
```
vilsbole, line 62: We should be able to dispense with IIFE here.
A simple guard should do the trick
```
if (tx.gasLimit) return {...tx}
try {
  ...
} catch {


}
```

PR #3409, opened 2020-08-03
Fix PTX abort notification styling

GuillaumeRx:


GuillaumeRx:
The border doesn't exist on the designs ;) @FrederikBolding

PR #3410, opened 2020-08-03
Move to AccountSelector

vilsbole:
[ch6576]
1. Rename and test AccountSelector
2. Move defaultAccount into the form. 
- [x] SendAssets
- [x] SwapAssets
- [x] DeployForm
- [x] ZapForm
- [x] BuyAssets
- [x] ReceiveForm

For:
- [x] MembershipPurchase
- [x] WriteForm
It appears that the form has another way to choose default value.

3. Misc: 
- create export of rambda functions from @vendor to reduce import declarations within components.
- remove SendAsset form reset on form change (reduce amount of rerenders for each setField)
- speed up send e2e test by chaining assertions (save page load time between each one)
- For SendAssets, place nonce calculation in hook instead of after account selection. Adapt unit test for the call. 
- Correct bug in SendAssets where defaultNetwork was always Ethereum instead of the network of the defaultAccount.


src/components/AccountSelector.test.tsx
```diff
@@ -0,0 +1,69 @@
...
+  value: sortByLabel(fAccounts)[0],
+  onSelect: jest.fn()
+};
+
+describe('AccountSelector', () => {
+  test('it displays account info by default', async () => {
+    getComponent(defaultProps);
+    const accounts = sortByLabel(defaultProps.accounts);
+    // Displays the account label
+    expect(screen.getByText(new RegExp(accounts[0].label, 'i'))).toBeDefined();
```
FrederikBolding, line 31: Is there a reason you don't use the `getByText` returned by `getComponent`?

(Goes for everywhere in the PR)

src/components/AccountSelector.tsx
```diff
@@ -85,7 +82,7 @@ function AccountDropdown({ accounts, asset, name, value, onSelect }: IAccountDro
           </>
         );
       }}
-      value={selected}
+      // value={selected || options[0]}
```
FrederikBolding, line 85: ?

src/features/DeployContracts/DeployContractsFlow.tsx
```diff
@@ -12,6 +12,7 @@ import { BREAK_POINTS } from '@theme';
 import { deployContractsInitialState, DeployContractsFactory } from './stateFactory';
 import { Deploy, DeployConfirm, DeployReceipt } from './components';
 import { DeployContractsState } from './types';
+import { StoreContext } from '@services';
```
FrederikBolding, line 15: Reorder imports

jest_config/__fixtures__/account.ts
```diff
@@ -1,18 +1,62 @@
...
+    mtime: 0,
+    uuid: '4ffb0d4a-adf3-1990-5eb9-fe78e613f70b' as TUuid,
+    network: fNetwork,
+    label: 'WalletConnect Account 1'
+  },
+  {
+    address: '0xfE5443FaC29fA621cFc33D41D1927fd0f5E0bB7c' as TAddress,
+    networkId: 'Ropsten',
+    wallet: 'LEDGER' as WalletId,
+    dPath: '',
```
FrederikBolding, line 37: This might cause issues in the future, perhaps add a valid DPath?

src/vendor/ramda.ts
```diff
@@ -0,0 +1,12 @@
+/**
+ *  Explicit list of helper functions in order to reduce bundle size
+ */
+
+export { default as sortBy } from 'ramda/src/sortBy';
```
FrederikBolding, line 5: Why is this necessary? Can't we just use regular tree-shaking?

src/components/AccountSelector.tsx
vilsbole: Removed

src/features/DeployContracts/DeployContractsFlow.tsx
vilsbole: Done

jest_config/__fixtures__/account.ts
vilsbole: Good point!

src/components/AccountSelector.test.tsx
vilsbole: It's from https://kentcdodds.com/blog/common-mistakes-with-react-testing-library#not-using-screen
> The benefit of using screen is you no longer need to keep the render call destructure up-to-date as you add/remove the queries you need. You only need to type screen. and let your editor's magic autocomplete take care of the rest.

src/vendor/ramda.ts
vilsbole: It seems like it doens't matter wether we `import * as R from 'ramda'` or `import { path } from 'ramda'. Both escape Tree shaking. https://stackoverflow.com/a/45746950/2057532
This is why we started importing from the `ramda/src`.
To avoid the multiple lines, I find it simpler to place all the imports in `vendor`. It would also reduce the amount of import changes whenever they decide to support Tree shaking.

__tests__/send-assets-page.po.js
```diff
@@ -15,15 +15,13 @@ export default class SendAssetsPage extends BasePage {
...
     await t
-      .click(
-        Selector('[data-testid="selector"]')
-          .find('div')
-          .withText(getTransValueByKey('ACCOUNT_LOOKUP_SELECTION_PLACEHOLDER'))
-      )
+      .click(getByText(getTransValueByKey('ACCOUNT_LOOKUP_SELECTION_PLACEHOLDER')))
       .click(getAllByText(FIXTURE_SEND_CONTACT).nth(1))
+      // .click(Selector('div[data-testid="contact-selector"]').withText(FIXTURE_SEND_CONTACT))
       .click(Selector('input[name="amount"]').parent())
```
FrederikBolding, line 21: ?

src/features/SendAssets/__tests__/SendAssets.test.tsx
```diff
@@ -58,7 +62,7 @@ describe('SendAssetsFlow', () => {
     return simpleRender(component(pathToLoad));
   };
 
-  test('Can render the first step (Send Assets Form) in the flow.', () => {
+  test('Can render the first step (Send Assets Form) in the flow.', async () => {
```
FrederikBolding, line 65: Is it necessary for it to be `async`?

src/services/EthService/nonce.ts
```diff
@@ -3,5 +3,5 @@ import { ProviderHandler } from './network';
 
 export function getNonce(network: Network, address: TAddress) {
   const provider = new ProviderHandler(network);
-  return provider.getTransactionCount(address);
+  return provider.getTransactionCount(address) || 0;
```
FrederikBolding, line 6: Is this safe?

__tests__/send-assets-page.po.js
vilsbole: Removed

src/features/SendAssets/__tests__/SendAssets.test.tsx
vilsbole: Removed

src/services/EthService/nonce.ts
vilsbole: Hmm, I added it since:
1. SendAssets unit test need mock `ethers`
2. If I mock `ethers` getNonce result is `undefined` and everything crumbles.
3. Setting a default nonce value seemed reasonable. Unsure if this is the right way to do so.

FrederikBolding: I'm unsure of whethyer a default nonce is a good idea or not. Can we not mock the `getTransactionCount`?

vilsbole: From slack conversation:
> Emmanuel: I didn't find a way to mock provider:
couldn't mock a relative import inside dependency from within the test file.
couldn't find how to create an implementation of provider from within an ethers mock implementation https://github.com/MyCryptoHQ/MyCrypto/blob/285f33ed5be4c61c3c42fef91d60277853e62411/src/services/EthService/network/providerHandler.ts#L24

> Frederik: Not entirely sure, just feels weird to return something that's probably not true
Like in case of an error your nonce would be set to 0 and the tx would fail
without notifying the user that something is wrong

@blurpesec Any idea how to mock Provider?

PR #3411, opened 2020-08-04
Test Add-Account flow

blurpesec:
### Description
Add tests for add account flow for electron-only wallet types. View Only is already done

#### Todos:
- [x] Keystore File
- [ ] Mnemonic Phrase - wait for dpaths
- [x] Private Key


src/features/AddAccount/stories.tsx
```diff
@@ -53,23 +54,35 @@ export const getStories = (): IStory[] => [
   },
   {
     name: WalletId.KEYSTORE_FILE,
-    steps: [NetworkSelectPanel, IS_DEV || IS_ELECTRON ? KeystoreDecrypt : InsecureWalletWarning],
-    isDisabled: IS_NOT_ELECTRON_AND_IS_NOT_DEV
+    steps: [
+      NetworkSelectPanel,
+      IS_DEV || IS_STAGING || IS_ELECTRON ? KeystoreDecrypt : InsecureWalletWarning
+    ],
+    isDisabled: IS_NOT_ELECTRON_AND_IS_NOT_DEV_OR_STAGING
```
vilsbole, line 61: We got to be careful that this doesn't get out of hand.
1. Staging is suppose to mirror Production. If the feature set or build process start to deviate from each other, then the whole point of staging becomes useless.
- Can we use a feature toggle to enable it for tests? (eg. PTX test in `send.test.js`)
- Or can we run the tests just for electron builds? 
2. The double negation confuses me:
` isDisabled: IS_NOT_ELECTRON_AND_IS_NOT_DEV_OR_STAGING`
can we simplify to something like? 
```
isDisabled: IS_WEB && IS_PROD
//or

isEnabled: IS_ELECTRON || !IS_PROD
```

src/features/AddAccount/AddAccountForm.reducer.tsx
```diff
@@ -39,9 +39,9 @@ const handleUnlock = (walletType: WalletId | undefined, payload: any) => {
       case WalletId.VIEW_ONLY:
         return [payload];
       case WalletId.KEYSTORE_FILE:
-        return [payload];
+        return [{ ...payload, address: payload.getAddressString(), derivationPath: '' }];
```
vilsbole, line 42: L83 we remap to `dPath`. Might as well set it directly to `dPath` here.

__tests__/addaccount.test.js
```diff
@@ -36,7 +40,47 @@ test('When a view only address is added should be displayed in dashboard metrics
...
+});
+
+test('When a keystore file address is added should be displayed in dashboard metrics', async (t) => {
+  await clearLocalStorage(FIXTURE_MYC_STORAGE_KEY);
+  await addAccountPage.addKeystoreFile();
+
+  await dashboardPage.waitPageLoaded(FIXTURES_CONST.TIMEOUT * 2);
+
+  FIXTURE_PRIVATE_KEY_TOKENS.forEach((t) => dashboardPage.expectBalanceInBalanceList(t));
+});
```
vilsbole, line 86: Each page load is reeeeally expensive since it reloads the entire app.
Prefer to write fewer and longer tests, where you follow the app flow and chain expectations.
We may need to remove the expectations from the dashboard page.

__tests__/addaccount-page.po.js
```diff
@@ -29,4 +30,42 @@ export default class AddAccountPage extends BasePage {
...
+    // Lose focus
+    await t.expect(Selector('button').withText(getTransValueByKey('ADD_LABEL_6_SHORT')).hasAttribute('disabled')).notOk('ready for testing', { timeout: FIXTURES_CONST.TIMEOUT });
+    await t.click(Selector('button').withText(getTransValueByKey('ADD_LABEL_6_SHORT')));
+  }
+
+  async addKeystoreFile() {
+    await this.navigateToPage();
+    await this.waitPageLoaded();
+    await t.click(getByText(getTransValueByKey('X_KEYSTORE2')));
+    this.waitForPage(PAGES.ADD_ACCOUNT_KEYSTORE);
```
vilsbole, line 55: Prefer to make the expensive page navigation explicit in the test file.
It will be easier to identify bottle necks.

__tests__/addaccount-page.po.js
```diff
@@ -1,17 +1,18 @@
...
 
 export default class AddAccountPage extends BasePage {
   async navigateToPage() {
     this.navigateTo(PAGES.ADD_ACCOUNT);
   }
 
-  async waitPageLoaded() {
-    await this.waitForPage(PAGES.ADD_ACCOUNT);
+  async waitPageLoaded(timeToWait) {
+    await this.waitForPage(PAGES.ADD_ACCOUNT, timeToWait);
```
vilsbole, line 15: naming

__tests__/addaccount-page.po.js
```diff
@@ -29,4 +30,42 @@ export default class AddAccountPage extends BasePage {
...
+    this.waitForPage(PAGES.ADD_ACCOUNT_PRIVATE_KEY);
+
+    await t
+      .typeText(Selector('div[data-testid="selector"]').find('input'), FIXTURE_ETHEREUM)
+      .click(Selector('div[data-testid="selector"]').find('span').withText(FIXTURE_ETHEREUM))
+      .click(Selector('button').withText(getTransValueByKey('ACTION_6')))
+      .typeText(Selector(`div`).withText(getTransValueByKey('PRIVATE_KEY_PLACEHOLDER')).find('input'), FIXTURE_TEST_PRIVATE_KEY)
+
+    // Lose focus
+    await t.expect(Selector('button').withText(getTransValueByKey('ADD_LABEL_6_SHORT')).hasAttribute('disabled')).notOk('ready for testing', { timeout: FIXTURES_CONST.TIMEOUT });
```
vilsbole, line 47: Applies for both.
Atm it seems better to place the assertions within the test file.
Do we really need it here?

__tests__/dashboard-page.po.js
```diff
@@ -9,8 +9,8 @@ export default class DashboardPage extends BasePage {
     this.navigateTo(PAGES.DASHBOARD);
   }
 
-  async waitPageLoaded() {
-    await this.waitForPage(PAGES.DASHBOARD);
+  async waitPageLoaded(timeToWait) {
+    await this.waitForPage(PAGES.DASHBOARD, timeToWait);
```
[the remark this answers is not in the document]
blurpesec: yup

__tests__/addaccount-page.po.js
blurpesec: give it a check now? I moved the expect's back inside the test file

__tests__/addaccount.test.js
blurpesec: I made a change to not delete them but to combine them into one test so we only need to load the dashboard once per wallet type instead of 2x. If we notice that being slow in the future, we can work something else out

PR #3414, opened 2020-08-05
Refactor SendAssets to use a reducer

FrederikBolding:


FrederikBolding:
[ch6512]

src/features/SendAssets/SendAssets.reducer.tsx
```diff
@@ -0,0 +1,153 @@
...
+
+export type ReducerAction = TAction<ValuesType<typeof sendAssetsReducer.actionTypes>, any>;
+
+// @ts-ignore
+export const initialState: State = { txConfig: {} };
+
+export const sendAssetsReducer = (state: State, action: ReducerAction) => {
+  switch (action.type) {
+    case sendAssetsReducer.actionTypes.FORM_SUBMIT: {
+      const txConfig = handleFormSubmit(action.payload);
```
vilsbole, line 45: Applies for the whole reducer:
- since the reducer is a pure function without side effects, it no longer 'handles' a flow. The 'handling' is done in the effects and the reducer just transforms and sets data. At first glance most of the transformations can be done directly in the reducer and we don't need a helper function for them. We're testing the reducer. If you really prefer a helper function give it an appropriate name.

src/features/SendAssets/SendAssets.reducer.tsx
```diff
@@ -0,0 +1,153 @@
...
+
+sendAssetsReducer.actionTypes = {
+  FORM_SUBMIT: 'FORM_SUBMIT',
+  SIGN: 'SIGN',
+  WEB3_SIGN: 'WEB3_SIGN',
+  SEND: 'SEND',
+  AFTER_SEND: 'AFTER_SEND',
+  RESUBMIT: 'RESUBMIT',
+  RESET: 'RESET'
+};
```
vilsbole, line 82: Naming is hard. When it comes to actionTypes the name should reflect the action taken.
By convention I suggest we use `REQUEST, SUCCESS, FAILURE` to reflect the action that happened.
eg. https://github.com/MyCryptoHQ/MyCrypto/blob/285f33ed5be4c61c3c42fef91d60277853e62411/src/services/WalletService/walletconnect/reducer.ts
If if we don't use all three in this case, it will be easier to move some of the handling into the flow and expand in the future.

src/features/SendAssets/SendAssets.tsx
```diff
@@ -113,17 +121,45 @@ function SendAssets() {
...
+      !isWeb3Wallet(reducerState.txConfig!.senderAccount.wallet)
+    ) {
+      const { txConfig, signedTx } = reducerState;
+      const provider = new ProviderHandler(txConfig!.network);
+
+      provider
+        .sendRawTx(signedTx)
+        .then((payload) => dispatch({ type: sendAssetsReducer.actionTypes.AFTER_SEND, payload }));
+    }
+  }, [reducerState.send]);
```
vilsbole, line 161: 😍

src/features/SendAssets/SendAssets.reducer.tsx
```diff
@@ -0,0 +1,153 @@
+import { hexlify, bigNumberify, Arrayish } from 'ethers/utils';
```
vilsbole, line 1: Prefer `bigify` from utils.
It has better handling of digits.

src/features/SendAssets/__tests__/SendAssets.reducer.test.tsx
```diff
@@ -0,0 +1,211 @@
+import { bigNumberify } from 'ethers/utils';
```
vilsbole, line 1: Prefer `bigify` from `@utils`

src/features/SendAssets/__tests__/SendAssets.reducer.test.tsx
```diff
@@ -0,0 +1,211 @@
...
+jest.mock('ethers/utils', () => {
+  const { mockFactory } = require('../__mocks__/utils');
+  return mockFactory({
+    gasPrice: '57000000000',
+    gasLimit: '21000',
+    nonce: '10',
+    data: '0x',
+    receiverAddress: '0xfE5443FaC29fA621cFc33D41D1927fd0f5E0bB7c',
+    value: '10000000000000',
+    to: '0xfE5443FaC29fA621cFc33D41D1927fd0f5E0bB7c'
```
vilsbole, line 39: Is this suppose to be the same as `defaultTxConfig`? Can we make it explicit?

FrederikBolding: It's tricky with Jest since you can't use anything out the scope of the mock function.

src/features/SendAssets/SendAssets.reducer.tsx
FrederikBolding: I've tried renaming them slightly, let me know if it's clearer.

src/features/SendAssets/__tests__/SendAssets.reducer.test.tsx
vilsbole: oh!

PR #3419, opened 2020-08-07
Prefer useContracts

vilsbole:
Update for #3276.

- Remove `ContractsProvider` in favour of `useContracts`.
- Add extra tests.
- Remove unused methods.
- move `uuid` creation inside hook.


src/types/contract.ts
```diff
@@ -1,10 +1,12 @@
 import { NetworkId } from './networkId';
 import { TUuid } from './uuid';
+import { TAddress } from './address';
 
 export interface Contract {
   name: string;
+  label?: string; // Added in InteractWithContracts/stateFactory.. !?
```
FrederikBolding, line 7: Can we remove this in favor of `name` then?

vilsbole: I only found where it was set but not where it was called and wanted to avoid touching/breaking something that isn't tested yet.

FrederikBolding: Okay, let's make sure it's removed soon though. Can we add a TODO?

PR #3425, opened 2020-08-11
Fix AccountSelector hard crash

FrederikBolding:


vilsbole:
Please include steps to reproduce the bug. Initial attempt with a MM account and 0 balance does not trigger an error on Request Asset

FrederikBolding:
Can't reproduce. Might have been an old config issue.

PR #3432, opened 2020-08-13
Update link for PTX KB article

Mrtenz:
This updates the link here:


src/features/ProtectTransaction/components/ProtectTxProtection.tsx
```diff
@@ -24,6 +24,7 @@ import { ProtectTxError } from '..';
 import { ProtectTxMissingInfo } from './ProtectTxMissingInfo';
 
 import bulletIcon from 'assets/images/icn-bullet.svg';
+import translate from '@translations/translate';
```
FrederikBolding, line 27: Respect import order and possibly we don't need the entire import? Isn't it part of `@translations`?

Mrtenz: Fixed. I just used the auto import function of my IDE. If we want to enforce a specific import order, maybe we should setup ESLint to check that?

FrederikBolding: Yeah, we probably should. Weird that the IDE didn't combine them automatically though.

PR #3435, opened 2020-08-14
Fixed label sort bug

blurpesec:


vilsbole:
Can we include a test to avoid future regressions?

vilsbole (approved):

PR #3439, opened 2020-08-17
Fix Asset Selector being searchable on swap

FrederikBolding:


src/components/AssetSelector.tsx
```diff
@@ -98,6 +98,7 @@ function AssetSelector({
         disabled={disabled}
         searchable={searchable}
         onChange={(option: TAssetOption) => onSelect(option)}
+        getOptionLabel={(option) => (showOnlySymbol ? option.ticker : option.name)}
```
vilsbole, line 101: Can we add a test for this case plz?
https://testing-library.com/docs/example-input-event

PR #3443, opened 2020-08-19
Tx Status (smol changes)

blurpesec:
### Description
Luker noticed a bug with the tx status link in recent txs panel. This pr fixes that. We also may mock up a quick icon or something for this in the design channel.


FrederikBolding (approved):
lgtm - my bad! 😄

PR #3456, opened 2020-08-21
Replace formik render with child callback function

vilsbole:
`Formik render={}` is deprecated.
Replace with child callback function e.g
```
<Formik>{(props) => (<>{//...etc}</>)}</Formik>
```


FrederikBolding:
@vilsbole Rebase please!

FrederikBolding (approved):

PR #3471, opened 2020-08-25
TX History v1

FrederikBolding:


FrederikBolding:
[ch6733]

src/services/ApiService/History/types.ts
```diff
@@ -0,0 +1,27 @@
...
+
+  readonly txType: ITxType;
+}
+
+export interface ITxHistoryERC20Transfer {
+  readonly from: TAddress;
+  readonly to: TAddress;
+  readonly contractAddress: TAddress;
+  readonly amount: number;
+}
```
blurpesec, line 27: Some of these props are of type `number`. I'm surprised this doesn't toss any overflow/underflow issues with the BN-esque values the api returns. Is this a potential problem?

src/services/Store/StoreProvider.tsx
```diff
@@ -337,6 +339,20 @@ export const StoreProvider: React.FC = ({ children }) => {
...
     }, {}) || {};
 
+  // TX HISTORY
+  const [txHistory, setTxHistory] = useState<ITxHistoryApiResponse[]>([]);
+
+  useEffectOnce(() => {
+    HistoryService.instance
+      .getHistory(accounts.filter((a) => a.networkId === DEFAULT_NETWORK).map((a) => a.address))
+      .then((history) => {
+        console.debug(history);
```
blurpesec, line 349: Remove this?

src/services/TxHistory/helpers.tsx
```diff
@@ -0,0 +1,63 @@
...
+  };
+};
+
+export const merge = (apiTxs: ITxReceipt[], accountTxs: ITxReceipt[]): ITxReceipt[] => {
+  // Prioritize Account TX - needs to be more advanced?
+  const filteredApiTxs = apiTxs.filter((tx) => !accountTxs.find((a) => a.hash === tx.hash));
+  return filteredApiTxs.concat(accountTxs).sort((a, b) => (a.timestamp || 0) - (b.timestamp || 0));
+};
+
+export const deriveTxType = (accountsList: StoreAccount[], tx: ITxReceipt): ITxHistoryType => {
```
blurpesec, line 41: This isn't tested anywhere, right?

src/services/TxHistory/helpers.tsx
```diff
@@ -0,0 +1,63 @@
...
+    gasUsed: bigNumberify(tx.gasUsed || 0),
+    value: parseEther(tx.value.toString()),
+    nonce: tx.nonce.toString()
+  };
+};
+
+export const merge = (apiTxs: ITxReceipt[], accountTxs: ITxReceipt[]): ITxReceipt[] => {
+  // Prioritize Account TX - needs to be more advanced?
+  const filteredApiTxs = apiTxs.filter((tx) => !accountTxs.find((a) => a.hash === tx.hash));
+  return filteredApiTxs.concat(accountTxs).sort((a, b) => (a.timestamp || 0) - (b.timestamp || 0));
```
blurpesec, line 38: I might be mistaken, but I believe the base table component already sorts the txs when a sortable column is specified. Do we need them to be available pre-sorted in the context for some other reason?

src/services/Store/StoreProvider.tsx
FrederikBolding: Will do before merge 👍

src/services/ApiService/History/types.ts
FrederikBolding: We need to solve this in the API, reached out on Slack.

src/services/TxHistory/helpers.tsx
FrederikBolding: Added test for `deriveTxType`

src/services/TxHistory/helpers.tsx
FrederikBolding: No, you are right! 👍

src/services/ApiService/History/types.ts
```diff
@@ -0,0 +1,27 @@
...
+  readonly hash: ITxHash;
+  readonly nonce: string; // Hex
+  readonly recipientAddress: TAddress;
+  readonly status: ITxStatus.PENDING | ITxStatus.SUCCESS | ITxStatus.FAILED | ITxStatus.UNKNOWN;
+  readonly timestamp?: number;
+  readonly to: TAddress;
+  readonly value: string; // Hex
+
+  readonly txType: ITxType;
+}
```
blurpesec, line 20: In my last pr i added these which i think would be good to stick here: https://github.com/MyCryptoHQ/MyCrypto/blob/master/src/types/transactionFlow.ts#L17-L23

blurpesec (approved):

PR #3473, opened 2020-08-25
Fix issue with adding Coinbase Wallet account

FrederikBolding:


src/services/WalletService/web3/web3.ts
```diff
@@ -20,6 +20,8 @@ export const unlockWeb3 = (onSuccess: (data: any) => void) => async (networks: N
     if (accounts) {
       onSuccess(network);
       return accounts.map((address) => new Web3Wallet(address, network.id));
+    } else {
+      throw new Error('Failed to get web3 accounts via wallet_getPermissions');
```
vilsbole, line 24: Add comment explaining case for Coinbase Wallet

vilsbole (changes requested):

PR #3475, opened 2020-08-26
Remove formik render from fields

vilsbole:
`render` is also deprecated for `Field`.
Replace with child callback.

### DIFF
--- a/src/features/DeFiZap/components/ZapForm.tsx
+++ b/src/features/DeFiZap/components/ZapForm.tsx
@@ -175,7 +175,6 @@ export const ZapFormUI = ({
                         value={field.value}
                         onBlur={() => {
                           form.setFieldTouched('amount');
-                          //handleGasEstimate();
                         }}
                         placeholder={'0.00'}
                       />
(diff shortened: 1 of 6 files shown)


src/features/DeFiZap/components/ZapForm.tsx
```diff
@@ -166,30 +166,27 @@ export const ZapFormUI = ({
...
+                <Field name="amount">
+                  {({ field, form }: FieldProps) => (
+                    <>
+                      <AmountInput
+                        {...field}
+                        asset={ethAsset}
+                        value={field.value}
+                        onBlur={() => {
+                          form.setFieldTouched('amount');
+                          //handleGasEstimate();
```
FrederikBolding, line 178: Can we remove this?

FrederikBolding (changes requested):

src/features/DeFiZap/components/ZapForm.tsx
vilsbole: Done

FrederikBolding (approved):

PR #3477, opened 2020-08-27
Reduce node calls from TX Status

FrederikBolding:
-  Fix issue with double fetching of txs from nodes


src/features/TxStatus/helpers.ts
```diff
@@ -1,43 +1,68 @@
...
   }
-  const fetchedTxConfig = makeTxConfigFromTxResponse(fetchedTx, assets, network, accounts);
-  return {
-    config: fetchedTxConfig,
-    receipt: makeUnknownTxReceipt(txHash as ITxHash)(ITxType.UNKNOWN, fetchedTxConfig)
-  };
+  return { fetchedTx, cachedTx: undefined };
+};
+
+export const makeTx = ({
```
vilsbole, line 38: Include test

vilsbole (changes requested):
Include a test plz

vilsbole (approved):

PR #3493, opened 2020-09-07
Bump ethereumjs-wallet

FrederikBolding:
Bump `ethereumjs-wallet` to get rid of `scrypt` dependency. 

Needs more testing


FrederikBolding:
[ch6798]

src/types/vendor/ethereumjs-wallet.d.ts
```diff
@@ -1,26 +1,15 @@
 /**
```
vilsbole, line 1: Can remove?

src/types/vendor/ethereumjs-wallet.d.ts
```diff
@@ -1,26 +1,15 @@
...
  * Incomplete -- Provider-engine needs its own typings
  */
-declare module 'ethereumjs-wallet/provider-engine' {
-  import { IPublicKeyOnlyWallet, IFullWallet } from 'ethereumjs-wallet';
-  class WalletSubprovider {
-    constructor(wallet: IPublicKeyOnlyWallet | IFullWallet, opts?: object);
-  }
-}
 
 declare module 'ethereumjs-wallet/thirdparty' {
```
vilsbole, line 5: Is the new version copied from `ethereumjs-wallet`?
Do you know if 0.6.5 includes the type changes, if so we can try to remove the declaration file.

vilsbole (changes requested):

src/types/vendor/ethereumjs-wallet.d.ts
FrederikBolding: We don't import `ethereumjs-wallet/provider-engine`, so I removed the type.

src/types/vendor/ethereumjs-wallet.d.ts
FrederikBolding: It is not copied from `ethereumjs-wallet`, and `0.6.5` does not have type information. I've just modified what we had to fit the new version.

src/types/vendor/ethereumjs-wallet.d.ts
vilsbole: I meant to ask if can remove the comment.
Apologies for the ambiguity.

vilsbole (approved):

PR #3496, opened 2020-09-08
Preserve document.domain for electron

vilsbole:
Building for electron renders a white page with the following error:
```
Uncaught DOMException: Failed to set the 'domain' property on 'Document': '' is an empty domain.
    at Object.<anonymous> (file:///Applications/MyCrypto.app/Contents/Resources/app.asar/main.7896c1edf3a5be136203.js:1:1587510)
```

Since electron builds have no `location.hostname` we remove the setting when building `electron`.

### DIFF
--- a/src/index.tsx
+++ b/src/index.tsx
@@ -9,7 +9,7 @@ import 'what-input'; // @todo: Investigate utility of dependency; Used in sass/s
 import React from 'react';
 import { render } from 'react-dom';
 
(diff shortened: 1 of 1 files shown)


vilsbole:
App builds correctly but:
1. `badBrowserScript` issue in `dev`.
2. `frame-src` issue in `production`.

Since electron is currently not supported closing this issue.

PR #3500, opened 2020-09-09
Add Multi TXs to TX History

FrederikBolding:
Steps to test:
- Go through a Multi TX flow (membership, rep migration etc)
- Refrain from refreshing until the test is done (to make sure you don't fetch the tx's from etherscan)
- Go to the dashboard
- Check whether the transactions you just made are in the tx history


FrederikBolding:
[ch6245]

src/utils/useTxMulti/useTxMulti.spec.tsx
```diff
@@ -0,0 +1,218 @@
...
+import { waitFor } from 'test-utils';
+
+import { fAccount, fAccounts, fAssets, fNetwork, fNetworks } from '@fixtures';
+import { ITxObject, ITxToAddress, ITxValue, ITxData, ITxHash, ITxType, ITxStatus } from '@types';
+import { AccountContext, DataContext, IDataContext, StoreContext } from '@services';
+
+import { useTxMulti } from './useTxMulti';
+
+const createTxRaw = (idx: number): Partial<ITxObject> => ({
+  to: ('address' + idx) as ITxToAddress,
```
vilsbole, line 13: Why `ITxToAddress` instead of `TAddress`?

src/utils/useTxMulti/useTxMulti.ts
```diff
@@ -30,6 +34,28 @@ export type TUseTxMulti = () => {
...
+
+  useEffect(() => {
+    if (
+      account &&
+      currentTx &&
+      currentTx.txResponse &&
+      currentTx.txHash &&
+      currentTx.status === ITxStatus.BROADCASTED
+    ) {
+      const network = getNetworkByChainId(currentTx.txRaw.chainId)!;
```
vilsbole, line 53: Isn't `network` already available in `state`?

src/utils/useTxMulti/useTxMulti.ts
```diff
@@ -30,6 +34,28 @@ export type TUseTxMulti = () => {
...
+      const txConfig = makeTxConfigFromTxResponse(currentTx.txResponse, assets, network, accounts);
+      const pendingTxReceipt = makePendingTxReceipt(currentTx.txHash)(ITxType.UNKNOWN, txConfig);
+      addNewTxToAccount(account, pendingTxReceipt);
+    }
+  }, [currentTx]);
```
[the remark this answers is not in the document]
vilsbole: It seems that you rely on `assets, accounts`. In which case they should be added to the dependencies.

src/utils/useTxMulti/useTxMulti.ts
```diff
@@ -30,6 +34,28 @@ export type TUseTxMulti = () => {
 export const useTxMulti: TUseTxMulti = () => {
   const [state, dispatch] = useReducer(TxMultiReducer, initialState, identity, 'TxMulti');
   const getState = () => state;
+  const { accounts } = useContext(StoreContext);
+  const { addNewTxToAccount } = useContext(AccountContext);
```
vilsbole, line 38: Naming: `addNew` is an oxymore. Prefer `addTxToAccount`.

src/utils/useTxMulti/useTxMulti.ts
```diff
@@ -30,6 +34,28 @@ export type TUseTxMulti = () => {
...
+      account &&
+      currentTx &&
+      currentTx.txResponse &&
+      currentTx.txHash &&
+      currentTx.status === ITxStatus.BROADCASTED
+    ) {
+      const network = getNetworkByChainId(currentTx.txRaw.chainId)!;
+      const txConfig = makeTxConfigFromTxResponse(currentTx.txResponse, assets, network, accounts);
+      const pendingTxReceipt = makePendingTxReceipt(currentTx.txHash)(ITxType.UNKNOWN, txConfig);
+      addNewTxToAccount(account, pendingTxReceipt);
```
vilsbole, line 56: It's really odd that we need so many manipulations between formats.
Can me converge the them?

vilsbole (commented):
I suspect this adds multiple txReceipts.

src/utils/useTxMulti/useTxMulti.spec.tsx
FrederikBolding: That's the type on `ITxObject`

src/utils/useTxMulti/useTxMulti.ts
FrederikBolding: I hadn't spotted that, but yeah!

src/utils/useTxMulti/useTxMulti.ts
FrederikBolding: Let's rename when we refactor `AccountProvider`

FrederikBolding: Is part of 3508

src/utils/useTxMulti/useTxMulti.ts
FrederikBolding: I think this is a general thing that we have to look into. But very much out of the scope of this PR.

src/utils/useTxMulti/useTxMulti.ts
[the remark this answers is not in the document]
FrederikBolding: Perhaps it doesn't need dependencies.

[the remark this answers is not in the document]
FrederikBolding: Actually it changes the amount of times run with/without the current dependency.

src/utils/useTxMulti/useTxMulti.spec.tsx
```diff
@@ -0,0 +1,223 @@
+import React from 'react';
+import { renderHook, act } from '@testing-library/react-hooks';
+import isEmpty from 'ramda/src/isEmpty';
```
vilsbole, line 3: from `@vendor`

vilsbole (approved):
Code lgtm;
Merge waiting for additional testing.

PR #3510, opened 2020-09-11
Disable Electron builds

FrederikBolding:


FrederikBolding:
[ch6842]

FrederikBolding:
Not sure if we just want to throw out the Electron build config for now?

blurpesec:
> Not sure if we just want to throw out the Electron build config for now?

I'd say go ahead and delete it. We should be able to re-add from git history it if we need it later.

vilsbole (approved):

PR #3515, opened 2020-09-14
Cancel TX Bug fix + Remove overwritten TX from History

FrederikBolding:


FrederikBolding:
[ch6181]

src/services/Store/StoreProvider.tsx
```diff
@@ -283,6 +286,30 @@ export const StoreProvider: React.FC = ({ children }) => {
...
+            t.nonce === pendingTxReceipt.nonce &&
+            t.asset.networkId === pendingTxReceipt.asset.networkId &&
+            t.hash !== pendingTxReceipt.hash &&
+            t.status === ITxStatus.SUCCESS
+        );
+
+        if (overwrittenTx) {
+          removeTxFromAccount(senderAccount, pendingTxReceipt);
+          return;
+        }
```
blurpesec, line 311: I think this whole bit can be handled in AccountProvider. The AccountProvider already has handling to replace a pending tx with a successful/failed one, so you can probable handle this there as well.

src/utils/queries.ts
```diff
@@ -34,7 +34,7 @@ export const constructCancelTxQuery = (txConfig: ITxConfig, newGasPrice: number)
   return queryString.stringify({
     ...unfinishedSpeedUpTxQueryParams,
     to: unfinishedSpeedUpTxQueryParams.from,
-    data: '0x0',
+    data: '0x',
```
blurpesec, line 37: Did this break somewhere?

src/utils/queries.ts
```diff
@@ -34,7 +34,7 @@ export const constructCancelTxQuery = (txConfig: ITxConfig, newGasPrice: number)
   return queryString.stringify({
     ...unfinishedSpeedUpTxQueryParams,
     to: unfinishedSpeedUpTxQueryParams.from,
```
blurpesec, line 36: Just noticed these, but can you rename these to a `cancelTxQueryParams` type var instead?

blurpesec (changes requested):

src/utils/queries.ts
FrederikBolding: I was playing around with it and couldn't cancel transactions without this change. Because the gas calculation would be off.

FrederikBolding: I reached out to you on Slack to verify that it was an issue.

src/utils/queries.ts
FrederikBolding: Sure

src/services/Store/StoreProvider.tsx
FrederikBolding: I'm not certain this should be the responsibility of the account hook. It will overcomplicate the logic in there. Reached out over Slack.

blurpesec (approved):

PR #3519, opened 2020-09-15
 Add ESLint rules to enforce consistent imports

Mrtenz:
This adds some `eslint-plugin-import` rules to enforce consistent imports, and to prevent duplicate imports/exports etc..


Mrtenz:
[ch6858]

GuillaumeRx:
Nice work ! I discussed with Emmanuel and we would just like `react` to be on top of everything. You can set it with
```
pathGroups: [
          {
            pattern: 'react',
            group: 'external',
            position: 'before',
          },
        ],
```
ref: https://dev.to/mjmaix/clean-eslint-import-management-for-reactjs-230h

Other than that it should be good to go :)

GuillaumeRx (approved):
LGTM ! ✨

PR #3520, opened 2020-09-15
Use date-fns for handling dates and times

Mrtenz:
## Description

`moment` is [being deprecated](https://momentjs.com/docs/#/-project-status/), so this replaces `moment` with `date-fns` everywhere. Currently this assumes a standard date format of `YYYY-MM-DD`, but this can be easily changed.

## Changes

* Created a new utility file with some functions for formatting and parsing dates.
* Replaced any usage of `moment` with the new utility functions.
* Added tests for the utility file.


src/components/TimeElapsed.tsx
```diff
@@ -1,18 +1,17 @@
 import React, { useState } from 'react';
 
-import { formatDistance } from 'date-fns';
+import { formatTimeDifference, useInterval } from '@utils';
 
-import { useInterval } from '@utils';
+interface Props {
+  value: number;
+}
```
GuillaumeRx, line 7: Since we don't export this and it's only one parameter you could directly type in the function parameters instead of defining an interface

src/database/versions.ts
```diff
@@ -26,16 +26,14 @@ export const dbVersions = {
   }
 };
 
-export const dbHistory = ['v1.0.0', 'v0.0.1'];
+export const dbHistory = ['v1.0.0', 'v0.0.1'] as const;
```
GuillaumeRx, line 29: why do you cast here ?

src/database/versions.ts
```diff
@@ -26,16 +26,14 @@ export const dbVersions = {
...
-// @ts-ignore
 export const getCurrentDBConfig = () => dbVersions[dbHistory[0]];
 
-// @ts-ignore
 export const getPreviousDBConfig = () => dbVersions[dbHistory[1]];
 
-export const getExportFileName = (currentDb: DBConfig, m: moment.Moment) => {
-  return `${SETTINGS_FILENAME}_${m.format('YYYY-MM-DD')}_${currentDb.version}.json`;
+export const getExportFileName = (currentDb: DBConfig, date: Date) => {
+  return `${SETTINGS_FILENAME}_${format(date, 'yyyy-MM-dd')}_${currentDb.version}.json`;
```
GuillaumeRx, line 36: Can you use only uppercase or lowercase ?

GuillaumeRx (approved):

GuillaumeRx (changes requested):
Miss clicked, meant to request changes.

src/components/TimeElapsed.tsx
Mrtenz: It looks a bit cleaner in my opinion to have it as separate interface, but I can revert.

src/database/versions.ts
Mrtenz: This is not a cast but a [const assertion](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-4.html#const-assertions). Basically it changes the type from `string[]` to `['v1.0.0', 'v0.0.1']`. This removes the need for `// @ts-ignore`, not really related to `date-fns` but I figured I'd just fix it since I was working on that file.

src/database/versions.ts
Mrtenz: `date-fns` uses `yyyy` for the full calendar year, `YYYY` is for the local week-numbering year. Not sure what is the difference is tbh. `dd` is for the day of month, `DD` for the day of year.

Mrtenz: > A common mistake is to use YYYY. yyyy specifies the calendar year whereas YYYY specifies the year (of “Week of Year”), used in the ISO year-week calendar. In most cases, yyyy and YYYY yield the same number, however they may be different. Typically you should use the calendar year.

So I'm assuming `yyyy` is correct.

GuillaumeRx: Thanks for the clarification ! 💯

src/database/versions.ts
GuillaumeRx: 😍👌

GuillaumeRx (approved):
LGTM 💯

PR #3539, opened 2020-09-22
Enable scanning of Ledger Live DPaths

FrederikBolding:
Re-enable scanning of Ledger Live DPaths


FrederikBolding:
@blurpesec Can you verify that this is good to enable? Why was it disabled initially?

blurpesec (approved):
I think this was probably just an oversight

PR #3542, opened 2020-09-24
Activate eslint ban-types

vilsbole:
[ch6856]

### DIFF
--- a/src/components/LegacyModal/Modal.tsx
+++ b/src/components/LegacyModal/Modal.tsx
@@ -26,7 +26,7 @@ const Fade = ({ ...props }: any) => (
   <CSSTransition {...props} timeout={300} classNames="animate-modal" />
 );
 
-export class Modal extends PureComponent<Props, {}> {
+export class Modal extends PureComponent<Props, TObject> {
   public modalParent: HTMLElement;
   public modalBody: ModalBody;
 
--- a/src/components/LegacyModal/Modal.tsx
+++ b/src/components/LegacyModal/Modal.tsx
@@ -26,7 +26,7 @@ const Fade = ({ ...props }: any) => (
   <CSSTransition {...props} timeout={300} classNames="animate-modal" />
 );
 
-export class Modal extends PureComponent<Props, TObject> {
+export class Modal extends PureComponent<Props, never> {
   public modalParent: HTMLElement;
   public modalBody: ModalBody;
 
--- a/src/components/PaperWallet.tsx
+++ b/src/components/PaperWallet.tsx
@@ -197,7 +197,7 @@ interface Props {
   isHidden?: boolean;
 }
 
-export default class PaperWallet extends React.Component<Props, {}> {
+export default class PaperWallet extends React.Component<Props, TObject> {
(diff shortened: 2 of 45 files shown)


src/components/LegacyModal/Modal.tsx
```diff
@@ -26,7 +26,7 @@ const Fade = ({ ...props }: any) => (
   <CSSTransition {...props} timeout={300} classNames="animate-modal" />
 );
 
-export class Modal extends PureComponent<Props, {}> {
+export class Modal extends PureComponent<Props, TObject> {
```
Mrtenz, line 29: If the component doesn't have state, I'd just use `null` here.

```suggestion
export class Modal extends PureComponent<Props, null> {
```

src/components/PaperWallet.tsx
```diff
@@ -197,7 +197,7 @@ interface Props {
   isHidden?: boolean;
 }
 
-export default class PaperWallet extends React.Component<Props, {}> {
+export default class PaperWallet extends React.Component<Props, TObject> {
```
Mrtenz, line 200: If the component doesn't have state, I'd just use `null` here.

```suggestion
export default class PaperWallet extends React.Component<Props, null> {
```

src/database/helpers.ts
```diff
@@ -2,11 +2,11 @@ import { TUuid } from '@types';
 
 import { FlowTransducer, GenObject } from './types';
 
-export const withUuid = <T extends {}>(fn: () => TUuid) => (x: T): T => ({ ...x, uuid: fn() });
+export const withUuid = <T extends TObject>(fn: () => TUuid) => (x: T): T => ({ ...x, uuid: fn() });
 
-export const toArray = (object: any): any[] => Object.values(object);
+export const toArray = (obj: any): any[] => Object.values(obj);
```
Mrtenz, line 7: Can we remove `any` here?

Mrtenz (changes requested):

src/components/LegacyModal/Modal.tsx
vilsbole: Done

src/components/PaperWallet.tsx
vilsbole: Done

src/database/helpers.ts
vilsbole: Changed to `unknown`

vilsbole: I wasn't able to find a type fix for this so reverted to renaming.

Mrtenz: How about

```ts
export const toArray = <T>(obj: T): T[] => Object.values(obj);
```

vilsbole: This raises 18 errors.
And the return value is an array of `values` rather than an array of `obj`s.
I tried similar declarations and each raise their own collection of issues:
- `<T>(obj: T): ValuesType<T>[] => Object.values(obj)` 
- `<T>(obj: <Record<string, T>>): T[]`

Mrtenz (approved):

PR #3543, opened 2020-09-24
Add UNI token claim detection

FrederikBolding:


FrederikBolding:
[ch6974]

blurpesec (commented):
There doesn't seem to be anything telling you which accounts your UNI tokens are claimable from? Or am i just missing that?

src/features/Dashboard/helpers.ts
```diff
@@ -1,10 +1,10 @@
-import { Asset } from '@types';
+import { State as StoreContextState } from '@services/Store/StoreProvider';
 
 import { Action } from './types';
 
-export const filterDashboardActions = (actions: Action[], assets: Asset[]) =>
+export const filterDashboardActions = (actions: Action[], state: StoreContextState) =>
```
blurpesec, line 5: Test

src/features/Dashboard/constants.ts
```diff
@@ -3,11 +3,14 @@ import receiveIcon from '@assets/images/icn-receive.svg';
 import sendIcon from '@assets/images/icn-send.svg';
 import swapIcon from '@assets/images/icn-swap.svg';
 import repIcon from '@assets/images/rep-logo.svg';
+import uniIcon from '@assets/images/uni-logo.png';
 import ledgerIcon from '@assets/images/wallets/ledger.svg';
 import trezorIcon from '@assets/images/wallets/trezor.svg';
 import { EXT_URLS, REPV1UUID, ROUTE_PATHS } from '@config';
+import { UNISWAP_LINK } from '@config/links';
```
blurpesec, line 10: import from `@config`

FrederikBolding:
> There doesn't seem to be anything telling you which accounts your UNI tokens are claimable from? Or am i just missing that?

It does not atm no. I am also not sure where we would show that.

blurpesec:
> It does not atm no. I am also not sure where we would show that.

How about a table like the one we use for ENS?

FrederikBolding:
> > It does not atm no. I am also not sure where we would show that.
> 
> How about a table like the one we use for ENS?

Afaik we wanted to have it as a card on the dashboard instead.

blurpesec (approved):

blurpesec:
I'm not able to test these because I don't have any UNI claims, but i'm assuming it's working ;)

PR #3544, opened 2020-09-25
Test NetworkSelector

vilsbole:
[ch6997]


FrederikBolding (changes requested):
There is a difference when adding a Ledger vs a Trezor
image

image

vilsbole:
The PR has the same behaviour as `app.mycrypto.com`.
This is related to the introduction of `LEDGER_NANO_S_NEW` and `TREZOR_NEW` wallet types.
I suggest we fix in a separate PR.

FrederikBolding (approved):

PR #3547, opened 2020-09-29
Fixes an issue with networks unindexable by name

blurpesec:
### Description
ethers.js allows you to add networks using network id for some networks, but not for networks that are unsupported. Noticed this issue first while looking into #3532.

### DIFF
--- a/src/services/EthService/network/helpers.ts
+++ b/src/services/EthService/network/helpers.ts
@@ -7,51 +7,61 @@ import { ETHERSCAN_API_KEY, INFURA_API_KEY } from '@config';
 import { DPathFormat, Network, NetworkId, NodeOptions, NodeType } from '@types';
 
 // Network names accepted by ethers.EtherscanProvider
-type TValidEtherscanNetwork = 'homestead' | 'ropsten' | 'rinkeby' | 'kovan' | 'goerli';
+type TValidEthersNetworkish = 'homestead' | 'ropsten' | 'rinkeby' | 'kovan' | 'goerli' | number;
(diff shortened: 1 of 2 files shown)


src/services/EthService/network/helpers.ts
```diff
@@ -7,51 +7,61 @@ import { ETHERSCAN_API_KEY, INFURA_API_KEY } from '@config';
 import { DPathFormat, Network, NetworkId, NodeOptions, NodeType } from '@types';
 
 // Network names accepted by ethers.EtherscanProvider
-type TValidEtherscanNetwork = 'homestead' | 'ropsten' | 'rinkeby' | 'kovan' | 'goerli';
+type TValidEthersNetworkish = 'homestead' | 'ropsten' | 'rinkeby' | 'kovan' | 'goerli' | number;
```
vilsbole, line 10: What does `networkish` refer to?!

vilsbole (changes requested):
Include tests for modified code.

src/services/EthService/network/helpers.ts
blurpesec: > What does `networkish` refer to?!

When setting a network using ethers.js they have string config specified for the above networks. But in the instance where it doesn't have a config for the network name, you need to use a chainid. This input type is called [Networkish](https://docs.ethers.io/v5/api/providers/types/#providers-Networkish) for some reason.

vilsbole (commented):
Code lgtm

vilsbole (approved):

PR #3553, opened 2020-10-02
Abstract steps of multi tx flows

blurpesec:
[ch7058]

### Description
Abstracts steps of multi tx flows by standardizing similar types shared between steps.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "MyCrypto",
   "author": "MyCryptoHQ",
-  "version": "2.2.2",
+  "version": "2.2.5",
   "main": "main.js",
   "description": "MyCrypto is a free, open-source interface for interacting with the blockchain",
   "repository": "https://github.com/MyCryptoHQ/MyCrypto",
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "MyCrypto",
   "author": "MyCryptoHQ",
-  "version": "1.19.1",
+  "version": "2.2.2",
   "main": "main.js",
   "description": "MyCrypto is a free, open-source interface for interacting with the blockchain",
   "repository": "https://github.com/MyCryptoHQ/MyCrypto",
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "MyCrypto",
   "author": "MyCryptoHQ",
-  "version": "2.2.2",
+  "version": "1.19.1",
   "main": "main.js",
   "description": "MyCrypto is a free, open-source interface for interacting with the blockchain",
   "repository": "https://github.com/MyCryptoHQ/MyCrypto",
--- a/src/components/TokenMigration/TokenMigrationStepper.tsx
+++ b/src/components/TokenMigration/TokenMigrationStepper.tsx
@@ -1,16 +1,14 @@
 import React, { useEffect, useReducer } from 'react';
 
-import { WALLET_STEPS } from '@components';
+import { createSignConfirmAndBroadcastTxSteps } from '@components';
 import { default as GeneralStepper, IStepperPath } from '@components/GeneralStepper';
 import { ROUTE_PATHS } from '@config';
 import { translateRaw } from '@translations';
 import {
-  ISimpleTxFormFull,
   ITokenMigrationConfig,
   ITokenMigrationFormFull,
   ITxStatus,
-  TokenMigrationState,
-  TxParcel
+  TokenMigrationState
 } from '@types';
 import { useTxMulti } from '@utils';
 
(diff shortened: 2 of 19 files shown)


src/components/TokenMigration/TokenMigrationStepper.tsx
```diff
@@ -54,48 +52,20 @@ const TokenMigrationStepper = ({ tokenMigrationConfig }: Props) => {
...
-      label: tokenMigrationConfig.receiptTitle,
-      component: TokenMigrationReceipt,
-      props: {
-        amount: reducerState.amount,
-        tokenMigrationConfig,
-        account,
-        transactions
-      }
-    }
+    ...createSignConfirmAndBroadcastTxSteps({
```
FrederikBolding, line 55: Isn't it more of `createSignConfirmAndReceiptSteps` or something like that?

src/components/TransactionFlow/helpers.spec.ts
```diff
@@ -13,3 +21,43 @@ describe('calculateReplacementGasPrice', () => {
...
+      flowConfig: repTokenMigrationConfig,
+      receiptTitle: translateRaw('REP_TOKEN_MIGRATION_RECEIPT'),
+      isSubmitting: false,
+      multiTxTitle: translateRaw('CONFIRM_TRANSACTION'),
+      multiTxComponent: TokenMigrationMultiTx,
+      receiptComponent: TokenMigrationReceipt,
+      prepareTx: noOp,
+      sendTx: () => new Promise(() => null)
+    });
+    expect(steps).toHaveLength(5);
```
FrederikBolding, line 61: Can we test the content a little bit too?

src/components/TransactionFlow/helpers.ts
```diff
@@ -13,6 +23,82 @@ type FieldValue = ValuesType<ISender>;
...
+  amount: string;
+  account: StoreAccount;
+
+  flowConfig: ITokenMigrationConfig | IMembershipConfig | IAssetPair;
+  transactions: TxParcel[];
+  isSubmitting: boolean;
+
+  multiTxTitle: string;
+  receiptTitle: string;
+  multiTxComponent(props: any): JSX.Element;
```
FrederikBolding, line 37: Can this be typed better?

src/components/TransactionFlow/helpers.ts
```diff
@@ -13,6 +23,82 @@ type FieldValue = ValuesType<ISender>;
...
+  receiptComponent(props: any): JSX.Element;
```
FrederikBolding, line 38: Same as above

src/components/TransactionFlow/helpers.ts
```diff
@@ -13,6 +23,82 @@ type FieldValue = ValuesType<ISender>;
 
 const preferValueFromSender = (l: FieldValue, r: FieldValue): FieldValue => (isEmpty(r) ? l : r);
 
+interface Props {
+  backStepTitle: string;
+  amount: string;
+  account: StoreAccount;
+
+  flowConfig: ITokenMigrationConfig | IMembershipConfig | IAssetPair;
```
FrederikBolding, line 31: Can we use `ITxMultiConfirmProps` as part of this prop construction so it isn't defined in two places?

src/components/TokenMigration/TokenMigrationStepper.tsx
```diff
@@ -54,48 +52,20 @@ const TokenMigrationStepper = ({ tokenMigrationConfig }: Props) => {
...
+      transactions,
+      backStepTitle: tokenMigrationConfig.formTitle,
+      amount: reducerState.amount!,
+      account: reducerState.account!,
+      flowConfig: tokenMigrationConfig,
+      receiptTitle: tokenMigrationConfig.receiptTitle,
+      multiTxTitle: translateRaw('CONFIRM_TRANSACTION'),
+      isSubmitting,
+      receiptComponent: TokenMigrationReceipt,
+      multiTxComponent: TokenMigrationMultiTx,
```
FrederikBolding, line 65: Can this never be a regular TX? Like membership is only 1?

FrederikBolding (changes requested):

src/components/TokenMigration/TokenMigrationStepper.tsx
blurpesec: yeah it can be, but it hasn't been in master yet. so we can address it when it pops up.

package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "MyCrypto",
   "author": "MyCryptoHQ",
-  "version": "1.19.1",
+  "version": "2.2.2",
```
FrederikBolding: Aren't we on `2.2.4`?

FrederikBolding (approved):

src/components/TokenMigration/TokenMigrationStepper.tsx
FrederikBolding: This is not the case and is a bug right now. There is only one tx when the approval is already done. I just discovered this when testing AAVE migration.

PR #3565, opened 2020-10-19
Fix swap display + REP migration styling issue

vilsbole:
### Fix design errors on Swap Form. 
- tooltip position
- font-size
- asset selector height
  - asset text font-weight

Screenshot 2020-10-19 at 17 44 41

Screenshot 2020-10-19 at 17 45 53

Screenshot 2020-10-19 at 17 46 02

### Fix input width in Migrate REP

Allow UITooltip child to be full width for the AmountInput in Migrate REP to fill the Panel.

Screenshot 2020-10-19 at 19 00 21


src/components/AssetSelector.tsx
```diff
@@ -32,32 +25,26 @@ export function AssetSelectorItem({
...
+      display="flex"
+      alignItems="center"
+      flexDirection="row"
+      padding="11px 10px"
       {...(onClick ? { onPointerDown: onClick } : null)}
       data-testid={`asset-selector-option-${ticker}`}
     >
       {showAssetIcon && <AssetIcon uuid={uuid} size={'1.5rem'} />}
       <Typography
-        bold={true}
```
FrederikBolding, line 41: Why the bold change?

FrederikBolding: I assume we have it bold to match the design of other Selectors. Fx. see Send Assets, where account names are bold

vilsbole: It seems that the bold is an implementation error so I corrected it.
If we look at the design specs the account names should also be 'normal'
https://www.figma.com/file/vsmfxvOp3JFKTEXdsSIFWI/MyCrypto-Web-App-Working-File?node-id=1257%3A89144

FrederikBolding (approved):

PR #3574, opened 2020-10-23
Fix external links

FrederikBolding:
🎉 🎉 🎉

## Description

Fix `openLink` not properly applying `noreferrer` to external links.

https://html.spec.whatwg.org/multipage/window-object.html#apis-for-creating-and-navigating-browsing-contexts-by-name

## Testing
Check that an external link doesn't have `window.opener` and `document.referrer` when opened.


Mrtenz (approved):
Tested it, seems to work fine :+1:

PR #3576, opened 2020-10-26
Delete Electron App

FrederikBolding:
Clean out all Electron and Desktop only code

Private keys, mnemonics and keystores.


src/components/SignTransactionWallets/index.ts
```diff
@@ -19,9 +16,9 @@ export const WALLET_STEPS: SigningComponents = {
   [WalletId.LEDGER_NANO_S_NEW]: SignTransactionLedger,
   [WalletId.TREZOR]: SignTransactionTrezor,
   [WalletId.TREZOR_NEW]: SignTransactionTrezor,
-  [WalletId.KEYSTORE_FILE]: SignTransactionKeystore,
-  [WalletId.MNEMONIC_PHRASE]: SignTransactionMnemonic,
-  [WalletId.MNEMONIC_PHRASE_NEW]: SignTransactionMnemonic,
+  [WalletId.KEYSTORE_FILE]: null,
+  [WalletId.MNEMONIC_PHRASE]: null,
+  [WalletId.MNEMONIC_PHRASE_NEW]: null,
```
vilsbole, line 21: Any reason to preserve these wallet ids ?

src/workers/fromV3.ts
```diff
@@ -1,23 +0,0 @@
-import { toBuffer } from 'ethereumjs-util';
```
vilsbole, line 1: 😍

src/utils/decrypt.spec.ts
```diff
@@ -1,174 +0,0 @@
-import {
```
vilsbole, line 1: Would these tests be useful for signer?

vilsbole (commented):
Flame out, you glorious skies,
Welcome our brave;
Kiss their exultant eyes;
Give what they gave.

----
This is wonderful! 

Can you check if we can also remove (now or later):
- icons: eg. icn-reload in `GeneralPhrasePanel.tsx`
- packages: eg. `@types/node-hid` or `ethereumjs-wallet`?
- translation keys
- global scss, and css definitions used in removed components?

src/utils/decrypt.spec.ts
FrederikBolding: Most of it is custom stuff that we don't need anymore as we will use more `ethersjs`.

src/components/SignTransactionWallets/index.ts
FrederikBolding: I wasn't sure what the exact ramifications of that would be.

vilsbole (approved):

vilsbole (approved):

codebling:
Is any additional info available regarding this change? 

Is desktop app no longer supported going forward? Is there an offline mode still supported?

FrederikBolding:
@codebling We are working on a replacement for the desktop app for the new version of MyCrypto. For now you can keep using the existing desktop app though for offline signing!

Let me know if you have further questions!

codebling:
@FrederikBolding thanks for the info! 

I am a (potential) new user and the desktop app appears to be the only way to create a wallet, but the current desktop app doesn't work for me due to the dependency on gconf (no longer available on Arch). For now I'll create a wallet elsewhere or spin up a virtual machine. I look forward to the new version!

PR #3583, opened 2020-10-27
Trigger onChange when submitting in lookup field using Enter key

FrederikBolding:
Fixes an issue with the contract interaction page where you couldn't enter an ABI if you input the contract address and pressed Enter to submit.


FrederikBolding:
[ch7196]

blurpesec (approved):

PR #3584, opened 2020-10-27
Configure redux & tools

vilsbole:
Configuration and DevTools demo store for:
- redux with redux-devtools
- redux-sagas middleware

Misc:
- Simplify and restyle DevTools


src/store/store.ts
```diff
@@ -0,0 +1,25 @@
...
+
+import { IS_DEV } from '@utils';
+
+import rootReducer from './reducer';
+import rootSaga from './sagas';
+
+const sagaMiddleware = createSagaMiddleware();
+const store = configureStore({
+  reducer: rootReducer,
+  middleware: [sagaMiddleware]
```
Mrtenz, line 12: Consider using the middleware built-in to Redux Toolkit, as it can help prevent some bugs.

```suggestion
  middleware: (getDefaultMiddleware) => getDefaultMiddleware().concat(sagaMiddleware)
```

src/features/DevTools/slice.ts
```diff
@@ -0,0 +1,40 @@
...
+export const getCount = path([sliceName, 'count']);
+
+/* Actions */
+export const increment = createAction<number | undefined>(`${sliceName}/increment`);
+export const reset = createAction<undefined>(`${sliceName}/reset`);
+
+/* Sagas */
+/** Saga - worker */
+function* onIncrement() {
+  yield call(console.log, 'Hello from Saga!');
```
Mrtenz, line 23: Just an fyi, there's no benefit in calling `console.log` using the `call` effect over simply writing `console.log('Hello from Saga!');`

src/features/DevTools/slice.ts
```diff
@@ -0,0 +1,40 @@
...
+
+const sliceName = 'demo';
+const initialState = {
+  greeting: 'Hello from Reducer',
+  count: 0
+};
+
+/* Selectors */
+export const getGreeting = path([sliceName, 'greeting']);
+export const getCount = path([sliceName, 'count']);
```
Mrtenz, line 14: I'm not sure what the benefit is of using these selectors over writing a simple function (like `state => state.demo.greeting`). These selectors also return a value with the type `unknown`.

src/store/index.ts
```diff
@@ -0,0 +1,3 @@
+export { useSelector, useDispatch } from 'react-redux';
```
Mrtenz, line 1: These hooks aren't type-safe by default, but you can use a simple wrapper function.

Example: https://github.com/MyCryptoHQ/signer/blob/master/src/app/store/utils/useSelector.ts

Mrtenz (changes requested):

src/store/store.ts
vilsbole: Done

src/features/DevTools/slice.ts
vilsbole: Yes, It's just a reminder to use `call` for testing side-effects.

src/store/index.ts
vilsbole: Ah great! I updated `src/store/index.tsx` to export the type-safe version.

src/features/DevTools/slice.ts
vilsbole: I'm still experimenting with them and suspect that the syntax can become cumbersome once all the reducers are combined. The type-safety is a problem though. I'll look into a better solution once I have a real store.

LuckyLuker:
[ch7005]

src/store/index.ts
```diff
@@ -0,0 +1,10 @@
+import { TypedUseSelectorHook, useDispatch, useSelector as useReduxSelector } from 'react-redux';
+
+import { default as store } from './store';
```
Mrtenz, line 3:
```suggestion
import store from './store';
```

src/features/DevTools/slice.ts
```diff
@@ -0,0 +1,40 @@
...
+
+/* Reducer */
+// 1. Toolkit sets .toString() to return the action type
+// 2. With the embeded Immer we can either modify the drat || return a new value
+const demoReducer = createReducer(initialState, {
+  [increment.type]: (s) => {
+    s.count = s.count + 1;
+  },
+  [reset.type]: () => initialState
+});
```
Mrtenz, line 38: This pattern isn't type-safe. You can use a `builder` instead to create reducers.

https://redux-toolkit.js.org/usage/usage-with-typescript#building-type-safe-reducer-argument-objects

Mrtenz (changes requested):

src/features/DevTools/slice.ts
vilsbole: Done.

src/store/index.ts
vilsbole: Done

Mrtenz (approved):

PR #3586, opened 2020-10-28
Fix Token Migration bugs

FrederikBolding:
- Fix not having a token migration screen for 1 tx, which caused it took look like the user was doing an approval but was actually doing the migration. This bug only occurs if the approval isn't necesarry.
- Fix loading state on all steps

### DIFF
--- a/src/components/TokenMigration/components/TokenMigrationConfirm.tsx
+++ b/src/components/TokenMigration/components/TokenMigrationConfirm.tsx
@@ -1,27 +1,28 @@
-import React, { useContext } from 'react';
+import React from 'react';
 
 import { ConfirmTransaction } from '@components';
-import { StoreContext, useAssets, useNetworks } from '@services';
-import { ITxType } from '@types';
-import { makeTxConfigFromRawTx } from '@utils/transaction';
+import { ITxType, StoreAccount } from '@types';
 
+import { makeTokenMigrationTxConfig } from '../helpers';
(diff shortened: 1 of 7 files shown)


src/utils/transaction.ts
```diff
@@ -282,6 +283,51 @@ export const makeTxConfigFromTxReceipt = (
   return txConfig;
 };
 
+export const makeTxConfigFromRawTx = (
```
blurpesec, line 286: I'm like 95% sure this exists somewhere already

src/components/TokenMigration/components/TokenMigrationConfirm.tsx
```diff
@@ -0,0 +1,34 @@
+import React, { useContext } from 'react';
+
+import { ConfirmTransaction } from '@components';
+import { StoreContext, useAssets, useNetworks } from '@services';
+import { ITxType } from '@types';
+import { makeTxConfigFromRawTx } from '@utils/transaction';
```
blurpesec, line 6: import from `@utils` instead

blurpesec (changes requested):

src/utils/transaction.ts
FrederikBolding: Yeah, I found and used a utility you had already made for token migration configs. Missed it at first.

blurpesec (approved):

PR #3587, opened 2020-10-28
Web3 E2E Mock

FrederikBolding:
Add web3 mock for e2e testing such that we can remove private keys entirely from the codebase.


FrederikBolding:
[ch7216]

src/utils/ethereumMock.ts
```diff
@@ -0,0 +1,69 @@
+import { ethers, Wallet } from 'ethers';
```
vilsbole, line 1: 1. I understand that you forked a project and adapted it to our repo.
Please include an explanation and source links in a comment to share the research. 
2. This is only used in tests. Please move out of `src/utils`. We can place it in `__tests__`?

vilsbole (changes requested):
Nice work!

src/utils/ethereumMock.ts
FrederikBolding: Jujst to clarify, there was no fork done for this PR. It was for the other experiment I did with Web3 E2E, which was dropped. Will move the file.

vilsbole (approved):

PR #3618, opened 2020-11-17
Remove ethereumjs-tx in favor of ethers

FrederikBolding:


src/services/WalletService/deterministic/ledger.ts
```diff
@@ -87,15 +85,17 @@ export class LedgerWallet extends HardwareWallet {
...
         r: addHexPrefix(result.r),
         s: addHexPrefix(result.s)
       };
 
-      return new EthTx(txToSerialize, { chain: txFields.chainId }).serialize();
+      const serializedTx = serializeTransaction(t, signature);
+
+      return Buffer.from(stripHexPrefix(serializedTx), 'hex');
     } catch (err) {
+      console.error(err);
```
Mrtenz, line 98: Remove?

src/services/WalletService/deterministic/trezor.ts
```diff
@@ -31,11 +31,15 @@ export class TrezorWallet extends HardwareWallet {
...
 
-  public signRawTransaction(tx: EthTx): Promise<Buffer> {
+  public signRawTransaction(tx: UnsignedTransaction): Promise<Buffer> {
     return new Promise((resolve, reject) => {
-      const { chainId, ...strTx } = getTransactionFields(tx);
+      const { chainId, ...rest } = tx;
+      if (!chainId) {
+        return reject(Error('Missing chainId on tx'));
+      }
+      const formattedTx = { ...rest, nonce: hexlify(rest.nonce!) };
```
Mrtenz, line 40: Is the nonce checked before this function is called? Otherwise maybe check if a nonce is set.

src/services/WalletService/non-deterministic/web3.ts
```diff
@@ -57,9 +57,9 @@ export default class Web3Wallet implements IFullWallet {
       data,
       nonce,
       chainId
-    };
+    } as unknown) as IHexStrWeb3Transaction;
 
-    if (!nodeLib) {
+    if (!nodeLib || !to) {
```
Mrtenz, line 62: Transactions don't always have `to`, see comment above.

src/utils/makeTransaction.ts
```diff
@@ -1,34 +1,20 @@
...
-  if (hasChainId(t)) {
-    return new Tx(t, { chain: t.chainId });
-  } else {
-    return new Tx(t);
+export const makeTransaction = (t: ITxObject | string): UnsignedTransaction => {
+  if (typeof t === 'string') {
+    return parseTransaction(t);
   }
+
+  return { ...t, nonce: parseInt(t.nonce) };
```
Mrtenz, line 17: You should always add a radix when using stuff like `parseInt`.

Mrtenz (changes requested):

FrederikBolding:
[ch7298]

Mrtenz (approved):

PR #3621, opened 2020-11-17
Add hiding of user actions

GuillaumeRx:
[CH7287]

### DIFF
--- a/src/components/ActionsPanel/ActionPanel.tsx
+++ b/src/components/ActionsPanel/ActionPanel.tsx
@@ -8,7 +8,7 @@ import { useUserActions } from '@services';
 import { StoreContext, State as StoreContextState } from '@services/Store/StoreProvider';
 import { COLORS, FONT_SIZE, SPACING } from '@theme';
 import { Trans } from '@translations';
-import { ActionTemplate } from '@types';
+import { ACTION_STATE, ActionTemplate } from '@types';
 
 import { Text } from '../NewTypography';
 import { ActionDetails, ActionsList } from './components';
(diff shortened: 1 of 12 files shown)


src/components/ActionsPanel/ActionPanel.tsx
```diff
@@ -68,14 +74,21 @@ export const ActionPanel = () => {
           <DetailsHeading>
             <Icon width={20} type={currentAction.icon} />
             <HeadingText>{currentAction.heading}</HeadingText>
+            <Icon width={20} type="closed-eye" onClick={dismiss} />
```
FrederikBolding, line 77: Can we have it use the clickable pointer? So it is obvious that it can be clicked.

src/components/ActionsPanel/components/ActionItem.tsx
```diff
@@ -20,7 +20,7 @@ const greenLightup = keyframes`
   }
 `;
 
-const Action = styled.div<{ state?: string }>`
+export const Action = styled.div<{ state?: string }>`
```
FrederikBolding, line 23: Why the export?

src/components/ActionsPanel/components/ActionItem.tsx
```diff
@@ -51,7 +51,7 @@ const IconContainer = styled.div`
   height: 28px;
 `;
 
-const TitleContainer = styled.div`
+export const TitleContainer = styled.div`
```
FrederikBolding, line 54: Why the export?

src/components/ActionsPanel/components/ActionsList.tsx
```diff
@@ -29,21 +44,44 @@ export const sortActions: (actionTemplates: ActionTemplate[]) => ActionTemplate[
...
     const userAction = findUserAction(a.name);
-    if (userAction && userAction.state === ACTION_STATE.COMPLETED) return false;
-    return true;
+    if (userAction && userAction.state === state) return true;
+    return false;
   };
 
-  const sortedActions = pipe(filter(isCompleted), sortActions)(actionTemplates);
-  const actions = union(sortedActions, actionTemplates);
+  const showedActions = filter((a: ActionTemplate) => !byState(a, ACTION_STATE.HIDDEN))(
```
FrederikBolding, line 55: Perhaps `visibleActions` is a better name?

FrederikBolding (changes requested):

src/components/ActionsPanel/components/ActionItem.tsx
GuillaumeRx: It was used for something I removed, fixing it

src/components/ActionsPanel/components/ActionItem.tsx
GuillaumeRx: It was used for something I removed, fixing it

FrederikBolding (approved):

PR #3622, opened 2020-11-18
Dynamic page titles

blurpesec:
[ch6456]

### Description
Set up dynamic page titles to improve SEO for the app.


src/routing/PrivateRoute.tsx
```diff
@@ -1,14 +1,30 @@
...
+interface PrivateRouteProps extends IAppRoute {
+  key: number;
 }
 
+interface PageTitleProps extends RouteComponentProps {
+  pageComponent: any;
+  title?: string;
+}
+
+const PageTitleRoute = ({ pageComponent: Component, title, ...props }: PageTitleProps) => {
```
vilsbole, line 20: have you tried with `react-helmet` ? https://dev.to/antibland/comment/140ag

src/features/ScreenLock/ScreenLockProvider.tsx
```diff
@@ -112,7 +112,7 @@ class ScreenLockProvider extends Component<
         this.props.history.replace(ROUTE_PATHS.DASHBOARD.path);
       });
       this.resetInactivityTimer();
-      document.title = translateRaw('SCREEN_LOCK_TAB_TITLE');
+      document.title = translateRaw('DEFAULT_PAGE_TITLE');
```
vilsbole, line 115: Since we route to dashboard afterwards we should be able to remove these and let the dashboard handle the change.

src/routing/PrivateRoute.tsx
blurpesec: I didn't do it that way because I noticed that we didn't already have it as a dependency and figured it'd be better not to add it for something so small.

src/types/routes.ts
```diff
@@ -5,7 +5,7 @@ export interface IRoutePath {
 }
 
 export interface IAppRoute extends IRoutePath {
-  component: React.ReactNode;
+  component: any;
```
FrederikBolding, line 8: Why?

src/routing/PrivateRoute.tsx
```diff
@@ -1,14 +1,30 @@
...
+import { useEffectOnce } from '@vendor';
 
-interface PrivateRouteProps {
-  [key: string]: any;
+interface PrivateRouteProps extends IAppRoute {
+  key: number;
 }
 
+interface PageTitleProps extends RouteComponentProps {
+  pageComponent: any;
```
FrederikBolding, line 16: Can this be a proper React type?

src/routing/PrivateRoute.tsx
```diff
@@ -1,14 +1,30 @@
...
+  title?: string;
+}
+
+const PageTitleRoute = ({ pageComponent: Component, title, ...props }: PageTitleProps) => {
+  useEffectOnce(() => {
+    document.title =
+      translateRaw('PAGE_TITLE_PREPEND') + title || translateRaw('DEFAULT_PAGE_TITLE');
```
FrederikBolding, line 23: Gives: "MyCrypto - undefined" if the title is undefined.

FrederikBolding (changes requested):

src/routing/PrivateRoute.tsx
vilsbole: I would prefer using a documented dependency to remove the burden of tests from our code.
But it's your call.

src/routing/PrivateRoute.tsx
blurpesec: nice catch :)

FrederikBolding (approved):

PR #3619, opened 2020-11-17
Use redux store

vilsbole:
[[ch7014]](https://app.clubhouse.io/mycrypto/story/7014/move-dbsync-to-redux-middleware)

- use react-redux over `useReducer`.
- use redux-persist over custom db storage.

To reduce the scope of the changes we introduce an intermediate step in our migration to redux by porting the existing reducer to redux.
In doing so we attempt to respect the current state structure (avoid changing selectors), as well as the existing model hooks api (to reduce changes to the callers).

PR #3630, opened 2020-11-20
Fix bug resulting in missing asset data (icons, rates & names) after screen lock, settings import, and settings reset

blurpesec:
[ch7317]

### Description
Fixed some bugs that result in missing asset data (icons, rates & other asset data) after screen lock & refresh, settings import, and settings hard reset.

### Changes
- Pass & handle existing assets as props when refreshing local storage.
- Refactor settings import to functional component just cause.
- Expanded `AssetLegacy` type to cover optional `ExtendedAsset` params.


FrederikBolding:
Closing this for now as we will find another solution when #3619 has been merged.

PR #3632, opened 2020-11-23
Revert db schema changes

FrederikBolding:


vilsbole (commented):
It still schema-change and I would prefer we add a valid migration over reverting (if version number is different execute each migration function)

FYI:
- redux persist will be handling the migration logic.

GuillaumeRx (approved):

PR #3637, opened 2020-11-23
MyCrypto Winter actions

GuillaumeRx:


FrederikBolding:
image
Looks like the tweet text is slightly weird.

src/types/userAction.ts
```diff
@@ -1,8 +1,11 @@
 import { ReactNode } from 'react';
 
-import { TIcon } from '@components';
-import { State as StoreContextState } from '@services/Store/StoreProvider';
+import { ActionButtonProps, TIcon, TweetButtonProps } from '@components';
+import { UniClaimResult } from '@services/ApiService/Uniswap/Uniswap';
```
FrederikBolding, line 4: Can we shorten this import a bit?

src/components/ActionsPanel/constants.ts
```diff
@@ -279,11 +338,107 @@ export const actionTemplates: ActionTemplate[] = [
...
+    heading: translateRaw('MYC_WINTER_MID_ACTION_HEADING'),
+    subHeading: GeneralSubHead,
+    props: {
+      content: translateRaw('MYC_WINTER_MID_ACTION_SUBHEAD')
+    },
+    body: [translate('MYC_WINTER_MID_ACTION_BODY_1'), translate('MYC_WINTER_MID_ACTION_BODY_2')],
+    priority: 0,
+    time: {
+      start: new Date('2020-12-12'),
+      end: new Date('2020-12-20')
```
FrederikBolding, line 387: Are the ends and starts supposed to overlap?

src/components/ActionsPanel/components/TweetButton.tsx
```diff
@@ -0,0 +1,31 @@
...
+
+export interface TweetButtonProps {
+  userAction: ExtendedUserAction;
+}
+
+export const TweetButton = ({ userAction }: TweetButtonProps) => {
+  const { memberships } = useContext(StoreContext);
+
+  const to = generateTweet(
+    translateRaw('MYC_WINTER_BONUS_TWEET', {
```
FrederikBolding, line 18: Can we add this as a prop so it is a more reusable component? If not, no worries.

src/components/ActionsPanel/ActionPanel.tsx
```diff
@@ -38,21 +40,35 @@ const HeadingText = styled.span`
...
+          filterUserActions(a, {
+            assets,
+            uniClaims,
+            ensOwnershipRecords,
+            accounts,
+            isMyCryptoMember
+          }),
+        filter((a: ActionTemplate) => (a.time ? dateIsBetween(a.time.start, a.time.end) : true))
+      )(actionTemplates),
+    [assets, uniClaims, ensOwnershipRecords, accounts, isMyCryptoMember]
```
FrederikBolding, line 70: Possibly this memo needs to be executed if the date changes? So the date between filter can be executed too.

FrederikBolding (changes requested):

src/components/ActionsPanel/constants.ts
GuillaumeRx: Yeah the next one should start when the previous one ends

FrederikBolding: But will a new one not start on the same day as another ends, effectively giving two actions?

src/components/ActionsPanel/components/TweetButton.tsx
GuillaumeRx: I can't actually - Otherwise I would've used the `ActionButton` but I need a wrapper to get the Membership address -

src/components/ActionsPanel/ActionPanel.tsx
GuillaumeRx: I prefer to wait for a app reload, it's pretty intensive to re-filter on every date change and I don't think it's useful

FrederikBolding (approved):
Good to merge when tests pass

PR #3653, opened 2020-12-02
Remove services dependant methods from utils

vilsbole:
Importing `@services` inside `@utils` raises circular dependency issues.
Move the relevant functions into `@helpers` and `@hooks`


src/features/SwapAssets/components/SwapTransactionReceipt.tsx
```diff
@@ -4,11 +4,11 @@ import pick from 'ramda/src/pick';
 
 import { MultiTxReceipt, TxReceipt } from '@components/TransactionFlow';
 import { getFiat } from '@config/fiats';
+import { makeTxItem } from '@helpers';
+import { TxParcel } from '@hooks';
```
FrederikBolding, line 8: Shouldn't this be imported from `@types`?

FrederikBolding (changes requested):
Please rebase

src/features/SwapAssets/components/SwapTransactionReceipt.tsx
vilsbole: Done.

FrederikBolding (approved):

PR #3664, opened 2020-12-09
Dashboard skeleton loading

FrederikBolding:
- [x] Move token scanning to Redux to reduce pop-in
- [x] Improve SVGs
- [x] Tests

### DIFF
--- a/jest_config/providersWrapper.tsx
+++ b/jest_config/providersWrapper.tsx
@@ -1,6 +1,7 @@
 import React from 'react';
 
-import { createStore } from '@store';
+import { DeepPartial } from '@reduxjs/toolkit';
+import { AppState, createStore } from '@store';
 import { Provider } from 'react-redux';
 import { ThemeProvider } from 'styled-components';
 
@@ -10,8 +11,14 @@ import { theme } from '@theme';
   Custom wrapper to enable rendered tests to consume providers data
   Ref: https://testing-library.com/docs/react-testing-library/setup
 */
-export const ProvidersWrapper = ({ children }: { children: React.ReactNode }) => (
-  <Provider store={createStore().store}>
+export const ProvidersWrapper = ({
+  children,
+  initialState
+}: {
+  children: React.ReactNode;
+  initialState?: DeepPartial<AppState>;
+}) => (
+  <Provider store={createStore(initialState).store}>
     <ThemeProvider theme={theme}>{children}</ThemeProvider>
   </Provider>
 );
--- a/src/features/AddAccount/AddAccountFlow.tsx
+++ b/src/features/AddAccount/AddAccountFlow.tsx
@@ -1,6 +1,6 @@
 import React, { useContext, useEffect, useReducer, useState } from 'react';
 
-import { fetchMemberships, useDispatch } from '@store';
+import { fetchMemberships, scanTokens, useDispatch } from '@store';
 import { withRouter } from 'react-router-dom';
 import { CSSTransition, TransitionGroup } from 'react-transition-group';
 
@@ -45,7 +45,7 @@ const AddAccountFlow = withRouter(({ history, match }) => {
   const dispatch = useDispatch();
   const [step, setStep] = useState(0); // The current Step inside the Wallet Story.
   const [formData, updateFormState] = useReducer(formReducer, initialState); // The data that we want to save at the end.
-  const { scanAccountTokens, scanTokens, addMultipleAccounts, accounts } = useContext(StoreContext);
+  const { addMultipleAccounts, accounts } = useContext(StoreContext);
   const { displayNotification } = useNotifications();
   const trackNewAccountAdded = useAnalytics({
     category: ANALYTICS_CATEGORIES.ADD_ACCOUNT,
@@ -94,12 +94,12 @@ const AddAccountFlow = withRouter(({ history, match }) => {
               displayNotification(NotificationTemplates.walletAdded, {
                 address: newAccounts[0].address
               });
-              scanAccountTokens(newAccounts[0]);
+              dispatch(scanTokens({ accounts: newAccounts }));
               dispatch(fetchMemberships([newAccounts[0]]));
             }
           : () => {
               displayNotification(NotificationTemplates.walletsAdded, { accounts: newAccounts });
-              scanTokens();
+              dispatch(scanTokens({}));
               dispatch(fetchMemberships(newAccounts));
             };
       handleAddition();
--- a/src/features/AddAccount/AddAccountFlow.tsx
+++ b/src/features/AddAccount/AddAccountFlow.tsx
@@ -99,7 +99,7 @@ const AddAccountFlow = withRouter(({ history, match }) => {
             }
           : () => {
               displayNotification(NotificationTemplates.walletsAdded, { accounts: newAccounts });
-              dispatch(scanTokens({}));
+              dispatch(scanTokens());
               dispatch(fetchMemberships(newAccounts));
             };
       handleAddition();
--- a/src/features/Dashboard/components/TokenPanel/TokenPanel.tsx
+++ b/src/features/Dashboard/components/TokenPanel/TokenPanel.tsx
@@ -29,7 +29,7 @@ export function TokenPanel() {
     .filter(isNotExcludedAsset(settings.excludedAssets));
 
   const handleScanTokens = async (asset?: ExtendedAsset) => {
-    dispatch(scanTokens({ assets: asset ? [asset] : undefined }));
+    dispatch(scanTokens({ assets: asset && [asset] }));
   };
 
   return showDetailsView && currentToken ? (
--- a/src/features/Dashboard/components/TokenPanel/TokenPanel.tsx
+++ b/src/features/Dashboard/components/TokenPanel/TokenPanel.tsx
@@ -9,13 +9,12 @@ import { TokenDetails } from './TokenDetails';
 import { TokenList } from './TokenList';
 
 export function TokenPanel() {
-  const { totals, currentAccounts, scanTokens } = useContext(StoreContext);
+  const { totals, currentAccounts, scanTokens, isScanning } = useContext(StoreContext);
   const { settings } = useSettings();
   const { getAssetRate } = useRates();
   const [showDetailsView, setShowDetailsView] = useState(false);
   const [showAddToken, setShowAddToken] = useState(false);
   const [currentToken, setCurrentToken] = useState<StoreAsset>();
-  const [isScanning, setIsScanning] = useState(false);
   const allTokens = totals(currentAccounts)
     .reduce((acc, a) => {
       if (a.contractAddress) {
@@ -26,13 +25,7 @@ export function TokenPanel() {
     .filter(isNotExcludedAsset(settings.excludedAssets));
 
   const handleScanTokens = async (asset?: ExtendedAsset) => {
-    try {
-      setIsScanning(true);
-      await scanTokens(asset);
-      setIsScanning(false);
-    } catch (e) {
-      setIsScanning(false);
-    }
+    await scanTokens(asset);
   };
 
   return showDetailsView && currentToken ? (
--- a/src/features/Dashboard/components/TokenPanel/TokenPanel.tsx
+++ b/src/features/Dashboard/components/TokenPanel/TokenPanel.tsx
@@ -1,5 +1,7 @@
 import React, { useContext, useState } from 'react';
 
+import { isScanning as isScanningSelector, scanTokens, useDispatch, useSelector } from '@store';
+
 import { StoreContext, useRates, useSettings } from '@services';
 import { isNotExcludedAsset } from '@services/Store/helpers';
 import { ExtendedAsset, StoreAsset } from '@types';
@@ -9,7 +11,9 @@ import { TokenDetails } from './TokenDetails';
 import { TokenList } from './TokenList';
 
 export function TokenPanel() {
-  const { totals, currentAccounts, scanTokens, isScanning } = useContext(StoreContext);
+  const { totals, currentAccounts } = useContext(StoreContext);
+  const isScanning = useSelector(isScanningSelector);
+  const dispatch = useDispatch();
   const { settings } = useSettings();
   const { getAssetRate } = useRates();
   const [showDetailsView, setShowDetailsView] = useState(false);
@@ -25,7 +29,7 @@ export function TokenPanel() {
     .filter(isNotExcludedAsset(settings.excludedAssets));
 
   const handleScanTokens = async (asset?: ExtendedAsset) => {
-    await scanTokens(asset);
+    dispatch(scanTokens({ assets: asset ? [asset] : undefined }));
   };
 
   return showDetailsView && currentToken ? (
--- a/src/features/Dashboard/components/WalletBreakdown/BreakdownChart.tsx
+++ b/src/features/Dashboard/components/WalletBreakdown/BreakdownChart.tsx
@@ -5,6 +5,7 @@ import equals from 'ramda/src/equals';
 import { Cell, Pie, PieChart, PieLabelRenderProps, Sector } from 'recharts';
 import styled from 'styled-components';
 
+import { SkeletonLoader } from '@components';
 import { Balance } from '@types';
 
 import { SMALLEST_CHART_SHARE_SUPPORTED } from './WalletBreakdownView';
@@ -22,6 +23,7 @@ interface BreakdownChartProps {
   balances: Balance[];
   selectedAssetIndex: number;
   isChartAnimating: boolean;
+  isScanning: boolean;
   shouldAnimate: boolean;
   handleMouseOver(index: number): void;
   handleMouseLeave(index: number): void;
@@ -127,6 +129,7 @@ const BreakdownChart = React.memo(
     handleMouseOver,
     handleMouseLeave,
     setIsChartAnimating,
+    isScanning,
     shouldAnimate,
     setShouldAnimate
   }: BreakdownChartProps) => {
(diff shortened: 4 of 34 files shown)


FrederikBolding:
[ch7440]

jest_config/providersWrapper.tsx
```diff
@@ -10,8 +11,14 @@ import { theme } from '@theme';
...
+  children,
+  initialState
+}: {
+  children: React.ReactNode;
+  initialState?: DeepPartial<AppState>;
+}) => (
+  <Provider store={createStore(initialState).store}>
     <ThemeProvider theme={theme}>{children}</ThemeProvider>
   </Provider>
 );
```
GuillaumeRx, line 24: Do we really need to pass the redux store to all the Components tests ?

src/features/AddAccount/AddAccountFlow.tsx
```diff
@@ -94,12 +94,12 @@ const AddAccountFlow = withRouter(({ history, match }) => {
...
                 address: newAccounts[0].address
               });
-              scanAccountTokens(newAccounts[0]);
+              dispatch(scanTokens({ accounts: newAccounts }));
               dispatch(fetchMemberships([newAccounts[0]]));
             }
           : () => {
               displayNotification(NotificationTemplates.walletsAdded, { accounts: newAccounts });
-              scanTokens();
+              dispatch(scanTokens({}));
```
GuillaumeRx, line 102: Why passing an empty object here ?

src/features/Dashboard/components/TokenPanel/TokenPanel.tsx
```diff
@@ -26,13 +29,7 @@ export function TokenPanel() {
...
 
   const handleScanTokens = async (asset?: ExtendedAsset) => {
-    try {
-      setIsScanning(true);
-      await scanTokens(asset);
-      setIsScanning(false);
-    } catch (e) {
-      setIsScanning(false);
-    }
+    dispatch(scanTokens({ assets: asset ? [asset] : undefined }));
```
GuillaumeRx, line 32: `assets: asset && [asset]` ?

src/features/Dashboard/components/WalletBreakdown/BreakdownChart.tsx
```diff
@@ -137,38 +140,41 @@ const BreakdownChart = React.memo(
...
-            cx={200}
-            cy={200}
-            innerRadius={0}
-            outerRadius={110}
-            label={(p) => CustomLabel(p as CustomLabelProps, selectedAssetIndex, handleMouseOver)}
-            labelLine={false}
-            dataKey="fiatValue"
-            onMouseEnter={(_: any, index: number) => handleMouseOver(index)}
+        {isScanning && <SkeletonLoader type="wallet-chart" width={400} height={350} />}
+        {!isScanning && (
```
GuillaumeRx, line 144: `isScanning ? ... : ...` would be easier to read ?

src/features/Dashboard/components/WalletBreakdown/BreakdownChart.tsx
```diff
@@ -137,38 +140,41 @@ const BreakdownChart = React.memo(
...
+            width={400}
+            height={350}
             onMouseLeave={() => handleMouseLeave(-1)}
-            animationDuration={800}
-            onAnimationStart={() => setIsChartAnimating(true)}
-            onAnimationEnd={() => {
-              setShouldAnimate(false);
-              setIsChartAnimating(false);
-            }}
+            onMouseEnter={() => handleMouseLeave(-1)}
```
GuillaumeRx, line 149: What does this ?

src/services/Store/store/membership.slice.ts
```diff
@@ -27,7 +27,7 @@ const slice = createSlice({
     setMembership(state, action: PayloadAction<MembershipStatus>) {
       state.record.push(action.payload);
     },
-    deleteMembership(state, action: PayloadAction<TAddress>) {
+    deleteMembership(state, action: PayloadAction<string>) {
```
GuillaumeRx, line 30: why changing this to a string ?

src/services/Store/store/tokenScanning.slice.ts
```diff
@@ -0,0 +1,129 @@
...
+  payload
+}: PayloadAction<{ accounts?: StoreAccount[]; assets?: Asset[] }>) {
+  const { accounts: requestedAccounts, assets: requestedAssets } = payload;
+
+  const networks = yield select(getNetworks);
+  const allAssets = yield select(getAssets);
+  const allAccounts = yield select(getAccounts);
+
+  const accounts: StoreAccount[] = requestedAccounts ? requestedAccounts : allAccounts;
+  const assets = requestedAssets ? [...allAssets, ...requestedAssets] : allAssets;
```
GuillaumeRx, line 118: can we only select assets and account if needed ?

GuillaumeRx (changes requested):

src/services/Store/store/membership.slice.ts
FrederikBolding: Because `Brand` is causing issues with TS :(

src/services/Store/store/tokenScanning.slice.ts
FrederikBolding: Assets are always needed, but I guess we could wait with accounts

jest_config/providersWrapper.tsx
FrederikBolding: What do you mean? The Redux store must be there if you are testing a component that gets data from it

src/features/Dashboard/components/WalletBreakdown/BreakdownChart.tsx
FrederikBolding: No idea, I did not change it

jest_config/providersWrapper.tsx
GuillaumeRx: Yes but it's not useful to load if the component don't use it :) I would prefer to wrap the component directly in the test file that needs it (like we do for `DataPriver` and `StoreProvider`) rather than loading this provider with every Component we test

src/features/Dashboard/components/WalletBreakdown/BreakdownChart.tsx
GuillaumeRx: Oh the line is highlighted green, I assumed this is something you added x) Nevermind then :)

src/services/Store/store/tokenScanning.slice.ts
GuillaumeRx: Oh yeah only for accounts then :)

jest_config/providersWrapper.tsx
FrederikBolding: You still have to wrap it in `ProvidersWrapper` manually to use it. I didn't add the store, I just added the extra parameter

GuillaumeRx: `ProvidersWrapper` is used inside `render` in `test-utils`, it automatically wraps all the test renders with ProvidersWrapper :)

FrederikBolding: Okay, but the we already did pass it to all components. I just added a prop

GuillaumeRx: Oh you're right ! I understand now why some of my test `console.warn` me on dumb components tests 😅 Nevermind then :)

src/services/Store/store/membership.slice.ts
GuillaumeRx: Then go with `string` unless you wanna cast :)

GuillaumeRx (approved):

src/features/PurchaseMembership/config.ts
```diff
@@ -36,7 +35,7 @@ export enum MembershipState {
 }
 
 export interface MembershipStatus {
-  address: TAddress;
+  address: string;
```
vilsbole, line 38: Conserve TAddress plz

src/services/Store/BalanceService.tsx
```diff
@@ -103,6 +112,20 @@ const tokenBalanceFetchWrapper = async (
...
+  contractList: string[],
+  options: any
+): Promise<EthScanBalanceMap<EthScanBalanceMap<EthScanBN>>> => {
+  try {
+    const tokenBalanceMap = await getTokensBalances(provider, addresses, contractList, options);
+    return tokenBalanceMap;
+  } catch {
+    return {} as EthScanBalanceMap<EthScanBalanceMap<EthScanBN>>;
+  }
+};
```
vilsbole, line 127: This simple sets the default response to to an object.
It's the responsibility of `getTokensBalances` to handle it's default value.

vilsbole (changes requested):
[1/2]

src/services/Store/store/tokenScanning.slice.ts
```diff
@@ -0,0 +1,133 @@
...
+  scanning: false
+};
+
+const slice = createSlice({
+  name: 'tokenScanning',
+  initialState,
+  reducers: {
+    scanTokens(
+      state,
+      _action: PayloadAction<{ accounts?: StoreAccount[]; assets?: Asset[] } | undefined>
```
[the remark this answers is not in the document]
FrederikBolding: How is the action typed if I remove the argument? It is required for the saga

src/services/Store/BalanceService.tsx
```diff
@@ -32,6 +33,14 @@ export const toBigNumberJS = (balances: EthScanBalanceMap): BalanceMap => {
   );
 };
 
+export const nestedToBigNumberJS = (
+  balances: EthScanBalanceMap<EthScanBalanceMap>
+): BalanceMap<BalanceMap> => {
+  return Object.fromEntries(
+    Object.keys(balances).map((key) => [key, toBigNumberJS(balances[key])])
```
vilsbole, line 40: 1. use `bigify` to harmonise out big number handling. 
2. Why not return an object instead of a tuple?
3. If so this can be simplified to
```
 .then(mapObjIndexed(bigify))
```

src/services/Store/BalanceService.tsx
```diff
@@ -199,15 +222,19 @@ export const getAccountsAssetsBalances = async (accounts: StoreAccount[]) => {
...
+  addresses: string[],
+  assets: Asset[]
+) => {
+  const provider = new ProviderHandler(network);
+  const assetsInNetwork = assets.filter((x) => x.networkId === network.id);
   const assetAddresses = getAssetAddresses(assetsInNetwork) as string[];
 
   try {
-    return tokenBalanceFetchWrapper(provider, account.address, assetAddresses, {
+    return tokenBalanceMultiFetchWrapper(provider, addresses, assetAddresses, {
```
vilsbole, line 235: `return` without an `await` renders the catch useless.

```
let apiFail = () => new Promise((resolve, reject) => {
  reject('failed')
})

let run = async () => {
  try {
    console.log('try')
    return apiFail().then(() => console.log('then'))
  } catch (err) {
    console.log('catch', err)
  }
}

run()
```

Screenshot 2020-12-14 at 15 39 46

Either `await` or use catch on the Promise.

src/services/Store/store/tokenScanning.slice.ts
```diff
@@ -0,0 +1,133 @@
...
+);
+
+/**
+ * Sagas
+ */
+export function* scanTokensSaga() {
+  yield takeLatest(scanTokens.type, scanTokensWorker);
+}
+
+export const getTokens = async (networks: Network[], accounts: StoreAccount[], assets: Asset[]) => {
```
vilsbole, line 56: Per conversation, separate api calls from transforms in order to test the later.

src/services/Store/store/tokenScanning.slice.ts
```diff
@@ -0,0 +1,133 @@
...
+
+      const unionedAssets = unionWith(eqBy(prop('uuid')), newAssets, existingAccount.assets);
+      return { ...acc, [uuid]: unionedAssets };
+    }, {});
+  }).then(mergeAll);
+};
+
+export function* scanTokensWorker({
+  payload
+}: PayloadAction<{ accounts?: StoreAccount[]; assets?: Asset[] } | undefined>) {
```
vilsbole, line 113: `StoreAccount` is designated for deprecation.
Can we use `IAccount` instead ?

vilsbole (changes requested):

src/services/Store/store/tokenScanning.slice.ts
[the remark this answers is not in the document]
vilsbole: It this case it may be best to create a separate `scanTokensRequest` action which is triggered by the saga to set the state.
We can choose to use either `status` as suggested in the summary https://redux.js.org/tutorials/essentials/part-5-async-logic or `request, success, failure` pattern.

src/features/Dashboard/components/WalletBreakdown/WalletBreakdown.test.tsx
```diff
@@ -0,0 +1,124 @@
...
+  it('can render', async () => {
+    const { getByText, getAllByText, container } = getComponent({
+      accounts: [{ ...fAccounts[0], assets: [...fAccounts[0].assets, ...fStoreAssets] }]
+    });
+
+    expect(getByText(translateRaw('WALLET_BREAKDOWN_TITLE'))).toBeInTheDocument();
+
+    // Renders Pie Chart
+    await waitFor(() =>
+      expect(container.querySelector('svg.recharts-surface')).toBeInTheDocument()
```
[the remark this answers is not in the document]
FrederikBolding: The problem is that the dependency doesn't support and render any other props. Hence why I had to find it using a query.

src/services/Store/BalanceService.tsx
FrederikBolding: Well, we changed this at one point to improve error handling: https://github.com/MyCryptoHQ/MyCrypto/commit/ee25ebe4bfcc3ba8d78ddcbb3ce50757adc37ca3

FrederikBolding: So I would prefer to keep that.

src/features/Dashboard/components/WalletBreakdown/WalletBreakdownView.tsx
```diff
@@ -275,7 +288,7 @@ export default function WalletBreakdownView({
           {translate('WALLET_BREAKDOWN_TITLE')}
           <BreakDownLabel>({label})</BreakDownLabel>
         </BreakDownHeading>
-        {parseFloat(totalFiatValue) === 0 ? (
+        {!isScanning && parseFloat(totalFiatValue) === 0 ? (
```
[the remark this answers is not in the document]
FrederikBolding: It was easier to set up this way. I tried initially with just one conditional but positioning the elements was a PITA.

src/services/Store/BalanceService.tsx
FrederikBolding: It can't error anyway as it is caught in the wrapper. So removing.

src/services/Store/BalanceService.tsx
FrederikBolding: We can't use `bigify` due to a mismatch in types of BN. See your comment on line 180.

FrederikBolding: Did simplify the helper functions though!

src/services/Store/BalanceService.tsx
vilsbole: The referenced commit adds 12LOC for what can be done in a single line on the caller. 
```
.then(...)
.catch(_ => {} as EthScanBalanceMap<...>)
```
I just really don't see any value added by preserving an extra layer in this case. The need to add an extra function name and remember that `tokenBalanceMultiFetchWrapper === getTokensBalances || {} when Error` is a small simplification that can help us improve the code base.

src/services/Store/store/tokenScanning.slice.ts
```diff
@@ -0,0 +1,159 @@
...
+  });
+};
+
+interface FetchBalancesResult {
+  network: Network;
+  tokenBalances: BalanceMap<BalanceMap<BigNumber>>;
+  baseAssetBalances: BalanceMap<BigNumber>;
+}
+
+const formatBalances = (assets: Asset[], accounts: IAccount[]) => ({
```
vilsbole, line 80: Please include a test for the formatter.

src/features/Dashboard/components/WalletBreakdown/WalletBreakdownView.tsx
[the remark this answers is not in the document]
vilsbole: Ok, was just checking 👍

src/features/Dashboard/components/WalletBreakdown/WalletBreakdown.test.tsx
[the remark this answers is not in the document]
vilsbole: Ok

src/services/Store/BalanceService.tsx
```diff
@@ -27,9 +29,13 @@ export const convertBNToBigNumberJS = (bn: EthScanBN): BN => {
...
-  return Object.fromEntries(
-    Object.keys(balances).map((key) => [key, convertBNToBigNumberJS(balances[key])])
-  );
+  return mapObjIndexed(convertBNToBigNumberJS, balances);
+};
+
+export const nestedToBigNumberJS = (
+  balances: EthScanBalanceMap<EthScanBalanceMap>
+): BalanceMap<BalanceMap> => {
+  return mapObjIndexed(toBigNumberJS, balances);
```
vilsbole, line 38: Continuing: https://github.com/MyCryptoHQ/MyCrypto/pull/3664#discussion_r543267452_
This is precisely why we `bigify` in order to harmonise the types.
The function should be able to handle the conversion from any bigish type to bignumber.js.
The important part is that all our big numbers within the application are the type of the return value of `bigify`.

src/services/Store/BalanceService.tsx
```diff
@@ -27,9 +29,13 @@ export const convertBNToBigNumberJS = (bn: EthScanBN): BN => {
 };
 
 export const toBigNumberJS = (balances: EthScanBalanceMap): BalanceMap => {
-  return Object.fromEntries(
-    Object.keys(balances).map((key) => [key, convertBNToBigNumberJS(balances[key])])
-  );
+  return mapObjIndexed(convertBNToBigNumberJS, balances);
```
vilsbole, line 32: use bigify for the conversion.

vilsbole (changes requested):

vilsbole (approved):
Great job!

PR #3674, opened 2020-12-15
Move Vault to Redux

FrederikBolding:


FrederikBolding:
[ch7469]

src/features/ScreenLock/ScreenLockLocked.tsx
```diff
@@ -45,7 +46,9 @@ const BottomActions = styled.div`
 `;
 
 const ScreenLockLocked: React.FC = () => {
-  const { decryptWithPassword, decryptError } = useContext(ScreenLockContext);
+  const decryptErrorRedux = useSelector(getDecryptionError);
```
vilsbole, line 49: Access by hooks requires additional mocking for tests and stories.
We're going to standardise redux access using `connect`.
This is a good moment to connect this component.

src/utils/safeJSONParse.ts
```diff
@@ -0,0 +1,10 @@
+export const safeJSONParse = (str: string) => {
```
vilsbole, line 1: Can you explain/comment the pattern please?
Since you later `throw` when calling `isValidJSON` it seems like it's just hiding the original error.

src/services/Store/store/vault.slice.test.ts
```diff
@@ -0,0 +1,88 @@
...
+  });
+
+  it('goes into error state if password is wrong', () => {
+    return expectSaga(decryptionWorker, decrypt(wrongPassword))
+      .withState({
+        legacy: { networks: fNetworks, assets: fAssets, accounts: fAccounts },
+        vault: {
+          data: encryptedData
+        }
+      })
```
vilsbole, line 83: use / introduce mockAppState from test-utils

src/services/Store/store/vault.slice.test.ts
```diff
@@ -0,0 +1,88 @@
...
+
+describe('decryptionWorker()', () => {
+  it('decrypts existing state and imports', () => {
+    return expectSaga(decryptionWorker, decrypt(hashedPassword))
+      .withState({
+        legacy: { networks: fNetworks, assets: fAssets, accounts: fAccounts },
+        vault: {
+          data: encryptedData
+        }
+      })
```
vilsbole, line 69: use / introduce mockAppState from test-utils

src/services/Store/store/vault.slice.test.ts
```diff
@@ -0,0 +1,88 @@
...
+const decryptedData =
+  '{"version":"v1.0.0","mtime":1608048741382,"accounts":{},"addressBook":{"a1acf1f2-0380-5bd6-90c3-2b4a0974a6fe":{"label":"MyCrypto Tip Jar","address":"0x4bbeEB066eD09B7AEd07bF39EEe0460DFa261520","notes":"Toss us a coin!","network":"Ethereum","uuid":"a1acf1f2-0380-5bd6-90c3-2b4a0974a6fe"}},"assets":{},"contracts":{},"networks":{},"notifications":{},"settings":{"fiatCurrency":"USD","darkMode":false,"dashboardAccounts":[],"excludedAssets":[],"inactivityTimer":1800000,"rates":{},"language":"en"},"password":"b9c950640e1b3740e98acb93e669c65766f6670dd1609ba91ff41052ba48c6f3","networkNodes":{},"userActions":{}}';
+
+Date.now = jest.fn(() => 1608048741382);
+
+describe('encryptionWorker()', () => {
+  it('encrypts existing state and resets app state', () => {
+    return expectSaga(encryptionWorker, encrypt(hashedPassword))
+      .withState({ legacy: marshallState(JSON.parse(decryptedData)) })
```
vilsbole, line 54: use / introduce mockAppState from test-utils

vilsbole (changes requested):
example of `mockAppState` here: https://github.com/MyCryptoHQ/MyCrypto/pull/3679/files#

src/utils/safeJSONParse.ts
FrederikBolding: I've added some comments, but in general we don't care about the error message as we show a generic one.

src/services/Store/store/vault.slice.test.ts
FrederikBolding: Getting massive circular dependency issues when trying to introduce the `mockAppState`. Would prefer fixing when one of the PRs are merged.

src/utils/safeJSONParse.ts
vilsbole: I was more thinking of preserving the original error for debugging purposes.
Do you have a link for this pattern?
The access by array to validate is offsetting.
It implies that to use function to access the parse result you need to write:
`safeJSONParse(str)[1]`.
Would something like this work ? 
```
try {
 return JSON.parse(data)
} catch (err) {
 return false
}
```

vilsbole (changes requested):

vilsbole (approved):

PR #3679, opened 2020-12-16
Replace legacy reducer

vilsbole:


src/services/ApiService/MembershipApi.ts
```diff
@@ -16,15 +16,15 @@ export const formatResponse = (timestamps: PromiseType<ReturnType<typeof getUnlo
   // Convert to Bigies so we can determine their expiry date.
   // @todo: prefer date-fns for time comparaisons.
   const expiries = pipe(
-    mapObjIndexed(mapObjIndexed(toString)),
-    mapObjIndexed(mapObjIndexed(bigify))
+    mapObjIndexed(mapObjIndexed(bigify)),
```
FrederikBolding, line 19: Can we keep these as BN's please? They are a result of a contract call and thus can theoretically be larger than what can fit in an int. (you parse these later as int)

src/services/ApiService/MembershipApi.ts
```diff
@@ -16,15 +16,15 @@ export const formatResponse = (timestamps: PromiseType<ReturnType<typeof getUnlo
...
+    mapObjIndexed(mapObjIndexed(toString))
   )(timestamps);
 
   return Object.keys(expiries)
     .map((address: TAddress) => ({
       address,
       memberships: Object.keys(expiries[address])
-        .filter((contract) => expiries[address][contract].isGreaterThan(bigify(0)))
+        .filter((contract) => expiries[address][contract] !== '0')
```
FrederikBolding, line 27: This is not equivalent

src/services/Store/DataManager/DataProvider.tsx
```diff
@@ -31,10 +37,7 @@ export const DataProvider: React.FC = ({ children }) => {
 
   // @todo: Redux create action for reset once legacy.reducer is replaced.
   const resetAppDb = (newDb = initialLegacyState) => {
-    dispatch({
-      type: ActionT.RESET,
-      payload: { data: newDb }
-    });
+    dispatch(appReset(newDb));
```
FrederikBolding, line 40: 👍

src/features/Dashboard/components/MembershipPanel.tsx
```diff
@@ -66,8 +67,8 @@ const Icon = styled.img<{ isMemberOrExpired: boolean }>`
 
 type Props = RouteComponentProps;
 function MembershipPanel({ history }: Props) {
-  const { membershipState, memberships, membershipExpirations } = useContext(StoreContext);
-
+  const { membershipState, memberships } = useContext(StoreContext);
+  const expiryDate = useSelector(membershipExpiryDate);
```
FrederikBolding, line 71: Should be passed via props, right?

FrederikBolding (changes requested):

src/services/ApiService/MembershipApi.ts
vilsbole: It doesn't have to be since we're just eliminating null values.
The contract api set's null as `0x00`.
`expiries` transforms it to a string '0' on which we can filter.

src/services/ApiService/MembershipApi.ts
vilsbole: The contract returns a BN for a unix timestamp
It's a contract which we control so I don't see the risk of an underflow/overflow.
Within the app, redux encourages us to only store serializable values.
The value is a date and should be treated as such.
This is why it seems best to store the value as a string and use date operations.

FrederikBolding: We don't control the contract the BN comes from. At the end of the day it is a returned result from an unlock contract.

GuillaumeRx (approved):

FrederikBolding (approved):

PR #3683, opened 2020-12-18
Broadcast tx styling and error handling fixes

FrederikBolding:


FrederikBolding:
[ch7456]

FrederikBolding:
[ch7454]

blurpesec (commented):
I've left you a message regarding this in slack. I wasn't able to get to the txReceipt because of anerror that appears on the confirm tx panel. But the transaction did succeed.

image

src/components/TransactionFlow/ConfirmTransaction.tsx
```diff
@@ -8,6 +8,7 @@ import sendIcon from '@assets/images/icn-send.svg';
 import walletIcon from '@assets/images/icn-wallet.svg';
 import { Amount, AssetIcon, Button, PoweredByText } from '@components';
 import ProtectIconCheck from '@components/icons/ProtectIconCheck';
+import { InlineMessage } from '@components/InlineMessage';
```
blurpesec, line 11: Can we import from `@components` here?

src/features/BroadcastTransaction/components/BroadcastTx.tsx
```diff
@@ -98,20 +94,25 @@ interface Props {
   handleNetworkChanged(network: NetworkId): void;
 }
 
-const BroadcastTx = ({ signedTx, network, onComplete, handleNetworkChanged }: Props) => {
+const BroadcastTx = ({
+  signedTx,
+  network: networkId,
```
blurpesec, line 99: Can you change this elsewhere to just be `networkId` so we don't confuse ourselves in the future?

src/features/BroadcastTransaction/components/BroadcastTx.tsx
```diff
@@ -132,34 +135,39 @@ const BroadcastTx = ({ signedTx, network, onComplete, handleNetworkChanged }: Pr
...
           placeholder="0xf86b0284ee6b2800825208944bbeeb066ed09b7aed07bf39eee0460dfa26152088016345785d8a00008029a03ba7a0cc6d1756cd771f2119cf688b6d4dc9d37096089f0331fe0de0d1cc1254a02f7bcd19854c8d46f8de09e457aec25b127ab4328e1c0d24bfbff8702ee1f474"
           onChange={handleChange}
-          onBlur={validateField}
-          inputError={!transaction ? inputError : ''}
+          inputError={userInput.length > 0 ? inputError : ''}
         />
-        {transaction && <IdenticonIcon address={userInput} />}
       </InputWrapper>
-      {transaction ? (
+      {isValid && transaction && (
```
blurpesec, line 141: `isValid` makes the `&& transaction` redundant

src/features/BroadcastTransaction/components/BroadcastTx.tsx
```diff
@@ -132,34 +135,39 @@ const BroadcastTx = ({ signedTx, network, onComplete, handleNetworkChanged }: Pr
...
         <React.Fragment>
           {!transaction.chainId && (
             <NetworkSelectWrapper>
-              <NetworkSelector network={network} onChange={handleNetworkChanged} />
-              {!network && (
+              <NetworkSelector network={networkId} onChange={handleNetworkChanged} />
```
blurpesec, line 145: Can you change this elsewhere to just be `networkId` so we don't confuse ourselves in the future?

src/features/BroadcastTransaction/stateFactory.tsx
```diff
@@ -70,7 +70,10 @@ const BroadcastTxConfigFactory: TUseStateReducerFactory<State> = ({ state, setSt
       cb();
     } catch (err) {
       console.debug(`[BroadcastTx] ${err}`);
-      displayToast(toastTemplates.failedTransaction);
```
blurpesec, line 73: Are we tossing out toasts from here on?

src/features/BroadcastTransaction/stateFactory.tsx
```diff
@@ -70,7 +70,10 @@ const BroadcastTxConfigFactory: TUseStateReducerFactory<State> = ({ state, setSt
...
+      setState((prevState: State) => ({
+        ...prevState,
+        error: err.message
+      }));
```
blurpesec, line 76: We need to add in some handling for whether the tx went through or not, so that we can display to the user the tx succeeded even if we failed to process the result & display a txReceipt.

blurpesec (changes requested):

src/features/BroadcastTransaction/stateFactory.tsx
FrederikBolding: I just found it to be unnecesary here when we already have an error message. We still use them elsewhere. If you prefer I can re-add?

src/features/BroadcastTransaction/stateFactory.tsx
FrederikBolding: Hmm, unsure of how that would work. In general if it succeeds we should be able to show a receipt. Probably out of scope for this anyway.

blurpesec (approved):
rebase, then good to go c:

PR #3684, opened 2020-12-18
Extras Tray

GuillaumeRx:


vilsbole (changes requested):
To review: 
- on mobile: Top nav should not overlay tray when scrolling.
- on desktop: When tray is open, a click on another nav item or outside the tray should close it.

src/features/Layout/Layout.tsx
```diff
@@ -139,6 +139,7 @@ export default function Layout({ config = {}, className = '', children }: Props)
   const { pathname } = useLocation();
 
   const [topHeight, setTopHeight] = useState(0);
+  const [isOpen, setIsOpen] = useState<boolean>(false);
```
vilsbole, line 142: The implementation is convoluted, Layout and other nav elements don't need to be aware of the Tray behaviour. Can we try using something like https://github.com/streamich/react-use/blob/master/docs/useClickAway.md (to import in `/vendor/react-use`)?

vilsbole (changes requested):
When scrolling the ExtrasTray on mobile, the bottom nav bar appears over the tray.

vilsbole (approved):

PR #3700, opened 2021-01-07
MultiTxReceipt fixes

FrederikBolding:


bowensanders:
This appears to have fixed the problem with tx receipt status not changing on GNT migration (the bug as reported)

Mrtenz (approved):

PR #3714, opened 2021-01-11
Update Accounts panel icons

GuillaumeRx:
[CH7360]


src/assets/icons/actions/edit.svg
```diff
@@ -0,0 +1,10 @@
+<svg width="16" height="16" viewBox="0 0 16 16" fill="none" xmlns="http://www.w3.org/2000/svg">
```
FrederikBolding, line 1: Remove height and width?

src/assets/icons/actions/add_bold.svg
```diff
@@ -0,0 +1,8 @@
+<svg width="16" height="16" viewBox="0 0 16 16" fill="none" xmlns="http://www.w3.org/2000/svg">
```
FrederikBolding, line 1: Remove width and height?

src/components/WalletUnlock/DeterministicAccountTable.tsx
```diff
@@ -372,7 +372,7 @@ const DeterministicTable = ({
             onClick={generateFreshAddress}
             disabled={!isComplete || freshAddressIndex >= DEFAULT_GAP_TO_SCAN_FOR}
           >
-            <Icon type="add" color="none" />
+            <Icon type="add" color="none" width="32px" />
```
FrederikBolding, line 375: This is not the same size as the blockies, is that intentional?

FrederikBolding (changes requested):

src/components/WalletUnlock/DeterministicAccountTable.tsx
GuillaumeRx: I just removed the sizing on the svg, I put the value that was previously set in the svg's code. It was this size before :)

FrederikBolding: I saw, just confirming that this is intentional.

GuillaumeRx: Not really part of the work but I initially wanted to use the `add` icon - it was slightly different from the one I needed so I left the removal of the width/height but didn't use the icon :)

FrederikBolding (approved):

PR #3717, opened 2021-01-12
Add better back navigation

FrederikBolding:
- Fix an issue where the back button would be broken if you navigated directly to a page


FrederikBolding:
[ch7568]

GuillaumeRx (approved):

PR #3718, opened 2021-01-12
Use debounce for gas estimation on SendAssets

FrederikBolding:


FrederikBolding:
[ch7559]

src/features/SendAssets/components/SendAssetsForm.tsx
```diff
@@ -442,9 +442,13 @@ const SendAssetsForm = ({ txConfig, onComplete, protectTxButton, isDemoMode }: P
     handleNonceEstimate(values.account);
   }, [values.account]);
 
-  useEffect(() => {
-    handleGasEstimate();
-  }, [values.account, values.address]);
+  useDebounce(
```
vilsbole, line 445: Can we migrate to saga instead?

FrederikBolding: Refactoring the Send flow is not the priority at the moment. Getting polished is. So I would say we wait with that until we refactor it further.

Mrtenz (approved):

PR #3734, opened 2021-01-15
Improve default account selection

FrederikBolding:


vilsbole (commented):
Nice, but it would be better to use a store selector which is aware of the current network.

src/utils/wallets.ts
```diff
@@ -1,6 +1,21 @@
-import { StoreAccount, TAddress, WalletId } from '@types';
+import { NetworkId, StoreAccount, TAddress, WalletId } from '@types';
 import { isSameAddress } from '@utils';
 
+export const getAccountsInNetwork = ({
```
vilsbole, line 4: Are accounts nested in Network? Maybe `getAccountsByNetwork` ?

src/features/BuyAssets/BuyAssetsForm.test.tsx
```diff
@@ -0,0 +1,52 @@
...
+
+import { fAccounts, fAssets, fNetworks, fSettings } from '@fixtures';
+import { DataContext, IDataContext, RatesContext, StoreContext } from '@services';
+import { translateRaw } from '@translations';
+
+import BuyAssetsForm from './BuyAssetsForm';
+
+function getComponent() {
+  return simpleRender(
+    <Router>
```
vilsbole, line 14: How about adding Router to `ProvidersWrapper`

vilsbole (changes requested):

FrederikBolding:
> Nice, but it would be better to use a store selector which is aware of the current network.

I don't know what you mean with current network. We don't have a globally selected current network, it is dependant on the flow.

blurpesec (commented):
Code looks fine to me, but can you rebase to have the e2e test changes in?

blurpesec (approved):

PR #3759, opened 2021-01-22
Remove TX Status limits, Bump Ethers, Fix formatting errors for some networks

FrederikBolding:


409H:
See CH7422

package.json
```diff
@@ -38,7 +38,7 @@
     "date-fns": "2.16.1",
     "ethereumjs-abi": "0.6.8",
     "ethereumjs-util": "6.0.0",
-    "ethers": "4.0.48",
+    "ethers": "5.0.26",
```
Mrtenz, line 41: Maybe explicitly install `@ethersproject/*` stuff rather than `ethers`.

src/helpers/transaction.ts
```diff
@@ -1,16 +1,14 @@
...
-  parseTransaction,
+  parse as parseTransaction,
   recoverAddress,
-  serializeTransaction,
+  serialize as serializeTransaction,
   Transaction
-} from 'ethers/utils';
+} from '@ethersproject/transactions';
+import { formatEther } from '@ethersproject/units';
+import { addHexPrefix, toChecksumAddress } from 'ethereumjs-util';
```
Mrtenz, line 11: We can use ethers.js here instead of ethereumjs.

src/services/EthService/utils/providerWrappers.ts
```diff
@@ -37,7 +37,8 @@ export const getTransactionReceiptFromHash = async (
   await provider
     .getTransactionReceipt(txHash)
     .then((receipt) => receipt)
-    .catch(() => {
+    .catch((err) => {
+      console.error(err);
```
Mrtenz, line 41: Did you forget to remove this?

Mrtenz (changes requested):

src/helpers/transaction.ts
FrederikBolding: What are these functions called in `ethers.js`?

package.json
FrederikBolding: Yeah, just don't know which ones we need at the moment 😄

src/helpers/transaction.ts
Mrtenz: * [`getAddress`](https://github.com/ethers-io/ethers.js/blob/master/packages/address/src.ts/index.ts#L77) in `@ethersproject/address`
* Couldn't find a proper replacement for `addHexPrefix` but if we want to remove `ethereumjs-util` fully at some point, it should be easy enough to make a util function for that.

package.json
Mrtenz: As of now:
* `@ethersproject/bignumber`
* `@ethersproject/address`
* `@ethersproject/providers`
* `@ethersproject/transactions`
* `@ethersproject/bytes`
* `@ethersproject/abstract-provider`
* `@ethersproject/units`
* `@ethersproject/wallet`

Mrtenz (dismissed):

FrederikBolding:
Thanks for the review @Mrtenz 

I think @bowensanders should give this a look over before merging though!

bowensanders:
I have just tested all available networks via tx's from their respective block explorers. 

I say this is OK to GO

Mrtenz (approved):

PR #3763, opened 2021-01-23
Handle dashboard state when all assets are hidden

blurpesec:
### Description
Introduced a new state of the dashboard for when all assets are hidden.

image


src/features/Dashboard/components/WalletBreakdown/WalletBreakdownView.tsx
```diff
@@ -281,7 +281,12 @@ export default function WalletBreakdownView({
       .dividedBy(new BigNumber(totalFiatValue))
       .multipliedBy(100)
       .toFixed(2);
-
+  console.debug(
+    'chartBalances',
+    chartBalances.length,
+    'breakdownBalances: ',
+    breakdownBalances.length
+  );
```
vilsbole, line 289: Do we need to keep this ?
If so prefer `logger` from `@utils`

src/features/Dashboard/components/WalletBreakdown/WalletBreakdownView.tsx
```diff
@@ -281,7 +281,12 @@ export default function WalletBreakdownView({
       .dividedBy(new BigNumber(totalFiatValue))
       .multipliedBy(100)
       .toFixed(2);
-
+  console.debug(
```
FrederikBolding, line 284: Remove?

FrederikBolding (changes requested):
Looks good except for one small thing

src/features/Dashboard/components/WalletBreakdown/WalletBreakdownView.tsx
FrederikBolding: Didn't see this 😆

FrederikBolding (approved):

PR #3765, opened 2021-01-25
Fix RequestAssets crash with non-ETH default account

FrederikBolding:


vilsbole (changes requested):
When clicking on the `Request` link in the nav menu, the app throws the following error: 

Screenshot 2021-01-25 at 14 21 49

src/utils/wallets.ts
```diff
@@ -7,12 +7,12 @@ export const getAccountsByNetwork = ({
...
 }: {
   accounts: StoreAccount[];
-  networkId: NetworkId;
+  networkId?: NetworkId;
   includeViewOnly?: boolean;
 }) =>
   accounts.filter(
     (acc) =>
-      acc.networkId === networkId &&
+      (acc.networkId === networkId || !networkId) &&
```
vilsbole, line 15: I looks like monkey patching.

vilsbole (changes requested):

FrederikBolding:
Can't reproduce your error, this is the error this should fix. @vilsbole

src/utils/wallets.ts
FrederikBolding: If you have suggestions I am open to it.

vilsbole (approved):

PR #3766, opened 2021-01-25
Use dynamic feature flags

vilsbole:
Move feature flags to redux so they are accessible within sagas.


src/services/Analytics/useAnalytics.test.tsx
```diff
@@ -45,12 +37,4 @@ describe('useAnalytics', () => {
     act(() => result.current.trackPage(params));
     expect(mockDispatch).toHaveBeenCalledWith(actionWithPayload(params));
   });
-  it('blocks calls to AnalyticsService when feature is inactive', async () => {
```
blurpesec, line 48: Do we not want this type of test anymore? I think it'd still be good to include something like this

blurpesec (dismissed):

src/services/Analytics/useAnalytics.test.tsx
vilsbole: Responsibility of respecting feature flag is moved to saga.
I cleaned up the hook test and added the an assertion to the saga test.

blurpesec (approved):

PR #3769, opened 2021-01-25
Humanize ENS address error

FrederikBolding:


FrederikBolding:
[ch7799]

src/components/DomainStatus.tsx
```diff
@@ -27,6 +30,13 @@ const withInlineError = (children: JSX.Element, type?: InlineMessageType) => (
   <InlineError type={type}>{children}</InlineError>
 );
 
+const humanizeEnsError = (error: ResolutionError, props: DomainStatusProps) => {
+  if (error.code === ResolutionErrorCode.RecordNotFound) {
```
blurpesec, line 34: Are there any other error codes we need to catch and display for, because all the error messages will likely suck if one does.

blurpesec (approved):

PR #3778, opened 2021-01-26
Simplify console plug

vilsbole:
Previous plug would overflow in default browser console width.


FrederikBolding (changes requested):
Text is invisible for me

FrederikBolding (approved):

PR #3799, opened 2021-01-30
Fix Trezor issues and types

FrederikBolding:


blurpesec (approved):
code approf - need qa from @bowensanders before release

bowensanders:
Mainnet transactions are working now.

There still seems to be an issue with testnet sends but I think this is a seperate issue.

FrederikBolding:
[ch7912]

PR #3803, opened 2021-02-01
Fix broken links

FrederikBolding:


FrederikBolding:
`about.mycrypto.com` is deprecated as we are moving towards the new landing page. Therefore ignore this PR.

PR #3806, opened 2021-02-02
Update to use react-refresh

blurpesec:
[ch7858]

### Description
Updates to use react-refresh, enabling hmr fixes on `<PrivateRoute />`


FrederikBolding (dismissed):
Can we remove `react-hot-loader` and other unused dependencies then?

vilsbole (approved):

PR #3809, opened 2021-02-02
Mobile membership fixes

FrederikBolding:
[ch7793]

### DIFF
--- a/src/components/InlineMessage/InlineMessage.tsx
+++ b/src/components/InlineMessage/InlineMessage.tsx
@@ -5,10 +5,9 @@ import styled from 'styled-components';
 import arrowSVG from '@assets/images/icn-arrow-purple.svg';
 import infoSVG from '@assets/images/icn-info.svg';
 import warningSVG from '@assets/images/icn-warning.svg';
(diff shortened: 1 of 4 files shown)


src/components/InlineMessage/InlineMessage.tsx
```diff
@@ -33,8 +32,9 @@ interface BannerTypographyProps {
   color: string;
 }
 
-const STypography = styled(Typography)<BannerTypographyProps>`
+const STypography = styled(Body)<BannerTypographyProps>`
   color: ${(props) => props.color};
+  vertical-align: middle;
```
GuillaumeRx, line 37: use inline styling instead of `styled-component`

GuillaumeRx (changes requested):

src/components/InlineMessage/InlineMessage.tsx
FrederikBolding: It complains when I use `verticalAlign="middle"` :(

GuillaumeRx (approved):

PR #3825, opened 2021-02-08
Center mobile nav icon

vilsbole:
Fix UI bug where Menu icon on mobile was no longer centered.

Before:
Screenshot 2021-02-08 at 17 42 32

After:
Screenshot 2021-02-08 at 17 38 51


GuillaumeRx (commented):
To avoid one time use of variants I would rather wrap the `Icon` and `Text` in `TopItem` with a `<Box variant='columnCenter`>`

GuillaumeRx (approved):

PR #3840, opened 2021-02-15
Fix Analytics Middleware

FrederikBolding:
Fix issue with not calling next middleware in middleware chain

https://redux.js.org/tutorials/fundamentals/part-4-store#writing-custom-middleware

### DIFF
--- a/src/services/Analytics/analytics.middleware.ts
+++ b/src/services/Analytics/analytics.middleware.ts
@@ -47,9 +47,8 @@ const analyticsMiddleware: Middleware<TObject, any, Dispatch<Action>> = (state) 
       );
       break;
     }
-    default:
-      next(action);
   }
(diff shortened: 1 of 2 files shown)


Mrtenz (approved):

src/services/Analytics/analytics.middleware.ts
```diff
@@ -47,9 +47,8 @@ const analyticsMiddleware: Middleware<TObject, any, Dispatch<Action>> = (state)
       );
       break;
     }
-    default:
-      next(action);
   }
+  return next(action);
```
vilsbole, line 51: How about adding a `default` label to avoid confusion such as:
```
default: // do nothing;
         break;
```

vilsbole (approved):

PR #3844, opened 2021-02-17
Fix serialization issues with transaction BNs

FrederikBolding:


src/services/Store/Account/useAccounts.tsx
```diff
@@ -1,19 +1,16 @@
...
 import {
   addAccounts,
   destroyAccount,
+  getAccounts,
   updateAccount as updateAccountRedux,
   updateAccounts as updateAccountsRedux,
   useDispatch
 } from '@store';
 import { Asset, IAccount, IRawAccount, ITxReceipt, NetworkId, StoreAccount, TUuid } from '@types';
+import { eqBy, prop, unionWith } from '@vendor';
```
vilsbole, line 12: 👍

src/services/Store/store/account.slice.ts
```diff
@@ -1,3 +1,4 @@
+import { BigNumber as EthersBN } from '@ethersproject/bignumber';
```
vilsbole, line 1: Any reason not to use `bigify` ?

src/services/Store/store/account.slice.ts
```diff
@@ -75,7 +76,19 @@ export default slice;
...
+export const getAccounts = createSelector([getAppState], (s) => {
+  const accounts = s[slice.name];
+  return accounts.map((a) => ({
+    ...a,
+    transactions: a.transactions.map((t) => ({
+      ...t,
+      value: EthersBN.from(t.value),
+      gasLimit: EthersBN.from(t.gasLimit),
+      gasPrice: EthersBN.from(t.gasPrice),
+      gasUsed: t.gasUsed && EthersBN.from(t.gasUsed)
```
vilsbole, line 88: With this, it because useful to include a selector test to assert that the value is `bigish`

src/services/Store/store/account.slice.ts
FrederikBolding: We use `EthersBN` for tx values at the moment.

src/services/Store/store/account.slice.ts
FrederikBolding: I don't think I understand entirely what you mean

vilsbole: It's a request to test the selector to ensure that the value is correctly transformed from a string to a `Bigish`.

vilsbole (approved):

PR #3857, opened 2021-02-23
Bump Lodash

FrederikBolding:
As per Snyk recommendation


vilsbole:
Would this be a good time to remove lodash?

FrederikBolding:
> Would this be a good time to remove lodash?

We still have Lodash because of deps using it, so I don't see this as a big priority.

vilsbole (approved):

PR #3860, opened 2021-02-24
Stop expecting ethscan to be supported

blurpesec:
### Description
Refactors the token scan to not expect ethscan to support the network.


FrederikBolding (changes requested):
Looks good, can we test the `BalanceService` functions?

FrederikBolding (approved):

PR #3862, opened 2021-02-24
Update gas notices copy

FrederikBolding:


FrederikBolding:
[ch8055]

vilsbole (changes requested):
Can we use `helpArticles.ts` instead of hard coding the link our translation file ?

vilsbole (approved):

PR #3864, opened 2021-02-25
Allows multiple assets to share the same coingeckoid

blurpesec:
### Description
Improves our handling of rates.

### Changes:
1) Improves processing of rates by decreasing scope of iterations to relevant entries (was previously looping through all assets, now it only loops through assets user has that have coingeckoids - This was an issue because this loop would occur 3-4 times on initial page load based on settings changes).
2) Allows multiple assets to use the same coingeckoid (previously didn't allow this).

### DIFF
--- a/src/services/Rates/RatesProvider.tsx
+++ b/src/services/Rates/RatesProvider.tsx
@@ -36,17 +36,22 @@ const buildAssetQueryUrl = (assets: string[], currencies: string[]) => `
 const fetchDeFiReserveMappingList = async (): Promise<ReserveMapping | any> =>
   DeFiReserveMapService.instance.getDeFiReserveMap();
 
-const destructureCoinGeckoIds = (rates: IRates, assets: ExtendedAsset[]): IRates => {
+const destructureCoinGeckoIds = (rates: IRates, coinGeckoIdMapping: GeckoIds): IRates => {
   // From: { ["ethereum"]: { "usd": 123.45,"eur": 234.56 } }
   // To: { [uuid for coinGeckoId "ethereum"]: { "usd": 123.45, "eur": 234.56 } }
   const updateRateObj = (acc: any, curValue: TTicker): IRates => {
-    const asset = assets.find((a) => a.mappings && a.mappings.coinGeckoId === curValue);
-    acc[asset!.uuid] = rates[curValue];
+    const relevantAssets = Object.entries(coinGeckoIdMapping).filter(
+      ([_, coinGeckoId]) => coinGeckoId === curValue
+    );
+    relevantAssets.forEach(([assetUuid, _]) => (acc[assetUuid] = rates[curValue]));
     return acc;
   };
 
   return Object.keys(rates).reduce(updateRateObj, {} as IRates);
 };
+interface GeckoIds {
+  [key: string]: string;
+}
 
 export const RatesContext = createContext({} as State);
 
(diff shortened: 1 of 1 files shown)


src/services/Rates/RatesProvider.tsx
```diff
@@ -36,17 +36,22 @@ const buildAssetQueryUrl = (assets: string[], currencies: string[]) => `
...
+const destructureCoinGeckoIds = (rates: IRates, coinGeckoIdMapping: GeckoIds): IRates => {
   // From: { ["ethereum"]: { "usd": 123.45,"eur": 234.56 } }
   // To: { [uuid for coinGeckoId "ethereum"]: { "usd": 123.45, "eur": 234.56 } }
   const updateRateObj = (acc: any, curValue: TTicker): IRates => {
-    const asset = assets.find((a) => a.mappings && a.mappings.coinGeckoId === curValue);
-    acc[asset!.uuid] = rates[curValue];
+    const relevantAssets = Object.entries(coinGeckoIdMapping).filter(
+      ([_, coinGeckoId]) => coinGeckoId === curValue
+    );
+    relevantAssets.forEach(([assetUuid, _]) => (acc[assetUuid] = rates[curValue]));
```
FrederikBolding, line 46: Can we do this using a `reduce`?

src/services/Rates/RatesProvider.tsx
```diff
@@ -36,17 +36,22 @@ const buildAssetQueryUrl = (assets: string[], currencies: string[]) => `
...
     return acc;
   };
 
   return Object.keys(rates).reduce(updateRateObj, {} as IRates);
 };
+interface GeckoIds {
```
FrederikBolding, line 52: Can we just use `Record` instead of defining a new type?

src/services/Rates/RatesProvider.tsx
```diff
@@ -96,7 +101,7 @@ export function RatesProvider({ children }: { children: React.ReactNode }) {
 
   useEffect(() => {
     worker.current = new PollingService(
-      buildAssetQueryUrl(geckoIds, Object.keys(Fiats)), // @todo: More elegant conversion then `DEFAULT_FIAT_RATE`
+      buildAssetQueryUrl(Object.values(geckoIds), Object.keys(Fiats)), // @todo: More elegant conversion then `DEFAULT_FIAT_RATE`
```
FrederikBolding, line 104: Can we filter by unique here?

src/services/Rates/RatesProvider.tsx
```diff
@@ -111,7 +116,7 @@ export function RatesProvider({ children }: { children: React.ReactNode }) {
       worker.current.stop();
       worker.current.close();
     };
-  }, [geckoIds.length]);
+  }, [Object.values(geckoIds).length]);
```
FrederikBolding, line 119: Can we filter by unique here?

FrederikBolding (changes requested):

GuillaumeRx:
This will enter in conflict with the rates migration to redux :(

FrederikBolding:
> This will enter in conflict with the rates migration to redux :(

Then make sure this is reflected in your version 😊

GuillaumeRx:
> Then make sure this is reflected in your version 😊

Well... this file doesn't exist and it'll already handle assets with no coingecko ids. Those functions have also been refactored

FrederikBolding:
> > Then make sure this is reflected in your version 😊
> 
> Well... this file doesn't exist and it'll already handle assets with no coingecko ids. Those functions have also been refactored

Okay, but the fix should still be reflected in the new version. So we just need to test that this behavior is in the redux version and add it if not. It doesn't make sense to redo rates without fixing this too.

GuillaumeRx:
> Okay, but the fix should still be reflected in the new version. So we just need to test that this behavior is in the redux version and add it if not. It doesn't make sense to redo rates without fixing this too.

I totally agree with you, it just doesn't make sense to add code that will be deleted soon :)

blurpesec:
This was fixed in #3869 !

PR #3870, opened 2021-02-25
Remove dead code

vilsbole:
Remove Favorite icon and associated hook calls.


FrederikBolding:
I thought we wanted to keep this around to use after launch?

vilsbole:
IMHO since we are currently reworking tables and there is no plan to reintroduce favorites it's easier to add the functionality later rather than maintain dead code.

blurpesec (approved):

PR #3876, opened 2021-03-01
Update donate contact label and desc

vilsbole:
[[ch8126]](https://app.clubhouse.io/mycrypto/story/8126/address-book-identicon-squished-w-long-label-note-update-tip-jar-copy)

1. Update default contact label and description
2. Limit column size to avoid squish
3. Add migration to update existing contact addresses

### DIFF
--- a/src/components/EditableText.tsx
+++ b/src/components/EditableText.tsx
@@ -84,7 +84,7 @@ function EditableText({
   };
 
   return (
-    <Box variant="rowCenter" height="100%" {...props}>
+    <Box variant="rowAlign" height="100%" {...props}>
       {editMode ? (
         <SInputField
           autoFocus={true}
--- a/src/components/EditableText.tsx
+++ b/src/components/EditableText.tsx
@@ -6,11 +6,7 @@ import editIcon from '@assets/images/icn-edit.svg';
 import { Typography } from '@components';
 import { COLORS, SPACING } from '@theme';
 
-const Wrapper = styled.div`
-  display: flex;
-  height: 100%;
-  align-items: center;
-`;
+import Box from './Box';
 
 const EditIcon = styled.img`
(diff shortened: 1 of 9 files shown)


src/components/EditableText.tsx
```diff
@@ -83,7 +84,7 @@ function EditableText({ saveValue, value, className, bold, truncate }: Props) {
   };
 
   return (
-    <Wrapper className={className}>
+    <Box variant="rowCenter" height="100%" {...props}>
```
FrederikBolding, line 87: This breaks the alignment on `AccountList`
image

src/services/Store/store/persist.config.spec.tsx
```diff
@@ -0,0 +1,17 @@
...
+import { values } from '@vendor';
+
+import { migrations } from './persist.config';
+
+describe('Persist Migrations', () => {
+  it('3: Updates the MYC_DONATE_ADDRESS in a users persistence layer', () => {
+    const migration = migrations['3'];
+    const contact = values(defaultContacts)[0];
+    const actual = migration({
+      [LSKeys.ADDRESS_BOOK]: [{ ...contact, label: 'Deprecated label' }]
```
FrederikBolding, line 12: Can we test that this keeps other entries in the address book, just as a sanity check?

FrederikBolding (changes requested):

src/services/Store/store/persist.config.spec.tsx
vilsbole: Updated.

src/components/EditableText.tsx
vilsbole: Fixed.

FrederikBolding (approved):

PR #3882, opened 2021-03-04
Fix message signing with WalletConnect

Mrtenz:
This fixes an issue where you are unable to sign messages when using WalletConnect. Before, the connection would be automatically killed because the WalletConnect component would be unmounted. This PR does a few things:

* Adds a `autoKill` prop to `useWalletConnect` (enabled by default, disabled for `SignMessage`).
* Changes the type of `WALLETS_CONFIG.WalletConnect` to `MISC`. This may break non-signing related things.

### DIFF
--- a/package.json
+++ b/package.json
@@ -37,7 +37,7 @@
     "@types/styled-system": "5.1.10",
     "@ungap/promise-any": "1.1.1",
     "@unstoppabledomains/resolution": "1.3.6",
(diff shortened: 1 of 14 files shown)


Mrtenz:
This should be tested by @bowensanders before merging.

vilsbole (approved):

PR #3883, opened 2021-03-05
Segment proxy env file change

blurpesec:
### Description
Changed the env file config for the segment proxy


blurpesec:
@vilsbole I checked both segment and amplitude to make sure that this is passing params through correctly. Can you poke around this just to validate my thoughts? If you approve - don't merge. I'll need to change the env variable that we use.

vilsbole (approved):

PR #3892, opened 2021-03-15
Changed to use sendMax checkbox instead of button

blurpesec:
[ch7069]

### Description
Changes to use a checkbox instead of a button for sendMax

### DIFF
--- a/src/features/SendAssets/components/SendAssetsForm.tsx
+++ b/src/features/SendAssets/components/SendAssetsForm.tsx
@@ -666,6 +666,7 @@ const SendAssetsForm = ({ txConfig, onComplete, protectTxButton, isDemoMode }: P
             onChange={(e) => {
               setFieldValue('amount', e.target.value);
             }}
+            disabled={isSendMax}
             asset={values.asset}
             value={values.amount}
             onBlur={() => {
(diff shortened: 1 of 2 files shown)


src/features/SendAssets/components/SendAssetsForm.tsx
```diff
@@ -265,6 +265,7 @@ const SendAssetsForm = ({ txConfig, onComplete, protectTxButton, isDemoMode }: P
   const [isEstimatingNonce, setIsEstimatingNonce] = useState(false); // Used to indicate that interface is currently estimating gas.
   const [isResolvingName, setIsResolvingDomain] = useState(false); // Used to indicate recipient-address is ENS name that is currently attempting to be resolved.
   const [fetchedNonce, setFetchedNonce] = useState(0);
+  const [isSendMax, toggleIsSendMax] = useState(false);
```
FrederikBolding, line 268: Can we disable the amount input while this is active? It seems weird to be able to change while I am supposed to be sending max

FrederikBolding (changes requested):

src/features/SendAssets/components/SendAssetsForm.tsx
blurpesec: i did so

FrederikBolding (approved):

PR #3896, opened 2021-03-17
Fix contract deployment issues

Mrtenz:


src/components/TransactionFlow/displays/FromToAccount.tsx
```diff
@@ -94,7 +94,7 @@ const FromToAccount = ({ networkId, fromAccount, toAccount, displayToAddress = t
           />
         </AddressWrapper>
       </AddressContainer>
-      {displayToAddress && (
+      {displayToAddress && toAccount.address && (
```
FrederikBolding, line 97: This should be unnecesary https://github.com/MyCryptoHQ/MyCrypto/blob/master/src/components/TransactionFlow/TxReceipt.tsx#L419

Perhaps the txType was missing causing the issue?

src/features/DeployContracts/stateFactory.tsx
```diff
@@ -68,9 +68,10 @@ const DeployContractsFactory: TUseStateReducerFactory<DeployContractsState> = ({
       chainId: network.chainId,
       nonce
     });
+
     // check if transaction fails everytime
     await getGasEstimate(network, transaction);
-    transaction.gasLimit = gasLimit;
+    transaction.gasLimit = addHexPrefix(bigify(gasLimit).toString(16));
```
FrederikBolding, line 74: Use `inputGasLimitToHex`

FrederikBolding (changes requested):

src/components/TransactionFlow/displays/FromToAccount.tsx
Mrtenz: There is no `txReceipt` (which is used to get the `txType`) at all it looks like. I'm not sure what a good way to solve that would be?

FrederikBolding: This should be set here?
https://github.com/MyCryptoHQ/MyCrypto/blob/master/src/features/DeployContracts/stateFactory.tsx#L99-L135

FrederikBolding (approved):

PR #3913, opened 2021-03-31
Remove UI flash in Suspense

vilsbole:
[[ch8440]](https://app.clubhouse.io/mycrypto/story/8440/fix-apploading-design)

Previously our AppLoading was using a DashboardPanel.
The flash would create a fold jump.
We smoothen the interaction by removing the panel.

Screenshot 2021-03-31 at 19 58 26


GuillaumeRx (approved):
LGTM;

PR #3915, opened 2021-04-01
Performance improvement when adding multiple accounts

blurpesec:
### Description

Batch create/update contacts to decrease the number of updates we make when adding multiple accounts.

### DIFF
--- a/src/services/Store/Contact/useContacts.tsx
+++ b/src/services/Store/Contact/useContacts.tsx
@@ -70,10 +70,9 @@ function useContacts() {
   const createContacts = (items: Contact[]) => {
     dispatch(
       createManyContacts(
-        items.map((item) => {
-          const uuid = generateDeterministicAddressUUID(item.network, item.address);
-          return { ...item, uuid };
-        })
+        items.map((item) => 
+          ({ ...item, uuid: generateDeterministicAddressUUID(item.network, item.address) })
+        )
       )
     );
   };
--- a/src/services/Store/Contact/useContacts.tsx
+++ b/src/services/Store/Contact/useContacts.tsx
@@ -5,9 +5,11 @@ import isEmpty from 'lodash/isEmpty';
 import { STATIC_CONTACTS } from '@config';
 import {
   createContact as createAContact,
+  createContacts as createManyContacts,
   destroyContact,
   selectContacts,
   updateContact as updateAContact,
+  updateContacts as updateManyContacts,
   useDispatch,
   useSelector
 } from '@store';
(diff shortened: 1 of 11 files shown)


src/services/Store/Contact/useContacts.tsx
```diff
@@ -65,10 +67,24 @@ function useContacts() {
...
 
+  const createContacts = (items: Contact[]) => {
+    dispatch(
+      createManyContacts(
+        items.map((item) => 
+          ({ ...item, uuid: generateDeterministicAddressUUID(item.network, item.address) })
+        )
+      )
+    );
+  };
```
vilsbole, line 78: Caller should be obnoxious of the uuid format.
Use action creator for setting the uuid.

```
type createContact(items: Contact[], generateUUID: (any) => TUuid = generateDeterministicUUID)
```

src/services/Store/Contact/useContacts.tsx
```diff
@@ -118,7 +134,9 @@ function useContacts() {
     contacts,
     contactRestore,
     createContact,
+    createContacts,
```
vilsbole, line 137: We are working to remove the hooks and let the components connect directly to the store.

src/services/Store/StoreProvider.tsx
```diff
@@ -120,7 +121,12 @@ export const StoreProvider: React.FC = ({ children }) => {
   const { assets } = useAssets();
   const { settings, updateSettingsAccounts } = useSettings();
   const { networks } = useNetworks();
-  const { createContact, contacts, getContactByAddressAndNetworkId, updateContact } = useContacts();
+  const {
+    createContacts,
```
vilsbole, line 125: Prefer `connect` over hook.

src/services/Store/StoreProvider.tsx
```diff
@@ -389,24 +395,41 @@ export const StoreProvider: React.FC = ({ children }) => {
...
+                  network: rawAccount.networkId,
+                  uuid: generateUUID()
+                }
+              ]
+            };
+          }
+          return acc;
+        },
+        { newContacts: [] as ExtendedContact[], updatedContacts: [] as ExtendedContact[] }
+      );
```
vilsbole, line 430: This looks like a create or update logic.
It should be the responsibility of the slice.
Move to redux.

FYI: there appears to already be a createOrUpdate defined

vilsbole (changes requested):

FrederikBolding:
Closing in favor of #4009

PR #3928, opened 2021-04-08
Optimise dev e2e

vilsbole:
Use `paste` option for form filling to decrease test time. 

Before
Screenshot 2021-04-08 at 16 32 46

After
Screenshot 2021-04-08 at 16 27 26


GuillaumeRx (approved):
LGTM 🚀

PR #3935, opened 2021-04-15
Speed up test runs

FrederikBolding:
Is roughly 60% faster at running tests on my machine compared to having `maxWorkers=2`


package.json
```diff
@@ -262,4 +262,4 @@
       "pre-push": "yarn test"
     }
   }
-}
+}
```
Mrtenz, line 265: Can you add a newline?

package.json
```diff
@@ -221,7 +221,7 @@
     "build": "cross-env TARGET_ENV=default webpack --config webpack_config/production.js",
     "build:staging": "cross-env TARGET_ENV=staging webpack --config webpack_config/production.js",
     "prebuild": "check-node-version --package",
-    "test": "jest --config=jest_config/jest.config.json --maxWorkers=2",
+    "test": "jest --config=jest_config/jest.config.json",
```
Mrtenz, line 224: Can we move `jest.config.json` to the root (like in signer)?

Mrtenz (changes requested):

vilsbole (approved):

vilsbole (dismissed):

package.json
FrederikBolding: Out of scope

Mrtenz (approved):

PR #3937, opened 2021-04-16
Use redux-persist migration on settings files

GuillaumeRx:


src/services/Store/store/helpers.ts
```diff
@@ -94,30 +90,6 @@ export const mergeAssets = (inbound: ExtendedAsset[], original: ExtendedAsset[])
...
-    const diff = difference(keys(store), keys(toImport));
-    return diff.length === 0;
-  }
-};
-
-export const migrateConfig = (toImport: Partial<LocalStorage>) => {
-  return {
-    ...toImport,
-    // @ts-expect-error rates are present in settings on data to be migrated, want to move it at root of persistence layer
-    rates: toImport.settings?.rates ? toImport.settings.rates : toImport.rates,
```
FrederikBolding, line 114: We need to support these cases in our new migrations, as with the current logic the `rates` fx. will be nuked when importing a settings file which has already migrated its rates.

src/services/Store/store/account.slice.ts
```diff
@@ -121,12 +121,12 @@ export const selectTxsByStatus = (status: ITxStatus) =>
...
   a
     .flatMap((a) => a.assets)
-    .reduce((acc, asset) => [...acc, getAssetByUUID(asset.uuid)(s)!], [] as ExtendedAsset[])
+    .reduce((acc, asset) => [...acc, getAssetByUUID(asset.uuid)(s)], [] as ExtendedAsset[])
 );
 
 export const getAccountsAssetsMappings = createSelector([getAccountsAssets], (assets) =>
   assets.reduce(
-    (acc, a) => ({ ...acc, [a.uuid]: a.mappings }),
+    (acc, a) => (a ? { ...acc, [a.uuid]: a.mappings } : acc),
```
FrederikBolding, line 129: Can an asset be undefined?

yarn.lock
```diff
@@ -1111,13 +1111,20 @@
...
+"@babel/runtime@^7.0.0", "@babel/runtime@^7.1.2", "@babel/runtime@^7.10.2", "@babel/runtime@^7.11.2", "@babel/runtime@^7.12.1", "@babel/runtime@^7.12.5", "@babel/runtime@^7.2.0", "@babel/runtime@^7.3.1", "@babel/runtime@^7.4.4", "@babel/runtime@^7.5.0", "@babel/runtime@^7.5.1", "@babel/runtime@^7.5.4", "@babel/runtime@^7.5.5", "@babel/runtime@^7.7.2", "@babel/runtime@^7.7.6", "@babel/runtime@^7.8.4", "@babel/runtime@^7.8.7", "@babel/runtime@^7.9.2":
   version "7.12.18"
   resolved "https://registry.yarnpkg.com/@babel/runtime/-/runtime-7.12.18.tgz#af137bd7e7d9705a412b3caaf991fe6aaa97831b"
   integrity sha512-BogPQ7ciE6SYAUPtlm9tWbgI9+2AgqSam6QivMgXgAT+fKbgppaj4ZX15MHeLC1PVF5sNk70huBu20XxWOs8Cg==
   dependencies:
     regenerator-runtime "^0.13.4"
 
+"@babel/runtime@^7.6.3":
```
FrederikBolding, line 1121: Why the yarn.lock change without any changes to deps?

src/services/Store/store/root.reducer.ts
```diff
@@ -59,17 +58,17 @@ export function* importSaga() {
 }
 
 function* importWorker({ payload }: PayloadAction<string>) {
-  const persistable = yield select(exportState);
   try {
-    const json = migrateConfig(JSON.parse(payload));
+    const migrated: LocalStorage = yield call(
+      // @ts-expect-error: bad type choice on call effect
+      migrate,
+      marshallState(JSON.parse(payload)),
```
FrederikBolding, line 65: We should probably change our export function to stop requiring `marshallState`. Out of scope for this PR though and will break some exported settings.

FrederikBolding (changes requested):

src/services/Store/store/account.slice.ts
GuillaumeRx: Yes it can if the account has some balance but the app doesn't have this asset in its list. Had this problem with the settings file we use in e2e

yarn.lock
GuillaumeRx: I think it's when we tried to bump redux-saga while debugging

src/services/Store/store/helpers.ts
GuillaumeRx: IMO if the rates are nuked it's not that bad they will be re-fetched right after the import. This problem will be fixed by adding the persist version to the export file

src/services/Store/store/helpers.ts
```diff
@@ -95,27 +94,11 @@ export const mergeAssets = (inbound: ExtendedAsset[], original: ExtendedAsset[])
...
-    trackedAssets: toImport.trackedAssets ? toImport.trackedAssets : {},
-    // @ts-expect-error rates are present in settings on data to be migrated, want to move it at root of persistence layer
-    settings: toImport.settings?.rates ? dissoc('rates', toImport.settings) : toImport.settings
-  } as LocalStorage;
+  const diff = difference(keys(toImport), keys(store));
+  return diff.length === 0;
```
FrederikBolding, line 101: This will cause older imports to not work because they have a `password` key.

src/services/Store/store/helpers.ts
FrederikBolding: I would prefer if we support these small migrations for stability 😄

yarn.lock
FrederikBolding: Can we remove it then?

FrederikBolding (changes requested):

FrederikBolding (approved):

PR #3939, opened 2021-04-20
Use Ethers FallbackProvider

FrederikBolding:
Needs testing before merge to make sure it makes a reasonable amount of requests to nodes and is working in all the ways we expect.


FrederikBolding:
Behaviour seems to be more along the lines of what we expect since version `5.1.1`

blurpesec (approved):

PR #3983, opened 2021-05-27
[WIP] Reduce token scans further

FrederikBolding:


FrederikBolding:
[ch8595]

jest_config/__fixtures__/transaction.ts
```diff
@@ -49,6 +49,7 @@ const toTxReceipt = (fixtureTxReceipt: any): ITxReceipt => {
   };
   return {
     ...result,
+    metadata: result.metadata ?? undefined,
```
Mrtenz, line 52: Why the `?? undefined`?

Mrtenz (changes requested):

jest_config/__fixtures__/transaction.ts
FrederikBolding: Because the fixtures may or may not include it. And we test with strict equality so if they aren't present as undefined it fails.

Mrtenz (approved):

PR #3987, opened 2021-05-31
Fix swap gas limit estimation

FrederikBolding:


src/services/ApiService/Dex/Dex.ts
```diff
@@ -118,6 +118,7 @@ export default class DexService {
       data.allowanceTarget !== AddressZero
         ? {
             ...formatApproveTx({
+              fromAddress: (account as TAddress) ?? undefined,
```
Mrtenz, line 121: `?? undefined` seems redundant here?

Mrtenz (changes requested):

src/services/ApiService/Dex/Dex.ts
FrederikBolding: No, because `null` and `undefined` aren't the same

FrederikBolding: `account` can be `string` or `null`

Mrtenz: Can't you change the type of account to `string | undefined` then?

FrederikBolding: Sure

Mrtenz (approved):

PR #4001, opened 2021-06-04
Move TX History to Redux

FrederikBolding:


src/services/Store/StoreProvider.tsx
```diff
@@ -240,11 +236,7 @@ export const StoreProvider: React.FC = ({ children }) => {
 
   useEffect(() => {
     if (mainnetAccounts.length > 0) {
-      HistoryService.instance.getHistory(mainnetAccounts).then((history) => {
-        if (history !== null) {
-          setTxHistory(history);
-        }
-      });
+      dispatch(fetchHistory());
```
GuillaumeRx, line 239: Can we move this conditional in the saga and dispatch the event at `handleRehydrateSuccess` in `persistence.slice.ts` ?

GuillaumeRx (changes requested):

FrederikBolding:
[ch8342]

src/services/Store/StoreProvider.tsx
FrederikBolding: We need it to run every time the accounts change though. So that is not enough.

FrederikBolding: Possibly. Will look into this.

GuillaumeRx (approved):

PR #4014, opened 2021-06-10
Stop users from using too many decimals

FrederikBolding:
[ch8002]

### DIFF
--- a/src/features/SendAssets/components/SendAssetsForm.tsx
+++ b/src/features/SendAssets/components/SendAssetsForm.tsx
@@ -85,6 +85,7 @@ import {
   formatSupportEmail,
   fromTokenBase,
   gasStringsToMaxGasBN,
+  getDecimals,
   isSameAddress,
   isVoid,
   sortByLabel,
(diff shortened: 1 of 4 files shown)


src/translations/lang/en.json
```diff
@@ -979,6 +979,7 @@
     "SWAP_EXPIRY_TOOLTIP": "Your quoted amount and rate will expire after a time. To refresh this, select \"Get New Quote.\"",
     "EXPIRED": "Expired",
     "SWAP_INSUFFICIENT_FUNDS": "Insufficient funds for transaction",
-    "ESTIMATED_COST": "Estimated Cost"
+    "ESTIMATED_COST": "Estimated Cost",
+    "TOO_MANY_DECIMALS": "The token you have selected supports a maximum of $decimals decimals"
   }
 }
```
Mrtenz, line 985: Can you add a new line

src/features/SendAssets/components/SendAssetsForm.tsx
```diff
@@ -318,6 +319,14 @@ export const SendAssetsForm = ({ txConfig, onComplete, protectTxButton }: ISendF
             const asset = this.parent.asset;
             if (!isEmpty(account)) {
               const balance = getAccountBalance(account, asset.type === 'base' ? undefined : asset);
+              const decimals = getDecimals(value);
+              if (decimals > asset.decimal) {
+                return this.createError({
```
Mrtenz, line 324: Can this be tested

src/utils/units.ts
```diff
@@ -101,6 +101,9 @@ const convertTokenBase = (value: TokenValue, oldDecimal: number, newDecimal: num
   return toTokenBase(fromTokenBase(value, oldDecimal), newDecimal);
 };
 
+export const getDecimals = (value: string) =>
+  value.includes('.') ? value.split('.')[1].length : 0;
```
Mrtenz, line 105: I don't know if we allow decimal values with `,` instead of `.`, but if we do, this won't always work

Mrtenz (changes requested):

src/utils/units.ts
FrederikBolding: We force `.`. Because my locale would normally use `,` 😄

GuillaumeRx: European powaaaaaaaaa

Mrtenz (approved):

PR #4017, opened 2021-06-10
Fix Ledger squished link

GuillaumeRx:


FrederikBolding (changes requested):
This component is only used for SignMessage it looks like.

And your changes break styling

Old
image

New
image

FrederikBolding (approved):

PR #4020, opened 2021-06-11
Remove tipjar from contacts

FrederikBolding:


FrederikBolding:
[ch8243]

GuillaumeRx (changes requested):
Please add a new migration in persist.config to remove the mycrypto tip jar if it exists in LS

FrederikBolding:
@GuillaumeRx I asked Jordan about this and we decided just to do it for new users. Do you think its important?

GuillaumeRx:
@FrederikBolding nope I tough it might be better if it was removed for everyone :) But if you prefer as it is I'm also fine with it

FrederikBolding:
@GuillaumeRx Not a big deal imo.

GuillaumeRx (approved):
We can always add the migration later, merging

PR #4027, opened 2021-06-15
Enhance polling saga tests

GuillaumeRx:


FrederikBolding (approved):

FrederikBolding (changes requested):
Can we test the final case?

FrederikBolding (approved):

PR #4035, opened 2021-06-17
Use own token list for Swaps

FrederikBolding:
Unleash the full potential of MyCrypto Swaps 🔥

### DIFF
--- a/src/features/SwapAssets/stateFormFactory.tsx
+++ b/src/features/SwapAssets/stateFormFactory.tsx
@@ -1,9 +1,11 @@
+import { useEffect } from 'react';
+
 import axios from 'axios';
 
 import { MYC_DEX_COMMISSION_RATE } from '@config';
 import { checkRequiresApproval } from '@helpers';
 import { DexService, getGasEstimate } from '@services';
-import { getAssetsByNetwork, selectNetwork, useSelector } from '@store';
+import { getAssetsByNetwork, getBaseAssetByNetwork, selectNetwork, useSelector } from '@store';
 import translate from '@translations';
 import { ISwapAsset, ITxGasLimit, Network, NetworkId, StoreAccount } from '@types';
 import {
@@ -34,7 +36,19 @@ const swapFormInitialState = {
 
 const SwapFormFactory: TUseStateReducerFactory<SwapFormState> = ({ state, setState }) => {
   const network = useSelector(selectNetwork(state.selectedNetwork)) as Network;
+  const baseAsset = useSelector(getBaseAssetByNetwork(network));
   const assets = useSelector(getAssetsByNetwork(state.selectedNetwork));
+  const sortedAssets = assets.sort((asset1, asset2) =>
+    (asset1.ticker as string).localeCompare(asset2.ticker)
+  );
+
+  useEffect(() => {
+    setState((prevState: SwapFormState) => ({
+      ...prevState,
+      fromAsset: baseAsset,
+      toAsset: sortedAssets.filter((a) => a.uuid !== baseAsset.uuid)[0]
+    }));
+  }, [network]);
 
   const setNetwork = (network: NetworkId) => {
     setState((prevState: SwapFormState) => ({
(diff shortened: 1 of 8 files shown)


src/features/SwapAssets/stateFormFactory.tsx
```diff
@@ -34,51 +34,26 @@ const swapFormInitialState = {
...
 
-const BASE_ASSET_ADDRESS = '0xeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee';
-
 const SwapFormFactory: TUseStateReducerFactory<SwapFormState> = ({ state, setState }) => {
   const network = useSelector(selectNetwork(state.selectedNetwork)) as Network;
+  const baseAsset = useSelector(getBaseAssetByNetwork(network));
+  const assets = useSelector(getSwapAssetsByNetwork(state.selectedNetwork));
+  const sortedAssets = assets.sort((asset1, asset2) =>
+    (asset1.ticker as string).localeCompare(asset2.ticker)
+  );
```
Mrtenz, line 43: Why is `asset1.ticker` not properly inferred here? Is this an issue with the selector again?

FrederikBolding: This was copied from elsewhere, the types can now be inferred. Will fix

src/features/SwapAssets/stateFormFactory.tsx
```diff
@@ -34,51 +34,24 @@ const swapFormInitialState = {
...
+  const assets = useSelector(getSwapAssetsByNetwork(state.selectedNetwork));
+  const sortedAssets = assets.sort((asset1, asset2) => asset1.ticker.localeCompare(asset2.ticker));
 
-  const setNetwork = (network: NetworkId) => {
+  useEffect(() => {
     setState((prevState: SwapFormState) => ({
       ...prevState,
-      selectedNetwork: network
+      fromAsset: baseAsset,
+      toAsset: sortedAssets.filter((a) => a.uuid !== baseAsset.uuid)[0]
```
blurpesec, line 47: Can you explain this?

src/services/ApiService/Dex/Dex.ts
```diff
@@ -95,8 +95,8 @@ export default class DexService {
     const { data }: { data: DexTrade } = await this.service.get('swap/v1/quote', {
       baseURL: this.getBaseURL(network.id),
       params: {
-        sellToken: sellToken.ticker,
-        buyToken: buyToken.ticker,
+        sellToken: sellToken.contractAddress ?? sellToken.ticker,
+        buyToken: buyToken.contractAddress ?? buyToken.ticker,
```
blurpesec, line 99: why contract address or ticker?

src/services/Store/store/asset.slice.ts
```diff
@@ -78,7 +78,14 @@ export default slice;
  * Selectors
  */
 
-export const getAssets = createSelector([getAppState], (s) => s[slice.name]);
+export const getAssets = createSelector([getAppState], (s) => s.assets);
+export const getSwapAssets = createSelector([getAssets], (assets) =>
+  assets.filter((a) => a.isCustom || a.isSwapRelevant)
```
blurpesec, line 83: are custom assets always swap-supported? would probably prefer to exclude unless we know for certain

FrederikBolding: The 0x API supports all ERC20s, it will simply fail if there isn't enough liquidity. This is why we landed on using the filtered list + custom tokens.

src/features/SwapAssets/stateFormFactory.tsx
FrederikBolding: We preselect `baseAsset` as the fromAsset. Then for the toAsset, we preselect the first value from the asset list that isn't the `baseAsset`, since we can't swap between two assets that are the same. It's just a way to preselect some assets before the user starts selecting things.

src/services/ApiService/Dex/Dex.ts
FrederikBolding: The API supports either addresses or tickers. Since all our ERC20 assets have an attached `contractAddress`, we can use that and for the very few assets that don't, we can use the ticker. The assets without `contractAddress` will be the base asset for the network.

blurpesec (approved):

PR #4041, opened 2021-06-21
Clean up ENS logic

### LINKED ISSUE #4042: ENS support incorrect, doesn't support all ENS names
### Description of Issue

When I try to send ETH to the ENS name [brantly.art](https://app.ens.domains/name/brantly.art), it doesn't recognize the name and throws the error "Domain brantly.art is not supported".

This likely means that MyCrypto has a hardcoded list of domain suffixes in order to recognize an ENS name. This is incorrect. You should instead treat any dot-separated name as a potential ENS name and attempt a look-up. ENS is about to fully integrate the DNS namespace (adding ~1300 suffixes to ENS), and the DNS suffix namespace itself expands over time. So a hardcoded list will be constantly out of date. If you use a library like etherjs, it'll work correctly and be future proof.

### What exactly were you doing when you encountered this issue?

Screen Shot 2021-06-22 at 8 32 05 PM

### How were you accessing your wallet?

MetaMask

- Browser: latest Chrome
- Browser version: latest
- Operating system: macos
- Device: Macbook Pro
- MyCrypto.com vs Desktop Application?: Website
- MyCrypto Version Number: latest

...

FrederikBolding:
Fixes #4042


GuillaumeRx (approved):

src/services/EthService/network/providerHandler.ts
```diff
@@ -150,6 +151,18 @@ export class ProviderHandler {
     );
   }
 
+  public resolveENSName(name: string): Promise<string | null> {
```
Arachnid, line 154: Would you consider swapping the order of these around? ENS is by far the largest distributed naming service, and is an open-source public good. Further, unstoppable has just issued a bunch of new TLDs which are practically guaranteed to conflict with ICANN's next allotment of DNS top level domains.

FrederikBolding: I was under the impression that the supported number of domains for Unstoppable was much smaller than ENS, hence we check whether the domain is supported by them and for all else go straight to ENS.

Can you elaborate on how that would be an issue? I don't see the error case 🤔

Arachnid: Unstoppable just launched 8 new TLDs.

At the moment none of Unstoppable's TLDs conflict with ENS or DNS, but when ICANN opens the next DNS TLD auction, some or all of them will certainly be bought, meaning that there will be conflicts between DNS domains and Unstoppable ones, with different people claiming ownership of the same name on different systems.

Since ENS integrates the DNS namespace, that will mean that ENS will also conflict with Unstoppable.

While obviously we're biased, we think ENS's approach is more sustainable - and if there's a conflict it's better to go with the global DNS root than Unstoppable's private for-profit service.

Here's a bit on our approach to TLDs and the root: https://medium.com/the-ethereum-name-service/why-ens-doesnt-create-more-tlds-responsible-citizenship-in-the-global-namespace-7e66658fe2b1

FrederikBolding: Great points! Will make sure to have a discussion with the team about this. Thanks for the heads up!

brantlymillegan: But why would you prioritize UD in the first place? ENS is far more widely integrated, 4.5x more integrations than ENS. In a potential future conflict (which is likely and entirely foreseen by UD's actions), why would you default to UD in that conflict?

brantlymillegan: I don't even really know why you support them at all (even as a secondary after ENS), given that they're just a for-profit knock-off of ENS with far less ecosystem support, and that will likely result in name collisions with ENS in the future (they know this). Name collisions mean user confusion, potential lost funds, phishing, etc, precisely the things that MyCrypto has worked hard to avoid.

jspence425: We support Unstoppable because they reached out and they did the legwork on it. We love ENS and we also love being open source and allowing open source contributions.

There was no decision about "prioritizing" one over the other - the ideal scenario is for both to work. There's no statement about prioritization and we're trying to accommodate the bug you pointed out without breaking any existing logic if possible. That's all.

Happy to discuss future handling at another time - for now the priority is to fix the bug you pointed out.

brantlymillegan: Hey jordan, what Nick is talking about is in your code you check UD first and then fall back to ENS (UD has priority). This is problematic for future name collisions between the two systems (likely given what UD is doing)

jspence425: Understood! And @FrederikBolding will make adjustments where necessary to make sure both work properly. I think that was solved after his discussion above with Nick and he brought it up to us internally.

brantlymillegan: If there are name collisions between the two systems in the future, which seems likely now given UD's chosen path, there is no way to make them both work properly, and it leads to the things i listed above: user confusion, potential lost funds, phishing, etc

jspence425: I have no issue revisiting this in the near future to make sure we address this potential collision so we can avoid those things!

FrederikBolding: Just to clarify there was no intention to give UD higher priority than ENS with this change. The idea was to keep the functionality as close to what we had before I started fixing the bug you reported. 

From my understanding the number of supported domains for UD was quite a lot fewer, so in my mind it was simpler to check whether it was one of the few supported domains and then always fall back to ENS for everything else. As you guys have brought up already, this can cause issues that we were unaware of. This is why I said we would have an internal discussion about this to come up with a solution. 

Thanks for making us aware of the problem 👍 But I want to stress that the information was already well received the first time you reached out 😄

FrederikBolding: Also, with the version of the UD library that we are using we are currently only resolving `.crypto` names. So the issue will only occur if/when the UD library updates and we update our dependency.

https://github.com/unstoppabledomains/resolution/blob/3f33b380777b5bd9c8041fa3552027267485fd80/src/Cns.ts#L117-L123

brantlymillegan: > Just to clarify there was no intention to give UD higher priority than ENS with this change. The idea was to keep the functionality as close to what we had before I started fixing the bug you reported.

Yes, I understand, and also I want to say thank you so much for responding to these issues so quickly! We appreciate the support MyCrypto has provided for many years. 👊

> So the issue will only occur if/when the UD library updates and we update our dependency.

FYI, .crypto names may also potentially conflict with DNS and ENS in the future too. UD doesn't own any of these TLDs or have them reserved in DNS. Since they are obviously desirable TLDs, they will most likely be created in the next gTLD round for DNS (in the next few years). Unless UD wins all of them (possible but highly unlikely in my mind), then there is (an entirely predictable) conflict between them and DNS and ENS (ENS is following the DNS namespace). So just .crypto is a problem by itself; all of their new TLDs just compound the problem further. In any event, can discuss further off Github.

FrederikBolding:
[ch8685]

PR #4046, opened 2021-07-01
Remove cached assets

FrederikBolding:


src/services/Store/store/asset.slice.test.ts
```diff
@@ -72,6 +72,27 @@ describe('AccountSlice', () => {
     expect(actual).toEqual(expected);
   });
 
+  it('addFromAPI(): merges current assets with new assets', () => {
+    const a1 = { uuid: 'random', contractAddress: '0x0', type: 'erc20' } as ExtendedAsset;
```
Mrtenz, line 76: Prefer `const a1: ExtendedAsset = { ... }` over casting with `as`.

Mrtenz (changes requested):

src/services/Store/store/asset.slice.test.ts
FrederikBolding: Will simplify by using `fAssets` instead.

Mrtenz (approved):

PR #4057, opened 2021-07-16
Normalize quotes when parsing signed message

Mrtenz:
Fixes an issue with macOS sometimes changing quotes, which makes the signed message invalid.


Mrtenz:
[ch8705]

FrederikBolding (approved):

PR #4082, opened 2021-08-11
Fix a few things for the upcoming Geth update

FrederikBolding:
Merge when Geth releases the next update and we update our nodes


FrederikBolding:
[ch8642]

Mrtenz (approved):

PR #4107, opened 2021-09-01
Actions panel > completed icon

409H:
[CH8845]

🎉 🎉 🎉

## Description

* Updated logic to display a green tick when an action has been completed instead of a low opacity tick
Screenshot 2021-09-01 at 02 07 45

## Changes

* New SVG into assets
* Code change to check state to show default icon or completed icon

## Steps to Test

1. Steps
2. to
3. test

## Quality Assurance

- [ ] If your code adds new text to the UI, the added text is [translatable](https://github.com/MyCryptoHQ/MyCrypto/wiki/Contributing---Translatable-strings)


src/components/ActionsPanel/components/ActionItem.tsx
```diff
@@ -89,15 +90,20 @@ export const ActionItem = ({
...
-        onClick={() =>
-          hidden &&
-          userAction &&
-          updateUserAction(userAction.uuid, { ...userAction, state: ACTION_STATE.NEW })
-        }
-        height={hidden ? '18px' : '24px'}
-      />
+
+      {userAction && userAction.state === 'completed' ? (
+        <img src={IconCompleted} title="Action is completed" />
```
FrederikBolding, line 95: Can we add this to `Icon` instead?

FrederikBolding (changes requested):

src/components/ActionsPanel/components/ActionItem.tsx
409H: Yep, good point - check 3912380

FrederikBolding (approved):

PR #4109, opened 2021-09-01
Feature/faucet eip1559 support

MrLuit:
## Changes

* Add support for new faucet API response (EIP-1559)

### DIFF
--- a/src/features/Faucet/helpers.tsx
+++ b/src/features/Faucet/helpers.tsx
@@ -52,25 +52,51 @@ export const makeTxConfig = (
    * getContactByAddressAndNetworkId() returns ExtendedContact, which is the closest we can get.
    * The result is casted to make it compatible with ITxConfig.
    */
-  return {
-    rawTransaction: {
-      to: txResult.to,
-      value: txResult.value as ITxValue,
-      gasLimit: txResult.gasLimit,
-      data: txResult.data,
-      gasPrice: txResult.gasPrice,
-      nonce: txResult.nonce.toString() as ITxNonce,
-      chainId: txResult.chainId,
-      from: txResult.from
-    },
-    amount: formatEther(txResult.value),
-    receiverAddress: txResult.to,
-    senderAccount: (senderContact as unknown) as StoreAccount,
-    from: txResult.from,
-    asset: baseAsset,
-    baseAsset,
-    networkId: network.id
-  };
+  if ('maxFeePerGas' in txResult && 'maxPriorityFeePerGas' in txResult) {
+    /* Type 2 tx */
+    return {
+      rawTransaction: {
+        to: txResult.to,
+        value: txResult.value as ITxValue,
+        gasLimit: txResult.gasLimit,
+        data: txResult.data,
+        maxFeePerGas: txResult.maxFeePerGas,
+        maxPriorityFeePerGas: txResult.maxPriorityFeePerGas,
+        nonce: txResult.nonce.toString() as ITxNonce,
+        chainId: txResult.chainId,
+        from: txResult.from,
+        type: 2
+      },
+      amount: formatEther(txResult.value),
+      receiverAddress: txResult.to,
+      senderAccount: (senderContact as unknown) as StoreAccount,
+      from: txResult.from,
+      asset: baseAsset,
+      baseAsset,
+      networkId: network.id
+    };
+  } else {
+    /* Type 1 tx */
+    return {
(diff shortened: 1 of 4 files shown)


src/features/Faucet/helpers.tsx
```diff
@@ -71,6 +70,27 @@ export const makeTxConfig = (
...
+    /* Type 1 tx */
+    return {
+      ...txConfig,
+      rawTransaction: {
+        ...txConfig.rawTransaction,
+        gasPrice: txResult.gasPrice
+      }
+    };
+  }
 };
```
Mrtenz, line 94: Maybe use [`makeTxConfigFromTx`](https://github.com/MyCryptoHQ/MyCrypto/blob/master/src/helpers/transaction.ts#L253).

src/features/Faucet/helpers.tsx
```diff
@@ -84,20 +104,37 @@ export const makeTxReceipt = (
...
+      type: 2
+    };
+  } else {
+    /* Type 1 tx */
+    return {
+      ...txReceipt,
+      status: ITxStatus.PENDING,
+      gasPrice: BigNumber.from(txResult.gasPrice)
+    };
+  }
```
Mrtenz, line 139: Maybe use [`makeTxConfigFromTx`](https://github.com/MyCryptoHQ/MyCrypto/blob/master/src/helpers/transaction.ts#L253).

Mrtenz (changes requested):

src/features/Faucet/helpers.tsx
FrederikBolding: Agreed. @MrLuit If the responses from the Faucet API are all transaction params as hexlified strings (except chainId and type) they match type `ITxObject`. I would then use `makeTxConfigFromTx`. 

Nevertheless you can maybe convert your existing object if it doesn't match `ITxObject` entirely and the pass to `makeTxConfigFromTx` - should simplify a lot.

src/features/Faucet/helpers.tsx
FrederikBolding: Is there a reason for not just constructing the receipt based on txConfig? Fx. using: https://github.com/MyCryptoHQ/MyCrypto/blob/master/src/helpers/transaction.ts#L99

src/services/ApiService/Faucet/types.ts
```diff
@@ -15,7 +15,7 @@ export interface FaucetChallengeResponse {
   };
 }
 
-export interface FaucetSolvedChallengeResponse {
+interface FaucetSolvedChallengeResponseV1 {
```
FrederikBolding, line 18: Is there a reason to keep the old version around?

FrederikBolding (dismissed):

src/services/ApiService/Faucet/types.ts
MrLuit: @FrederikBolding If we only include V2 the faucet will break until the server is upgraded. We could probably remove it after that.

src/features/Faucet/helpers.tsx
MrLuit: Unfortunately `makeTxConfigFromTx` only seems to work for outgoing transactions (sender must be in `accounts: StoreAccount[]`). I'd say another helper for this funtionality would be nice to have but not a high priority as long as it's only used in the faucet.

src/features/Faucet/helpers.tsx
MrLuit: See other comment

src/features/Faucet/helpers.tsx
FrederikBolding: I would just say use it anyway and rewrite the incorrect data afterwards? That should still simplify your code a lot! Right?

src/features/Faucet/helpers.tsx
FrederikBolding: As soon as the tx config is created correctly shouldn't the other function work fine to adapt it to a receipt?

src/services/ApiService/Faucet/types.ts
FrederikBolding: Cool! Please make a card to do so 👍

LuckyLuker:
[ch 8947]

LuckyLuker:
oops [ch8947]

Mrtenz (approved):

PR #4139, opened 2021-09-18
Fix RSK EIP-1191 checksum errors

7alip:
## Description
There are a couple of issues based on ETH checksum validation for the following scenarios:

- **Send Asset**
  - It doesn't allow users to proceed sending assets through RSK network unless you provide lowercase address value. Because RSK uses EIP-1191 checksum validation.
 
- **Add Custom Token**
  - It doesn't automatically fill form fields and throws error when you submit the form unless you provide lowercase address value for RSK network tokens even if they  exist in assets api. The error most likely comes from [Ethers BaseProvider validation](https://github.com/ethers-io/ethers.js/blob/f2a32d0d5b4ea3721d3f3ee14db56e0519cf337f/packages/address/src.ts/index.ts#L92) when it gets [rawTokenBalance](https://github.com/MyCryptoHQ/MyCrypto/blob/95c041fd63e80220467d2e174610fb2ae0998435/src/services/EthService/network/providerHandler.ts#L67). 

- **Add Address to Address Book**
  - Similar checksum validation errors occur here as well.

## Changes

1. Updated `isValidETHRecipientAddress` validator to make it support RSK checksum errors.
2. Changed `isValidETHAddress` validator with `isValidAddress` which support also RSK checksums in Formik schemas.
3. Provided lowercase address to `providerHandler` service when user adds token on RSK network.
4. Created RSK validation error/warning messages


FrederikBolding:
@7alip Thanks for your submission! I'll try to make time to take a look at it this week! 😄

7alip:
@FrederikBolding thank you! Looking forward to your review 😊

package.json
```diff
@@ -267,4 +267,4 @@
       "pre-push": "yarn test"
     }
   }
-}
\ No newline at end of file
+}
```
FrederikBolding, line 270: Please revert whitespace changes

src/features/SendAssets/components/SendAssetsForm.tsx
```diff
@@ -200,6 +201,14 @@ export const SendAssetsForm = ({ txConfig, onComplete, protectTxButton }: ISendF
   const [isResolvingName, setIsResolvingDomain] = useState(false); // Used to indicate recipient-address is ENS name that is currently attempting to be resolved.
   const [fetchedNonce, setFetchedNonce] = useState(0);
   const [isSendMax, toggleIsSendMax] = useState(false);
+  const [selectedNetwork, setSelectedNetwork] = useState<Network>(getNetworkById(DEFAULT_NETWORK));
```
FrederikBolding, line 204: There shouldn't be a need to add more state for this. It is already present in `values`.

src/features/SendAssets/components/SendAssetsForm.tsx
```diff
@@ -200,6 +201,14 @@ export const SendAssetsForm = ({ txConfig, onComplete, protectTxButton }: ISendF
...
+
+  const TO_FIELD_ERROR = useMemo(() => {
```
FrederikBolding, line 206: I think we can add the network to `TO_FIELD_ERROR` as so: `Enter a valid $network address, ENS name, or blockchain domain`, to reduce the clutter here?

src/features/Settings/components/AddToAddressBook.tsx
```diff
@@ -59,12 +60,20 @@ export default function AddToAddressBook({ toggleFlipped, createContact }: Props
   const { getContactByAddress } = useContacts();
   const { getNetworkById } = useNetworks();
   const [isResolvingDomain, setIsResolvingDomain] = useState(false);
+  const [selectedNetwork, setSelectedNetwork] = useState<Network>(getNetworkById(DEFAULT_NETWORK));
```
FrederikBolding, line 63: Same applies here as in Send, we shouldn't need to add more state 😄

src/features/Settings/components/AddToAddressBook.tsx
```diff
@@ -59,12 +60,20 @@ export default function AddToAddressBook({ toggleFlipped, createContact }: Props
...
+
+  const TO_FIELD_ERROR = useMemo(() => {
```
FrederikBolding, line 65: Same as Send

src/services/EthService/validators.ts
```diff
@@ -83,9 +84,23 @@ export function isValidETHLikeAddress(address: string, isChecksumValid: boolean)
 
 export const isValidETHRecipientAddress = (
```
FrederikBolding, line 85: Perhaps we can rename this to remove `ETH`?

src/services/EthService/network/providerHandler.ts
```diff
@@ -163,8 +174,16 @@ export class ProviderHandler {
    */
   public async getTokenInformation(tokenAddress: TAddress): Promise<TokenInformation | undefined> {
     return this.injectClient(async (client) => {
+      let _tokenAddress = tokenAddress;
+      /**
+       * ethersjs doesn't support EIP-1191 checksum validation.
+       * This will allow to {Add Custom Token} form fields to be filled.
+       */
+      if (client.network.chainId === 30 || client.network.chainId === 31) {
+        _tokenAddress = tokenAddress.toLowerCase() as TAddress;
```
FrederikBolding, line 183: Heh, I don't like this. Have to think more about whether there is a better solution. 🤔  cc @Mrtenz

src/services/EthService/network/providerHandler.ts
```diff
@@ -62,15 +62,26 @@ export class ProviderHandler {
...
+    return this.injectClient((client) => {
+      let contractAddress = token.contractAddress;
+
+      /**
+       * It will fail when you add custom token on RSK network with EIP-1191 checksum address.
+       * Considered sending lowercase address value to pass ethersjs checksum validation
+       * https://github.com/ethers-io/ethers.js/blob/f2a32d0d5b4ea3721d3f3ee14db56e0519cf337f/packages/address/src.ts/index.ts#L92
+       */
+      if (client.network.chainId === 30 || client.network.chainId === 31) {
+        contractAddress = token.contractAddress?.toLocaleLowerCase();
```
FrederikBolding, line 74: Same as below 🤔

src/services/EthService/validators.ts
```diff
@@ -83,9 +84,23 @@ export function isValidETHLikeAddress(address: string, isChecksumValid: boolean)
...
+    // Is a invalid EIP-1191 checksum when trying to send through RSK network
+    if (!isValidRSKAddress(address, network.chainId)) {
+      return {
+        success: false,
+        name: 'ValidationError',
+        type: InlineMessageType.ERROR,
+        message: translate('CHECKSUM_RSK_ERROR', { $network: network.name })
+      };
+    } else {
+      return { success: true };
```
FrederikBolding, line 101: Could these two cases be a ternary to reduce the nesting?

FrederikBolding (changes requested):
Gave it an initial look over, there is a few things here that I feel that there might be a better solution to, but will think more about it. Thanks for your contribution!

cc @Mrtenz

FrederikBolding:
[sc-7907]

src/services/EthService/network/providerHandler.ts
Mrtenz: The better solution would be to have support for it in Ethers.js. Maybe we should open an issue about it.

src/features/SendAssets/components/SendAssetsForm.tsx
7alip: I investigated about lifting Formik state to outside of the form or to make validation schema dynamic but I couldn't find an easier way other than adding a new state. As you have noticed, I need to update the `Schema` according to `network field changes`. I will appreciate if you may suggest something regarding that issue.

src/services/EthService/network/providerHandler.ts
7alip: Let me revert the changes in this file then and discuss it with Ethers.js project.

src/features/SendAssets/components/SendAssetsForm.tsx
7alip: I found out the solution by using `when` in yup validation.

https://github.com/formium/formik/issues/1228#issuecomment-453747566

src/services/EthService/network/providerHandler.ts
7alip: Another temporary solution (until Ethers.js supports EIP-1191 validation) might be the case that if a user provides a valid EIP-1191 checksummed address, then convert it with `ethUtils.toChecksumAddress`, otherwise invalidate the input.

PR #4151, opened 2021-09-30
Fix RSK EIP1191 checksum issues

7alip:
## Description

_I created new PR according to requesting changes in #4139_

## Changes

- Renamed `isValidETHRecipientAddress` with `isValidRecipientAddress` which now takes chainId parameter.
- Removed unnecessary network state and updated schema validation.
- Added `network` parameter to `TO_FIELD_ERROR` translation.
- Reverted changes in `providerHandler` to find a better solution.

This changes are supposed to fix checksum errors in the following scenarios: 

- Send Asset through RSK and RSK Testnet networks and
- In Settings add RSK and RSK Testnet addresses to Address Book

## Issues

There are more scenarios where EIP1191 checksum validation needs to be fixed.

- Interact with Contracts
- Add Custom Token

We discussed for the better solution as [to have support for it in Ethers.js](https://github.com/MyCryptoHQ/MyCrypto/pull/4139#discussion_r717492373).


src/features/InteractWithContracts/components/Interact.tsx
```diff
@@ -154,7 +155,7 @@ const FormSchema = object().shape({
   address: object({
     value: string().test(
       'check-eth-address',
-      translateRaw('TO_FIELD_ERROR'),
+      translateRaw('TO_FIELD_ERROR', { $network: DEFAULT_NETWORK }),
```
7alip, line 158: I provided `network` to avoid displaying it as `Enter a valid $network address, ENS name, or blockchain domain`

FrederikBolding:
[sc-7907]

src/features/InteractWithContracts/components/Interact.tsx
FrederikBolding: Couldn't you move the schema inside the component so it can access the `network` prop and use that instead?

src/features/SendAssets/components/SendAssetsForm.tsx
```diff
@@ -279,9 +280,13 @@ export const SendAssetsForm = ({ txConfig, onComplete, protectTxButton }: ISendF
...
     address: object()
       .required(translateRaw('REQUIRED'))
-      .test('valid', translateRaw('TO_FIELD_ERROR'), function (value) {
-        return value && value.value && isValidETHAddress(value.value);
-      })
+      .when('network', (network: Network, schema: ObjectSchema) =>
+        schema.test('valid', translateRaw('TO_FIELD_ERROR', { $network: network.name }), function (
+          value
+        ) {
+          return value && value.value && isValidAddress(value.value, network.chainId);
```
FrederikBolding, line 287: This looks to work, but I am not sure it will be possible to send without fixing the issues in Ethers. Balance scanning etc seems to be failing on RSK right now. Have you talked to ricmoo or is there an open issue?

FrederikBolding (changes requested):
Looks good - a few small comments!

FrederikBolding:
Don't close. @7alip have you looked more at this?

kmk142789 (approved):

PR #4163, opened 2021-10-14
Add Unlock referral on sidechains

FrederikBolding:


FrederikBolding:
Need to check that this works 👍

PR #4177, opened 2021-10-22
Improve node stability

FrederikBolding:
- Reverting back to custom FallbackProvider
  - Removes wrapped quorum errors
  - Stops spamming eth_blockNumber without need
  - Solves some issues where not all providers will be tried
  - Solves an issue where `sendTransaction` wouldn't resolve because of rate limits on one node in the node list
- Reduce retry limit for nodes - fixes Infura node spam in case of hitting rate limits


LuckyLuker:
[sc - 9154]

LuckyLuker:
[sc-9154]

blurpesec (approved):
Lgtm

PR #4197, opened 2021-11-15
Fix pending state gas increase calculation

FrederikBolding:


src/utils/index.ts
```diff
@@ -114,3 +114,4 @@ export * from './fileExtensions';
 export * from './nft';
 export * from './eip1559';
 export * from './currency';
+export { default as useEffectAllDepsChange } from './useEffectAllDepsChange';
```
Mrtenz, line 117: Why not just export it normally in the first place?

src/utils/useEffectAllDepsChange.ts
```diff
@@ -0,0 +1,30 @@
+import { useEffect, useRef } from 'react';
+
+import { usePrevious } from '@vendor';
+
+// https://stackoverflow.com/questions/62974717/useeffect-when-all-dependencies-have-changed
+function useEffectAllDepsChange(fn: () => void, deps: unknown[]) {
```
Mrtenz, line 6: Can you use an arrow function?

Mrtenz (changes requested):

Mrtenz (approved):

FrederikBolding:
[sc-9232]

Mrtenz (approved):

PR #4213, opened 2021-11-29
Fix MetaMask breaking change

### LINKED ISSUE #4211: Upcoming Breaking Change to MetaMask wallet_getPermissions
In an upcoming version of MetaMask (likely shipping in the next 7 to 14 days), we are making a breaking change to the return value of `wallet_getPermissions` and `wallet_requestPermissions`. Specifically, caveat names will be removed, and the type of the caveat that contains the complete list of the user's exposed accounts will change. MyCrypto makes use of the current return value e.g. [here](https://github.com/MyCryptoHQ/MyCrypto/blob/master/src/services/EthService/web3/node.ts#L98-L100).

The good news is that migration can occur at any time. For example, to fix the above snippet, simple change it from:

```typescript
  const exposedAccounts = walletPermissions.find(
    (caveat: any) => caveat.name === 'exposedAccounts'
  ) as IExposedAccountsPermission | undefined;
```

to:

```typescript
  const exposedAccounts = walletPermissions.find(
    (caveat: any) => caveat.name === 'exposedAccounts' || caveat.type === 'restrictReturnedAccounts'
  ) as IExposedAccountsPermission | undefined;
```

and it will be compatible with both the current state of affairs and the upcoming breaking change.

We apologize for the inconvenience. If you have any questions, please let us know!

FrederikBolding:
Fixes #4211

### DIFF
--- a/src/services/EthService/web3/node.ts
+++ b/src/services/EthService/web3/node.ts
@@ -96,7 +96,7 @@ export const requestAccounts = (web3: Web3Provider): Promise<Web3RequestPermissi
 const deriveApprovedAccounts = (walletPermissions: IWeb3Permission[] | undefined) => {
   if (!walletPermissions) return;
   const exposedAccounts = walletPermissions.find(
-    (caveat: any) => caveat.name === 'exposedAccounts'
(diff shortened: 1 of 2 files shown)


FrederikBolding:
[sc-9305]

blurpesec (approved):

PR #4265, opened 2022-01-20
Fix send assets data bug

FrederikBolding:
Fixes an issue where data would be stick around if you previously tried to send an ERC20

Steps to test:

1. Go to send flow
2. Create an ERC20 tx
3. Go to confirm screen
4. Go back
5. Switch to ETH transfer and fill out form
6. Go to confirm screen
7. See that the data field is now `0x`


blurpesec (approved):
Tested and it fixes the issue :)

PR #4277, opened 2022-02-08
Bump Node.js to 16

Mrtenz:
Node.js 12 is currently maintenance only, and is soon EOL. Node.js 16 is the current LTS:


samisbakedham:
This code looks good to me.

kmk142789 (approved):

PR #4363, opened 2023-03-07
[FIX] `RNS` Resolution

ahsan-javaiid:
## Description
This PR fixes RNS resolution issue on send screen. There are two reasons due to which RNS names are not resolving.
1) Network is hardcoded in lookup field component here: https://github.com/MyCryptoHQ/MyCrypto/blob/457ec2f0a1cce537275c3b67d7e871f1ae3ae162/src/components/GeneralLookupField.tsx#L142
2) In case of RSK network RPC provider does not know RNS registry address: https://github.com/MyCryptoHQ/MyCrypto/blob/457ec2f0a1cce537275c3b67d7e871f1ae3ae162/src/services/EthService/network/helpers.ts#L28 which in case of eth, points to correct address.

## Changes
1) Remove hardcoded network in lookup field component
2) Set RNS registry address inside provider in case of RSK network

## Demo

https://user-images.githubusercontent.com/104683677/223435969-c759ed23-a148-4425-96f3-5b73f7f718a7.mp4

## Steps to test
1) Select RSK network
2) Enter any `rns` name in the address field
3) Name should resolve to address

### DIFF
--- a/src/components/GeneralLookupField.tsx
+++ b/src/components/GeneralLookupField.tsx
@@ -10,7 +10,6 @@ import {
   ProviderHandler
 } from '@services/EthService';
 import { createENSResolutionError, isResolutionError } from '@services/EthService/ens';
-import { selectDefaultNetwork, useSelector } from '@store';
 import { ErrorObject, IReceiverAddress, Network } from '@types';
 
 import AddressLookupSelector, { LabeledAddress } from './AddressLookupSelector';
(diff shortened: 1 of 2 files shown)


alepc253:
@FrederikBolding could you please take a look?

thisisyou33-publishing-networks (approved):

Phum7895 (approved):

